PR #9, opened 2016-05-07
allow to set path to via cmake.cmakePath config

stanionascu:
Sometimes you might not have cmake in the path, having such an option is convenient and allows to use and test different cmake versions without modifying the path.


vector-of-bool:
Thanks!

PR #10, opened 2016-05-10
add support for Visual Studio Win64/ARM generators

stanionascu:
As per https://cmake.org/cmake/help/v3.5/manual/cmake-generators.7.html#visual-studio-generators

support all generators starting from 2005 (due to regexp check) :)
this way no need for compilers/make-tools in the path


vector-of-bool:
Thanks!

PR #12, opened 2016-05-11
Fix MSVC output parsing

twxs:
I've fixed the regex in parseMSVCDiagnostic
There was a bad indice used to capture the `message`, the regex also reject line that begins with `\d+>` because errors were reported twice.

As MSVC never gives the column, I have added a function to adjust the error range to underline the line content instead of the the first char.

### DIFF
--- a/src/cmake.ts
+++ b/src/cmake.ts
@@ -388,8 +388,38 @@ export class CMakeTools {
         };
     }
 
+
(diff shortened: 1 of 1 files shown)


vector-of-bool:
Thanks! I'll release with these changes, and then subsequent releases will fix any issues with the parsers.

PR #15, opened 2016-06-02
allow configuration of the source directory.

yeswalrus:
Handy for multi-language projects where your workspace root and the location of the root cmake lists file may differ.


vector-of-bool:
This is a pretty useful change, thanks!

Issue #20, opened 2016-08-16
Add support for install


rtbo:
My `CMakeLists` scripts use the install directive to group all necessary DLLs and executables in the same folder, ready for deployment.
I also need this to run my product (project with exe and multiple dlls) as windows won't find the dlls in the build tree.
Install command in vscode along with `${cmake.installPrefix}` setting would be very handy.

vector-of-bool:
Sorry to be slow to respond. Would invoking the ``install`` target be insufficient?

MinnieTheMoocher:
nearly.
what is needed is also a configuration setting
`"cmake.buildDirectory": "${workspaceRoot}/install",`
analogously to
"cmake.buildDirectory": "${workspaceRoot}/build",

That setting should have the effect of adding the cmake command line option
-DCMAKE_INSTALL_PREFIX=...
with the "..."'s replaced by the contents of that configuration variable

rtbo:
This is exactly what I did in the pull-request #21 (see last commit 2bbbc27d03cee340b1a2030f16fabbd6f62a399e)

rtbo:
in fact I used `cmake.installPrefix` for consistency with `-DCMAKE_INSTALL_PREFIX`. I can replace by `cmake.installDirectory` for consistency with `cmake.buildDirectory`. I have no preference for this concern.

vector-of-bool:
Finished!

MinnieTheMoocher:
`cmake.installPrefix` is fine. the analogy to the cmake command line is good this way.

Issue #19, opened 2016-08-16
Add support for build type variable substitution in settings


rtbo:
Not relevant for every setting entry, but would allow to write
```json
"cmake.buildDirectory": "${workspaceRoot}/build/${buildType}"
```
which would substitute `${buildType}` to either `Debug` or `Release` and allow to use different build trees.

rtbo:
this is relevant only for generators that do not support multiple configurations. (visual studio)
For such, a warning should come to the user saying `${buildType}` makes no sense in `buildDirectory`.

vector-of-bool:
Done!

PR #21, opened 2016-08-21
support for ${buildType} and install

rtbo:
a few commits to solve ${buildType} (#19) and install (#20).
As `${buildType}` main purpose is to be used in `buildDirectory`, it is necessary to store `selectedBuildType` in a new persistent cache file on disk (.vscode/cmaketools.json), otherwise it is not possible to find the last build configuration from the cmake cache.
Alternative to this additional persistent cache would be to test for a cmake cache with each possible value of `${buildType}` and find the one with youngest mtime.


vector-of-bool:
Sorry to be so slow to respond. Been busy. How thoroughly have you tested with these changes? There are a few things I'd tweak, but it looks pretty good.

src/cmake.ts
```diff
@@ -317,24 +358,14 @@ export class CMakeTools {
...
-            self.selectedBuildType = 'None';
-        if (await self.isMultiConf())
-            self.selectedBuildType = 'Debug';
-        else
-            self.selectedBuildType = (await self.cache.get('CMAKE_BUILD_TYPE')).as<string>();
+        if (await self.cache.exists() && await self.isMultiConf())
+            if (self.selectedBuildType === 'None')
+                self.selectedBuildType = 'Debug';
+
         return self.selectedBuildType;
```
vector-of-bool: The logic here seems to be very different from the old logic. It should probably check for the existence of ``.cmaketools.json``, as well as defaulting to the build type in ``CMakeCache.txt`` if the ext-cache is not found, and 'None' in the case that ``CMakeCache.txt`` is also missing.

src/cmake.ts
```diff
@@ -736,9 +769,22 @@ export class CMakeTools {
...
+                .replace('${buildType}', this.selectedBuildType);
             settings_args.push("-D" + key + "=" + value);
         }
+        let prefix = self.config<string>("installPrefix");
+        if (prefix && prefix !== "") {
+            prefix = prefix
+                .replace('${workspaceRoot}', vscode.workspace.rootPath)
+                .replace('${buildType}', this.selectedBuildType);
+            if (prefix.includes(' '))
+                prefix = '"' + prefix + '"';
```
vector-of-bool: Quoting strings is probably incorrect behavior. It should handle arguments with special characters just fine.

rtbo:
Sorry for having let this stale. been busy too.
I have tested each new addition with GLFW source code on a linux laptop and also checked that I was not breaking anything (of course I can't tell if I've been exhaustive with this). Also tested on Windows.

src/cmake.ts
rtbo: Yes, that is correct. Special characters are handled by `child_process.spawn`.

vector-of-bool:
In the interim, [this issue](https://github.com/vector-of-bool/vscode-cmake-tools/issues/27) appeared! It seems that ``child_process`` does not correctly handle arguments/commands with spaces. I'll need to do some more testing.

src/cmake.ts
rtbo: Yes I agree. In fact `.cmaketools.json` is already checked for in the constructor and defaults `selectedBuildType` to `None` (check the line where _`extCache` is constructed).
I pushed ca0cec417e5dd62a021434c66a7d6a09ae18e65e to load `selectedBuildType` from `CMakeCache.txt` if it is `None`

vector-of-bool:
I'll merge this into a branch and play around with it to see if there are any behaviors that might need tweaking.

rtbo:
great!

PR #52, opened 2016-10-14
Reload CMakeCache when build changes (Fix #51)

### LINKED ISSUE #51: Switch the generator
I'm not sure that is a bug and even that I can reproduce this each time 
but I see this from time to time, this is related with a command : "Delete Cached settings and reconfigure".

I don't specify the generator explicitly  (by the way - how to do this?), and for my environment (Win10/VC) nmake is in use. however, if I'm performing the command above, VisualStudio/MSBuild generator is in use then.

bbosnjak:


bbosnjak:
I just made new commit. I did not expected it to be part of pull request.

Please check following output (note there is "\r" after clean):

[vscode] Executing cmake command: cmake --build c:/_hud/ctest/build/Release --target "clean
" --config Release -- -j 6
make: *** No rule to make target `clean
'.  Stop.
[vscode] CMake exited with status 2

vector-of-bool:
It's part of the PR since it was a commit to the same branch. In the future, start a new git branch just for a single PR.

It looks pretty good. How thoroughly have you tried out these changes?

bbosnjak:
1) (Fix #51) I am configuring embedded cross-compiler on windows with MinGW like make. I am new in CMake and my use case is not typical , this means lot of reconfiguration (delete everything in build folder and configure again). It was quite annoying with constantly getting NMake selected. I am using change for one week now and re-configuring at least 10 times per day.

bbosnjak:
2) End line (\r\n) on Windows. This is actually by mistake committed too early. Only change in _refreshTargetList is actually needed. All others are done not in Windows or followed by trim() which will remove extra '\r'. I am pushing commit that revers this (split with regex is slower then split with char).
Without this change I can only execute "all" build. All others result with make error.

Also I just realized that you implemented error parser for CMake. This does not works on Windows:
                const diag_re = /CMake (.*?) at (.*?):(\d+) \((.*?)\):\s+((?:.|\n)*?)\s*\n\n\n((?:.|\n)*)/;
You have to replace every '\n' with '\r?\n' to get it working.

vector-of-bool:
This looks good to me.

Thanks for the tip on the cmake diagnostic parsing. I didn't even notice that it wasn't working on my Windows box.

Issue #35, opened 2016-09-20
"Stop" button does not work on Windows


vector-of-bool:
Seems to be relating to the way that CMake generates processes on Windows?

MinnieTheMoocher:
Yes and no.
Of course, cmake itself is responsible for killing all its child processes it spawns.
You should only be responsible for properly killing cmake.
However, there is a trick you can do on Windows to make sure that
cmake AND all its children die when you kill it:

http://stackoverflow.com/questions/3342941/kill-child-process-when-parent-process-is-killed

bbosnjak:
To stop process on windows you have to use "taskkill". 
See: http://stackoverflow.com/questions/32705857/cant-kill-child-process-on-windows

vector-of-bool:
Thanks @bbosnjak! This will be fixed in the next release.

ykchong45:
Sorry to ask, but I don't see a stop button when configuring my project. How can I stop in the middle of configuration process?
I'm using CMake tool v1.9.1

bobbrow:
There currently is no stop button for configuration, only for build.  You can open a feature request issue if you would like us to consider adding it.

PR #60, opened 2016-10-19
Fidex stop on Windows (Issue #35)

bbosnjak:


vector-of-bool:
Thanks! I'll try this out later and see about getting it released. I've wanted to fix this bug for a while but never got around to it...

PR #63, opened 2016-10-23
Feature/green hills compiler error parser

bbosnjak:


vector-of-bool:
Thanks for this! I don't have access to certain tools, so I wouldn't have been able to do this myself.

One request, since I don't have access to test this myself: could you add a test case for this? For example, just take some lines of GHC output and test that they are parsed correctly. No need to actually invoke a compilation, just save the output strings into the test case. That way I can be sure that future refactoring doesn't spontaneously break this.

bbosnjak:
Done!

vector-of-bool:
Looks good! Should be in next release.

PR #64, opened 2016-10-24
Enhancement/optimize diagostic parser

bbosnjak:
I did some work to optimize diagnostic parser:
1) Parsing is done line by line. Only cmake diagnostics are multiline and parser is changed.
2) Diagnostics are set to DiagnosticCollection as soon one file diagnostics are collected.
3) Handling of duplicate diagnostics (error from .h included from lots of .c files).
Parser functions are moved in new class but code is same as before (except cmake error parsing).
I am using changes for few days with GHS compiler, mingw-make and ninja.


vector-of-bool:
Thanks! This looks like an awesome set of changes!

Since this is a much larger set of changes than anything I've received previously, I hope you won't mind if I ask a few questions about it and do some review. It's looking pretty good, though!

src/cmake.ts
```diff
@@ -1070,243 +1516,18 @@ export class CMakeTools {
         this.statusMessage = 'Refreshing targets...';
         const generator = this.activeGenerator;
         if (generator && /(Unix|MinGW|NMake) Makefiles|Ninja/.test(generator)) {
+            let parser = new TargetParser();
```
vector-of-bool: Could you stick with using `const` as much as possible?

src/cmake.ts
```diff
@@ -422,6 +421,453 @@ export class ConfigurationReader {
...
+
+interface IParserCollection {
+    [parser: string]: DiagnosticParser;
+}
+
+class ErrorParser extends OutputParser {
+    private buildDiagnostic: Maybe<vscode.DiagnosticCollection>;
+    private diagDiagnostic: Maybe<vscode.DiagnosticCollection>;
+    private binaryDir: string;
+    private diags_acc: Object;
```
vector-of-bool: Can we stick with `mixedCasePropertyNames`?

src/cmake.ts
```diff
@@ -422,6 +421,453 @@ export class ConfigurationReader {
...
+    }
+}
+
+interface IParserCollection {
+    [parser: string]: DiagnosticParser;
+}
+
+class ErrorParser extends OutputParser {
+    private buildDiagnostic: Maybe<vscode.DiagnosticCollection>;
+    private diagDiagnostic: Maybe<vscode.DiagnosticCollection>;
```
vector-of-bool: `diagDiagnostic` is a confusing name. What does it actually refer to?

src/cmake.ts
```diff
@@ -422,6 +421,453 @@ export class ConfigurationReader {
...
+            filepath: abspath,
+            key: diag.full,
+            diag: vsdiag,
+        }];
+    }
+}
+
+interface IParserCollection {
+    [parser: string]: DiagnosticParser;
+}
```
vector-of-bool: Instead of using an object to store things by strings, could we go with a `Map<string, DiagnosticParser>`? It is better checked by the compiler and better expresses intent at usage sites.

src/cmake.ts
```diff
@@ -422,6 +421,453 @@ export class ConfigurationReader {
...
+
+/**
+ * Parses a diagnostic message from Green Hills Compiler.
+ * Use single line error reporting when invoking GHS compiler (--no_wrap_diagnostics --brief_diagnostics).
+ *
+ * @class GHSDiagnosticParser
+ * @extends {DiagnosticParser}
+ */
+class GHSDiagnosticParser extends DiagnosticParser {
+    public parse(line: string): [boolean, Maybe<FileDiagnostic>] {
```
vector-of-bool: It seems unnecessary to return both a `boolean` and a `Maybe<...>`, as the `Maybe<T>` encodes whether something is null as part of its type. Using the `Maybe<T>` helps the compiler detect bad usages (It will produce an error if you use it without checking for `null`)

tsconfig.json
```diff
@@ -9,7 +9,7 @@
         "sourceMap": true,
         "rootDir": ".",
         "noImplicitReturns": true,
-        "strictNullChecks": true
+        "strictNullChecks": false
```
bbosnjak: I agree with you. I also prefer strictNullChecks. Problem is Throttler class in async.ts. ts compiler will generate errors when strictNullChecks.
I copied Throttler from vscode sources. I don't like to change it to be compatible with strictNullChecks. It would be best to turn off strictNullChecks per file / function but I do not know how to do it.
Any suggestion how to handle it without changing Throttler code?

src/cmake.ts
bbosnjak: I agree

src/cmake.ts
bbosnjak: Same thing as before:
diagDiagnostic = CMakeTools._diagnostics
buildDiagnostic =  CMakeTools._buildDiags

You split diagnostic to different DiagnosticCollection, I kept same design.

src/cmake.ts
bbosnjak: I agree.
Actually, I tried to do as few changes as possible 'diags_acc' was in previous code doing same thing (as local function variable)

src/cmake.ts
bbosnjak: When I volunteered to improve parser I asked for help to 'polish' changes.
I sent you Collaboration invite.

tsconfig.json
vector-of-bool: I don't think it can be done without tweaking the code at least a little bit. You could try adding the postfix-`!` operator in places where the code fails to compile to assert the non-nullness of values.

src/cmake.ts
```diff
@@ -1703,7 +1897,8 @@ export class CMakeTools {
             {
                 silent: false,
                 environment: this.config.configureEnvironment,
-            }
+            },
+            new ErrorParser(this.binaryDir, this._diagnostics)
```
bbosnjak: It is not not clear to me what do you mean. "set of diagnostics" ?

bbosnjak:
"Could the parsing be done using functions that return values instead of by populating an object?"

I do not know how to do it. For parsing I need "history". In my world (embedded C programming) functions does not have history, so Object was obvious way.
We need history for:
1) Collect all errors from one 'c' file before sending it to DiagnosticCollection.
2) Remember include file errors and send them again if there are new errors. Errors from 'h' files are generated when 'c' file is compiled, and due to the preprocessor (#ifdef) different part of 'h' file can be used by differernt 'c' files.
3) Remember lines for multiline error parser

vector-of-bool:
Now that I think about it, I'm wondering if the optimized parsing could be done using a generator function instead of parsing the entire output (which, as you've discovered, can be extremely large) or using a stateful accumulator.

Such a generator would `yield` lines from the command output, and the parser would iterate on the lines, performing the parsing. That way, diagnostics are parsed line by line, as they come in, without storing the entirety of output in a giant string. I haven't used generators in TypeScript before, so it may be a bit of learning...

I'm going to be away for a few days and won't be able to get a lot of coding done, so I won't be very active for a while. I'll still see notifications on this thread and be able to answer any questions, so I won't be completely absent.

Thanks again for your contributions!

tsconfig.json
bbosnjak: Done

src/cmake.ts
bbosnjak: SORRY:
On first 'n' lines it will return [true, null] indicating that it matched line but FileDiagnostic are **NOT** competently collected,** we will not try other parsers**.

vector-of-bool:
I am returned. I'll take a bit of a closer look at the entire changes and maybe make some tweaks. It's looking closer to getting in.

vector-of-bool:
I've pushed a few large changes that you can look at.

(You may want ot look at teach individual commit rather than all the changes at once, since moving of a lot of code was done at one point, so most of the file will be marked as changed, which makes it harder to see what actually changed).

I'm also probably going to add a few more tests tomorrow. I realized that the MSVC diags have been broken for a while and it would be great to have some test cases for the parsers.

bbosnjak:
LineParseResult is now misleading. Places where you put "PARSER_NEEDS_MORE" are actually "PARSER_FAIL" meaning that parser has failed and we should try some other parser.
"PARSER_NEEDS_MORE" are "{ status: LineParseStatus.Done, diagnostic: null }".
I do not see point of using new "status", boolean is sufficient (it can be clearly seen from how response is used).

vector-of-bool:
Ah! It finally clicked in my mind what is actually going on with the return values here. I think I got close with what I've done, but it isn't really correct. Later today I'll try adding a few more changes, and then you can see if I got it down.

I think using the TS tagged unions is the right choice here, as it provides more type safety than returning a tuple with a `Maybe<FileDiagnostic>`.

src/cmake.ts
```diff
@@ -464,17 +456,24 @@ class ErrorParser extends util.OutputParser {
     private _activeParser: Maybe<DiagnosticParser>;
     private _parserCollection: Set<DiagnosticParser>;
 
-    constructor(binaryDir: string) {
+    constructor(binaryDir: string, parsers: Maybe<string[]>) {
```
vector-of-bool: Perhapse a variadic list of parsers? Such as:

~~~typescript
constructor(binaryDir: string, ...parsers: string[])
~~~

May be simpler, and doesn't require users to pass in an explicit `null`.

vector-of-bool: Also, maybe a literal type to represent the names of parsers, rather than any arbitrary string. Like:

~~~typescript
type ParserID = ('cmake', 'gcc', 'gnuld', 'msvc', 'ghs');
~~~

Then the compiler would even be able to check for typos.

bbosnjak: You are fast! Thanks!

Let me explain changes...

Second thing that is broken with your changes:
a) parseBuildDiagnostics is not taken in account
b) cmake() function originally only handled cmake errors

My latest changes is proposal how this could be handled.
Also this fits to the one (of two) "improvements" I would like to see implemented.
If I use msvc compiler I know that gcc or other parsers will newer have match, so there is no point to execute them. This could be part of "cmake.diagParsers" (ore something similar) setting.
Default value for "cmake.diagParsers" (if not set) would be null/undefined => execute all implemented parsers.

bbosnjak:
I am confused. I answered to your comments to the "ErrorParser is constructed with parser list" but now I can not see yours and may posts?!?

vector-of-bool:
I can see a comment starting "You are fast!" Is that your comment?

I can make some tweaks at the moment and push them up. Will that cause any conflicts?

bbosnjak:
Yes, this is my comment. Feel free to push changes when you want!

bbosnjak:
Second improvement I would like to see implemented is how status() is working. Now every line is sent to vscode.window.setStatusBarMessage(). This does not help lot to monitor compiling progress.
mingw-make and ninja have extra output with progress. We could math only this lines and set status based on them.

bbosnjak:
Is there any reason why you do not use tslinter?

vector-of-bool:
I haven't used tslinter simply because I haven't. Never really crossed my mind.

I have thought about replacing the status stuff with an actual status message parser, ie. one that would read the percent and progress messages from Ninja/Make and show a more interesting representation of build progress. It is surprisingly easy to make a kind of pseudo-progress bar using the `-` and `|` characters in the status bar, as those two characters have the same width.

bbosnjak:
Added build progress parser (make & ninja) is tested with ninja and mingw-make on windows

vector-of-bool:
Thanks for all the work you've been doing!

I've recently added some greater automated end-to-end tests that go from a blank project to a final build. This adds a lot of confidence that refactors will not cause any regressing.

I will try to take another look later today to get this merged.

bbosnjak:
I merged your latest changes.
If you decide to merge changes I would like to have parser list from "ErrorParser is constructed with parser list" configurable. I can do this change once you merged this pull request.
BTW, in latest release is "server-client.js" in out folder (and no source for this file).

bbosnjak:
One more thing.
I added TrottledOutputChannel because it looks like perfect pattern to solve hangup on lots of lines to vscode.OutputChannel. It works (without it I have vscode hangs) but I do not understand why.
I expected that _accumulatedData in append() are sumetimes != empty, but this never happen.
Maybe you can look at it if I implemented use case correctly.

vector-of-bool:
Okay... I think it's all good to merge. I even got a little progress bar-type thing in there. If you feel like it's good to merge, I'd love to get this in, and maybe get it released soon.

vector-of-bool:
Ah! Those are much better characters. Should've thought of that...

bbosnjak:
In windows status bar use none monospace font (looks like Segoe UI). '-' and '|' does not have same width and progress bar change its length with updates.
I added one alternative (works with most fonts on windows). There are more options on: https://changaco.oy.lc/unicode-progress-bars/

bbosnjak:
"Added cmake.buildDiagnosticsParsers configuration" => please check and update, configuration name, spelling, grammar, etc...

This is last change I need to do my work effectively with vscode. I will be happy if you include this branch in release.

PR #86, opened 2016-11-22
Fix GNULD error regex parsing

rmekarni:
I could really use that fix... I haven't tested it !
The problem is that the :lineNumber is added to the file name so vscode can't open it when I check markers
Only Link, gcc is fine

### DIFF
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -16,7 +16,7 @@ export interface RawDiagnostic {
 
 
 export function parseGCCDiagnostic(line: string): Maybe<RawDiagnostic> {
-    const gcc_re = /^(.*):(\d+):(\d+):\s+(?:fatal )?(warning|error|note):\s+(.*)$/;
+    const gcc_re = /^(.*):(\d+):(\d+):\s+(?:fatal )?(.*):\s+(.*)$/;
     const res = gcc_re.exec(line);
     if (!res)
         return null;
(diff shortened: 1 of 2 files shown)


rmekarni:
Ok I figure it out how to test and I adjust fix, this is good (first time I test an extension, this is really good)
But I'm not sure if I can add a test. If you're going to do it, I will looking at your commit to learn.
Check mine, you will see that it's nothing, just added some space \s before ':' in some ld stdout syntax (mine actually) and replace warning|error by .* to match all language (french in my case)
You can actually notice that with my try-hard English !
Have a good day :)

vector-of-bool:
I'm unsure of what you might be trying to fix/change about the parsing just by looking at your commits. I think adding a test specifically to check what you are trying to do would be useful to help me understand. Here's the link error parser test from `extension.test.ts` (Around line 152):

~~~typescript
test('Parsing linker error', () => {
    const line = "/some/path/here:101: undefined reference to `some_function'";
    const diag = diagnostics.parseGNULDDiagnostic(line);
    assert(diag);
    if (diag) {
        assert.strictEqual(diag.line, 100);
        assert.strictEqual(diag.message, "undefined reference to `some_function'");
        assert.strictEqual(diag.file, '/some/path/here');
        assert.strictEqual(diag.severity, 'error');
        assert.strictEqual(path.posix.normalize(diag.file), diag.file);
        assert(path.posix.isAbsolute(diag.file));
    }
});
~~~

You could use this as a template for a new test case. In this case, I have a hardcoded example output from the linker that I want to check is parsed correctly. I call `parseGNULDDiagnostic` with the string and then check that the various properties on the resulting diagnostic object have the values extracted from the linker output line properly.

It's most likely that your linker output looks different, and that is why it is causing it to fail. You can copy the above test and replace the test string with the output from your linker and check that the parser extracts all the properties from the message properly.

rmekarni:
Thank you very much for helping me doing tests.
I made some tests for french support and add a test for warnings
I checked the debug and all tests succeeded
Waiting for your return

vector-of-bool:
Thanks! Looks good! I may make a few tweaks after I merge, but it looks good to merge now.

PR #100, opened 2016-12-28
Legacy cmake tools should detect clang-cl as msvc for correct debugge…

dutow:
…r choice


dutow:
Or maybe this should be added to the master branch, where only this configuration method exists?

vector-of-bool:
Sorry for the delay. I've been absent for a while. Have you tested that this all works correctly for you? Also, one comment: could the `COMPILER_ID` variable be change to `_compiler_id`? I'd rather not introduce variables that may be misconstrued as something provided by CMake. `COMPILER_ID` looks a bit to similar to `CMAKE_<LANG>_COMPILER_ID`, and it tricked me at first into looking at the CMake docs to try and find out what it was.

dutow:
Yes, I tested it locally, and also this condition is used in CMake internally (in the compiler detection code for the Visual Studio generator)

Issue #136, opened 2017-03-27
Allow override cmake-variants.json linkage with BUILD_SHARED_LIBS


mloskot:
AFAIU, the standard CMake option `BUILD_SHARED_LIBS`
* by default, is implicitly set to `OFF` (`FALSE`) in `InitializeCache.cmake` file.
* can be controlled using `linkage` setting in [cmake-variants.json](https://github.com/vector-of-bool/vscode-cmake-tools/blob/baeb727957e6145a1ee24ed551c5fdfbf15080c0/docs/build_variants.md),

Surprisingly, this option can not be controlled using `cmake.configureSettings` (somewhat related discussions are #47 and #90).

Now, I'm not sure why, but `cmake.configureSettings` seems like the "first thing to try" to customise a build.

Second, 'polluting' a working copy with extra file `cmake-variants.json`, which by the way can not live in `.vscode`, seems like a bit of an overkill.

If one can run `cmake -DBUILD_SHARED_LIBS=ON {srcdir}`, why not to allow to replicate the same in more straightforward manner using VSCode CMake Tools?

BTW, [cmake-variants.json](https://github.com/vector-of-bool/vscode-cmake-tools/blob/baeb727957e6145a1ee24ed551c5fdfbf15080c0/docs/build_variants.md#cmake-variantsjson) docs say:
> Here is the default variants created by CMake Tools when one is not present in a project directory

Is this still up to date? I know close to nothing about TypeScript, but [src/variants.ts](https://github.com/vector-of-bool/vscode-cmake-tools/blob/2c848739df04ff03ce910c29c5f83d7b83c11d18/src/variants.ts) looks like it never creates the file, does it?

p.s. Thanks a ton for the [awesome CMake Tools](https://twitter.com/mloskot/status/846067946380251139) folks. I've been waiting for such simply fantastic IDE experience for almost a decade.

vector-of-bool:
There is another request similar to this made only a few days ago (#133), so this will probably get in soon.
Also, the variants file will not be generated, it just uses that as a default if the file does not exist.

mloskot:
Thanks. I did a bit of docs tidy-up in #140

PR #140, opened 2017-03-28
Clarify cmake-variants.json is not created if one is not present

mloskot:
The presented variants configuration is just used as default
if the file does not exist.
Related to discussion in #136.


vector-of-bool:
Thanks. That wording is better.

PR #152, opened 2017-05-07
Updated README.md to add information about extension dependencies

ingmarlehmann:
I never managed to get this extension working without installing the twxs.cmake extension as well. VS code listed twxs.cmake as a dependency for vscode-cmake-tools. That is the only reason I figured it out. It was not mentioned anywhere obvious in the documentation. Then again, I'm a TLDR person, perhaps it was possible to get this extension working without installing twxs?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,7 @@
 powerful configure+build workflow for CMake-based projects within the
 Visual Studio Code editor.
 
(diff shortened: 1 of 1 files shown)


ingmarlehmann:
Awesome extension by the way, thank you!

PR #168, opened 2017-06-07
User-facing logging.

### LINKED ISSUE #161: Better diagnostics instead of command 'cmake.*' not found
I've got some time I could contribute to improving this nice extension.
After looking at questions on stack overflow I think now would be beneficial if cmake-tools could provide better diagnostics, especially with new CMake server mode.
Looking at the code I'd like to do something like this:
 1. Refactor CMakeToolsWrapper to be always present and implement commands registration.
 1. Add logging to some of output windows.

So from user perspective if the extension can not be initialized for some reason, a sane error message is shown (like 'The CMake Tools extension failed to initialize, please see logs for details) for any command, with logs really providing helpful information.

Please let me know if it something useful or you have different ideas/progress in this direction (or if there other ways to help).

ytimenkov:
* CMakeToolsWrapper doesn't inherit from API but rather aggregates it
   and provides access.
 * Added logger which writes to the same output as CMakeTools to enable
   users fix problems with the envrironment (e.g. missing CMake binary).
 * Disable CMakeTools commands and key bindings if CMake Server has
   failed to initialize. This avoids 'command not found' messages.
 * Minor fixes to make tslint happy.

Fixes #161

### DIFF
--- a/src/wrapper.ts
+++ b/src/wrapper.ts
@@ -6,158 +6,131 @@ import * as api from './api';
 import * as legacy from './legacy';
 import * as client from './client';
 import * as util from './util';
-import {config} from './config';
+import { config } from './config';
+import { log } from './logging';
 
-export class CMakeToolsWrapper implements api.CMakeToolsAPI {
-  private _impl: Promise<api.CMakeToolsAPI>;
+export class CMakeToolsWrapper {
+  private _impl?: api.CMakeToolsAPI = undefined;
 
-  constructor(private _ctx: vscode.ExtensionContext) {}
+  constructor(private _ctx: vscode.ExtensionContext) { }
 
-  async dispose() {
+  public async dispose() {
     await this.shutdown();
     this._reconfiguredEmitter.dispose();
   }
 
-  private async _sourceDir() {
-    return (await this._impl).sourceDir;
-  }
-  get sourceDir() {
-    return this._sourceDir();
-  }
-
-  private async _mainListFile() {
-    return (await this._impl).mainListFile;
-  }
-  get mainListFile() {
-    return this._mainListFile();
+  public get toolsApi(): api.CMakeToolsAPI {
+    if (this._impl)
+      return this._impl;
+    else
+      throw new Error("CMakeTools is not initialized");
   }
 
-  private async _binaryDir() {
-    return (await this._impl).binaryDir;
-  }
-  get binaryDir() {
-    return this._binaryDir();
+  public async configure(extraArgs?: string[], runPrebuild?: boolean) {
+    if (this._impl)
+      await this._impl.configure(extraArgs, runPrebuild);
   }
 
-  private async _cachePath() {
-    return (await this._impl).cachePath;
-  }
-  get cachePath() {
-    return this._cachePath();
+  public async build(target?: string) {
+    if (this._impl)
+      await this._impl.build(target);
   }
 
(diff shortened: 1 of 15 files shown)


ytimenkov:
I looked at other more popular extensions like gitlens or powershell for inspiration.

2 Further improvements to this are:
 1. Move some commands which don't require interacting with CMake (like quickstart) from CommonCMakeToolsBase directly into wrapper. This allows some functionality even before CMake is initialized. (plus inheritance vs aggregation).
 1. Make client more 'restartable', even through a dedicated command. Maybe required in some scenarios, like when paths to tools change (i.e. user doesn't have cmake and/or make in the PATH), switching generator and/or build directory.

I do think that most of problems people experiencing now with Server mode are caused by cmake being picky. I have these:
 1. I don't have make command in path => pickGenerator fails.
 1. Even after this CMake responds with 'project is not configured' to 'cache' request => whole extension fails to initialize again.

src/wrapper.ts
```diff
@@ -6,158 +6,131 @@ import * as api from './api';
...
-    return this._diagnostics();
-  }
-
-  private async _targets() {
-    return (await this._impl).targets;
-  }
-  get targets() {
-    return this._targets();
+  public async configure(extraArgs?: string[], runPrebuild?: boolean) {
+    if (this._impl)
```
ytimenkov: Alternatively this.toolsApi may be used here with some try/catch logic in top-level command execution.
E.g. configure() has it because cmake server may fail to (re)start.

vector-of-bool:
Thanks for this! I'll get it in for 0.9.7, hopefully with other changes to fix the persistent 0.9.5 issues.

vector-of-bool:
Going over it in a local copy of the branch:

The reason `this._impl` was a promise to the actual backend is that a user can invoke a command _while_ CMake Server is reloading. If they do so, then the command will execute once CMake Server is loaded on `this._impl` is resolved. It's common to need to reload it, since CMake Server doesn't allow you to modify certain things while it is running, such as environment variables.

Also, there's no need to expose `toolsApi` if the wrapper provides an identical API (That's the idea of the wrapper, to handle the difference between the two backends with one API).

I like the idea of being able to restart more easily and readily. A lot of the tweaks in this PR work towards that. There may be some additional work to get it all working properly, though.

Is there a specific motivation in making `this._impl` a non-promise?

vector-of-bool:
So I've done a refactor of `wrapper.ts` that makes it much more reliable and repeatable. It brings back async-everything, but I'm wrapping all the APIs so that errors during setup are caught as soon as possible and display an error.

ytimenkov:
@vector-of-bool,

> Is there a specific motivation in making this._impl a non-promise?

It's just another way to handle 'uninitialized' state. As you mentioned, it's either ignore user command or wait. I don't have strong opinion about it.

What I do like is your refactoring to use decorators. I was thinking about same approach, but wasn't sure if it's available.

 I saw some other extensions go beyond and add some timeout (few seconds) with option to cancel the request. E.g. if I press F7 multiple times I'll get all commands queued and executed later...

> Also, there's no need to expose toolsApi if the wrapper provides an identical API (That's the idea of the wrapper, to handle the difference between the two backends with one API).

That was as shortcut to not do many changes to tests. Essentially what I had in mind is exactly what you did by separating front- and backend APIs.

PR #169, opened 2017-06-07
Fixed version parsing and comparison

### LINKED ISSUE #163: trouble with Version 0.9.5
Today I updated to version 0.9.5. All of the commands are not working anymore. Also build and debug symbols at the bottom of vs code lost. My vs code version is 1.12.2

ytimenkov:
Fixes #163

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -189,12 +189,12 @@ export interface Version {
   patch: number;
 }
 export function parseVersion(str: string): Version {
-  const version_re = /(\d+)\.(\d+).(\d+)/;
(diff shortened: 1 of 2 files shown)


ytimenkov:
Initially I thought it was a typo in `lhs.minor == rhs.major`, but it appears that parser doesn't work correctly: mat[0] is a full match.

vector-of-bool:
D'oh! I should have written tests for this since the beginning. This should also fix some of the issues when CMake Server was made the default. Thanks for catching it!

PR #171, opened 2017-06-08
Bugfix/ts updates

ytimenkov:
After I pulled latest develop branch my `npm install` command started to fail, and then build fails.

I've made few commits, please cherry-pick if some of them are not desired.


vector-of-bool:
Thanks. I always get tripped up when I update the dev packages and everything breaks.

PR #172, opened 2017-06-08
Improve cmake-server initialization for pre-configured project

ytimenkov:
CMake source suggests that if binaryDir points at one containing CMakeCache.txt then all other handshake parameters are optional and will be read by CMake itself, thus avoiding potential mismatches between generator, toolset or similar.


vector-of-bool:
Good find. It's unfortunate that the current cmake-server docs are so far removed from the actual implementation. Maybe I should send them a PR for the docs.

PR #176, opened 2017-06-12
Reuse CMake output windows.

### LINKED ISSUE #175: Extension creates multiple output panels with same name
When CMakeTools extension starts it creates "CMake/Build" output with detection information.
After running build it creates another output with the same name, but VS code can select only the last one.

ytimenkov:
createOutputChannels were scattered across the extension which caused VSCode to create duplicate panes which were not accessible.

Fixes #175


ytimenkov:
Having OutputChannelManager may be overkill and a single global variable is sufficient.

vector-of-bool:
I've noticed this sometimes. Thanks for finding a fix!

Issue #70, opened 2016-11-10
Doesn't pass environment variables


magpier84:
Function getenv("HOME") returns empty string when application is launched via target debugging.

CMakeTools 0.7.4 (it worked on 0.7.3)
VSCode 1.7.1

vector-of-bool:
I'm not able to reproduce the issue. Do you have anything in your `debugConfig`?

magpier84:
Platform: OSX, llvm 3.9.0
I have a such settings only:
{
    "cmake.buildDirectory": "${workspaceRoot}/../build",
    "cmake.experimental.enableTargetDebugging": true
}

magpier84:
So, this is what I have investigated at the moment.

1. Running the target without "debugConfig" generates default launch configuration. This configuration is launched with empty environment. May be, because of absence of "environment" variable, needed for "cppdbg". But I am using other debugger.
2. But the main problem for me is not well formed (for 0.7.4) debugConfig string

Instead of:

"cmake.debugConfig": {
        **"all":** {
            "request": "launch",
            "type": "lldb",
            "env": {"QT_PLUGIN_PATH": "/Users/admin/Qt/5.5/clang_64/plugins"}
        }
    }

I should use:

"cmake.debugConfig": {
        "request": "launch",
        "type": "lldb",
        "env": {"DYLD_IMAGE_SUFFIX": "_debug"}
  }

vector-of-bool:
At the moment, I've only got in support for the `cpptools` debugging engine from Microsoft, which wants environment variables in `environment` rather than `env`, and they are passed in a different format (a list of objects rather than a single object of key-value paris). Are you trying to use a different debugger?

Also, yes. I did change the format of the `debugConfig`.

magpier84:
Yes, I am using "LLDB Debugger" extension. It works for me after fixing my debugConfig.
But default debug configuration (cppdbg) is still launched with missing environment variables.

vector-of-bool:
I should probably make it clearer that target debugging is specifically written against `cpptools`. So, if you fix up the `debugConfig`, is it all working well?

magpier84:
It doesn't work correctly for cppdbg (lose system environment variables) and works for lldb.
Note, that lldb - is debugger extension for cpptools.

clangen:
FWIW I have this exact same issue on macOS (but it works fine in Linux).

[...]

PR #177, opened 2017-06-13
Updated Target Debugging documentation

ytimenkov:
* Reorganized sections in the document to reflect recent  changes related to CMake Server.
 * Added example of launch.json
 * Picking debug target returns a path to selected program, which may be used in launch.json

Fixes: #70


ytimenkov:
I'm not a native speaker so please feel free to fix the language. I just had spell checker :)

vector-of-bool:
Thanks! Having more docs is super helpful.

Issue #186, opened 2017-06-20
Feature: Use Visual Studio generator Windows if available


ytimenkov:
Current implementation of CMake Tools suggest using either Ninja or NMake Makefiles generators on Windows. Unfortunately if only Visual Studio is installed none of them available by default: Ninja requires manual installation, NMake is usually not in the default PATH.

This especially hurts when CMake runs in Server mode because in Legacy mode CMake may run without generator at all and pick first available on the system (usually falling back to VS family which it detects via registry), however Server Mode requires generator to be specified.

## Proposal ##
 1. Add ability to EnvironmentManager to provide extra generators suitable for particular environment. This includes Visual Studio family and MinGW Makefiles.
 1. When detecting the availability of MAKE_PROGRAM, tests should be performed in modified environment (i.e. merged with Active Environment).

This should reduce cases when extension fails to initialize.

vector-of-bool:
Did this issues get finished with your PR?

Randshot:
**Closed**.
If this issues still persists in the current **_0.11.0-beta4_**, feel free to reopen this one or create a new one.

PR #187, opened 2017-06-20
Feature/advanced generator selection

ytimenkov:
Addresses #186 

Created issue so it can be referred in the changelog.

### DIFF
--- a/package.json
+++ b/package.json
@@ -238,9 +238,7 @@
                     "type": "array",
                     "default": [
                         "Ninja",
-                        "Unix Makefiles",
-                        "MinGW Makefiles",
-                        "NMake Makefiles"
+                        "Unix Makefiles"
                     ],
                     "description": "The preferred CMake generator(s) to use when configuring (tried in order of listing)"
                 },
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -39,7 +39,7 @@ export class Entry implements api.CacheEntry {
       advanced: boolean) {
     this._key = key;
     this._type = type;
-    if (type == api.EntryType.Bool) {
+    if (type === api.EntryType.Bool) {
       this._value = util.isTruthy(value);
     } else {
       this._value = value;
--- a/src/client.ts
+++ b/src/client.ts
@@ -291,6 +291,7 @@ export class ServerClientCMakeTools extends common.CommonCMakeToolsBase {
             (prog.progressMaximum - prog.progressMinimum);
           this.statusMessage = prog.progressMessage;
         },
+        pickGenerator: () => this.pickGenerator(),
       });
   }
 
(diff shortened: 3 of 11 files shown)


vector-of-bool:
Thanks for all these changes! The generator code has been problematic in the past, with probably the most issues (especially with Visual Studio).

A few questions:

1. Does this fulfill  #186 completely?
2. Do all the tests still pass?
3. Have you poked around with this running in the debugger, to see if it all behaves as expected?
4. I don't think the `cmake.preferredGenerators` option is used anymore in this PR. Is that correct?

ytimenkov:
> Does this fulfill #186 completely?

Yes. The issue is a placeholder could be mentioned in the changelog.

> Do all the tests still pass?

I had issues with running smoke tests because I need to have some environment active, otherwise even if I put ninja into path configure fails because it won't find compiler. Found that I can add it to test_project/.vscode/.cmaketools.json. (or probably running under vcvars would work too).

I have 4 tests failing atm. Will update comment after done fixing them.

> Have you poked around with this running in the debugger, to see if it all behaves as expected?

Yes, that's how I did it. Although I prefer to have tests - poking every time is expensive.

> I don't think the cmake.preferredGenerators option is used anymore in this PR. Is that correct?

No, it has precedence over 'suggested' generator. Look at `getPreferredGenerators()` in `common.ts`. It first takes generators from config and then appends ones from EnvironmentProvider. I decided to pass whole Generator objects around, so both toolset and architecture are not lost.
I can't say that populating toolset and platform is consistent in all scenarios, but I definitely think that platform from EnvironmentProvider should take precedence over one in configuration. Otherwise there will be inconsistency like mentioned in one of recent issues.

package.json
```diff
@@ -238,9 +238,7 @@
                     "type": "array",
                     "default": [
                         "Ninja",
-                        "Unix Makefiles",
-                        "MinGW Makefiles",
-                        "NMake Makefiles"
+                        "Unix Makefiles"
```
ytimenkov: I removed NMake and MinGW generators because they should be provided by Environment provider.
However it seems NMake is preferred over msbuild by some.
Benefit of MSBuild is that it works even without activating environment or from plain command line.

ytimenkov:
ok, test updates:
1. *Configure respects environment overrides* - I don't see how it could pass before for Server Mode, because CMake is ran as server before configuration is altered...

vector-of-bool:
Yeah... There are a few flaky tests. I need to work a bit towards making the tests more robust, but they cover a lot ATM (just not as much as I'd like).

Thanks for the catch of `configureEnvironment`. That's one that's bugged me and I never realized the problem was CMake Server. Oops.

Anyway, this all looks pretty good. I'll get it in and see about releasing it. Hopefully it will clear up some generator problems people have been having...

Issue #188, opened 2017-06-21
Run without building


NathanW2:
I suspect this might be an option I overlooked but is it possible to run a target without doing a build first?

ytimenkov:
Yes, https://github.com/vector-of-bool/vscode-cmake-tools/blob/develop/docs/target_debugging.md#running-program-from-launchjson
However it will run in the debugger (see note at the end of the section).

If you need to run without build and without debugger then you have to create a task, and for this you need to specify path to the built executable. VS Code doesn't allow `${command:}` substitutions in tasks.

Randshot:
**Closed**.
If this issues still persists in the current **_0.11.0-beta4_**, feel free to reopen this one or create a new one.

PR #189, opened 2017-06-21
Add option to build before run (buildBeforeRun)

NathanW2:
Adds options for #188 

Adds `cmake.buildBeforeRun` default `true` to allow running without building.


NathanW2:
I was also considering adding an extra option for running for once off: `Run target without building`

vector-of-bool:
Thanks!

NathanW2:
Thanks to you for making a handy plugin! :)

Issue #190, opened 2017-06-21
Proposal: Some workflow improvements


ytimenkov:
I think that one of best thing with this extension is that it facilitates very low entry barrier compared CMake's configure->generate->build approach.

If it's not clear the goal is to make CMake Tools extension more robust and behave more expected.

### Limitations of current implementation, especially with 'Server Mode': ###
(not in any particular order)
1. Changing variant actually requires clean configure, because it assumes different compiler and CMake forbids this. Now it may be just silently ignored, depending on generator and compiler used.
1. Same for generator: it is either fails to start/configure or will be silently ignored.
1. Opening a configured is very different form opening an unconfigured one: a very different sets of operations available.
1. Changing Binary dir in any way (by changing settings or any of variables being substituted) may require clean configuration and now mostly silently ignored by the extension.
1. CMake Server should be restarted in some of cases above, e.g. when having multiple binary directories with differently configured projects.

#### Introduce clear separation between 'configured' and 'unconfigured' projects. ####
This is both in API and UI level, for example:
1. *Unconfigured*: build & debug targets selectors should not be visible, only "Configure" is possible.
1. *Configured*: should be possible to build and debug, but changing variant shouldn't be allowed.

#### Binary directory should be sufficient to activate project ####
CMake Tools should "activate" backend with binary directory, which then starts providing metadata.
This means that CMake should store some extra data in binary dir, like environment and variant (likely move .cmaketools.json there).

When changing binary dir CMake Tools should reinitialize all metadata.

It should be also possible to work with externally-configured project.

#### Introduce simple/advance mode ####
If opening unconfigured project CMake Tools should ask something like "Hey, it looks like this project is CMake-based. Would you like me to initialize build system for you? (Yes/No)" and then experience should be not different from the current one. This is "simple mode".

In "advanced mode" CMake Tools just keeps registry of known build directories, allowing to switch between them or initialize new. In the latter case UX could be similar to CMake-GUI tool.

Another option is to store registry and essential settings at the same time in json configuration. Something like variants.json or [CMakeSettings.json](https://blogs.msdn.microsoft.com/vcblog/2016/11/16/cmake-support-in-visual-studio-the-visual-studio-2017-rc-update/). This is somewhat close to tasks.json.

In case of using json, once initial configuration is done, it is used only to get a user-friendly name for binary directory. As a stretch, CMake Tools may track changes to it and offer clean reconfigure when it detects changes.

#### Configuration selector ####
[...]

PR #192, opened 2017-06-22
Add configureBeforeBuild option

NathanW2:
Adds another option that I find handy. `configureBeforeBuild` default true.  Doing a configure each time normally rebuilds targets that haven't always changed at least with the project I work on.   Turning off `configureBeforeBuild` normally gives faster builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -219,6 +219,11 @@
                     "default": true,
                     "description": "Save open files before building"
                 },
+                "cmake.configureBeforeBuild": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Always configure the project before building."
+                },
                 "cmake.clearOutputBeforeBuild": {
                     "type": "boolean",
                     "default": true,
--- a/src/common.ts
+++ b/src/common.ts
@@ -995,7 +995,7 @@ export abstract class CommonCMakeToolsBase implements CMakeToolsBackend {
     if (!ok) {
       return -1;
     }
-    if (this.needsReconfigure) {
+    if (this.needsReconfigure && config.configureBeforeBuild) {
       const retc = await this.configure([], false);
       if (!!retc) return retc;
     }
--- a/src/config.ts
+++ b/src/config.ts
@@ -31,6 +31,10 @@ export class ConfigurationReader {
     return this._readPrefixed<string>('sourceDirectory') as string;
   }
 
+  get configureBeforeBuild(): boolean {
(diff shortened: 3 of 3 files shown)


ytimenkov:
I consider running configure to be a bug (at least in server mode): it doesn't reset dirty flag and therefore runs configure all the time.
It is not needed at all if build system is initialized, because CMake Tools invokes `cmake --build` which takes care of it: for VS generators it will do configure if really needed and will rely on make/ninja to check dependencies otherwise.
Was thinking to fix it with #190.

NathanW2:
That is what I was thinking as well because running `cmake --build` should know if something is different and then run configure anyway.

NathanW2:
This gives issues with the buildBeforeRun option so I will revisit later this week if it's something people want.

vector-of-bool:
It's not safe to let it re-run CMake when I call `cmake --build` while using CMake Server, since the external CMake process might touch and modify data outside the scope of the CMake Server process, and it will become out-of-date.

As such: The only good solution is to fix the non-resetting "dirty" flag. I've noticed this for a while but never done anything as my own projects never take long to configure. For projects that have a long configure step, it's important that the dirty flag be checked properly.

NathanW2:
I'm not stressed on this fix, it was only just to make my builds a little
quicker, but I can always have a work around for it.

vector-of-bool:
Doing more investigation, a big challenge is a bug in CMake Server that causes it to become "dirty" immediately after configuring finishes. It has a race on its own files in that it sees a file change _after_ it's already marked itself as clean, and the files have already been changed.

ytimenkov:
Is it because it watches for file changes? Did you report this to CMake team?

vector-of-bool:
@ytimenkov, yep it's using `libuv` to watch files. I'll open a ticket with them, I only just discovered the problem while looking into this issue yesterday.

vector-of-bool:
[Reported](https://gitlab.kitware.com/cmake/cmake/issues/17014). We'll see what happens with it.

vector-of-bool:
I realize that this PR has been open for a stupidly long time. I apologize, truly.

A recent version added more reliable dirty-checks, which may make this PR obsolete. Are you still seeing configure run even when it isn't expected?

Randshot:
Since I don't see any progress here, I will close this PR.
If you still want this feature implemented, then please reopen this PR and merge the latest *develop* branch into your PR branch.

Issue #193, opened 2017-06-22
Handle ${workspaceRoot} in cmake.environment option


dcourtois:
Hi there ! First, thanks a lot for the really great extension !

And as for this feature request : it would be great if tokens such as `${workspaceRoot}` would be correctly replaced in the `cmake.*` options.

Our application first compile a small tool, which is then used to generate code from definition files, before actually compiling the application. And this tool needs a few environment variables, with paths relative to the output binary folder to work. Since `cmake.environment` doesn't handle `${workspaceRoot}` I need to use the absolute path and so my `settings.json` becomes non-portable.

On my installed extension, I quickly patched `common.js`, in `executionEnvironmentVariables` I just replace all occurrences of `${workspaceRoot}` by the value of `vscode.workspace.rootPath` in the `config.environment` object. It works fine. The problem is when I wanted to clone your extension to submit a pull request : I noticed that it's written in TypeScript, and I have absolutely no clue how to code in this language :D

Anyway, this should not be too difficult, but it would be pretty handy :) (if I find the time to get into TypeScript, I might make the modification and submit the pull request, but not sure if / when I will)

dcourtois:
In the end I was able to take a look at the code, and here the pull request (sorry if I'm  a bit 'verbose' in both this  issue and the PR)  :  #194

dcourtois:
#194 was merged, closing this issue ! thanks again !

Issue #128, opened 2017-03-08
Build type selection not honoring `CMAKE_CONFIGURATION_TYPES`


seanmiddleditch:
I have set `CMAKE_CONFIGURATION_TYPES` in my `CMakeLists.txt` but cmake-tools does not seem to be honoring it entirely in the UI. In the build type selection dropdown, for instance, I always see the four default CMake configurations.

This is with the cmake-server enabled. The build configuration dropdown is always empty for me with it turned off.

vector-of-bool:
The extension used to respect `CMAKE_CONFIGURATION_TYPES`, but it rare for users to customize that setting. With the addition of the variants feature, it dropped by the wayside. Currently, if there is no `cmake-variants` file in a project, the four common defaults are hardcoded in. It would be useful to fall back on `CMAKE_CONFIGURATION_TYPES` instead, but I'm not sure how much work would be required.

Randshot:
**Closed**.
If this issues still persists in the current **_0.11.0-beta4_**, feel free to reopen this one or create a new one.

tonka3000:
Hey @bobbrow,

This is still an issue in 1.14.0. It would be very nice when the dropdown menu will have only valid configurations after configure (should be available via cmake-server).

Can you please reopen this issue because for me that is a feature-request.

## Usecase
I know many projects which use these values to force single config in multi config generators, including myself :smile: . A example-code for that would be
```cmake
if(NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
endif()
if(CMAKE_BUILD_TYPE)
  set(CMAKE_CONFIGURATION_TYPES ${CMAKE_BUILD_TYPE})
endif()

set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)

foreach(_config_type ${CMAKE_CONFIGURATION_TYPES})
    string(TOUPPER ${_config_type} _config_type)
    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_${_config_type} "${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}")
[...]

PR #194, opened 2017-06-22
handle vscode special vars in cmake.*environment options

dcourtois:
Related to issue #193 
I had the time to take a closer look at the source code, and I noticed that a function was already existing : common.replaceVars. So I put that in util.ts, and in config.ts I created a private function to apply this to all the values in environment objects.

Unfortunately, I couldn't test anything because I don't know how to setup VSCode to compile this extension and then install / test it ... so I hope you can take a look and tell me how to proceed from here :) I'm willing to rework the pull request (if I don't follow the coding style, if I did something not recommended, etc.) so don't hesitate to tell me if there's something wrong !

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -533,7 +533,7 @@ export abstract class CommonCMakeToolsBase implements CMakeToolsBackend {
    * @brief Read the source directory from the config
    */
   get sourceDir(): string {
-    const dir = util.replaceVars(config.sourceDirectory);
+    const dir = this.replaceVars(config.sourceDirectory);
     return util.normalizePath(dir);
   }
 
@@ -549,7 +549,7 @@ export abstract class CommonCMakeToolsBase implements CMakeToolsBackend {
    * @brief Get the path to the binary dir
    */
   public get binaryDir(): string {
-    const dir = util.replaceVars(config.buildDirectory);
+    const dir = this.replaceVars(config.buildDirectory);
     return util.normalizePath(dir, false);
   }
 
@@ -954,7 +954,7 @@ export abstract class CommonCMakeToolsBase implements CMakeToolsBackend {
     await util.writeFile(init_cache_path, init_cache_content);
     let prefix = config.installPrefix;
     if (prefix && prefix !== '') {
-      prefix = util.replaceVars(prefix);
+      prefix = this.replaceVars(prefix);
       args.push('-DCMAKE_INSTALL_PREFIX=' + prefix);
     }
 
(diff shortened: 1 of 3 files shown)


vector-of-bool:
Thanks for the input! The PR won't work as-is (There's a usage of `this` in `replaceVars` that won't play nice). I'll see about adding the functionality in the next release version.

dcourtois:
Hi, thanks for taking the time to check this ! Is there any documentation on how to work on the extension ? (what tool to install, how to compile, run tests, etc.)
I would love to help (I'm also in need of feature #128 :p)

ytimenkov:
Fairly simple. I think it's better to install nodejs, then typescript globally:
~~~
npm install -g typescript
~~~
then after checking out cmake tools repo, locally
~~~
npm install
~~~
this downloads packages mentioned in package.json (think of nuget restore).

then just select "run extension" in vs code.

I suggest checking [VS Code documentation](https://code.visualstudio.com/docs/extensions/example-hello-world), it's good.

For typescript I like [this one](https://basarat.gitbooks.io/typescript/content/docs/getting-started.html) much more compact than [official handbook](https://www.typescriptlang.org/docs/handbook/basic-types.html), and just gives answers.

That what I used. Didn't know anything about typescript 2 weeks ago :D

dcourtois:
Aha ! was missing the `npm install` instruction, thanks !! Thanks to that, I could check my modification, and it seems it works fine: I only had a small fix to replaceVars to make, and it compiles correctly, and works well on my test project.

ntwerdochlib:
@ytimenkov Perhaps the setup you described previously regarding the steps required to build the code from the repo should be documented somewhere?  Not a TypeScripter myself but want to make some changes and was unable to build the code until I stumbled upon your reply.

vector-of-bool:
@dcourtois, I think it's odd that it compiled for you... since the unbound `this` in `replaceVars` should have made sadness. Huh...

I think having a new `CONTRIBUTING.md` doc would be great.

TypeScript isn't hard to learn at all. With a good editor like VSCode, it pretty much holds your hand the whole way until you get used to it on your own. This extension is actually my first TypeScript project, and it seems to have turned out pretty well :smile:

dcourtois:
Agreed for the `contributing.md` file. And for the problem in replaceVars, I probably didn't have the error because I'm not using the `${buildType}` which is what is using `this`.

dcourtois:
Hi again, see my last commit for a way of fixing this : since some variable depends on the actual cmake stuff (such as buildType) and other do not (variables accessing generic config or workspace stuff) I split the replaceVars in 2 functions: one in util.ts and one in common.ts

This way, no more problems, and the util.ts one can be used everywhere needed.

I ran the tests also, but it generated an exception in client.ts, at a place that seemed totally unrelated to the modifications I did, so I'm not sure what to do of that.

dcourtois:
Hi, any news on this ?? I'm willing to work on that, so don't hesitate to tell me if there's anything wrong (from code convention to naming, architecture, whatever) :)

vector-of-bool:
@dcourtois, sorry for the delay. I'll see about getting this released before the day is done.

dcourtois:
Howesome, thanks a lot !
(and now worries, I was just asking to know if I needed more work on the PR, I know time can be a problem :))

Issue #198, opened 2017-06-30
CMake Tools and cpptools integration


dcourtois:
Hi there, there's been quite a lot of stuff going on in the cpptools extension recently (the official C/C++ extension, that everyone using CMake is probably using :p) and I was wondering if we could make CMake Tools and cpptools work a bit better together.

For the moment, to use intellisense, you need to create a `c_cpp_properties.json` file, and in this file you need to specify the `includePath` and the `defines` you want to use. Both things that CMake knows perfectly well. So, that would be reeeeeaaally nice if CMake Tools could have the possibility to take over this file and update it depending on the currently selected build type ! Especially on big projects like the ones I'm working on where I have something like 50+ include paths to maintain and tens of defines, changing for each build type !

What do you think ? Does it seems feasible ? (as usual, I'll try to implement that as a POC, but in the meantime, wanted to share the idea)

ytimenkov:
There is another extension which does what you're asking for, CMake Tools Helper by @maddouri.
🤔 Was thinking about it this morning: should they be merged or kept separately?

On one hand they may have different responsibilities: CMake Tools is a language-agnostic, brings functionality of running CMake to VS Code, while Helper introduces support for particular language (C++) and integrates with particular extension (from cpptools from Microsoft).

One the other hand, there are already some assumptions on cpptools and having single complete solution provides much better user experiencer than need to install 4 different extensions.
Besides, CMake is used for C++ project in 90% cases (my wild guess. other languages just complementary).
Detecting presence of cpptools dynamically is fairly easy.

ytimenkov:
Related discussion is going in #174.

dcourtois:
Ah, didn't know about CMake Tools Helper. Currently testing this. If it works, I think I'll close this issue.
I think it's still reasonable to let another extension handle this, and keep CMake Tools as simple as possible.

dcourtois:
Hmm, I've been trying to get CMake Tools Helper to work correctly, and I didn't have much chance.
Also the fact that it doesn't support the "all" target (which is the target I'm usually using) is a big show stopper for me.

Adding basic `c_cpp_properties.json` generation is pretty easy (I've done it already) and works fine. The problem is for the `"all"` target. On my project, I simply join the include paths and defines of all my targets in this case, but I can see how this would be really bad on other setup :P

[...]

PR #201, opened 2017-07-02
Feature/cpptools integration

dcourtois:
Here is the pull request for issue #198 

I've tested this on my work production projects (which have something like 150+ targets, lots of META type ones, etc.) and I didn't have any issue. Everything is configurable so it should be quite easy to tailor to anyone's use.

Let me know if you think this needs more work.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -360,11 +360,11 @@ export abstract class CommonCMakeToolsBase implements CMakeToolsBackend {
     var selectedTargets: any[] = [];
     if (targetName && codeModel) {
       const targets = codeModel.configurations[0].projects[0].targets;
-      if (targetName === "__mergeEverything") {
+      if (targetName === "__merge__") {
         for (const target of targets) {
           selectedTargets.push(target);
         }
-      } else if (targetName === "__firstValid") {
+      } else if (targetName === "__first__") {
         if (targets.length > 0) {
           selectedTargets.push(targets[0]);
         }


s-martin:
+1

vector-of-bool:
Sorry for being slow on this one... Got busy. Pinging @maddouri to take a look as well.

dcourtois:
Not a problem, I totally understand, I'm also maintaining a few things on GitHub being used by other people (far less than cmake-tools :p) and I know that sometimes personal life do not leave us enough time :)

And I see that there's a conflict. I'm going to fix it.

vector-of-bool:
No word from maddouri yet. I'll take a look into merging this. His extension offers a few more things than just cpptools integration. I'll go over to an open ticket on his project and discuss a bit more.

maddouri:
Hi guys!

@vector-of-bool Sorry for the delay. _(I had to take care of few things IRL ;) )_

@dcourtois I won't have time to review+test the PR, but from https://github.com/vector-of-bool/vscode-cmake-tools/issues/198#issuecomment-312486844 and the diff, I can say that we have different goals and approaches to the "cpptools integration" feature:

* This PR:
    1. allows "fine grained" control of `c_cpp_properties.json`
    1. allows the user to select (in the VS status bar) cpptools target that might not correspond to the currently selected CMake target
* https://github.com/maddouri/vscode-cmake-tools-helper
    1. takes control of the content of `c_cpp_properties.json` in order to have things "just work" (at least for non-meta targets... for now)
    1. automatically makes cpptools' current target the same as CMake's

My goal is to be close to what is allowed by [CLion](blog.jetbrains.com/clion/) i.e.:
1. Opening a CMake-based project
1. Starting to code immediately (without having to configure anything... Ok, maybe set some CMake Cache variables if needed, but that's all)

I do not want to have to spend any amount of time configuring anything. (I want the extension to take care of 90% of my usecases) Also, I want cpptools' target to be the same as CMake (without me having to set it manually)
For now, I've managed to get pretty close. (IMHO) The only real annoying problems that remain are:
1. Automatically adding the "standard" includes/defines (e.g. `/usr/lib/gcc/x86_64-linux-gnu/5/include`, `_LINUX`, etc.) but I will develop a small workaround soon (and, quite frankly, cpptools should detect those by itself IMHO)
1. The default CMake Tools target is always "all" ... which causes a "null" config in `c_cpp_properties.json` ... unless we find a solution/agreement in https://github.com/vector-of-bool/vscode-cmake-tools/issues/174 , I will have to develop some workaround to force CMake Tools to select another target by default

In conclusion, feel free to correct me if I'm wrong, but reading through this PR, I have the impression that it does not do what I want.
Of course, this is **totally fine** :) I do not claim that my usecases/needs reflect those of the rest of this extension's users. That is why I have developed a separate extension ;)

bobbrow:
@maddouri, cpptools does have logic to detect for system includes, but unfortunately does not currently expose the results of that detection in a way that can be consumed directly by another extension.  You can indirectly see the results when a new c_cpp_properties.json file is created by our extension because we add these paths to the "includePath" and "browse.path" at that time.

We haven't spent much time thinking about how we might expose the system includes for other extensions to read from, but if that would be of value to you, we might consider it (unfortunately our backlog is huge and the team is small).  Standard defines like `__WIN32`, `__APPLE__`, `__linux__`, etc are already embedded in the IntelliSense engine and don't need to be redefined.  I haven't seen `_LINUX` around that much so I don't know how prevalent its usage is.

- I apologize for barging in on your PR. Carry on.  :)

maddouri:
@bbosnjak Thank you for taking the time to provide such insights. I appreciate it.

> We haven't spent much time thinking about how we might expose the system includes for other extensions to read from, but if that would be of value to you, we might consider it

It would indeed benefit [cmake-tools-helper](https://github.com/maddouri/vscode-cmake-tools-helper) to get the standard includes from cpptools. That would solve [a few issues](https://github.com/maddouri/vscode-cmake-tools-helper/issues/4) ^^

I would like to suggest exposing them in the extension's API (something akin to what this extension is doing in 
https://github.com/vector-of-bool/vscode-cmake-tools/blob/develop/src/api.ts using the CMake-Server-based backend). This way, even if `c_cpp_properties.json` has not (yet) been created, other extensions can still get the information from cpptools. (also, I find this approach it more convenient to implement :)

_IMHO, ideally, we should be able to "communicate" (e.g. get/set the includes, get/set current "target" for intellisense, etc.) with cpptools through an extension's API rather than files. (sorry for the digression) _

> Standard defines ... are already embedded in the IntelliSense engine

Great!

dcourtois:
@maddouri I just noticed something: 

> allows the user to select (in the VS status bar) cpptools target that might not correspond to the currently selected CMake target

You wrote that about this pull request (based on the comment I wrote on the associated Issue, if I'm not mistaken) but I'm not sure I understand what you mean. I just handled the problem of the `all` CMake target by allowing the user to set 1 or more targets to use as a fallback whenever CMake's code model is empty. To clarify:

- when the user select a valid target in the status bar (an executable, library, anything that produces a valid code model) this target is used to add the include paths and defines, no matter what the user set in the settings
- when the user selects a targets which does not produce a valid code model (or on the first run, because `all` is the default target) then the targets specified by the users in their settings are used as a fallback.

So the only case the selected target is not the one used to fill the `c_cpp_properties.json` file is when `all` (or other targets producing a null codeModel) target is selected.

bobbrow:
@dcourtois, we are aware of some of the shortcomings of our include path detection. Our Linux testing has focused primarily on Ubuntu and Mint which are the top distros we see in our telemetry and survey data.  Like you said, the paths are compiler dependent, but we do have a backlog item to attempt to improve our defaults by using the command you posted.  We may not get the path exactly right since we would have to rely on whatever compiler is in the $PATH, but since we're not trying to build an actual binary with our IntelliSense engine, there is usually some room for error here.  And users are always free to update the path after we've attempted to detect it.  We won't overwrite your changes.

dcourtois:
@bobbrow Yep understand, I just wanted to clarify things :) And as a side note, even if you were to rely on env vars, you wouldn't be guaranteed to get the correct compiler, because CMake, make and other tools all have ways to override the default env var compiler. My point is that this standard include peaths problem is not cpptools' responsibility. Providing default generic paths for the users to extend is fine, but trying to cover every use cases possible would be a bit difficult I think :)

vector-of-bool:
So sorry to leave this sitting around. Life, uh, gets in the way. I should be able to look over this in the next few days.

It may be interesting to consider this PR in the face of cpptools officially starting work on compilation database support.

dcourtois:
Compilation database support ? What is that ?

ytimenkov:
This one: http://clang.llvm.org/docs/JSONCompilationDatabase.html

MathiasMagnus:
Just as a sidenote, there was a recent notice that compilation database support will ["soon"](https://github.com/Microsoft/vscode-cpptools/issues/156#issuecomment-332835018) land in the [vscode-cpptools](https://github.com/Microsoft/vscode-cpptools) extension. Does cmake-server mode emit a JSON that is compatible with this database format? How much change is required to vscode-cmake-tools in order to make ends meet with vscode-cpptools?

dcourtois:
From what I understand, CMake can generate this database based on the option `CMAKE_EXPORT_COMPILE_COMMANDS` since version 2.8.5. So it's "just" a matter of adding this option when configuring the project (can be done very easily vscode-cmake-tools' side)

So we just have to wait for cpptools extension to detect this database and use it to feed its intellisense engine.

vector-of-bool:
I'm going to have a discussion with the cpptools team regarding future directions with extension integrations. Stay tuned!

MathiasMagnus:
Just to let you know, the feature landed in cpptools yesterday. Now it's only a matter of making ends meet.

WOOT

dcourtois:
Is there any doc, link, anything to see how it will be supported by cpptools ? Will it simply be a setting like "compilation_database": "some/path/to/the/file" ?? (that would be awesome)

mika-fischer:
@dcourtois See https://github.com/Microsoft/vscode-cpptools/releases/tag/v0.14.0

Flamefire:
The support of the compilation database is in VSC now. What about making this PR much simpler by "just" adding the compilation database to the cpp-properties?

bobbrow:
cpptools actually just published something better than compile_commands.json support. We [export an API](https://www.npmjs.com/package/vscode-cpptools) that lets extensions provide the configurations directly.  When CMake Tools releases support for that API, there should no longer be a need to write to `c_cpp_properties.json` anymore.

Flamefire:
>Cpp-tools already knows this. You can set default values for various c_cpp_properties.json entries in user-wide settings.json and set it to the place where cmake-told places it. Set once, forget.

How? The build folders for different projects are different. What would I set there?

MathiasMagnus:
@bobbrow Awesome stuff! This is just the kind of thing the LaTeX extension would need. A place took hook in LaTeX-specific invocations. Or how does this work? Who's responsible for mustering out of the CMake Server json protocol what entries concern them? Is this something CMake tools has to support, to cater all the language extensions' APIs and predigest the json?  

@Flamefire ${workspaceRoot} works inside settings.json.

bobbrow:
@MathiasMagnus, see PR #474

PR #203, opened 2017-07-05
Add support for detecting Emscripten environments

vvuk:
[Emscripten](https://github.com/kripken/emscripten) ships with a CMake toolchain file, and it's handy to be able to switch to it as an Environment using vscode-cmake-tools.  This adds an environment detector for Emscripten, using both the `$EMSCRIPTEN` env var as well as a list of configured dirs.

I had to add the ability to merge in settings in addition to env vars from environments for this in order to set `CMAKE_TOOLCHAIN_FILE`; seems to work, but I'm not sure if there's a more canonical way of doing this.

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -21,6 +21,7 @@ export interface Environment {
   description?: string;
   mutex?: string;
   variables: Map<string, string>;
+  settings?: Object;
   preferredGenerator?: Generator;
 }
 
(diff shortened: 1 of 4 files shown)


src/environment.ts
```diff
@@ -235,6 +236,35 @@ async function tryCreateMinGWEnvironment(dir: string): Promise<Environment | und
...
+    const ret: Environment = {
+      name: `Emscripten - ${version}`,
+      mutex: 'emscripten',
+      description: `Root at ${emscripten}`,
+      settings: {
+        'CMAKE_TOOLCHAIN_FILE': cmake_toolchain
+      },
+      variables: new Map<string, string>([]),
+      preferredGenerator: {
+        name: 'Ninja'
```
ytimenkov: `preferredGenerator` may be a bad name, something environment-specific generator is more accurate.
That is generator for build tool which usually shipped with particular compiler. That's why MSBuild and MinGW Makefiles are populated here. Ninja and Unix Makefiles are generic and searched first, based on `cmake.preferredGenerators` setting.

So if Emscripten doesn't ship Ninja I suggest leave this field undefined.

vector-of-bool:
Looks good. I'll make a few tweaks and get it out.

Issue #200, opened 2017-07-02
Possibility to move `cmake-variants.json` into .vscode folder 


dcourtois:
Hi,

I've been wondering if it would be possible to have the option to move `cmake-variants.json` into the .vscode folder. The reasons why I'd like to do that :

* This file contain settings / configuration of CMake Tools. As such it should go into the dedicated folder for VSCode's extensions, eg. `.vscode` (CppTools extension put its `c_cpp_properties.json` in this folder for instance)
* That would keep the main project folder clean. Currently this file is the only one non-related to the project. For example, If I were to send my project to someone working with CMake on the command line, he wouldn't want to see this `cmake-variants.json` files. Putting it in the .vscode folder makes it easier.

That's not very important I know, and I also know that you can't change th path since lots of people are probably already using the current path. But that would be nice to have the option (and maybe by default use the .vscode folder so that new users have a clean folder)

What do you think ?

ytimenkov:
As a workaround it's possible to create `cmake.defaultVariants` setting in workspace settings.

Now I wonder whether both are necessary?

dcourtois:
I'm not sure I understand : are you suggesting that the *content* of `cmake-variants.json` could go directly in the settings ???

dcourtois:
See #204 for an implementation proposition.

dcourtois:
Merged (f3fd2e78e851676cc54340e2f9b4d72e689f04e2)

PR #204, opened 2017-07-05
allow loading cmake-variants from .vscode folder

dcourtois:
As the title says :) This is the request following issue #200

The way it works is pretty simple: it tries to look for the `cmake-variants.json` file in the usual location (workspace root) and if it doesn't find it here, it tries to look in the `.vscode` folder.

If needed, I could update the documentation so that the loading priority is explicitly defined somewhere for users (and unrelated, but GitHub, VSCode and most Markdown viewers handle relative links in Markdown files so in the doc you can use them :))


vector-of-bool:
Thanks!

PR #205, opened 2017-07-10
Fixed incorrect VS 2015 detection

oxdradium:
The default algorithm for detecting VS 2015 installations searches for a missing "vcvarsall.bat" in the "VS140COMNTOOLS" path.
So for detecting older versions of VS we also can use the new detection routine by using "vswhere" with the "-legacy" option.


vector-of-bool:
Thanks!

PR #211, opened 2017-07-14
added property "cmakePath" to configuration

Seppone:
I think this should be available :-)
At least it is listed in the docs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,6 +199,11 @@
             "type": "object",
             "title": "CMake Tools configuration",
             "properties": {
+                "cmake.cmakePath": {
+                    "type": "string",
+                    "default": "",
+                    "description": "The Path to CMake binary"
+                },
                 "cmake.buildDirectory": {
                     "type": "string",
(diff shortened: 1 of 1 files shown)


vector-of-bool:
Hehe. This option was in the original version of the extension several months ago. It was removed when the `twxs.cmake` extension also updated to include a `cmake.cmakePath` configuration option. If a user has both extensions installed (which most do), VSCode won't be happy when it tries to read the config option (because there will be two entries in the configuration file). Until that conflict is resolved (Not sure how or when), I can't include the option myself.

@twxs, what to do? Having CMake functionality split between two extensions is a bit exasperating for users.

twxs:
Maybe we could reconsider the merging option...

Seppone:
ok. I think merging your extensions would be great for your user. :-) Well...at least for me. :-)
So :+1:  for merging
:-1:  for my pull request

@vector-of-bool and @twxs : By the way, thanks for all your effort :1st_place_medal:

Issue #216, opened 2017-07-29
Failing to start debugging MSVC 14.0


BJMH:
Hey,

I'm running CMake Tools 0.10.2 on Windows 10 and using Visual C++ 2015 and 2017. In CMake Tools 0.10.1 and 0.10.2 I've had trouble starting debugging, getting the error: 

> Unable to start debugging. Launch options string provided by the project system is invalid. Unable to determine path to debugger. Please specify the "MIDebuggerPath" option.

Im able to start debugging with a launch.json, using:

```
{   
    "version": "0.2.0",  
    "configurations": [  
        {  
            "name": "Launch CMake-built program",  
            "type": "cppvsdbg",  
            "request": "launch",  
            "program": "${command:cmake.launchTargetProgramPath}",
            "args": [],
            "stopAtEntry": false,
            "cwd": "${workspaceRoot}",
            "environment": [],
            "externalConsole": true
        }
    ]
}
```

It appears to be related to line 895 in `common.ts`

`type: (this.compilerId && this.compilerId.includes('MSVC')) ? 'cppvsdbg' : 'cppdbg',`

For whatever reason `this.compilerId.includes('MSVC')` mustn't be returning true for MSVC 14.0. No idea why or how wide spread this is, because it worked on previous versions of CMake Tools where I was using 14.0. Though I have replicated this issue on two PCs. I can fix the problem for myself just by statically making the debugger `type: 'cppvsdbg'`.

Thanks

dcourtois:
This is a C++ extension related problem, not a CMake Tools related one ...

And as for your problem, it's not a problem. On windows, you need to use `"cppvsdbg"` to debug things compiled with the Visual Studio compiler (`"cppvsdbg"` is just the option telling the C++ extension to use the Visual Studio debugger)

Using `gdb` or any other debugger with the MSVC compiler will obviously not work (I'm assuming that's what you were using since you get the `Please specify the "MIDebuggerPath" option` which is related to `gdb` debugger, if I remember correctly)

BJMH:
I'm fairly certain its a CMake Tools problem, because it is line 895 in `common.ts` of the CMake Tools code that is erroneously returning `'cppdbg'` when it should be returning `'cppvsdbg'`. I am using MSVC, so `(this.compilerId && this.compilerId.includes('MSVC'))` should be `true` and therefore the type set to `'cppvsdbg'`.

[...]

PR #218, opened 2017-08-01
Fix for #216

dcourtois:
`CMAKE_CXX_COMPILER` is not always written in the CMake cache file (I don't know why, but check #216 , it seems to happen quite a lot) so I added a `linkerId` property (only in server mode, I don't know how to implement this for the legacy version) which is more reliable (it seems to always be specified)

This way the debug target command will correctly set the debugger type even when `CMAKE_CXX_COMPILER` is not in the cache.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -71,6 +71,19 @@ export class ServerClientCMakeTools extends common.CommonCMakeToolsBase {
     return null;
   }
 
+  get linkerId() {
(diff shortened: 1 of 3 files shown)


ytimenkov:
> `CMAKE_CXX_COMPILER` is not always written in the CMake cache file

Because it is written into `CMakeFiles/<version>/CMakeCXXCompiler.cmake` by some of CMake bootstrapping modules. (extensive documentation on this is in the `enable_language` command).

PR #263, opened 2017-11-26
Fix #256 - Parsing MSVC diagnostics

### LINKED ISSUE #256: MSVC multi-process build logs parsed incorrectly
Hi,

When building a project using the `Visual Studio 15 2017 Win64` generator, `vscode-cmake-tools` starts the build with the following command:

`cmake --build c:/users/foo/bar/build --target ALL_BUILD --config Debug -- /m /property:GenerateFullPaths=true`

This invokes `MSBuild.exe` with the `/m` option, which requests a multiprocess build, which in turn causes MSBuild to prefix all output with the build process it came from (simplified for demonstration purposes):

```
1>Project "C:\foo\bar\build\ALL_BUILD.vcxproj" on node 1 (default targets).
     1>Project "C:\foo\bar\build\ALL_BUILD.vcxproj" (1) is building "C:\foo\bar\build\ZERO_CHECK.vcxproj" (2) on node 1 (default targets).
     2>Done Building Project "C:\foo\bar\build\ZERO_CHECK.vcxproj" (default targets).
     1>Project "C:\foo\bar\build\ALL_BUILD.vcxproj" (1) is building "C:\foo\bar\build\bar.vcxproj" (4) on node 2 (default targets).
     4>Done Building Project "C:\foo\bar\build\bar.vcxproj" (default targets).
     1>Project "C:\foo\bar\build\ALL_BUILD.vcxproj" (1) is building "C:\foo\bar\build\example.vcxproj" (3) on node 3 (default targets).
     3>InitializeBuildStatus:
         Touching "example.dir\Debug\example.tlog\unsuccessfulbuild".
       CustomBuild:
         All outputs are up-to-date.
       ClCompile:
         C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.11.25503\bin\HostX86\x64\CL.exe /c /IC:\foo\bar\include /Zi /nologo /W3 /WX- /diagnostics:classic /Od /Ob0 /D WIN32 /D _WINDOWS /D "CMAKE_INTDIR=\"Debug\"" /D _MBCS /Gm- /EHsc /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /GR /Fo"example.dir\Debug\\" /Fd"example.dir\Debug\vc141.pdb" /Gd /TP /errorReport:queue C:\foo\bar\examples\example.cpp
         example.cpp
     3>C:\foo\bar\include\bar.hpp(67): error C2429: language feature 'init-statements in if/switch' requires compiler flag '/std:c++latest' 
```

`vscode-cmake-tools` seems to parse the file path as including the `3>`, but later parses it correctly in MSBuild's error summary as such:

capture

go2sh:
Change the parsing of the build process number from a negative look-ahead to a non-capturing group, if present.

### DIFF
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -274,7 +274,7 @@ export class MSVCDiagnosticParser extends DiagnosticParser {
 
   public parseLine(line: string): LineParseResult {
     const msvc_re =
-        /^\s*(?!\d+>)\s*([^\s>].*)\((\d+|\d+,\d+|\d+,\d+,\d+,\d+)\):\s+((?:fatal )?error|warning|info)\s+(\w{1,2}\d+)\s*:\s*(.*)$/;
+        /^\s*(?:\d+>)?\s*([^\s>].*)\((\d+|\d+,\d+|\d+,\d+,\d+,\d+)\):\s+((?:fatal )?error|warning|info)\s+(\w{1,2}\d+)\s*:\s*(.*)$/;
     const res = msvc_re.exec(line);
     if (!res) return PARSER_FAIL;
     const full = res[0];


vector-of-bool:
Thanks! I'll add a test for this.

Issue #265, opened 2017-11-28
would be very handy for substitutions to respect the ${env:VARIABLE} syntax


rolandking:
Our cmake is run locally by a makefile which substitutes in various environment variables (eg CONDA_PREFIX which we use for package management). It would be very handy to be able to substitute environment variables into paths using the ${env:VARIABLE} syntax which is used in other extensions.

Randshot:
**Closed**.
If this issues still persists in the current **_0.11.0-beta4_**, feel free to reopen this one or create a new one.

PR #266, opened 2017-11-28
support for ${env:...} syntax

rolandking:
This adds support for issue #265. It adds a regex match for ${env:...} and a replace with the value from the process environment. That's automatically picked up by many of the properties already but also an escapeArray() method has been added to allow the arrays of config/build/install parameters to pass through substitution as well.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -29,6 +29,16 @@ export class ConfigurationReader {
         `${platform}.${key}`, this.readConfig<T>(`${key}`));
   }
 
+  private _escapeArray(obj:Array<string>){
+    const acc: Array<string> = [];
+    return obj.reduce(
+      (acc,key:string) =>{
+        acc.push(util.replaceVars(key));
+        return acc;
+    },acc)
+
+  }
+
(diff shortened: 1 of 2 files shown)


Randshot:
Since this PR is obsolete, I will close it.
Please target any future PRs to the refactor-1.0 branch.
**Closed.**

rolandking:
was that issue ever addressed, I saw no follow-up on it at all, no discussion and the pull request was just ignored. That didn't make it feel like contributions are actually welcomed.

vector-of-bool:
@rolandking. I have to apologize for the way I treated this PR, and it's not the only one. Life got very busy for me around the time you sent it and the email fell to the bottom of my inbox (among many others, unfortunately).

I've recently pulled in some new collaborators (Randshot being one of them) to help make sure this doesn't happen again.

Of course, contributions are always welcome and will (hopefully) not be ignored in the future like this one was.

To answer your question, though: Yes, environment variable substitution is now implemented in the latest beta release and will be pushed to public 0.11.0 in the near future.

PR #277, opened 2017-12-20
Feature/refactor 1.0 Default choice for variant

KoeMai:
### This changes visible behavior of initial load of a project 

- small refactoring for reuse of code for variant selection for initial load
- check of default variant choice on parsing of variant configuration file
- load of default choice on start

## The purpose of this change

This modification implements the initial load of default selection of the variant choices.
On start of the extension for a unknown project with cmake_variant.json or yaml are the default selection not loaded. This modification loads the variant configuration defined default choices.
If there is a invalid default choice set then the modification will give a warning and replace the value by the first found choice.

### DIFF
--- a/src/variant.ts
+++ b/src/variant.ts
@@ -11,6 +11,7 @@ import rollbar from './rollbar';
 import {fs} from "./pr";
 import * as util from './util';
 import {MultiWatcher} from './watcher';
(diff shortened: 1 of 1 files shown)


vector-of-bool:
Thanks! I'll make a few tweaks and push it out in beta2, along with some other fixes.

Issue #279, opened 2017-12-21
Start of extension fails with state of incompatible variant


KoeMai:
### Brief Issue Summary

On testing the new beta on my windows PC I can not use the extension.
The developer mode shows a exception "Unexpected missing variant setting.".
In my last loaded project I used a other variant file then in the new project.

### Expected:

I have to exceptions:
1. The extension should always be accessible, so that I can reset the internal state for troubleshooting or reset/edit the kit selection.
2. The extension should fix a invalid internal history state without user interaction.

### Apparent Behavior:

1. Open project with variant configuration
2. select a variant
3. close project
4. open project with different variant configuration or edit variant configuration of present project  before opening
5. --> extension get exception

### Developer Tools Log

```
console.ts:123 [Extension Host] activating extension
console.ts:123 [Extension Host] starting language server
console.ts:123 [Extension Host] [CMakeTools] 2017-12-21T07:05:25.305Z [debug] [main] Safe constructing new CMakeTools instance
console.ts:123 [Extension Host] [CMakeTools] 2017-12-21T07:05:25.306Z [debug] [kit] Constructing KitManager
console.ts:123 [Extension Host] [CMakeTools] 2017-12-21T07:05:25.307Z [debug] [variant] Constructing VariantManager
console.ts:123 [Extension Host] [CMakeTools] 2017-12-21T07:05:25.309Z [debug] [main] Constructing new CMakeTools instance
console.ts:123 [Extension Host] [CMakeTools] 2017-12-21T07:05:25.310Z [debug] [main] Starting CMakeTools second-phase init
console.ts:123 [Extension Host] [CMakeTools] 2017-12-21T07:05:25.310Z [debug] [rollbar] Checking Rollbar permissions
console.ts:123 [Extension Host] [CMakeTools] 2017-12-21T07:05:25.310Z [debug] [rollbar] Rollbar enabled?  false
console.ts:123 [Extension Host] [CMakeTools] 2017-12-21T07:05:25.347Z [info] [variant] Loaded new set of variants
console.ts:123 [Extension Host] Fehler beim Aktivieren der Extension "vector-of-bool.cmake-tools": Unexpected missing variant setting.
t.log @ console.ts:123
t._logExtensionHostMessage @ extensionHost.ts:391
(anonymous) @ extensionHost.ts:211
emitTwo @ events.js:106
emit @ events.js:194
process.nextTick @ internal/child_process.js:766
[...]

PR #280, opened 2017-12-21
#279 Fix internal state history incompatiblity with variant definition.

KoeMai:
This modification handle the exception and use temporary the default variant choices.
### This changes behavior 

- Refactoring activeVariantOptions getter 
- Added new error handling if activeVariantSettings is incompatible to variants
- Added more details to error output.

## Other Notes/Information

I develop this patch on a linux machine and triggered the error cases manual.
I think it would solve my windows machine problem but there is no node.js/npm installed so I can not test it there.

### DIFF
--- a/src/variant.ts
+++ b/src/variant.ts
@@ -214,6 +214,56 @@ export class VariantManager implements vscode.Disposable {
 
   get haveVariant(): boolean { return !!this.stateManager.activeVariantSettings; }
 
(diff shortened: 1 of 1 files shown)


vector-of-bool:
Thanks for the PR. I'll take a look soon.

vector-of-bool:
Thanks! I'll make a few tweaks and get it in beta 3

PR #297, opened 2018-01-06
Prepare testing on CI

KoeMai:
Thank you for developing this vscode extension, I like the old version and the new version. The VsCode is a modern IDE that allows to easily extend it. I hope to get a nice embedded development enviroment in the future. I would like to help you advance this extension. In the last few weeks I have tried to fix some problems, I hope it was helpful. 

One problem what I had on contributing to the project was that I didn´t want to break something. I have different OS present but it is hard to synchronice them. With this Pull request I started to create a CI based on Travis and AppVeyor. It allows you to test the extension with different compilers such as CLang, GCC, different Visual Studio version and to test different operating systems. Some of the reported problems are also found on the CI.

### This changes contribution quality

- Integrate CI ( Travis and AppVeyor) to get a wide test coverage
- Replace scripts in fakebin with C++ Project to make it more realistic and portable between different operating systems
- Clean up some smaller issues


test/fakeOutputGenerator/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+
+include(CTest)
+enable_testing()
+
+add_executable(FakeOutputGenerator main.cpp)
+set_property(TARGET FakeOutputGenerator PROPERTY CXX_STANDARD 98)
+
+set(CPACK_PROJECT_NAME ${PROJECT_NAME})
+set(CPACK_PROJECT_VERSION ${PROJECT_VERSION})
+include(CPack)
```
vector-of-bool: You can probably remove the CPack and CTest lines, just to be clean

scripts/prepare_test.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+current_dir=$(pwd)
```
vector-of-bool: You can use `pushd` and `popd` to go in and out of directories more easily, even in Windows `.bat` files.

src/util.ts
```diff
@@ -42,17 +42,13 @@ export function removeAllPatterns(str: string, patterns: string[]): string {
  * case of the path.
  *
  * @param p The input path
- * @param normalize_case Whether we should normalize the case of the path
  * @returns The normalized path
  */
-export function normalizePath(p: string, normalize_case = true): string {
+export function normalizePath(p: string): string {
```
vector-of-bool: Why remove path case normalization for Windows?

KoeMai: Is was a bugfix for some test in suite "Diagnostics". 
I replaced the source directory from config by a `dummyPath` to make it independent from the config. It is out of scope of the test. 
But the test failed because on windows, the path "dummyPath" is translated to "dummypath/..." by the normalize_case parameter. I did not see the need. 

Was there a error case which makes it necessary?

test/fakeOutputGenerator/CMakeLists.txt
KoeMai: Done

scripts/prepare_test.sh
KoeMai: Done

src/util.ts
vector-of-bool: In general, not normalizing paths when you work on a filesystem with normalized paths can cause trouble. I was seeing issues where I'd generate two different file URIs and VSCode would treat them as two files, even though they actually referred to the same file.

I'd rather do the `normalizePath` on the test value just to be safe.

KoeMai: I have revert the modification and update test.

vector-of-bool:
I've merged into another branch and will set up some CI integration and we'll see what happens.

KoeMai:
Hey, 
I'm using CI from travis and appveyor. I think you have chosen the best approach building an own branch. Before merging the branch back to the refactor-1.0 branch, all tests should run successfully. I saw some failing tests on [travis] (https://travis-ci.org/KoeMai/vscode-cmake-tools)  .

PR #300, opened 2018-01-11
Support MSVC 2008

Mi-La:
### This changes

- Do not add build flags for MSVC 2008, because 2008 doesn't support these
   - '/m', '/property:GenerateFullPaths=true'

## The purpose of this change

Without this it's not possible to build within VS Code using MSVC 2008 compiler.

## Other Notes/Information

I noticed that the new beta has an TODO for this, so I beg for have it done in the final release ;-).


vector-of-bool:
Looks like a simple fix, thanks! (Sorry for the delay)

PR #302, opened 2018-01-21
Fix generator selection problem with order and ninja

KoeMai:

- Change order of prefered generators
   + Old: 
      1. Generators from settings
      2. Generators from kit
    + New:
      1. Generators from kit
      2. Generators from settings
- selector will throw only when no matching generator was found an error to the user.

- Add handling for error code


KoeMai:
Sorry, this was to early to publish. I had missunderstood some use case for preferedGenerator in the extension.

Issue #220, opened 2017-08-09
Failed to parse json generated by vswhere


kimura-synthesis:
I got following error with Windows  10(Japanese) and MSVC 2017 Community.

```
Unexpected token � in JSON at position 368: SyntaxError: Unexpected token � in JSON at position 368
    at JSON.parse (<anonymous>)
    at Object.<anonymous> (C:\Users\kkimura\.vscode\extensions\vector-of-bool.cmake-tools-0.10.2\out\src\environment.js:271:39)
    at Generator.next (<anonymous>)
    at fulfilled (C:\Users\kkimura\.vscode\extensions\vector-of-bool.cmake-tools-0.10.2\out\src\environment.js:4:58)
```
Here is output of `vswhere.exe -all -format json -product *  -legacy`.
```json
[
  {
    "instanceId": "01eb69bb",
    "installDate": "2017-07-31T06:40:09Z",
    "installationName": "VisualStudio/15.2.0+26430.16",
    "installationPath": "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community",
    "installationVersion": "15.0.26430.16",
    "displayName": "Visual Studio Community 2017",
    "description": "�w���A�I�[�v�� �\�[�X�A����ьX�̊J���҂̂��߂̖����Ŋ��S�ȋ@�\������� IDE",
    "enginePath": "C:\\Program Files (x86)\\Microsoft Visual Studio\\Installer\\resources\\app\\ServiceHub\\Services\\Microsoft.VisualStudio.Setup.Service",
    "channelId": "VisualStudio.15.Release",
    "channelPath": "C:\\Users\\kkimura\\AppData\\Local\\Microsoft\\VisualStudio\\Packages\\_Channels\\4CB340F5\\catalog.json",
    "channelUri": "https://aka.ms/vs/15/release/channel",
    "releaseNotes": "https://go.microsoft.com/fwlink/?LinkId=660284#15.1.26430.16",
    "thirdPartyNotices": "https://go.microsoft.com/fwlink/?LinkId=660300"
  }
]
```

Output encoding is CP932. I think this is vswhere's bug.

vector-of-bool:
Just opened a ticket for vswhere. We'll see what happens.

kimura-synthesis:
I've succeeded generate json file encoded with UTF-8.

Following typescript code is for your reference.
```typescript
import * as proc from 'child_process'

function runVSWhere(){
[...]

PR #306, opened 2018-01-27
parse vswhere output from text format due to i18n encoding problem.

itn3000:
### This changes

- parse vswhere output from "text" format instead of JSON.parse()

## The purpose of this change

As a premise, JSON.parse()  throws error if invalid character is found in input string.
Invalid character is almost in "Description" element.
but VsWhereItem only needs  "installationVersion", "installationPath", "displayName",
so "Description" element can be just ignored.

JSON.parse does not have "ignore invalid sequence" option, so I changed to use text format.

## Other Notes/Information

this is little bit unstable way(because it depends on vswhere output).
Please tell me if there is more efficient and stable way.

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -262,6 +262,46 @@ async function tryCreateEmscriptenEnvironment(emscripten: string): Promise<Envir
   return;
 }
 
+function parseVsWhereItemFromTextFormat(output: string): VSWhereItem[] {
+  var lines = output.split(/\r?\n/);
+  var installationPath: string = "";
+  var installationVersion: string = "";
+  var displayName = "";
+  var ret: VSWhereItem[] = [];
+  var currentItem: VSWhereItem;
+  for(var i = 0;i < lines.length;i++) {
+    var line = lines[i];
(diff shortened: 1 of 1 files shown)


vector-of-bool:
Sorry for the month of silence. I'm now getting back into the flow of things and cleaning up the project.

I remember a much earlier issue involving the output locale when invoking VSWhere.exe. I made a comment somewhere in the vswhere repo asking about this and was told that setting a certain option would fix the JSON output to contain the properly encoded characters.

itn3000:
Actually, I made the PR(https://github.com/Microsoft/vswhere/pull/132) to solve the #220, but I haven't been able to give the persuasive explanation to merge the PR.
For accepting the PR, I must make good explanation. It may be needed to take more time.
So, I decide to make adhoc PR for this extension to use in non-ascii environment anyway.

Randshot:
Since this is out of date, I will close this PR.

vector-of-bool:
@itn3000, we have a possible fix that will be available in beta5

PR #310, opened 2018-02-03
Leave launch terminal open, reuse it if not closed

chatziko:
### This changes the terminal behaviour for `cmake.launchTarget`

- the terminal stays open after the launch target executes
- on the next launch the terminal is reused if it is still open

## The purpose of this change

Previously the terminal closed immediately after the launch target finished. There's no much point in this behaviour: executing the target (instead of debugging) is usually done to see its output, and I don't think that having an extra integrated terminal open adds any annoyance.

So this PR leaves the terminal open, and reuses it in the future to avoid ending up with lots of them.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -190,6 +190,8 @@ export abstract class CommonCMakeToolsBase implements CMakeToolsBackend {
 
   private _nagManager = new NagManager(this._context);
 
+  protected _terminal: Maybe<vscode.Terminal>;
+
   get compilerId() {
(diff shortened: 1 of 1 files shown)


vector-of-bool:
Thanks! It looks good, but could you rebase the work on the `feature/refactor-1.0` branch? If not, I can rework your changes into the branch myself since it isn't a significant amount of code.

chatziko:
Here it is: #311, closing this one.

PR #317, opened 2018-02-18
[Beta4] More substitutions

### LINKED ISSUE #232: cmake.buildToolArgs doesn't support replacing of VS Code vars (eg ${workspaceRoot})
This may be an oversight or by design, but the cmake.buildToolArgs doesn't support using vars like ${workspaceRoot}.

My CMake project has some .NET projects in it, so I need to pass additional arguments and it would be nice to do something like this:

"cmake.buildToolArgs": [
        "-p:Platform=x86",
        "-p:Configuration=Release",
        "-p:SolutionDir=${workspaceRoot}"
    ]

Right now I must hard code the "p:SolutionDir" value.

### LINKED ISSUE #272: Support for other internal variables than ${buildType} in cmake.buildDirectory
CmakeTools has great feature which allows to separately define settings for buildType (debug, release etc) and also for project variants by `cmake.defaultVariants` definition. Then it is possible to choose all their combinations to build.
For complex projects where such feature is useful, it could be convenient to make `cmake.buildDirectory` value also dependent on chosen `buildType` and projectVariant.
Let's allow to reference all variants defined in cmake.defaultVariants object as variables, similary to buildType.
Example:
```
"cmake.defaultVariants": {
   "buildType": {
      "default$": "debug",
      "description$": "the build type to use",
      "debug": {
         "oneWordSummary$": "DebugClang",
         "description$": "Emit debug information without performing optimizations",
         "buildType": "Debug",
         "settings": {
         "CMAKE_CXX_COMPILER":"clang++",
         "CMAKE_C_COMPILER": "clang"
         }
      },
      "release": {
         "oneWordSummary$": "Release",
         "description$": "Enable optimizations, omit debug info",
         "buildType": "Release"
      },
      "reldeb": {
         "oneWordSummary$": "RelWithDebInfo",
         "description$": "Perform optimizations AND include debugging information",
         "buildType": "RelWithDebInfo"
      }
   },
   "projectVariant": {
      "default$": "FreeVersion",
      "description$": "Free version of application",
      "FreeVersion": {
         "oneWordSummary$": "Free",
         "description$": "Settings specific to free version of app",
         "settings": {
            "ENABLE_PAID_FEATURES": "OFF"
          }
      },
      "PaidVersion": {
         "oneWordSummary$": "Paid",
         "description$": "Settings specific to paid version of app",
         "settings": {
            "ENABLE_PAID_FEATURES": "ON"
          }
      }
   }
   },
   "cmake.buildDirectory": "${workspaceRoot}/build/${buildType}/${projectVariant}"
```
For example, after project configuration buildDirectory will point to <br/> `<workspace-root>/build/Debug/PaidVersion`

### LINKED ISSUE #286: Kit name and variant short name accessable by variable substitution
### Brief Issue Summary

I use for development different compilers with the same root CmakeList.txt File. It would be helpful to have a separate build directory for each compiler + variant.  This speed up the switch between the variants or compiler build environments, because no complete CMake regeneration is needed, only a configure.

### Expected:

The build directory path is changable by the settings. I would like to Set into the build directory path over variable substitution the following information:

- the kit name
- the short of the variants

### Platform and Versions

- **Operating System**: Windows 10
- **CMake Version**: 3.9
- **VSCode Version**: 1.19
- **CMake Tools Extension Version**: Plan to us it
- **Compiler/Toolchain**:  GCC 6.3 Windows and GCC for ARM

### LINKED ISSUE #292: 0.11.0-beta4: Variable substitution in "cmake.configureSettings" dysfunct
- **CMake Tools Extension Version**: 0.11.0-beta4

### LINKED ISSUE #305: 0.11.0-beta4: cmake.configureArgs ignored?
### Brief Issue Summary

cmake.configureArgs setting in the workspace .vscode/settings.json file appears to be neglected.

For example, my settings.json file in a project looks like:

{
   // Additional arguments to pass to CMake when configuring
   "cmake.configureArgs": [
      "-DUSE_FORTRAN=OFF"
   ]
}

And I placed on the first line of the project root's CMakeLists.txt:

message("USE_FORTRAN=" ${USE_FORTRAN})

When I build the project, it prints

"USE_FORTRAN="

When I configure the same project in terminal with the argument 

"cmake .. -GNinja -DUSE_FORTRAN=OFF"

I get the expected output:

"USE_FORTRAN=OFF"

### Platform and Versions

- **Operating System**: Windows 7
- **CMake Version**: 3.10.2
- **VSCode Version**: 1.19.2
- **CMake Tools Extension Version**: 0.11.0-beta4
- **Compiler/Toolchain**: Ninja + Visual C++ 2017

### LINKED ISSUE #329: ${workspaceRoot} is not substituted if is used in kit "toolchainFile"
### Brief Issue Summary

If ${workspaceRoot} is used in cmake-kit.json:
    {
        "name": "Emscripten",
        "toolchainFile": "${workspaceRoot}/emscripten/toolchain.cmake"
    }
It will not be substituted.

- **Operating System**: Windows
- **CMake Version**: 3.10.2
- **VSCode Version**: 1.20.1
- **CMake Tools Extension Version**:  0.11.0-beta4
- **Compiler/Toolchain**: [[Visual C++ 2018, GCC 5.3, GHS PPC, GHS V850, ARM 6.6.1, Clang 5.0.1, GCC 7.2.1 (arm)]]

Randshot:
##  Related Issues
Closes #292.
Closes #286.
Closes #272.
Closes #305.
Closes #329.
Closes #232.

## Changes
- Enable the use of `buildArgs` and `buildToolArgs`
- Enable the use of `installPrefix`
    - if the user specified `CMAKE_INSTALL_PREFIX` manually via either `configureArgs` or `configureSettings`, the `installPrefix` option will be ignored
- Enable the use of `environment`, `configureEnvironment` and `buildEnvironment`
    - `configureEnvironment` should only get applied when *configuring*
    - `buildEnvironment`, which should only get applied when *building*
    - `environment` get applied when *configuring* and when *building*
- Provide variable substitutions *for* more variables and also *in* more variables:
    - `configureSettings`, `configureArgs`, `buildArgs`, `buildToolArgs`, variant identifiers
    - special case for variant identifier `buildType`, which is accessible via `buildLabel`
    - `environment`, `configureEnvironment`, `buildEnvironment`
    - `installPrefix`
    - maybe some more that I forgot to mention...
- Add support for `${env.VARNAME}` syntax
- Update the documentation for configuration options a bit
- Add tests for variable substitution behavior
- Add tests for defining environment variables via:
    - `environment`, `configureEnvironment` and `buildEnvironment`
    - also making sure that they don't get applied to the wrong scope

## Other Notes/Information
If I have missed anything, please comment below.


vector-of-bool:
I recently merged some changes for automated testing and CI. Would you mind adding some automated tests that check the variable expansion?

src/driver.ts
```diff
@@ -198,6 +219,14 @@ export abstract class CMakeDriver implements vscode.Disposable {
       subs.set(full, repl);
     }
 
+    const env2_re = /\$\{env.(.+?)\}/g;
```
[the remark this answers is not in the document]
Randshot: It does support both. Look at the one above.

src/driver.ts
```diff
@@ -323,9 +357,15 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
    */
-  get binaryDir(): string { return this._binaryDir }
+  get binaryDir(): string { return this._binaryDir; }
   private _binaryDir = '';
 
+  /**
+   * Directory where the targets will be installed.
+   */
+  get installDir(): string|null { return this._installDir; }
+  private _installDir: string|null = '';
```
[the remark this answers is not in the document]
Randshot: It is null when the user doesn't set `cmake.installPrefix`.

src/state.ts
```diff
@@ -66,6 +66,26 @@ export class StateManager {
   }
   set projectName(s: string|null) { this.extensionContext.workspaceState.update('projectName', s); }
 
+  /**
+   * The replacements for the current workspace
+   */
+  get replacements(): {[key: string] : string | undefined} {
```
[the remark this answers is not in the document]
Randshot: I just needed a place to store the replacement object where different modules can access it.
I originally planned to add the variant replacements from the variant class, but then decided to do it in the driver class.

src/state.ts
```diff
@@ -66,6 +66,26 @@ export class StateManager {
...
+    const replacements = this.extensionContext.workspaceState.get<string>('replacements');
+    if (replacements) {
+      return JSON.parse(replacements);
+    } else {
+      return {};
+    }
+  }
+  set replacements(repl: {[key: string] : string | undefined}) {
+    if (repl) {
```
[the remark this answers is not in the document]
Randshot: I just copied the type from your original implementation. But I can change it if you want me to.

Randshot:
I will add some tests in the next commit.

src/driver.ts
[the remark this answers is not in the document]
vector-of-bool: Oops. Sorry.

src/driver.ts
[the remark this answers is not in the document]
vector-of-bool: I don't see it ever assigned to `null`, just the empty string.

vector-of-bool:
None of the tests currently do any mocking. Maybe someday, but for now the tests just run against the "real thing," which is sufficient and effective for now. All the tests are contained in a single file in the tests directory, and just copying from another one should be sufficient. Don't sweat the details.

Also, ignore the AppVeyor failure. Somethings awry with the VS 2017 Preview test.

Randshot:
So, I have taken a closer look at the already existing tests, but I honestly don't know how I am supposed to test the variable expansion, since it all happens within the `CMakeDriver`.
I would have to create a new instance of the driver and then let it do the configure stuff, after which I can read the `CMakeCache.txt` file to check to results.
Am I missing something here ❓

Randshot:
Hm, I somehow can't install this (PR) version via .vsix.
I packaged it with `vsce package` and used the same installation command as for your Beta4 release.
Anyway, how about we just merge this PR and add the unit test later, since I don't know how I can test the variable expansion.

KoeMai:
You are right. It is hard to write a unit test for the complete modification, but for some sub parts it is possible, like environment replacement or the expandString function.
The test for this feature on extension test level is really complex. 
I work at the moment on unit test for the cmakeserverclient. I hop this allows us to fake the client later and make it easier to write tests for the cmakedriver.

src/driver.ts
```diff
@@ -198,6 +224,17 @@ export abstract class CMakeDriver implements vscode.Disposable {
       subs.set(full, repl);
     }
 
+    const env2_re = /\$\{env\.(.+?)\}/g;
+    while ((mat = env2_re.exec(instr))) {
+      const full = mat[0];
+      const varname = mat[1];
+      const repl = process.env[varname.toLocaleLowerCase()] || '';
```
vector-of-bool: Is `varname.toLocaleLowerCase()` correct? On Windows env vars are case insensitive, but Unix respects case.

src/state.ts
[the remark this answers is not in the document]
vector-of-bool: If no class other than the driver needs access to the replacements (I can't find any other use), can the replacement stuff be taken off of the `StateManager` class?

src/driver.ts
Randshot: I can add platform detection, but I really just copied the already existing one and modified the regex.

vector-of-bool: Then _I_ probably did a wrong. May be prudent to move reading of environment variables to `util.ts` so that we can switch on case-sensitivity.

Randshot: I will add a commit for that after I have gotten some sleep.
Have been working on hypervisor stuff for a couple hours and my mind isn't working anymore.

Randshot:

Randshot:
@vector-of-bool @KoeMai  
I have started writing tests for variable substitution behavior.
Since there will be many tests, I have decided to move them into a separate file (`variable-substitution.test.ts`).
I have pushed my changes so far, maybe you could take a look at it, so that I don't work in a wrong direction.

At the moment, I check each substitution in an own test case, which has the side effect, that a `configure` is run for each case. It would be way more efficient to only configure it one time and then read out the CMake cache in the test cases without re-configuring it each time. I don't know how to do a one time setup step though, so if you know how to, please say so.

I also plan on merging all my other PRs into this branch/PR, so that I can write tests for them.
I think that will be better because my PRs mostly rely on this PR to really work well.

test/extension-tests/successful-build/test/variable-substitution.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  teardown(async function(this: Mocha.IBeforeAndAfterContext) {
+    this.timeout(30000);
+    await cmt.asyncDispose();
+    testEnv.teardown();
+  });
+
+  test('Check substitution for "workspaceRoot"', async () => {
+    // Set fake settings
+    testEnv.setting.changeSetting('configureSettings', {workspaceRoot: '${workspaceRoot}'});
```
KoeMai: This settings is missing in the new documentation. Please add it there.

https://github.com/vector-of-bool/vscode-cmake-tools/blob/feature/refactor-1.0/docs/configuring.rst

test/extension-tests/successful-build/test/variable-substitution.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    expect(cacheEntry.type).to.eq(api.CacheEntryType.String, '[userHome] unexpected cache entry type');
+    expect(cacheEntry.key).to.eq('userHome', '[userHome] unexpected cache entry key name');
+    const user_dir = process.platform === 'win32' ? process.env['HOMEPATH']! : process.env['HOME']!;
+    expect(cacheEntry.as<string>()).to.eq(user_dir, '[userHome] substitution incorrect');
+    expect(typeof cacheEntry.value).to.eq('string', '[userHome] unexpected cache entry value type');
+  }).timeout(60000);
+
+  test('Check substitution for variant names', async () => {
+    // Set fake settings
+    testEnv.setting.changeSetting('configureSettings', {buildLabel: '${buildLabel}', otherVariant: '${otherVariant}'});
```
KoeMai: Variable Substitution has no describtion for '${buildLabel}' and '${otherVariant}'.

Randshot: @KoeMai `${buildLabel}` is for the `buildType` variant option. It is available for substitution via `buildLabel`. `${otherVariant}` is the name of a variant option I defined in the test project folder.

test/extension-tests/successful-build/test/variable-substitution.test.ts
Randshot: I will add it after I have fixed my current issues with `xxxEnvironment`.

test/extension-tests/successful-build/test/variable-substitution.test.ts
KoeMai: I think we need in the documentation a description for the replacement. Do I understand this right?
The buildLabel and otherVariant are the identifiers for the variants. The ${idenfier} is replaced on configuration by the selected choice.

Randshot: Some users requested for the variant names to be available for variable substitution, which is why I added it.

I have defined the following variant config in the test project folder:
```json
{
    "buildType": {
        "default": "debug-label",
        "choices": {
            "debug-label": {
                "short": "debug-label short",
                "buildType": "Debug"
            },
            "not-debug": {
                "short": "not-debug short",
                "buildType": "Release"
            }
        }
    },
    "otherVariant": {
        "default": "option1",
        "choices": {
            "option1": {
                "short": "option1 short"
            },
            "option2": {
                "short": "option2 short"
            }
        }
    }
}
```

And it is pretty much how you explained it.
I had to replace the `buildType` variant with another name to avoid name collisions.

test/extension-tests/successful-build/test/variable-substitution.test.ts
Randshot: Just on the note of documentation.. atm A LOT is missing from it.

test/extension-tests/successful-build/test/variable-substitution.test.ts
Randshot: I have added some documentation about the variant identifier substitution.

test/extension-tests/successful-build/test/variable-substitution.test.ts
Randshot: Done.

Randshot:
Ah, it's so nice when all tests pass ;)
Anyway, @KoeMai I have added some documentation for various configure options and also about the variant identifier substitution.

vector-of-bool:
I recommend: https://github.com/SethMichaelLarson/trytravis

Randshot:
Certainly looks interesting :D

Randshot:
@vector-of-bool Is there a reason why 'squash and merge' is not enabled for this repo?
I think that in some cases, it is quite sensible to do it.

vector-of-bool:
I've disabled squash and merge as well as FF merging because I strongly prefer to keep all of Git history (at least for PRs, feel free to rebase/squash on your local copy). Even if there are silly "typo" commits, I prefer to see them in the history.

Give trytravis a go. It works beautifully.

src/driver.ts
```diff
@@ -153,24 +153,76 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
-      userHome: user_dir,
-    };
+  async expandString(instr: string, env?: proc.EnvironmentVariables): Promise<string> {
+    // Update the replacements and get the updated values
+    const replacements = this._replacements;
+
+    // Setup temporary environment
+    const old_env = process.env;
+    const custom_env = util.mergeEnvironment(old_env as proc.EnvironmentVariables, config.environment, env ? env : {});
+    process.env = custom_env;
```
vector-of-bool: I may be missing some context, but why do we modify the extension's own environment variables?

src/driver.ts
vector-of-bool: Can you move out the case-normalization to `util.ts`?

There's an itching in the back of my mind that lower-casing environment variables on Linux may be to blame for one or more bug reports I've seen in the past.

src/driver.ts
```diff
@@ -534,10 +615,25 @@ export abstract class CMakeDriver implements vscode.Disposable {
       flags.push(...util.objectPairs(this._kit.cmakeSettings).map(([key, val]) => _makeFlag(key, util.cmakeify(val))));
     }
 
+    // Expand all flags
     const final_flags = flags.concat(settings_flags);
-    log.trace('CMake flags are', JSON.stringify(final_flags));
+    const expanded_flags_promises = final_flags.map(
+        async (value: string) => this.expandString(value, await this.getExpandedConfigureEnvironment()));
```
vector-of-bool: If you lift `this.getExpandedConfigureEnvironment` into a variable before you expand the flags, we'll only call the function once. `getExpandedConfigureEnvironment()` could be expensive, depending on what they have in their configure environment.

src/driver.ts
```diff
@@ -534,10 +615,25 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+    log.trace('CMake flags are', JSON.stringify(expanded_flags));
+
+    // Setup temporary environment for the configure step
+    const old_env = process.env;
+    const configure_env = util.mergeEnvironment(old_env as proc.EnvironmentVariables,
+                                                await this.getExpandedEnvironment(),
+                                                await this.getExpandedConfigureEnvironment());
+    process.env = configure_env;
+
+    const retc = await this.doConfigure(expanded_flags, consumer);
```
vector-of-bool: The setting the environment then resetting is spooky to me. In this case in particular, because if `doConfigure` throws we won't reset the environment.

src/state.ts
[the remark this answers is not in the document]
vector-of-bool: I don't see `StateManager.replacements` used anywhere other than the root extension. Can it be removed safely?

test/extension-tests/successful-build/test/variable-substitution.test.ts
vector-of-bool: Where's the doc change? I don't see it.

src/driver.ts
Randshot: I added an optional parameter to the `expandString()` method, which gets applied there, and reset later. But you are right, I could probably just do the regex stuff directly with the passed environment, without modifying the actual process one.

src/driver.ts
Randshot: The only reason I am setting it here is because of the legacy driver. The environment for the cmake-server gets set when starting it. I could add an (optional) environment argument to the `doConfigure()` though.

src/state.ts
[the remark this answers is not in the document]
Randshot: Ok, I will just remove it, since I didn't end up doing it like I originally planned to.

test/extension-tests/successful-build/test/variable-substitution.test.ts
Randshot: @vector-of-bool I added them in the docs files `docs/configuring.rst`

src/driver.ts
Randshot: Done.

Randshot:
@vector-of-bool 
About that `trytravis` thing, I don't see a use for me at the moment, since all the configure/build tests are skipped if `process.env.HasVs` is not `true`, which means that they are also skipped on travis.
It looks useful though, so I will give it a try later.

PR #325, opened 2018-03-01
[Beta4] Add setting field `cmake.autoRestartBuild`

### LINKED ISSUE #318: Suggestion: Restart build by invoking cmake.build twice
Running build when another build is running results in an error pop up with this message:

> A CMake task is already running. Stop it before trying to build.

This is reasonable, but a bit annoying as this pupup must be closed by clicking the "Close" button. The only way that I know of that stops the build is the little "stop" icon at the bottom.

I suggest that if you invoke the cmake.build (F7 by default) command twice in a short interval such as one second, the current build is cancelled, the popup is closed, and a new build starts.

Randshot:
### Changes
I have added a setting field named `cmake.autoRestartBuild`.
When this field is turned on, running CMake tasks will get stopped if the user starts a new one.

Closes #318.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -72,6 +72,8 @@ class ConfigurationReader {
 
   get clearOutputBeforeBuild(): boolean { return !!readPrefixedConfig<boolean>('clearOutputBeforeBuild'); }
 
+  get autoRestartBuild(): boolean { return !!readPrefixedConfig<boolean>('autoRestartBuild'); }
+
   get configureSettings(): any { return readPrefixedConfig<Object>('configureSettings'); }
 
   get initialBuildType(): string|null { return readPrefixedConfig<string>('initialBuildType'); }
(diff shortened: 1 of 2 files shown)


vector-of-bool:
Have you done manual validation here? I remember there being something wrong with the ext not detecting the already-running proc and simply starting another build simultaneously.

Randshot:
I can do a manual check, but I am honestly waiting for @KoeMai s PR regarding CMakeDriver tests. With that it will be possible to test such cases pretty easily.

Randshot:
@vector-of-bool I have done some manual testing and it turned out that the `_isBusy` flag didn't get set anywhere, so I set/reset it when/after running CMake.
As far as I could tell, it worked pretty well.

Randshot:

vector-of-bool:
@Randshot, for future reference: Adding a user-visible setting requires also adding it to the manifest in `package.json`, otherwise VSCode won't give the user any help text while editting their `settings.json`. I've added the setting to the manifest in the main refactor branch

Randshot:
Ok, I will keep that in mind.

PR #326, opened 2018-03-01
[Beta4] Add support for buildArgs and buildToolArgs

### LINKED ISSUE #232: cmake.buildToolArgs doesn't support replacing of VS Code vars (eg ${workspaceRoot})

Randshot:
### Changes
I have added support for `cmake.buildArgs` and `cmake.buildToolArgs`.
Variable substitution should also work.

Closes #232.

### DIFF
--- a/src/driver.ts
+++ b/src/driver.ts
@@ -627,6 +627,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
     if (!ok) {
       return null;
     }
+
     const gen = this.generatorName;
     const generator_args = (() => {
       if (!gen)
@@ -641,10 +642,15 @@ export abstract class CMakeDriver implements vscode.Disposable {
       else
         return [];
     })();
+
     const args = [ '--build', this.binaryDir, '--config', this.currentBuildType, '--target', target, '--' ]
-                     .concat(config.buildArgs, generator_args, config.buildToolArgs);
+                     .concat(generator_args, config.buildToolArgs, config.buildArgs);
+    const expanded_args_promises = args.map(async (value: string) => await this.expandString(value));
+    const expanded_args = await Promise.all(expanded_args_promises);
+    log.trace('CMake build args are', JSON.stringify(expanded_args));
+
     const cmake = await paths.cmakePath;
-    const child = this.executeCommand(cmake, args, consumer);
+    const child = this.executeCommand(cmake, expanded_args, consumer);
     this._currentProcess = child;
     await child.result;
     this._currentProcess = null;
--- a/src/driver.ts
+++ b/src/driver.ts
@@ -440,7 +440,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
     const candidates = this.getPreferredGenerators();
     for (const gen of candidates) {
       const gen_name = gen.name;
-      const generator_present = await (async(): Promise<boolean> => {
+      const generator_present = await(async(): Promise<boolean> => {
         if (gen_name == 'Ninja') {
           return await this.testHaveCommand('ninja-build') || await this.testHaveCommand('ninja');
         }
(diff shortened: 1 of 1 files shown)


src/driver.ts
```diff
@@ -439,7 +439,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
     const candidates = this.getPreferredGenerators();
     for (const gen of candidates) {
       const gen_name = gen.name;
-      const generator_present = await (async(): Promise<boolean> => {
+      const generator_present = await(async(): Promise<boolean> => {
```
vector-of-bool: Not a problem, but a notice tangentially related to this whitespace change.: I added a `.clang-format` for the codebase. The latest Clang releases have _excellent_ support for TypeScript and I've been running code through it as I go. In the future, it would be nice to ensure we always run formatting before commits.

src/driver.ts
```diff
@@ -640,9 +641,14 @@ export abstract class CMakeDriver implements vscode.Disposable {
       else
         return [];
     })();
-    const args = [ '--build', this.binaryDir, '--config', this.currentBuildType, '--target', target, '--' ].concat(
-        generator_args);
-    const child = this.executeCommand(config.cmakePath, args, consumer);
+
+    const args = [ '--build', this.binaryDir, '--config', this.currentBuildType, '--target', target, '--' ]
+                     .concat(generator_args, config.buildToolArgs, config.buildArgs);
```
vector-of-bool: IIRC, `config.buildArgs` goes before the `--` option? I might be wrong, I never used the option myself.

vector-of-bool:
I've already hit "Approve," but is it possible to add any tests here? I'm honestly not sure. We need a better smoke testing solution than what we already have.

src/driver.ts
Randshot: I think I have clang-format running? The editor formatted it this way when I saved the file.

src/driver.ts
Randshot: I didn't know what the `--` was for, so I let it just be there, but I can swap them around. At least the variable expansion worked like it was supposed to. Like in my other PR, I am waiting for driver testing before merging this one.

Randshot:
I have answered in the comment above.

Randshot:

vector-of-bool:
The `--` in the build command tells CMake to forward the arguments to the underlying build command, ie. `ninja`, `make` or `msbuild`.

What version of clang-format do you have? I recently upgraded to a newer one because older ones barf on some TypeScript constructs.

Randshot:
@vector-of-bool This is the output:
```
PS C:\Program Files\LLVM\bin> .\clang-format --version
clang-format version 5.0.1 (tags/RELEASE_501/final)
```

vector-of-bool:
I _was_ using 3.9 and it was pretty unhappy. I switch to 6.0.0 SVN and it started working again. I'd imagine 5.0 should work just fine.

I have a pending change to do a lot of formatting and linting work.

vector-of-bool:
@Randshot, if you fix the conflict, you should be ready to merge this, unless you want to wait on a better testing situation.

Randshot:
@vector-of-bool I am waiting for the testing PR, so that I can write tests for the substitution behavior.

Randshot:
Moved to #317.
Closed.

PR #331, opened 2018-03-03
Fix ${workspaceRoot} in kit "toolchainFile" Issue #329

bbosnjak:


vector-of-bool:
It looks like this will already be caught with #317. That PR has been pending for a while, but it should be in for the next release/beta.

Randshot:
Closed, since this issue has already been dealt with in PR #317.

PR #332, opened 2018-03-03
More TSLint and clang-format

vector-of-bool:
### This changes the linting and formatting rules

- Run tslint on Travis
- Adds several stylistic _and_ behavioral lint rules.
- `.clang-format` is now a bit more sane.

## Other Notes/Information

Unfortunately, this PR touches almost every single file. May create a bit of a merge nightmare, but I hope that once this is done once it'll never be required to touch every single file again.

I chose a subset of the TSLint rules that I found compelling. Please feel free to suggest adding/removing any rules.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -61,9 +61,11 @@ SpacesInAngles:  false
 SpaceInEmptyParentheses: false
 SpacesInCStyleCastParentheses: false
 SpaceAfterCStyleCast: false
-SpacesInContainerLiterals: true
+SpacesInContainerLiterals: false
 SpaceBeforeAssignmentOperators: true
 ContinuationIndentWidth: 4
 SpaceBeforeParens: ControlStatements
+JavaScriptWrapImports: true
+JavaScriptQuotes: Single
 DisableFormat:   false
 ...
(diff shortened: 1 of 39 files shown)


.clang-format
```diff
@@ -61,9 +61,11 @@ SpacesInAngles:  false
 SpaceInEmptyParentheses: false
 SpacesInCStyleCastParentheses: false
 SpaceAfterCStyleCast: false
-SpacesInContainerLiterals: true
+SpacesInContainerLiterals: false
```
vector-of-bool: I honestly thought clang-format was broken because it kept adding these annoying spaces, but it turns out I had this wonky setting turned on the entire time. :angry:

Randshot: lol.. :D

Randshot:
This is really weird. My clang-format keeps removing the space between `await` and parentheses.

vector-of-bool:
It may be that clang is parsing `await` as a function instead of a unary operator. Maybe they fixed it by 6.0?

Randshot:
Ok, I have updated LLVM to 7.0 and now it seems to work.

Issue #337, opened 2018-03-09
No prefered generator are in the kit set for VS2015, 2013


KoeMai:
### Brief Issue Summary

The kit file contains typical for the visual studio projects a prefered generator. 
Vs2017 works, but vs2015 and VS2013 does not have one. 

### Expected:

All visual studio's should have a prefered generator. 
Non failing test on travis. 

### Platform and Versions

- **Operating System**: Windows
- **CMake Version**: 3.8
- **VSCode Version**: [[1.16.0, 1.17.2, etc.]]
- **CMake Tools Extension Version**: [[beta5 dilev
- **Compiler/Toolchain**: [[Visual C++ 2017, 2015, 2013

### Other Notes/Information

[[Write whatever you feel is relevant]]

KoeMai:
Will be fixed in 0.11.0-beta5.

PR #338, opened 2018-03-09
Bugfix/issue 337 contains add_build_test environment

KoeMai:
### This changes fix missing generator for vs v11 and introducts an extension test environment

- #337 Appveyor fails because the default generator for Visual studio V11.0 is missing
- Introduces a test environment for extension tests.
  - Add helper classes to access 
    - CMake project environment
    - Vscode parts settings, extension context (states) and extension typical parts like selection of kits.

## The purpose of this change

The idea of this pull request was to introduct a test environment for extension tests in the user view.
It is the first working implementation which allows to use the extension and control extension provided files. It use an example project which allows to generate a program. There is a test program which outputs a JSON file. This JSON file can be used to get information from the build as feedback into the test.

## Other Notes/Information

If it goes into the refactoring branch, then it is possible to test
- get the different compiler, generators and cmake versions tested.
- prefered generators
- different CMakeTools settings from settings.json


test/helpers/vscodefake/workspaceconfiguration.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export class CMakeToolsSettingFile {
+
+  readonly originalValues: vscode.WorkspaceConfiguration;
+  readonly filepath: string;
+  readonly fakeValues: CMakeToolsWorkspaceConfiguration;
+  readonly originalFunction: any;
+
+  constructor(sandbox: sinon.SinonSandbox) {
+    this.originalValues = vscode.workspace.getConfiguration('cmake');
```
vector-of-bool: Lift these property initializations to be inline with the declaration since they do not depend on the value of a constructor parameter.

test/helpers/vscodefake/extensioncontext.ts
```diff
@@ -0,0 +1,18 @@
...
+  workspaceState: vscode.Memento;
+  globalState: vscode.Memento;
+  extensionPath: string;
+
+  asAbsolutePath(relativePath: string): string { return relativePath; }
+  storagePath: string|undefined;
+
+  constructor() {
+    this.globalState = new TestMemento();
+    this.workspaceState = new TestMemento();
```
vector-of-bool: Lift these initializers

test/helpers/vscodefake/extensioncontext.ts
```diff
@@ -0,0 +1,18 @@
+import * as vscode from 'vscode';
+import {TestMemento} from './memento';
+
+export class FakeContextDefinition implements vscode.ExtensionContext {
+
+  subscriptions: {dispose(): any;}[];
```
vector-of-bool: This field is not initialized

test/helpers/cmake/builddirectoryhelper.ts
```diff
@@ -0,0 +1,20 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as rimraf from 'rimraf';
+
+export class BuildDirectoryHelper {
+
+  private readonly location: string;
+
+  public constructor(location: string) { this.location = location; }
```
vector-of-bool: A shorthand:

```typescript
public constructor(private readonly location: string) {}
```

will declare and set the property `location` to the first argument passed to the constructor. No need to declare the field or do an explicit assignment. It's used in a few places in the extension.

vector-of-bool:
I'll be glad to finally have reliable end-to-end tests, even if the individual components still need more testing. I'm hopeful that @ytimenkov's refactors will be helpful in testing efforts.

I don't like mocking: not because _mocking_ is bad on its own, but that it indicates a more fundamental problem in the way the code is structured. Without any big changes, there's currently no way to test the extension code without some mocking. The fact that certain operations _rely_ on user input is a hairy/fragile design that I need to get fixed. I think @ytimenkov's changes will help get there.

I was too hasty in the 1.0-refactor to fix the UI-dependence, so it stayed.

src/kit.ts
```diff
@@ -386,10 +386,12 @@ const VsArchitectures: {[key: string]: string} = {
  * Preferred CMake VS generators by VS version
  */
 const VsGenerators: {[key: string]: string} = {
-  14: 'Visual Studio 14 2015',
-  15: 'Visual Studio 15 2017',
+  11: "Visual Studio 11 2012",
```
Randshot: Use `''` for strings (also applies to this one `"Visual Studio 12 2013",` below)

test/extension_tests/successful_build/test/configure_and_build.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  teardown(async function(this: Mocha.IBeforeAndAfterContext) {
+    this.timeout(30000);
+    await cmt.asyncDispose();
+    testEnv.teardown();
+  });
+
+  test('Configure ', async() => {
+    expect(await cmt.configure()).to.be.eq(0);
+
+    expect(testEnv.projectFolder.BuildDirectory.IsCMakeCachePresent).to.eql(true, 'no expected cache presetruent');
```
Randshot: Little type here: `'no expected cache present'`

src/kit.ts
vector-of-bool: `clang-format` has a fix for this one.

test/extension_tests/successful_build/test/configure_and_build.test.ts
KoeMai: Done

test/helpers/cmake/builddirectoryhelper.ts
KoeMai: Done

test/helpers/vscodefake/extensioncontext.ts
KoeMai: Add throwing exception for all unused parts of the interface.

test/helpers/vscodefake/extensioncontext.ts
KoeMai: Done

test/helpers/vscodefake/workspaceconfiguration.ts
KoeMai: Done

vector-of-bool:
Spurious failure on Travis. I kicked it and the build succeeded.

Looks like a good start. I do want to hear what @ytimenkov thinks, though.

ytimenkov:
I only briefly looked at the changes.

As I understand test just simulates user opening a project and building it. Which is what smoke tests are for.
And I agree that it is good start and we can improve on it.

About mocks... I ponder if test/suite is (or could be) driven from .bat file, why do we need to fake Code's classes? In other words, create a set of scenarios, each starting a new instance of Code.

I share the concern about using mocks for this purpose and better way would be to encapsulate interaction with Code into set of interfaces and/or classes ("glue"), and mock them (I.e. mocking extension-specific `CMakeToolsPersistency` vs general `vscode.Memento`).
(IIRC when I tried to run tests without using Code at all, I failed because of some convenient utility classes like `Event`).

vector-of-bool:
I feel good for this. Once these are in place, I'll feel safer taking some of the other big PRs and we can make good progress to the beta5.

@KoeMai, go ahead and merge if/when you are ready.

scripts/run_extension_tests.bat
```diff
@@ -3,12 +3,30 @@ setlocal
```
vector-of-bool: I am unable to make the Linux tests to fail. A malformed CMakeLists.txt still allows the tests to pass.

Randshot: @vector-of-bool I think that is because the build/configure tests don't get executed on Linux at the moment: https://github.com/vector-of-bool/vscode-cmake-tools/pull/338/files#diff-a73238827bcc789254199f160d6f5d9aR17

vector-of-bool: Oof. I missed that. Still, I'd like to see them running. @KoeMai, is there a reason they are disabled on Linux?

KoeMai: I disabled linux and mac os. I first wanted feedback about the test way.
If you want to see the benefit of this way then look into 'add_prefered_generator_tests'. 

My open problem is the way to switch machine dependencies like vs or clang.. Or the make system. If you look on appveyor then you see it use only visual studio 11. 
My first idea for the different compiler support is on add_prefered_generator_tests.

Additional this PR contains the basic infrastructur. It is ready to use.

KoeMai:
About the test infrastructure I did not found a way to get control over the extension and handle the expension response. I think we can only mock on this level. All other extensions I looked on did it in the simular way or have only unit tests (no Integration tests) . But I think this tests are necessary because they are near to the user and test the integration of the software units.

scripts/run_extension_tests.bat
vector-of-bool: Let's get this merged. I branched from this branch and I have some changes to do some Linux and macOS tests.

PR #339, opened 2018-03-10
Initial version of contributing guidelines.

ytimenkov:
I think it's time to have them, so sharing my experience.

From top of my head and recent experience in setting things up and running (left out doc section, with switch to RST much more prerequisites are required).

I think could also be added references to generic style guides / best practices for TypeScript.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -1,0 +1,45 @@
+Building extension
+==================
+As with most VS Code extensions you need you need `Node.JS <https://nodejs.org/en/>`_ installed.
+
+The process if failry straightforward:
+
+1. Install dependencies
+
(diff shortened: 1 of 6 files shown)


docs/development.rst
```diff
@@ -0,0 +1,45 @@
+Building extension
+==================
+As with most VS Code extensions you need you need `Node.JS <https://nodejs.org/en/>`_ installed.
+
+The process if failry straightforward:
```
vector-of-bool: `failry` -> `fairly`.

You'll need to add this to the toctree in `index.rst` as `development`. Also, it will be useful to move the `Developer Reference` section from `index.rst` to the top of this file.

PR #344, opened 2018-03-18
Set CC and CXX to cl.exe when using a VS kit (Fix #322)

### LINKED ISSUE #322: 0.11 beta4 - Misbehaving Generator Kit Selection
### Brief Issue Summary

Selected kit ignored by CMake/Ninja

### Expected:

I just installed Strawberry Perl on my system, and apparently it has its own GCC C compiler (found at C:/Strawberry/c/bin/). It is now detected as one of available kits along with MSVC 2017 amd64/x86.

I expect to be able to select any of these 3 kits to compile and CMake to configure accordingly, but...

### Apparent Behavior:

Selecting MSVC 2017 amd64 on vscode UI, leave the UI to show MSVC 2017 amd64 as the active kit, but when CMake configures, it uses the GCC compiler in the Strawberry subfolder.

Furthermore, after it configures the project with the Strawberry compiler, if I tried ">CMake: Select A Kit"  and pick MSVC 2017 amd64, it hangs with log:

```
[driver] Switching to kit: Visual Studio Build Tools 2017 - amd64
[proc] Executing command: C:\Users\tikuma\AppData\Local\Temp\vs-cmt-0.8133370168873524.bat
```
No further message after this unlike before.

### Platform and Versions

OS: Windows 7
CMake Version: 3.10.2
VSCode Version: latest
CMake Tools Extension Version: 0.11.beta4

### Other Notes/Information

[[Write whatever you feel is relevant]]

vector-of-bool:
The code comment explains the why/how.

I don't feel like there's a reliable way to test this via AppVeyor, since the behavior is baked really far down into CMake itself.


vector-of-bool:
@Randshot, VSCode has events for when the user updates their configuration. We can add a hook to that and restart CMake server if relevant settings are updated.

Randshot:
Ok, then we would have to restart the CMake server when the user changes at least one of the following:
- cmake.environment
- cmake.configureEnvironment

PR #345, opened 2018-03-18
Support project-specific Kits.

### LINKED ISSUE #293: Change "cmake-kits.json" location
### Brief Issue Summary

The kit approach is nice if you change the compiler on the same development environment.
I have in my embedded projects different compiler with cross compilation. In most projects it is not possible to switch between all my installed compilers, because not all compilers support all language features. I would use the kit file to describe the possible compiler and build environments (PC or Target) of my project. This requires the possibility to make the kits project dependent.

### Expected:

Way 1: Different order of kits search 
1. "cmake-kits.json" in project/.vscode
2. "cmake-kits.json" in workspace,
3. "cmake-kits.json" in user profile
4. "cmake-kits.json" in system

Way 2: Define "cmake-kits.json" location by settings.json

### Platform and Versions

- **Operating System**: Windows, Linux
- **CMake Version**: 3.9
- **VSCode Version**:  1.19
- **Compiler/Toolchain**: Visual C++ 2015, GCC 5.3, Clang

vector-of-bool:
Loads both user-local and project specific, and both are available from the UI.

Closes #293

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -603,7 +603,7 @@ export class KitManager implements vscode.Disposable {
   private _kits = [] as Kit[];
 
   /**
-   * The path to user-specific the `cmake-kits.json` file
+   * The path to the user-specific `cmake-kits.json` file
    */
   private readonly _userKitsPath: string;
 
@@ -757,7 +757,7 @@ export class KitManager implements vscode.Disposable {
 
     const new_kits = Object.keys(new_kits_by_name).map(k => new_kits_by_name[k]);
 
-    log.debug('Saving news kits to', this._userKitsPath);
+    log.debug('Saving new kits to', this._userKitsPath);
     await fs.mkdir_p(path.dirname(this._userKitsPath));
     const stripped_kits = new_kits.map((k: any) => {
       k.type = undefined;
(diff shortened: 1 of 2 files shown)


src/kit.ts
```diff
@@ -540,14 +603,19 @@ export class KitManager implements vscode.Disposable {
   private _kits = [] as Kit[];
 
   /**
-   * The path to the `cmake-kits.json` file
+   * The path to user-specific the `cmake-kits.json` file
```
Randshot: Just the sentence structure:
`The path to the user-specific `cmake-kits.json` file´

src/kit.ts
```diff
@@ -688,8 +757,8 @@ export class KitManager implements vscode.Disposable {
 
     const new_kits = Object.keys(new_kits_by_name).map(k => new_kits_by_name[k]);
 
-    log.debug('Saving news kits to', this._kitsPath);
-    await fs.mkdir_p(path.dirname(this._kitsPath));
+    log.debug('Saving news kits to', this._userKitsPath);
```
Randshot: `news` -> `new`

Issue #346, opened 2018-03-18
Environment variables don't get applied after the cmake-server got started


Randshot:
## Brief Issue Summary
Because of the way the (configure) environment variables get applied for the cmake-server, they don't get applied if the user changes them in the `settings.json` __*after*__ the extension started the cmake-server.
The user can fix this by restarting the cmake-server (e.g. by restarting vscode), but I don't think that the user should have to worry about that.

@vector-of-bool has suggested that we use vscode hooks to watch for any relevant changes to the project-specific and user-specific settings file. In that case, we would have to watch for any changes to the following option fields:
- `cmake.environment`
- `cmake.configureEnvironment`

I quite like the approach @vector-of-bool suggested, so I will also suggest that we do it that way.
If you have any other suggestions or I have missed something, then feel free to comment them below.

Tag @KoeMai, @ytimenkov, @vector-of-bool.

vector-of-bool:
Marking done in the project since the PR was merged.

PR #347, opened 2018-03-18
Install event hook for environment and configureEnvironment options

Randshot:
This PR addresses issue #346.

### DIFF
--- a/src/cms-driver.ts
+++ b/src/cms-driver.ts
@@ -5,7 +5,7 @@ import * as api from './api';
 import {CacheEntryProperties, ExecutableTarget, RichTarget} from './api';
 import * as cache from './cache';
 import * as cms from './cms-client';
-import cfg from './config';
+import config from './config';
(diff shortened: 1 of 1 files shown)


src/cms-driver.ts
```diff
@@ -5,6 +5,7 @@ import * as api from './api';
 import {CacheEntryProperties, ExecutableTarget, RichTarget} from './api';
 import * as cache from './cache';
 import * as cms from './cms-client';
+import cfg from './config';
```
vector-of-bool: Can you change `cfg` -> `config`, just to match the rest of the project.

Randshot: Then I will have to change every local varname with the `config` identifier. There are a couple of them in the `cms-driver.ts` file.
@vector-of-bool Should I do that?

vector-of-bool: Go for it.

Randshot: Done.

PR #355, opened 2018-03-26
Pass environment variables to configure from variant definitions

vector-of-bool:


test/extension-tests/successful-build/test/variant-envs.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  test('Check for environment variables being passed to configure', async () => {
+    // Set fake settings
+    // Configure
+    expect(await cmt.configure()).to.be.eq(0, '[variantEnv] configure failed');
+    expect(testEnv.projectFolder.buildDirectory.isCMakeCachePresent).to.eql(true, 'expected cache not present');
+    const cache = await CMakeCache.fromPath(await cmt.cachePath);
+
+    const cacheEntry_ = cache.get('variantEnv');
+    expect(cacheEntry_).to.not.be.eq(null, '[variantEnv] Cache entry was not pressent');
```
Randshot: `pressent` -> `present`

vector-of-bool:
I think vars should be ordered by the "scope" of what they're doing. That's barely even the right word. I'm not sure how to convey it properly. The smaller the scope, the higher priority. How about this ordering (Higher -> lower):

1. Variant variables. Defined per-project, therefore smallest in scope.
2. `settings.json` `configureEnvironment`. Specifically used for running configure.
3. `settings.json` variables. Defined by a user, not specific to a project. (Not considering differences between project-settings and user-settings.)
4. Kit variables. Defined for a specific build environment, not tied to a particular project, user, or even VSCode itself.
5. System environment variables. Not related to CMake Tools at all.

I don't think that users will stomp over env vars unintentionally, and if they were to somehow break the kits by overriding things in `settings.json` or `cmake-variants.yaml`, their environment variables are bad. Full stop. Either the env vars break the kits, or the kits break their env vars. There's no solution if their settings cause conflicts.

Regarding expansions in variants: I'm inclined to say that _no_, expansions should not occur in this context. The first problem is that values in a variant are available for expansion in other contexts, so we can't use a variant to expand itself recursively. Second is that I'd like to keep the `cmake-variants.yaml` file CMake Tools-agnostic.

Randshot:
@vector-of-bool I like that. I was trying to convey the same order, but i got the high/low wrong, it should have been the other way round.

PR #356, opened 2018-03-27
[beta4] Possible vswhere UTF-8 fix & Differentiate between release and pre-release installations

### LINKED ISSUE #220: Failed to parse json generated by vswhere

Randshot:
## Related Issues
Fixes #220.

## Changes
This PR tries to fix the vswhere utf8 bug, where the JSON parser throws because of encoding/code page issues. The method I used is the one mentioned by heaths in the issue above.
I added a separate function that uses `execFile()` instead of `spawn()` in the hope that it fixes the issues other users encountered. Sadly, I couldn't really verify whether this fixes the problem at hand though.

The other thing that this PR does is to differentiate between release and pre-release versions of Visual Studio. For pre-release versions, the `Preview` identifier is appended to the display name.
Since I don't have the **Community** edition of VS installed, but the **Professional** one, pretty much all the `preferred-generators` tests failed. I fixed this by getting the known kits from the extension (I had to add a getter for that) and checking whether the default kit name in question is present. I redid the check for preferred generators the same way, instead of using a regex.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -299,8 +299,9 @@ export async function vsInstallations(): Promise<VSInstallation[]> {
   const installs = [] as VSInstallation[];
   const inst_ids = [] as string[];
   const vswhere_exe = path.join(thisExtensionPath(), 'res/vswhere.exe');
-  const vswhere_args = ['-all', '-format', 'json', '-products', '*', '-legacy', '-prerelease'];
-  const vswhere_res = await proc.execute(vswhere_exe, vswhere_args, null, {silent: true, encoding: 'utf8'}).result;
+  const vswhere_args =
+      ['/c', 'chcp 65001 |', vswhere_exe, '-all', '-format', 'json', '-products', '*', '-legacy', '-prerelease'];
+  const vswhere_res = await proc.execute('cmd.exe', vswhere_args, null, {silent: true, encoding: 'utf8'}).result;
 
   if (vswhere_res.retc !== 0) {
     log.error('Failed to execute vswhere.exe:', vswhere_res.stdout);
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -465,7 +465,7 @@ async function tryCreateNewVCEnvironment(inst: VSInstallation, arch: string, pr?
         platform: VsArchitectures[arch] as string || undefined,
       };
     }
-    log.debug(` Selected Prefered Generator Name: ${generatorName}`);
+    log.debug(` Selected Preferred Generator Name: ${generatorName}`);
   }
 
   return kit;
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -300,11 +300,8 @@ export async function vsInstallations(): Promise<VSInstallation[]> {
   const inst_ids = [] as string[];
   const vswhere_exe = path.join(thisExtensionPath(), 'res/vswhere.exe');
   const vswhere_args = ['-all', '-format', 'json', '-products', '*', '-legacy', '-prerelease'];
-  const vswhere_res = await proc.executeFile(vswhere_exe, vswhere_args).result;
-  if (vswhere_res.retc !== 0) {
-    log.error('Failed to execute vswhere.exe:', vswhere_res.stdout);
-    return [];
-  }
+  const vswhere_res = await proc.executeFile(vswhere_exe, vswhere_args);
+
   const vs_installs = JSON.parse(vswhere_res.stdout) as VSInstallation[];
   for (const inst of vs_installs) {
     if (inst_ids.indexOf(inst.instanceId) < 0) {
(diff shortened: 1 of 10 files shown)


src/proc.ts
```diff
@@ -74,6 +72,14 @@ export interface ExecutionOptions {
   cwd?: string;
 }
 
+export interface ExecFileOptions {
```
vector-of-bool: You can have this interface extend `ExecutionOptions`, just to remove duplication

src/proc.ts
```diff
@@ -163,3 +169,94 @@ export function execute(command: string,
...
+  const final_env = util.mergeEnvironment(process.env as EnvironmentVariables, options.environment || {});
+
+  const exec_opts: proc.ExecOptionsWithStringEncoding
+      = {env: final_env, shell: options.shell, encoding: options.encoding || 'utf8'};
+
+  if (options && options.cwd) {
+    exec_opts.cwd = options.cwd;
+  }
+
+  const child: proc.ChildProcess = proc.execFile(command, args, exec_opts);
```
vector-of-bool: Use the `callback` signature of [the `child_process.execFile()` API](https://nodejs.org/api/child_process.html#child_process_child_process_execfile_file_args_options_callback). We don't need the output streaming abilities and `OutputConsumer` stuff.

src/kit.ts
```diff
@@ -299,11 +300,13 @@ export async function vsInstallations(): Promise<VSInstallation[]> {
   const inst_ids = [] as string[];
   const vswhere_exe = path.join(thisExtensionPath(), 'res/vswhere.exe');
   const vswhere_args = ['-all', '-format', 'json', '-products', '*', '-legacy', '-prerelease'];
-  const vswhere_res = await proc.execute(vswhere_exe, vswhere_args).result;
+  const vswhere_res = await proc.execute(vswhere_exe, vswhere_args, null, {silent: true, encoding: 'utf8'}).result;
```
vector-of-bool: What happened to using `cmd.exe` and `chcp`?

Randshot: Ah, I haven't added that yet, since I am still  puzzled by the VS problem.

src/kit.ts
```diff
@@ -299,8 +299,9 @@ export async function vsInstallations(): Promise<VSInstallation[]> {
   const installs = [] as VSInstallation[];
   const inst_ids = [] as string[];
   const vswhere_exe = path.join(thisExtensionPath(), 'res/vswhere.exe');
-  const vswhere_args = ['-all', '-format', 'json', '-products', '*', '-legacy', '-prerelease'];
-  const vswhere_res = await proc.execute(vswhere_exe, vswhere_args, null, {silent: true, encoding: 'utf8'}).result;
+  const vswhere_args =
+      ['/c', 'chcp 65001 |', vswhere_exe, '-all', '-format', 'json', '-products', '*', '-legacy', '-prerelease'];
```
vector-of-bool: Does this work? I'd think `child_process` would quote the arguments before calling CreateProcess and we'd end up with something like

```bash
cmd.exe /c "chcp 65001 |" vswhere.exe -all -format -json -products * -legacy -prerelease
```

That'll behave weird unless cmd.exe is doing its own goobery command-line parsing and not using `argv` from the CRT

vector-of-bool: Just tested, and can confirm that cmd.exe is doing something goobery:

```bash 
> cmd.exe /c "echo hi" echo foo
hi echo foo
```

Randshot: hm, it does work for me

Issue #366, opened 2018-04-01
Documentation Targets for 0.11.0


vector-of-bool:
Similar to the "Important Test Cases" ticket, this one lists aspects that need to be documented. Thankfully, a good initial listing can be obtained just by looking at the `contributions` section of `package.json`.

## Aspects that Must Be Documented:

- [x] Kits:
  - [x] User-local vs. project-local kits
  - [x] How kits work
  - [x] How kits are automatically detected
  - [x] Common kit options
  - [x] Toolchain kits
  - [x] VS kits
  - [x] Compiler kits
- [x] Variants:
  - [x] The default variants
  - [x] Where variant files can go
  - [x] How CMake Tools applies variants
  - [x] Examples of complex variant matrices
- [x] Configure
  - [x] Pre-reqs to running a configure
  - [x] Effects of running a configure
  - [x] Clean-configuring
  - [x] How changing the kit can automatically cause a clean-configure
  - [x] Controlling configuration via `settings.json` and variants
  - [x] Selecting a CMake generator
- [x] Build
  - [x] Running a build
  - [x] Pre-reqs of running a build
  - [x] Selecting a default build target
  - [x] Building a specific target only once
  - [x] How CMake Tools executes builds
  - [x] Cleaning build output (Running `clean`)
  - [x] A clean rebuild
  - [x] Stopping a running build
- [ ] Install (Not sure if this should be included since the default install requires privileged execution.)
- [ ] The cache editor (**Needs some serious TLC**. May want to drop the feature until we get custom WebViews in VSCode, which will make this a lot more sane and manageable.)
- [x] Quick-start!
- [x] Debugging a named target
  - **Question**: I have an internal debate over how to support debugging targets. Users (including myself) don't want to faff with `launch.json`, but it's very flexible. The initial goal of this feature was a zero-configuration debug setup. It's worked with a few kinks but been especially troublesome when users want to tweak the debug config. Perhaps we should wait for a custom WebView to manage more debugging?
  - [x] Debugging via `launch.json` and `${command:cmake.launchTargetPath}`
- [x] Launching a target *without* debugging (in the terminal)
- [x] Troubleshooting
- [x] General getting started
- [ ] **DEV DOCS**

Tag @ytimenkov, @KoeMai, @Randshot.

I've already tackled a few of the above items, but want to go over them with a keen eye to make sure they're still up-to-date.

[...]

Issue #340, opened 2018-03-11
Important Test Cases for 0.11.0


vector-of-bool:
For the pending 0.11.0 release, I'd like to get test cases for most of our big tickets, as well as tests for basic functionality. I'll keep this checklist here and we can add to it and reference it for the upcoming release 

## Cases which must be tested:

- [x] `${workspaceRoot}` in toolchain kit.
- [x] Can configure a project 
- [x] Can build a project
- [x] Environment variables in variants (#248)
- [x] The Visual Studio generator matching the active VS kit should be selected if no other generator is available.
- [x] Substitutions for `buildArgs` and `buildToolArgs`
- [x] Testing for all `*Environment` setting fields
- [x] Substitutions for `configureSettings` and `configureArgs`
- [x] Substitutions using variant option names

Yeah... the list is short for now, but we'll grow it. Comment below on this that should be tested.

Tag @Randshot, @ytimenkov, @KoeMai.

Randshot:
I plan to write tests for all my PRs:
- Variable substitution for `buildArgs` & `buildToolArgs`
- Testing for `*Environment` setting fields
- Variable substitution for `configureSettings` & `configureArgs`
- Variable substitution for variant option names

KoeMai:
I write at the moment tests for the preferred generator. With this case I will introduce different compilers into the test environment.

vector-of-bool:
[There's now a Gitter chat](https://gitter.im/vscode-cmake-tools/Lobby). Maybe a real badge someday.

vector-of-bool:
@Randshot did we write a test for `${workspaceRoot}` in Toolchain kits?

Randshot:
@vector-of-bool I don't think so, at least I didn't.

vector-of-bool:
The case I'm referring to is a toolchain kit where the path to the toolchain file contains `${workspaceRoot}` such that the path to the toolchain file is dependent on the path to the current workspace. This helps when a project defines its own toolchain file and a user wishes to use it as a kit. This is even more relevant with changes to support workspace-local kit definitions.

The path to the toolchain file only needs to be expanded at the last moment before passing `CMAKE_TOOLCHAIN_FILE` to the CMake configure process.

vector-of-bool:
BetWith beta5 released, what more test cases do we need to fill out?

vector-of-bool:
## Test Case Refactoring

[...]

PR #368, opened 2018-04-01
Refactor/quickstart

KoeMai:
### This changes refactors cmake command quickstart 

- Separate UI from business logic
- Add tests for business logic of quick start
- Add tests for quick start UI

- [ ] Clearify use of `cmake.sourceDirectory`  in quick start (Chicken or the egg Problem, because there is no CMakeFile which could be used by the driver. The driver is needed for substitution yet.)
- [ ] Tests for UI without folder
- [x] Tests for open source code
- [ ] Tests for check correctness for project templates

### DIFF
--- a/scripts/ci.ps1
+++ b/scripts/ci.ps1
@@ -59,7 +59,7 @@ if (! $NoTest) {
         -TestsPath "$REPO_DIR/out/test/unit-tests" `
         -Workspace "$REPO_DIR/test/unit-tests/test-project-without-cmakelists"
 
-    foreach ($name in @("successful-build"; "without-cmakelist-file"; )) {
+    foreach ($name in @("vs-preferred-gen"; "successful-build"; "without-cmakelist-file"; )) {
         Invoke-VSCodeTest "CMake Tools: $name" `
             -TestsPath "$REPO_DIR/out/test/extension-tests/$name" `
             -Workspace "$REPO_DIR/test/extension-tests/$name/project-folder"
--- a/scripts/ci.ps1
+++ b/scripts/ci.ps1
@@ -59,7 +59,7 @@ if (! $NoTest) {
         -TestsPath "$REPO_DIR/out/test/unit-tests" `
         -Workspace "$REPO_DIR/test/unit-tests/test-project-without-cmakelists"
 
-    foreach ($name in @("successful-build"; )) {
+    foreach ($name in @("successful-build"; "without-cmakelist-file"; )) {
         Invoke-VSCodeTest "CMake Tools: $name" `
             -TestsPath "$REPO_DIR/out/test/extension-tests/$name" `
             -Workspace "$REPO_DIR/test/extension-tests/$name/project-folder"
--- a/src/quickstart.ts
+++ b/src/quickstart.ts
@@ -1,0 +1,111 @@
+import * as fsNode from 'fs';
+import * as path from 'path';
+
+import {fs} from './pr';
+
+export enum ProjectType {
+  Library,
+  Exectable
+}
+
+export interface ProjectTypeDesciptor {
+  label: string;
+  type: ProjectType;
+  description: string;
+}
+
+export const projectTypeDescriptions: ProjectTypeDesciptor[] = [
+  {label: 'Library', type: ProjectType.Library, description: 'Create a library'},
+  {label: 'Executable', type: ProjectType.Exectable, description: 'Create an executable'}
+];
+
+export class CMakeQuickStart {
+
+  public readonly cmakeFilePath: string;
+  public project_name: string|undefined;
+  private _type: ProjectType|undefined;
+  private _sourceCodeFilePath: string;
+
+  public get sourceCodeFilePath() { return this._sourceCodeFilePath; }
+
+  public set type(type: ProjectType|undefined) {
+    if (type == ProjectType.Exectable) {
+      this._sourceCodeFilePath = path.join(this.workingPath, 'main.cpp');
+    } else {
+      this._sourceCodeFilePath = path.join(this.workingPath, `${this.project_name}.cpp`);
+    }
+    this._type = type;
+  }
+
+  public get type(): ProjectType|undefined { return this._type; }
+
+  constructor(readonly workingPath: string) {
+
+    this.cmakeFilePath = path.join(this.workingPath, 'CMakeLists.txt');
+
+    if (fsNode.existsSync(this.cmakeFilePath)) {
+      throw Error('Source code directory contains already a CMakeLists.txt');
+    }
+  }
+
+  private async createCMakeListFile() {
+    const init = [
+      'cmake_minimum_required(VERSION 3.0.0)',
+      `project(${this.project_name} VERSION 0.1.0)`,
+      '',
+      'include(CTest)',
+      'enable_testing()',
+      '',
+      this.type == ProjectType.Library ? `add_library(${this.project_name} ${this.project_name}.cpp)`
+                                       : `add_executable(${this.project_name} main.cpp)`,
+      '',
+      'set(CPACK_PROJECT_NAME ${PROJECT_NAME})',
+      'set(CPACK_PROJECT_VERSION ${PROJECT_VERSION})',
+      'include(CPack)',
+      '',
+    ].join('\n');
+
+    return fs.writeFile(this.cmakeFilePath, init);
+  }
+
+  private async createLibrarySourceFile() {
+    return fs.writeFile(this.sourceCodeFilePath, [
+      '#include <iostream>',
+      '',
+      'void say_hello(){',
+      `    std::cout << "Hello, from ${this.project_name}!\\n";`,
+      '}',
+      '',
+    ].join('\n'));
+  }
+
+  private async createMainSourceCodeFile() {
+    return fs.writeFile(this.sourceCodeFilePath, [
+      '#include <iostream>',
+      '',
+      'int main(int, char**) {',
+      '   std::cout << "Hello, world!\\n";',
+      '}',
+      '',
+    ].join('\n'));
+  }
+
+  private async createExampleSourcecodeFile() {
+    if (!(await fs.exists(this.sourceCodeFilePath))) {
+      switch (this.type) {
+      case ProjectType.Library:
+        return this.createLibrarySourceFile();
+      case ProjectType.Exectable:
+        return this.createMainSourceCodeFile();
+      }
+    }
+  }
+
+  public async createProject(project_name: string, type: ProjectType) {
+    this.project_name = project_name;
+    this.type = type;
+
+    await this.createCMakeListFile();
+    await this.createExampleSourcecodeFile();
+  }
+}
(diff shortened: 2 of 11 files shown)


scripts/ci.ps1
```diff
@@ -59,7 +59,7 @@ if (! $NoTest) {
         -TestsPath "$REPO_DIR/out/test/unit-tests" `
         -Workspace "$REPO_DIR/test/unit-tests/test-project-without-cmakelists"
 
-    foreach ($name in @("successful-build"; )) {
+    foreach ($name in @("successful-build"; "without-cmakelist-file"; )) {
```
vector-of-bool: Oh noes. Did I accidentally commit out the tests? Can you make sure that all the extension tests are listed here?

src/quickstart.ts
```diff
@@ -0,0 +1,111 @@
...
+  type: ProjectType;
+  description: string;
+}
+
+export const projectTypeDescriptions: ProjectTypeDesciptor[] = [
+  {label: 'Library', type: ProjectType.Library, description: 'Create a library'},
+  {label: 'Executable', type: ProjectType.Exectable, description: 'Create an executable'}
+];
+
+export class CMakeQuickStart {
```
vector-of-bool: It would be simpler to have a free function that implements `createProject()` and also takes the `workingPath` as a parameter, since the only use of this class is to be created and immediately call `createProject()`.

vector-of-bool: Try this design:

Have an `interface QuickStartParams` that includes things like the `ProjectType`, the `name` string, the source directory, etc. Have an `async` function that takes the params and performs the whole quickstart. You can then test the quickstart without opening any project directory or mocking any UI elements.

As a similar change, you can create an `async getUserQuickStartParams()` function that performs all the UI requests to generate a `QuickStartParams` object. If you make this function simple enough, no testing of it will be necessary.

@ytimenkov, opinions?

KoeMai: yes, I will move the path generation into the create functions for cmake and executeable or library. This allows to remove the class and add them as namespace functions.

KoeMai: I don't understand the need of QuickStartParams. I think we could move  quickstart UI parts out of the CMakeTools class to arrive the single responsibility. At the moment it is still part because we need to call the configure command.  do we have a interface for all commands? Then we code use dependency injection. 

The extension tests I have created are only used to test the UI integration. The unit tests need a workspace to simplify clean up.
I do not see a way to reduce the tests. The strong types have already reduce the count of tests.

KoeMai: I tried to redesign it as a function collection but in that case each function will get a quickstart parameter object. This is like writing C function with structurs to get a object oriented design. I would leave it as a class.

KoeMai: - [ ] Extract Quickstart UI workflow from CMakeTools-Class into separate function 
This function is using the interface CMakeToolsApi and class CMakeQuickstart to access needed backend behavior. 
All UI interactions could be moved into callbacks. This allows to test the workflow without UI interaction. The workflow class will not replace the UI tests, it will reduce only the count of tests and the order of user interaction:

1. CMakeList present
2. project name
3. project type
4. Open source file
5. configure project

vector-of-bool: Not really. Object orientation is more than just composing attributes together. Having a parameter object is more of a "Named-Parameter Idiom" for JS/TS. There are no invariants on the fields of the parameter collection (other than their names and types), neither are there methods acting upon it. Usage might look like this:

```typescript
const qs_res = await qs.quickStart({
  sourceDir: "<some-path>",
  type: ProjectType.Library,
  name: "MyProject",
});
```

And the usage with the UI might look like this:

```typescript
// Show the UI to get the parameters for a quickstart
const qs_params = await qs.getQuickStartOptionsViaUI({sourceDir: this.sourceDir});
if (!qs_params) {
  // Use cancelled
  return;
}
const qs_res = await qs.quickStart(qs_params);
await vscode.window.showTextDocument(qs_res.mainListFile);
await this.configure();
```

KoeMai: I changed the quickstart class by quickstart.createProject.

Do we want to separate UI from UI workflow? UI can only be tested by vscode integration test. The workflow chould be tested without vscode intergation. If we use always the same UI elements we could check our expectations against vscode by small tests for the UI elements. This would allow to reduce the needed extension tests.

src/quickstart.ts
```diff
@@ -1,149 +1,163 @@
...
+
+  let sourceFile: string = '';
+  try {
+    if (projectInformation.type == ProjectType.Exectable) {
+      sourceFile = await createMainSourceCodeFile(projectInformation);
+    } else if (projectInformation.type == ProjectType.Library) {
+      sourceFile = await createLibrarySourceFile(projectInformation);
     }
-  }
+  } catch (err) { log.debug(err); }
```
vector-of-bool: Be more careful with error handling here. Instead of catching and throwing for errors from the `createLibrarySourceFile` functions, have them return a union type that contains the error information:

```typescript 
enum CreateFileError {
  FileExists = 1,
}

async function createMainSourceCodeFile(info): Promise<CreateFileError|string> {
  // ... 
}
```

Usage like this:

```typescript 
const result = await createMainSourceCodeFile(projectInformation);
if (result == CreateFileError.FileExists) {
  return result;
}
// It succeeded, so we have a string:
sourceFile = result;
```

src/quickstart.ts
```diff
@@ -1,149 +1,163 @@
...
 
-  public async createProject() {
-    await this.createCMakeListFile();
-    await this.createExampleSourcecodeFile();
-  }
+  return {cmakeListFile: cmakeFile, sourceFiles: [sourceFile]} as GeneratedProjectFiles;
 }
 
-export interface QuickStartCallbacks {
+export interface UiControlCallbacks {
```
vector-of-bool: `UiControlCallbacks` -> `UIControlCallbacks`

src/quickstart.ts
```diff
@@ -1,149 +1,163 @@
...
   onProjectNameRequest: () => Promise<string|undefined>;
-  onProjectTypeRequest: (items: ProjectTypeDesciptor[]) => Promise<ProjectTypeDesciptor|undefined>;
-  onOpenSourceFiles: (filePaths: string) => void;
+  onProjectTypeRequest: (items: ProjectTypeDesciption[]) => Promise<ProjectTypeDesciption|undefined>;
+  onOpenSourceFiles: (filePaths: GeneratedProjectFiles) => Promise<void>;
   onError: (message: string) => void;
+  onWarning: (message: string) => void;
```
vector-of-bool: These are a bit too close to event names, even though they are not.

Change `onProjectNameRequest` to `getProjectName`, and so on.

src/quickstart.ts
```diff
@@ -1,149 +1,163 @@
...
-  private async createLibrarySourceFile() {
-    return fs.writeFile(this.sourceCodeFilePath, [
-      '#include <iostream>',
-      '',
-      'void say_hello(){',
-      `    std::cout << "Hello, from ${this.projectName}!\\n";`,
-      '}',
-      '',
-    ].join('\n'));
+async function createMainSourceCodeFile(projectInformation: CreateProjectInformation) {
```
vector-of-bool: Separate the "create file" functions from a "create content" function. Have the content-creation function generate the string content of the file, and have the file-creation function just call the content creator and then write it to a file.

If you do this, you will be able to test the content-creation functions without having to write anything to the filesystem.

test/unit-tests/quickstart.test.ts
```diff
@@ -23,25 +23,32 @@ suite('QuickStart test', async () => {
   test('CMakeLists.txt exists', async () => {
     await fs.writeFile(path.join(working_path, 'CMakeLists.txt'), 'Dummy');
 
-    expect(() => new CMakeQuickStart(working_path, "", ProjectType.Exectable)).to.throws(
-        'Source code directory contains already a CMakeLists.txt');
+    const project: quickstart.CreateProjectInformation
+        = {name: '', type: quickstart.ProjectType.Exectable, sourceDirectory: working_path};
+    quickstart.createProject(project)
```
vector-of-bool: Need to `await` on this line

vector-of-bool:
Regarding `cmake.sourceDirectory`: We can do one of two things:

1. Ignore it.
2. Do an expansion anyway. We don't _need_ the driver to do an expansion. You can call the `expand.expandString` function and fill out the parameters manually. It will have to make some assumptions, though.

For checking project templates, I've added a review comment regarding how they can be separated from the functions that write the file.

src/quickstart.ts
```diff
@@ -0,0 +1,149 @@
...
+      '#include <iostream>',
+      '',
+      'int main(int, char**) {',
+      '   std::cout << "Hello, world!\\n";',
+      '}',
+      '',
+    ].join('\n'));
+  }
+
+  private async createExampleSourcecodeFile() {
```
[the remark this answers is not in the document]
KoeMai: +1

KoeMai:
I think we have in this review many discussions about the design and methodic. 
@vector-of-bool and @ytimenkov please give me a response, I would then try too redesign the code according your feedbacks.

# Testing 
The test I have written are from the first draft. The UI tests are not up to date. Before I clean it up I try to understand the interest how we want to test the different parts.  I am with you for testing the backend independent from the vscode instances with mocha. At the moment we have for backend no vscode independed test runner: I think it is to early for introduce a separate backend test, there are to many dependencies to vscode.
May be we can prepare it and move all vscode free component unit tests into the new test runner.
Should we create a new issue for this? 

# UI testing
In the UI part there is a mix of UIview , UIControl and the backend logic. The extension test are at the moment in use to avoid breaking the present implementation. I think we could split them into tests the UIControl and backend logic independent. 

# Warning and errors

If I look into the code base on UI parts there are different approaches. The UI control returns error codes, they are needed by the present realization. It returns different error codes for different actions. I have introduced a enum for them but I think we could reduce this to good and bad. 

For the integration into vscode and usability we should follow one strategy to publish error and warnings. At the moment I realized for quickstart the onWarning and onError callbacks in the uicontrol. I like to get your response. I think the errors can be handled by exceptions and the warnings by a list of warnings objects as function return, then the callbacks are obsolete. And we could create a resuable ui error handler. I think this is independenfrom the rollback parts.

# Behavior changes

## SourceDirectory

I think we could use the expand.expandString().

vector-of-bool:
I'm going to do some quick big bugfixes for a 0.11.1, then I'll jump on this if I get the time.

Randshot:
Sry, accidentally deleted the branch while cleaning up my local git fork.

CwanyJest95:
> ## This change addresses item #340 and #366 and tests for #363
> ### This changes refactors cmake command quickstart
> The following changes are proposed:
> 
> * Separate UI from business logic
> * Add tests for business logic of quick start
> * Add tests for quick start UI
> * [ ]  Clearify use of `cmake.sourceDirectory`  in quick start (Chicken or the egg Problem, because there is no CMakeFile which could be used by the driver. The driver is needed for substitution yet.)
> * [ ]  Tests for UI without folder
> * [x]  Tests for open source code
> * [ ]  Tests for check correctness for project templates

0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2

PR #369, opened 2018-04-01
Feature/workspace folder

bjosa:
### This changes variable substitution and corresponding tests

- Add variable workspaceFolder to expansion Options
- Added test for the change

## Other Notes/Information

Also replaced deprecated `workspace.rootPath` with `workspace.workspaceFolders[0]`

### DIFF
--- a/src/driver.ts
+++ b/src/driver.ts
@@ -190,15 +190,28 @@ export abstract class CMakeDriver implements vscode.Disposable {
   }
 
   /**
+   * Get the vscode root workspace folder.
+   *
+   * @returns Returns the vscode root workspace folder. Returns `null` if no folder is open or the folder uri is not a `file://` scheme.
+   */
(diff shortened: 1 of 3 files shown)


src/driver.ts
```diff
@@ -189,16 +189,29 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
   }
 
+  /**
+   * Get the vscode root workspace folder.
+   *
+   * @returns Returns the vscode root workspace folder. Returns `null` if no folder is open or the folder uri is not a `file://` scheme.
+   */
+  private get _workspaceRootPath() {
+    if (!vscode.workspace.workspaceFolders || vscode.workspace.workspaceFolders[0].uri.scheme !== 'file') {
+      return null
```
Randshot: A semicolon is missing here.

PR #381, opened 2018-04-13
[windows] Use quotes to allow spaces in path

tusharpm:

- Quotes around `vswhere_exe` path.

## The purpose of this change

For the case where `vswhere_exe` path has a whitespace character.

## Other Notes/Information

On Windows, my username has a whitespace ("Tushar Maheshwari" in my case).
VS Code stores the installed extensions, and their resources in the home directory, which has the username in the path.
The extension uses `vswhere` without quotes, which fails with:
```
'C:\Users\Tushar' is not recognized as an internal or external command, operable program or batch file.
```
Tried locally with the changes, the extension succeeds in executing `vswhere` and continues execution.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -303,13 +303,13 @@ export interface VSInstallation {
 export async function vsInstallations(): Promise<VSInstallation[]> {
   const installs = [] as VSInstallation[];
   const inst_ids = [] as string[];
-  const vswhere_exe = path.join(thisExtensionPath(), 'res/vswhere.exe');
+  const vswhere_exe = path.join(thisExtensionPath(), 'res', 'vswhere.exe');
   const sys32_path = path.join(process.env.WINDIR as string, 'System32');
 
   const vswhere_args =
-      ['/c', `${sys32_path}\\chcp 65001 | ${vswhere_exe} -all -format json -products * -legacy -prerelease`];
+      ['/c', `${sys32_path}\\chcp 65001 | "${vswhere_exe}" -all -format json -products * -legacy -prerelease`];
   const vswhere_res
-      = await proc.execute(`${sys32_path}\\cmd.exe`, vswhere_args, null, {silent: true, encoding: 'utf8'}).result;
+      = await proc.execute(`${sys32_path}\\cmd.exe`, vswhere_args, null, {silent: true, encoding: 'utf8', shell: true}).result;
 
   if (vswhere_res.retc !== 0) {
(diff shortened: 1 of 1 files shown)


src/kit.ts
```diff
@@ -303,13 +303,13 @@ export interface VSInstallation {
...
-  const vswhere_exe = path.join(thisExtensionPath(), 'res/vswhere.exe');
+  const vswhere_exe = path.join(thisExtensionPath(), 'res', 'vswhere.exe');
   const sys32_path = path.join(process.env.WINDIR as string, 'System32');
 
   const vswhere_args =
-      ['/c', `${sys32_path}\\chcp 65001 | ${vswhere_exe} -all -format json -products * -legacy -prerelease`];
+      ['/c', `${sys32_path}\\chcp 65001 | "${vswhere_exe}" -all -format json -products * -legacy -prerelease`];
   const vswhere_res
-      = await proc.execute(`${sys32_path}\\cmd.exe`, vswhere_args, null, {silent: true, encoding: 'utf8'}).result;
+      = await proc.execute(`${sys32_path}\\cmd.exe`, vswhere_args, null, {silent: true, encoding: 'utf8', shell: true}).result;
```
Randshot: Is shell mode necessary here?

tusharpm: I tried without setting it. It reported something like:
```
'\"C:\Users\Tushar Maheshwari\[...]\vswhere.exe\"' is not recognized as an internal or external command, operable program or batch file.
```
It escaped the quotes because of the missing `windowsVerbatimArguments` (default: false).
The existing `proc.execute` does not pass that option to `child_process.spawn`, like it does `shell`.
According to the [documentation](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options), the former "is set to `true` automatically when `shell` is specified."

Therefore, although the `shell` mode might not exactly be necessary, it helps in the fix here.
The alternative is to modify `proc.ts:execute` to pass `windowVerbatimArguments` also.
Thoughts?

vector-of-bool: Sorry for the slow response. Does setting `windowsVerbatimArguments` work? I'm disinclined to ever use `shell`, but we'll have to if nothing else works.

I'm hoping to push 0.11.1 today and want to get this in the release.

tusharpm: Oh, I tested that change in a different branch. This one should be alright to merge if `shell` is acceptable.

vector-of-bool:
@Randshot could you follow up on your review? I can override and do the merge without your review, but I'd rather not.

PR #386, opened 2018-04-16
Implement better configure-dirty checks.

### LINKED ISSUE #385: "No build system was generated yet" errors from CMake Server
It looks like we try to get the codemodel from CMake Server before the generation is finished. This is the most common rollbar issue ATM, even though it doesn't seem to cause any real trouble?

vector-of-bool:
CMake's dirty signal is a bit unreliable, but it does let us query all files
that are inputs to the build. We can use this information to implement
our own dirty checks.

This commit also tweaks some auto-saving so that we can be sure that
configure output is written to the terminal at the proper time. It also
moves auto-saving out of the driver logic into the business logic code.

[Fix #385], possibly?

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -419,10 +419,40 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
   }
 
   /**
+   * Save all open files. "maybe" because the user may have disabled auto-saving
+   * with `config.saveBeforeBuild`.
+   */
+  async maybeAutoSaveAll(): Promise<boolean> {
+    // Save open files before we configure/build
+    if (config.saveBeforeBuild) {
+      log.debug('Saving open files before configure/build');
+      const save_good = await vscode.workspace.saveAll();
+      if (!save_good) {
+        log.debug('Saving open files failed');
+        const chosen = await vscode.window.showErrorMessage<
+            vscode.MessageItem>('Not all open documents were saved. Would you like to continue anyway?',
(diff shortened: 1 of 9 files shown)


src/cms-driver.ts
```diff
@@ -18,6 +18,20 @@ import * as util from './util';
 
 const log = createLogger('cms-driver');
 
+class CMakeInputFile {
```
KoeMai: Please move the cmake dirty check into a separate file, it is an encapsulated unit.

src/cmake-tools.ts
```diff
@@ -484,8 +514,13 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
    * Implementation of `cmake.build`
    */
   async build(target_?: string): Promise<number> {
-    // First, reconfigure if necessary
-    if (await this._needsReconfigure()) {
+    // First, save open files
+    if (!await this.maybeAutoSaveAll()) {
+      return -1;
```
KoeMai: Is this error code unique and is it needed?

vector-of-bool: Not unique, but its purpose is to signal to the caller that the build "failed" (because saving files failed).

KoeMai:
Please add an unit test for file change detection, dirty generation and cmake server cmakeInputs.

vector-of-bool:
Want to give it another look? I'm almost ready to merge, and it will seal up the things I want for 0.11.1

PR #395, opened 2018-04-22
Bugfix/remove lowercase in diagnostic path

bbosnjak:

OS: windows

Path in diagnostic output are converted to lowercase. If there is name in path not in lowercase, when clicking on issue in "PROBLEMS" window vscode will open new editor window with same file just lowercase name.


vector-of-bool:
I think this relates to needing a proper canonicalization of Windows paths. I'm just not sure how to go about doing it.

LGTM.

Issue #371, opened 2018-04-03
If "cmake.cmakePath" is set wrong, cmake-tools fails with unhandled exception


s-martin:
### Brief Issue Summary

I had from another platform set the `cmake.cmakePath` to `cmake3`, which didn't exist on my current plattform.

### Expected:

I would expect that cmake-tools would return some meaningful error, e.g. "No cmake file in cmake.cmakePath found".

### Apparent Behavior:

cmake-tools threw an unhandled exception:
```
console.ts:136 [Extension Host] Error: spawn cmake3 ENOENT
	at exports._errnoException (util.js:1050:11)
	at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)
	at onErrorNT (internal/child_process.js:367:16)
	at _combinedTickCallback (internal/process/next_tick.js:80:11)
	at process._tickCallback (internal/process/next_tick.js:104:9)
t.log @ console.ts:136
t._logExtensionHostMessage @ extensionHost.ts:400
(anonymous) @ extensionHost.ts:213
emitTwo @ events.js:106
emit @ events.js:194
process.nextTick @ internal/child_process.js:766
_combinedTickCallback @ internal/process/next_tick.js:73
_tickCallback @ internal/process/next_tick.js:104
```

### Platform and Versions

- **Operating System**: Windows 7
- **CMake Version**: 3.9.4
- **VSCode Version**: 1.21.1
- **CMake Tools Extension Version**: 0.11-beta5
- **Compiler/Toolchain**: Visual C++ 2017 x86

Randshot:
@s-martin CMT doesn't check whether a file exists before trying to execute it, so that is why an exception is thrown in your case. It's trivial to add a check, but I don't know what @vector-of-bool s opinion is on this topic.

KoeMai:
Should be fixed in next version.

PR #397, opened 2018-04-22
Fix and clean up cmake missing error message creation for #371

KoeMai:
The cmake missing error message was
The cmake path variable had many dependency to other components.
This refactoring introduce a single interface which represents the
cmake executable. It provides all information about the
cmake executable (version, serverModeSupport, and paths).

This modification allows to make the cmake executable check
in one function, separate ui from check and reduces the hide
dependencies. This ensures that only a valid cmake path is used
in a consistent way inside the drivers.

- Add new function for version and file present check.
- Refactored drivers to use cmakeexecutable object for path access.
- Add error message on missing cmake.
- Avoid throwing exception in _getCMakePath()
- Reformat source code with node_modules clang-format (7.0.0)

grafik

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -318,7 +318,9 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
       log.debug('Not starting CMake driver: no kits defined');
       return null;
     }
-    const cmakePath = await paths.cmakePath;
+    let cmakePath = await paths.cmakePath;
+    if (cmakePath === null)
+      cmakePath = '';
     const cmake = await getCMakeExecutableInformation(cmakePath);
     if (!cmake.isPresent) {
       vscode.window.showErrorMessage(`Bad CMake executable "${
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -9,6 +9,7 @@ import * as ws from 'ws';
 import * as api from './api';
 import {ExecutionOptions, ExecutionResult} from './api';
 import {CacheEditorContentProvider} from './cache-editor';
+import {CMakeExecutable, getCMakeExecutableInformation} from '@cmt/cmake/cmake_executable';
 import {CMakeServerClientDriver} from './cms-driver';
 import config from './config';
 import {CTestDriver} from './ctest';
@@ -21,11 +22,9 @@ import * as logging from './logging';
 import {NagManager} from './nag';
 import paths from './paths';
 import {fs} from './pr';
-import * as proc from './proc';
 import rollbar from './rollbar';
 import {StateManager} from './state';
 import {StatusBar} from './status';
-import * as util from './util';
 import {VariantManager} from './variant';
 
 const log = logging.createLogger('main');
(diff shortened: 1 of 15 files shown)


src/cmake/cmake_executable.ts
```diff
@@ -0,0 +1,40 @@
+import * as proc from '../proc';
```
vector-of-bool: Re-format this file, and use `kebab-case` in the filename.

src/cmake-tools.ts
```diff
@@ -167,32 +166,25 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
-      if (util.versionGreater(version, '3.7.1')) {
-        drv = await CMakeServerClientDriver.create(this._stateManager, kit);
+      if (cmake.isServerModeSupported) {
+        drv = await CMakeServerClientDriver.create(cmake, this._stateManager, kit);
       } else {
-        log.info(
-            'CMake Server is not available with the current CMake executable. Please upgrade to CMake 3.7.2 or newer.');
-        drv = await LegacyCMakeDriver.create(this._stateManager, kit);
+        log.info(`CMake Server is not available with the current CMake executable. Please upgrade to CMake ${
+            cmake.minimalServerModeVersion} or newer.`);
```
vector-of-bool: `cmake.minimalServerModeVersion` will not print correctly, since it's not a simple string. It will print as `[object Object]`.

vector-of-bool: While here, can you change `log.info` to a warning? Just to make it more clear in the logs.

src/cmake/cmake_executable.ts
```diff
@@ -0,0 +1,40 @@
...
+        if (version_ex.retc === 0 && version_ex.stdout) {
+        cmake.isPresent = true;
+
+        console.assert(version_ex.stdout);
+        const version_re = /cmake version (.*?)\r?\n/;
+        cmake.version = util.parseVersion(version_re.exec(version_ex.stdout)![1]);
+
+        // We purposefully exclude versions <3.7.1, which have some major CMake
+        // server bugs
+        if (util.versionGreater(cmake.version, cmake.minimalServerModeVersion)) {
```
vector-of-bool: Change this branch to a ternary:

```typescript
cmake.isServerModeSupported = util.versionGreater(cmake.version, cmake.minimalServeModeVersion);
```

src/driver.ts
```diff
@@ -2,15 +2,15 @@
...
-import * as path from 'path';
-import * as vscode from 'vscode';
-
-import * as api from './api';
-import config from './config';
-import * as expand from './expand';
-import {CMakeGenerator, CompilerKit, getVSKitEnvironment, Kit, kitChangeNeedsClean, ToolchainKit, VSKit} from './kit';
-import * as logging from './logging';
-import paths from './paths';
+ import * as path from 'path';
```
vector-of-bool: A bunch of random spaces appeared here. Re-format.

src/legacy-driver.ts
```diff
@@ -3,17 +3,17 @@
  * Can also talk to newer versions of CMake via the command line.
  */ /** */
 
-import * as path from 'path';
-import * as vscode from 'vscode';
+ import * as path from 'path';
```
vector-of-bool: More mystery spaces

src/paths.ts
```diff
@@ -95,8 +95,7 @@ class Paths {
             }
           }
         }
-        // We've got nothing...
-        throw new Error('No CMake found on $PATH');
+        return '';
```
vector-of-bool: Instead of empty string, return `null` and tweak the return type of this function. Change callers to check for `null` and handle appropriately.

src/cmake/cmake-executable.ts
```diff
@@ -9,13 +9,13 @@ export interface CMakeExecutable {
   minimalServerModeVersion: util.Version;
 }
 
-export async function getCMakeExecutableInformation(path: string): Promise<CMakeExecutable> {
+export async function getCMakeExecutableInformation(path: string|null): Promise<CMakeExecutable> {
```
vector-of-bool: Keep this accepting a `string` and instead `console.assert(path.length != 0)`, that is to prevent a bad executable from going through. Change code that gets the path from the config to handle the error at that level, rather than at this level.

src/paths.ts
```diff
@@ -64,20 +64,20 @@ class Paths {
...
-  get cmakePath(): Promise<string> { return this._getCMakePath(); }
+  get cmakePath(): Promise<string|null> { return this._getCMakePath(); }
   get ctestPath(): Promise<string> { return this._getCTestPath(); }
 
   private async _getCTestPath(): Promise<string> {
     const ctest_path = config.raw_ctestPath;
     if (!ctest_path || ctest_path == 'auto') {
       const cmake = await this.cmakePath;
-      return path.join(path.dirname(cmake), 'ctest');
+      return path.join(path.dirname(cmake!), 'ctest');
```
vector-of-bool: Don't do a non-null assertion here. Instead, return `null` for the CTest path in the same way as cmakePath.

src/cmake/cmake-executable.ts
KoeMai: I would change it to accept only string.
The assert has the problem it will rise an `AssertException`, this would broke the control flow, but this function should provide a valid executable information by returning the cmake struct. In cause of a problem it would be `cmake.isPresent = false`, and the executable is invalid.

src/paths.ts
KoeMai: Done

src/ctest.ts
```diff
@@ -346,22 +346,28 @@ export class CTestDriver implements vscode.Disposable {
...
-                                new CTestOutputLogger(),
-                                {environment: config.testEnvironment, cwd: driver.binaryDir});
-
-    const res = await child.result;
-    await this.reloadTests(driver);
-    if (res.retc === null) {
-      log.info('CTest run was terminated');
-      return -1;
+    const ctestpath = await paths.ctestPath;
+    if (ctestpath) {
```
vector-of-bool: Swap the if-else here. Check if `ctest_path` is null and return early, then the larger code branch can be unindented and there is not need for an `else` branch.

PR #403, opened 2018-04-23
In scanForKits allow for arm-none-eabi-gcc

schultetwin1:
### This changes scanForKit 

- Scan for kits allows for a target triple prefixing gcc in the binary name.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -139,7 +139,7 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): MaybeCo
       return null;
     }
     const version = version_match[1];
-    const gxx_fname = fname.replace(/^gcc/, 'g++');
+    const gxx_fname = fname.replace(/gcc/, 'g++');
     const gxx_bin = path.join(path.dirname(bin), gxx_fname);
     const name = `GCC ${version}`;
     log.debug('Detected GCC compiler kit:', bin);
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -324,7 +324,7 @@ export async function vsInstallations(): Promise<VSInstallation[]> {
 
   const vs_installs = JSON.parse(vswhere_res.stdout) as VSInstallation[];
   for (const inst of vs_installs) {
-    if ((inst_ids.indexOf(inst.instanceId) < 0) && (inst.displayName)) {
+    if (inst_ids.indexOf(inst.instanceId) < 0) {
       installs.push(inst);
       inst_ids.push(inst.instanceId);
     }
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -118,7 +118,7 @@ type MaybeCompilerKitPr = Promise<CompilerKit|null>;
 export async function kitIfCompiler(bin: string, pr?: ProgressReporter): MaybeCompilerKitPr {
   const fname = path.basename(bin);
   // Check by filename what the compiler might be. This is just heuristic.
-  const gcc_regex = /^gcc(-\d+(\.\d+(\.\d+)?)?)?(\.exe)?$/;
+  const gcc_regex = /^((\w+-)*)gcc(-\d+(\.\d+(\.\d+)?)?)?(\.exe)?$/;
   const clang_regex = /^clang(-\d+(\.\d+(\.\d+)?)?)?(\.exe)?$/;
   const gcc_res = gcc_regex.exec(fname);
   const clang_res = clang_regex.exec(fname);
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -141,7 +141,13 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): MaybeCo
     const version = version_match[1];
     const gxx_fname = fname.replace(/gcc/, 'g++');
     const gxx_bin = path.join(path.dirname(bin), gxx_fname);
-    const name = `GCC ${version}`;
+    const target_triple_re = /((\w+-)+)gcc.*/;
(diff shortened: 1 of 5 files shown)


vector-of-bool:
Looks like the regex change caused it to pick a different Kit to test with on Linux. @KoeMai do you want to take a look at this?

schultetwin1:
Ah, that's a good catch. When I run gcc -v there is a line listed as Target:
> Target: arm-none-eabi

What if I prepend the target to the name (if it exists)? So the name would now be arm-none-eabi-gcc.

According to the [cmake docs](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html) using the toolchain file will set a flag to mark that cmake is cross compiling. I'm unsure where else that does but there does appear to be some difference between using a toolchain file and just setting CC and CXX without a toolchain file.

vector-of-bool:
Setting that flag is important in cases where CMake will try to execute compiled code as part of the build process. It's also important to specify the behavior of the `find_*` commands. I'm wondering it would be possible to auto-generate toolchain files from these cross-compilers... hmm...

KoeMai:
I miss a test for the regex, like the [Fake compilers](https://github.com/vector-of-bool/vscode-cmake-tools/blob/ccdfcd072cdf67e8bc09ef20429d952a19deffcc/test/unit-tests/kit-scan.test.ts#L50). Could you add one?

This modification let travis find the gcc for cross compiling to windows before the native compiler. We need to look on this problem the build fails but is should work. The second problem is that the executable is needed on linux but it is generated for windows. I will try to catch them.

schultetwin1:
I've added a test for this new regex. After changing the replace for gcc -> g++ things are passing in apple land. It looks like the failed tests in the linux unit tests is due to the first found compiler being a cross compiler (as @KoeMai said). Any thoughts on how to fix that? Even though we picked a different compiler should the build still succeed?

KoeMai:
The easiest way is reusing the excludeKitLabel, so that the kit picker (quick-picker.ts) checks additional the description of a kit. With this solution we could exclude `mingw` for the linux target inside the test environment configuration (KITS_BY_PLATFORM). We need a native compiler in the tests to check the correct configuration.

schultetwin1:
Unfortunately, currently mingw in not in the kit label. I can add that so that the label is not just "GCC {Version}" but "binary name {VERSION}"?

vector-of-bool:
You can use another regex to rename the kit. Like: `<triple>-gcc` -> `GCC for <triple> <version>`.

I don't think having them show up would be at all confusing to most users. Most would still see their regular GCC in the list and use that.

schultetwin1:
Ah, I like that idea, I'll do that.

schultetwin1:
Somehow those changes made my builds pass... I'm unsure if that should've happened. How do I run these tests at my desk? I've tried running the CI script but that seems to blow up. I also tried some of the VSCode builds but some of them relying on cmake running first.

vector-of-bool:
For travis you can use `trytravis`, which is was I do to test changes before pushing. What errors are you seeing from the CI scripts?

schultetwin1:
Ah, I figured out why the ci.ps1 script wasn't working for me. It needs to be run in powershell 6. I've put out a quick PR to check that so other windows developers don't run into the same issue (Powershell 5 is what ships with Windows by default). I'm running tests locally now and will also give trytravis a go.

vector-of-bool:
So something's definitely going wrong with the AppVeyor build. @KoeMai do you want to look into it?

schultetwin1:
I tried taking a look into it but couldn't repro the issue on my machine and didn't have enough info from the logs to figure out what was going on. @vector-of-bool or @KoeMai, I can look into it farther but I'm very new to typescript, mocha, and the tooling around it. Is there anyway to get more verbose logs?

KoeMai:
I took a look on Appveyor Vs 2015 visual studio, I don´t know whats happend but it does not found visual studio kits.

vector-of-bool:
I haven't been to my windows box for a few days. Any discoveries? It may be useful to add some additional logging around the `vsInstallations()` function in `kit.ts` to check what vswhere is doing and saying.

vector-of-bool:
@KoeMai what do you think about just removing the VS 2015 test? I think we get fine coverage from the other two environments.

KoeMai:
I think the displayName check creates the error and this may i create an error in the kit representation.

vector-of-bool:
@KoeMai Would you look into trying to get this fixed? I'd like to get this one merged for 1.0.0

src/kit.ts
```diff
@@ -318,7 +324,7 @@ export async function vsInstallations(): Promise<VSInstallation[]> {
 
   const vs_installs = JSON.parse(vswhere_res.stdout) as VSInstallation[];
   for (const inst of vs_installs) {
-    if (inst_ids.indexOf(inst.instanceId) < 0) {
+    if ((inst_ids.indexOf(inst.instanceId) < 0) && (inst.displayName)) {
```
KoeMai: This line makes the problem. Visual studio 2015 and older do not provide a displayName. So all kit other then VS2017 are dropped.  In a later step are the displayName generated if necessary.

I think this modification does not belong to the PR.

vector-of-bool: Nice find. I remember a PR several months ago that fixed this exact `displayName` issue. It's good then that we didn't just pass it through.

@schultetwin1 would you look into this?

schultetwin1: @KoeMai  you are correct, taking that out now. Sorry about that.

schultetwin1:
That seemed to do it, thanks @KoeMai!

PR #404, opened 2018-04-25
Make sure ci.ps1 is running in powershell 6

schultetwin1:
ci.ps1 uses built in variables which are only present in powershell 6. $PSVersionTable does not have an os entry in powershell version 5. This will cause ci.ps1 to break in cryptic ways.

This adds quick break to the powershell script if its running on an old version of powershell.


vector-of-bool:
I thought I had fixed it to read the version from package.json. Oops.

PR #416, opened 2018-05-05
Merge Kit Types

vector-of-bool:
This PR addresses several kit-related requests and issues since 0.11.0.

1. There are no longer distinct kit types. Each kit can have any of the parameters. This addresses using `clang-cl` in a Visual Studio environment, which is a long-standing issue.
2. There is no a special implicit kit created in memory called `__unspec__`. It's never written to disk, and it's display name varies by context, but it has no options defined. This allows users to fall-through to CMake's own compiler deduction.

Still on the docket is to refactor out `KitManager` to not be so intrusive and more easily testable.


schemas/kits-schema.json
```diff
@@ -1,148 +1,74 @@
...
-                    },
-                    "required": [
-                        "name"
-                    ]
+            },
+            "cmakeSettings": {
+                "type": "object",
+                "patternProperties": {
+                    ".*": {
+                        "description": "Value for the CMake Setting"
```
KoeMai: Are this only -D values does this could be a -C or -T?

src/driver.ts
```diff
@@ -262,20 +229,14 @@ export abstract class CMakeDriver implements vscode.Disposable {
     if (this._kit.environmentVariables) {
       util.objectPairs(this._kit.environmentVariables).forEach(([k, v]) => this._kitEnvironmentVariables.set(k, v));
     }
-    switch (this._kit.type) {
-    case 'vsKit': {
+    if (this._kit.visualStudio && this._kit.visualStudioArchitecture) {
       const vars = await getVSKitEnvironment(this._kit);
```
KoeMai: Could we move this visual studio dependend environmentVariables to code which loads kits.
The driver makes no difference between VSKitEnvironment variables and kitEnvironementVariables.

The created error is a more important information on loading of kits.

src/driver.ts
```diff
@@ -262,20 +229,14 @@ export abstract class CMakeDriver implements vscode.Disposable {
     if (this._kit.environmentVariables) {
       util.objectPairs(this._kit.environmentVariables).forEach(([k, v]) => this._kitEnvironmentVariables.set(k, v));
     }
-    switch (this._kit.type) {
```
KoeMai: I like this PR, it reduces the complexity of the code and increases the flexibility for the user.

schemas/kits-schema.json
vector-of-bool: These are only for `-D` options. Passing in a `-C` initializer is a separate topic, but has come up.

vector-of-bool:
I slapped down the AppVeyor builds because it's wasting my time. I waited for at least one to finish green, so I feel good about this.

Issue #417, opened 2018-05-06
Specify path to cmake executable


habemus-papadum:
### Brief Issue Summary
There doesn't seem to be a config property for the cmake executable location.

### Expected:
A way to specify cmake executable location and clear instructions as to which property is to be used.

### Other Notes/Information
I have multiple versions of cmake on my system to support various projects

KoeMai:
There is a way but it is missing inside the documentation. The settings file support it.
`"cmake.cmakePath": "<filepath of executable>"`

The `<filepath of executable>` has the default value `cmake`. This will search inside the path variable but you could use an absolute path to access the specific cmake versions.

habemus-papadum:
Thanks!

PR #418, opened 2018-05-06
Add cmakePath variable description to documentation.

KoeMai:
## This change addresses item  #417


vector-of-bool:
I'm forcing it. The Travis build is aborting abnormally, but is otherwise OK.

Issue #375, opened 2018-04-07
CMake: Debug Target does not find GDB


KoeMai:
### Brief Issue Summary

I try to use the extension in different projects. At the moment the debugging does not work out of the box. If I use the GCC then I get the following message, see figure below. I am not sure if there is a relation between debugger and kit needed. Is there some manual configuration needed? Or should it work out of the box?

### Expected:

1. Debugger should start out of the box (for quick start projects)
2. Possibility to place a debug configuration name to Kit or Variant, so that it switch automatic debug target dependent parts for different kits/variants on start of debugging.
3. May be a automatic generation of debug configuration of Kits, so that it use the debugger which is needed by the compiler (GCC, visual studio, cross compiler GCC will not work there are more information needed)

### Apparent Behavior:

1. Select MinGw GCC as kit
2. Create new executable project with quick start 
3. Build
4. Run debugger
![unabletoconfiguredebugger](https://user-images.githubusercontent.com/21360085/38452200-a68143cc-3a3f-11e8-8657-f738256b6fb5.png)

### Platform and Versions

- **Operating System**: Windows 10
- **CMake Version**: 3.10.0
- **VSCode Version**: 1.21.1
- **CMake Tools Extension Version**: 0.11-beta5
- **Compiler/Toolchain**: mingw-w64/msys2 + gcc 6.3

KoeMai:
I see the missing part. But I am not sure how we want to find gdb.
Below you can see my modification. It works but the kit contains at the moment no debugger path. 
This makes it necessary to convert the compiler path to it. 
  ```
...
  const debug_config: vscode.DebugConfiguration = {
    type: is_msvc ? 'cppvsdbg' : 'cppdbg',
    name: `Debug ${target_path}`,
    request: 'launch',
    cwd: '${workspaceRoot}',
    args: [],
    MIMode: mi_mode,
  };
  if (mi_mode == 'gdb') {
      if(this._kitManager.activeKit!.type == "compilerKit") {
        const kit: CompilerKit = this._kitManager.activeKit as CompilerKit;
        const compilers: {[lang: string]: string} = kit.compilers;
        // \todo error handling for missing "C" compiler
        const gdb = compilers["C"].replace("gcc", "gdb");
        // \todo find gdb path is a user transparent way
        debug_config["miDebuggerPath"] = gdb;
      } else {
        // \todo error handling and handling for toolchain files
      }
...
```

KoeMai:
[...]

Issue #242, opened 2017-10-26
Debug doesn't set MIDebuggerPath


jovere:
The CMake Tools plugin works for the most part after selecting the environment (the "Working in..." part in bottom right).  It correctly builds everything when I select MinGW.  However, if I try to select Debug, using the default CMake Tools debugging options, I get the following message:
```Unable to start debugging. Launch options string provided by the project system is invalid. Unable to determine path to debugger. Please specify the "MIDebuggerPath" option.```
If the MIDebuggerPath was set to the environment (ie. the `activeEnvironments` in `.cmaketools.json` plus the expected debugger, (ie. `bin/gdb`), manual creation of the `launch.json` would not be required.

For example, my present MinGW environment is as follows in `.cmaketools.json`:
```
  "activeEnvironments": [
    "MinGW - C:/Program Files (x86)/mingw-w64/i686-7.2.0-posix-dwarf-rt_v5-rev0/mingw32"
  ],
```

It would be nice if the CMake Tools was smart enough to set `MIDebuggerPath` to:
```
C:/Program Files (x86)/mingw-w64/i686-7.2.0-posix-dwarf-rt_v5-rev0/mingw32/bin/gdb
```

**Comment by FelikZ** [comment]
Try this:

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch CMake-built program",
            "type": "cppdbg",
            "request": "launch",
            "program": "${command:cmake.debugTarget}",
            "stopAtEntry": true,
            "cwd": "${workspaceRoot}"
        }
    ]
}
```

KoeMai:
This should work in the next version. the gdb  path is generated out of the c compiler/ cxx compiler from cmake cache.

If it does not work like expected then reopen it.

PR #420, opened 2018-05-07
Bugfix miDebuggerPath problem

KoeMai:
## This change addresses item #375 and #242

This refactoring extracts the logic to generate
debugconfigurations from the debug command into a new module.
This new module generates debugConfigurations out of
the cmakecache and support now the generation of miDebuggerPath.
This module is reintegrated into the debug command.


src/debugger.ts
```diff
@@ -0,0 +1,105 @@
+import {CMakeCache} from '@cmt/cache';
+
+
+export namespace Debugger {
```
vector-of-bool: Namespaces are a strange beast. I used them in `vscode-gitflow`, but I didn't find them very useful and just made things cluttered. I'd prefer to just use regular imports instead.

KoeMai:
I removed the not required namespace.
Additional I restore the apple gcc support on mac, but I am not able to try it out (I have no mac).

Do we need some extension tests or integration tests?

KoeMai:
I have no idea how I can make travis happy again. :/

vector-of-bool:
Merge if you are ready.

KoeMai:
I made this modification on windows, there it works. For linux I need to add a fallback scenarios. 
- Clang and gcc-7 does not work, lldb is there a separate package (clang->lldb->gdb)
- gcc-7  has know not it own gdb7 it use the gdb (gcc-7->gcc-7->gdb). 

The integration test does not start the real debugger, thats way travis and appveyor are happy.

KoeMai:
should we add a Setting to set Thema fallback debugger? or should we add a debugger to the kits?

PR #422, opened 2018-05-10
Add search compiler in mingw dir

KoeMai:
### This changes behavior

- Add generic approach to add additional search path for kit scan

- [ ] Unit tests
- [ ] Integration tests


vector-of-bool:
I believe I know what is happening on AppVeyor:

The kit scan is finding MinGW and adding it to the kit list, even though you have an empty kits file. This causes the error to show up as a QuickPick asking you to pick a kit when it should just error out as the test expects. The timeout is the QuickPick waiting for user input that will never happen.

PR #427, opened 2018-05-13
Extension Manager Refactors

vector-of-bool:
This is a pretty huge set of refactors that are going to support a few upcoming features, additions, and other changes:

1. All the "interesting work" and "business logic" is being ripped out of mutable global state.
2. None of the "interesting work" interacts with the VSCode UI.
3. Multiple instances of the "interesting work" classes can be created simultaneously for different directories: Even directories not open in VSCode.
    - This will support greater test flexibility. No need to open a new VSCode directory for every test.
    - Multi-root support as each workspace directory can be treated and worked with separately.
    - Nothing assumes that `vscode.workspace.workspaceRoot` is the actual workspace.
4. The `KitManager` class is dead. Long live `ExtensionManager`.

I know @ytimenkov doesn't like `FooManager` classes, and I'm inclined to agree. `ExtensionManager` doesn't really improve on the downsides of `Manager` classes, but I want this one to be _different_:

1. It sits between the VSCode command API and the real "interesting work."
2. It controls and switches between CMake Tools instances open for each workspace folder.
3. It will (in a future PR) control the CMake Tools status bar elements.
4. All "Manager" functionality will eventually be merged into `ExtensionManager`. This PR starts with Kits.
5. It will never be doing any "interesting work:" Only act as an intermediary between the VSCode UI and the real extension instance(s). For this purpose it's code must never be an area of interest: Its methods should be so dumb and simple that observation with the naked eye can assure its correctness: No testing should ever be _required_. The comment density should be extremely high.

@ytimenkov I think you'll find this PR particularly interesting.


src/extension.ts
```diff
@@ -21,29 +22,442 @@ import CMakeTools from './cmake-tools';
...
+ * A class to manage the extension.
+ *
+ * Yeah, yeah. It's another "Manager", but this is to be the only one.
+ *
+ * This is the true "singleton" of the extension. It acts as the glue between
+ * the lower layers and the VSCode UX. When a user presses a button to
+ * necessitate user input, this class acts as intermediary and will send
+ * important information down to the lower layers.
+ */
+class ExtensionManager implements vscode.Disposable {
```
KoeMai: I think this class is responsible for three issues.

-  Workspace coordinator
   It monitors changes on the workspace and forwards commands to the workspace relatee cmake instances. 

- cmaketool instances management
 Each workspace has a direct relation to one cmaketool instance. This means the Information and objects are tidy coupled to each other. I would separate them into a class which is used by the workspacecoordinator (ExtrnsionManager) 

- Kit file management
  The kit file management is a big topic and it has two levels user related kits and workspace related kits. The scan part and the collection of user kits is a user related task and is independent from the workspace. I would separate it from the workspace coordination. The workspace use only user kits by the select kit commands. But the workspace oder cmaketools instance need to manage the actual kit. I would forward the kit selection to the cmaketools. this would reduce the responsabilities for the extension manager.

vector-of-bool:
Seeing as 1.0.0 is so late, I'm moving this to post-1.0

vector-of-bool:
I've marked Clang for Windows as an invalid kit (unless found via `clang-cl`). CMake has no support for Clang targeting MSVC using the GNU CLI.

vector-of-bool:
@KoeMai I'm feeling ready to merge.

With the death of KitManager, do you want to look into simplifying the configure-and-build tests? Now that the QuickPick is out of the way, you should be able to insert kits into CMakeTools instances without any mocking. This should speed up tests and make them much easier to understand.

This also removes the need of kit scanning. Kits can be generated by hand and passed in to CMakeTools via `setKit`.

PR #429, opened 2018-05-13
Bugfix miDebuggerPath problem with fallback behavior

KoeMai:
## This change addresses item #375 and #242

Same as  #420.

Additional it has a test debugger command mechanism and a fallback to gdb in search path. 

Fallback behavior
```
[proc] Executing command: /usr/bin/lldb-4.0 --version
[proc] Executing command: /usr/bin/gdb-4.0 --version
[proc] Executing command: gdb --version
```

Removed debugger integration into kit file.


vector-of-bool:
Can #420 be closed?

KoeMai:
The testing fails on windows, it needs the extension cpptools for the cppvdbg debug launch.
Do we want this kind of extension test?

vector-of-bool:
Hmm... what about separating the generation of the `DebugConfiguration` and testing the generation of that config without launching a debugger?

Issue #426, opened 2018-05-12
Preferred generators are ignored


ghuser404:
Fields listed below are ignored. The only way I found to specify a generator is via `cmake.generator` in `settings.json`.

Field `preferredGenerator` in `cmake-kits.json` is ignored.
Field `cmake.preferredGenerators` is `settings.json` is ignored.

vector-of-bool:
Can I get some more information on what generators you are trying to use? What platform are you one? What CMake version are you using?

ghuser404:
Sure.

Let's say "Ninja" generator: `"preferredGenerator": { "name": "Ninja" }` in `cmake-kits.json`
I'm on Windows 10 64-bit with latest updates.
CMake version 3.9.3.

vector-of-bool:
There's a priority the CMake Tools will decide on a generator:

1. The `cmake.generator` setting.
2. The `cmake.preferredGenerators` setting
3. The `preferredGenerator` on the active kit
4. If nothing, an error occurs.

What do you see from CMake Tools when you try to configure? Is it failing to choose a generator? Is it choosing a generator that you don't want? Can you show the debug log from when you try to configure and it fails?

ghuser404:
It just selects whatever it wants. When I didn't use `cmake.generator`, it used to select the Visual Studio one. After I used `cmake.generator` with `Ninja` (and then removed), it now defaults to `Ninja`.

For example:
> [driver] Switching to kit: Visual Studio Community 2017 - x86
> [cms-client] Configuring using the "Ninja" CMake generator

vector-of-bool:
@KoeMai has any of your generator work affected this issue?

KoeMai:
my unit test for the 'not configured yet'  pr is using different generators. but there was a race condition. I think if PR is mainline then it should work.

zenakuten:
I'm also having this problem.  I'm on Windows 10 x64.  The generator option is ignored in all situations for me.  I've tried setting preferredGenerator in cmake-tools.json, cmake.preferredGenerator in settings.json, and cmake.generator in settings.json.  All are ignored.  

When executing cmake to configure, the -G option is never passed.

vector-of-bool:
Sorry for the delay on this issue. I've been busy with a few other projects.

Once #437 is in a release, I'll check back on the state of this issue.

LU15W1R7H:
[...]

Issue #415, opened 2018-05-05
Mystery Error: This instance was not yet configured


vector-of-bool:
I'm getting a huge number of Rollbar issue reports for "This instance was not yet configured" from CMake Server. This issue appears when you call `cmakeInputs` before properly configuring the project.

I've been completely unable to reproduce the issue and can't see in the code any obvious reason why it would occur. No user has reported the issue, as it seems that it's occurrence doesn't actually block users.

If anyone reading this is seeing this error message, please comment below! (**Edit:** Please also include information about your platform, CMake version, OS, etc.)

DoddzyCodes:
Had this problem just pop up.

I only noticed it start happening after enabling the C/C++ extension.  I had already configured and built the project.

After installing the C/C++ extension and reloading, the instance not configured error would pop up on almost any cmake-server operation. Disabling the C/C++ extension stopped the error showing up, and weirdly enough, I haven't been able to reproduce the error after re-enabling it.

Will let you know if I can find something a bit more reproducible.

vector-of-bool:
@DoddzyCodes, what CMake version and OS are you running?

DoddzyCodes:
@vector-of-bool Ubuntu 18.04, Cmake 3.10.2

KoeMai:
I think there error is created by switch between kit files.
I test at the moment the new concept for the debugger integration, I get the error often when I switch between MinGw and Visual studio.

vector-of-bool:
@KoeMai could you find a way to reliably reproduce the issue? I want to attach a debugger to CMake at the same time to see what is going on

magpier84:
Moreover, following log line appear in some cases:

[rollbar] Unhandled exception: Unhandled Promise rejection: build [cmake-server] This instance was not yet configured. {}

chetferry:
MacOS 10.13.4
make 3.11.
arm-eabi-none-gcc compiler

Getting the same errors but only intermittently.  About 1 our of 3 times that I build.

johnf9896:
I get the same error after running CMake: Build. The same log line appears.

Windows 10 Home
CMake 3.11.2

magpier84:
Unfortunately this popup + error message about already running cmake during launching app made me switch to another IDE for some time. Now both errors are not reproduced :(

vector-of-bool:
@magpier84 Okay. Does the other error message need another GitHub issue?

magpier84:
I will try to reproduce it.

[...]

PR #437, opened 2018-05-25
Bugfix/not configured yet

KoeMai:
## This change addresses item #426, #415 

Fix some problem. But the tests are not ready for CI.


src/cms-driver.ts
```diff
@@ -104,13 +107,13 @@ export class CMakeServerClientDriver extends CMakeDriver {
   async doPreBuild(): Promise<boolean> { return true; }
 
   async doPostBuild(): Promise<boolean> {
-    await this._refreshPostConfigure();
+    await this._refreshPostConfigure(); // <-- 2. Call this one
```
vector-of-bool: Remove these comments.

src/cms-driver.ts
```diff
@@ -197,15 +201,22 @@ export class CMakeServerClientDriver extends CMakeDriver {
 
   get cmakeCacheEntries(): Map<string, CacheEntryProperties> { return this._cacheEntries; }
 
+
   async doSetKit(need_clean: boolean, cb: () => Promise<void>): Promise<void> {
+    this._clientChangeInProgress = new Promise(async resolve => {
```
vector-of-bool: This is dangerous (the promise can go unresolved). Here's my suggested refactor:

1. Move the inner body of the promise callback to a separate private function (Name something like `_setKitAndRestart`).
2. Have `doSetKit()` simply do `return this._clientChangeInProgress = this._setKitAndRestart(needClean, cb);`

The `_clientChangeInProgress` will be immediately set to the `Promise` for the inner function, and it will properly handle exceptions and resolve itself when `_setKitAndRestart` returns.

src/driver.ts
```diff
@@ -551,13 +554,18 @@ Please install or configure a preferred generator, or update settings.json or yo
     const expanded_flags = await Promise.all(expanded_flags_promises);
     log.trace('CMake flags are', JSON.stringify(expanded_flags));
 
+    // This resets for the cms-driver the client
+    // clears the configuration of cmake server
+    await this._refreshExpansions();
```
vector-of-bool: Could you put a better explanation of why this call was moved here? I don't quite understand the purpose of it.

src/cms-driver.ts
KoeMai: in this case I had the problem in the execution order the promises. the used promise is used to block handling an driver call until client change is done.

I will refactor it later.

src/driver.ts
KoeMai: Somethings I should go to bed before commits. 
The 'not configure yet' throws an error after restart of cmake client. 
The refreshexpansions triggers such a restart. If I move this line in front of the configure step then the configure step use the same expansions like the build step. It ensures the configure/codemodel/cmakeInputs are present in the build step and consistent. 

Does this descript help more? I would remove the sentence.

KoeMai: Done

src/cms-driver.ts
KoeMai: Done

src/cms-driver.ts
KoeMai: Done

PR #439, opened 2018-05-28
Feature/build on launch

KoeMai:
### This changes behavior of launch and debug steps

- Add new configuration option `buildBeforeRun`

## The purpose of this change

This modification allow to configure the launchTargetPath, lanuchTarget or debugTarget to
build the target before execution.

## Open steps

- [x] Add option to documentation


src/cmake-tools.ts
```diff
@@ -758,7 +758,43 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
       log.warning('No program will be executed');
       return null;
     }
-    return chosen.path;
+    return executable.path;
+  }
+
+  async PrepareLaunchTargetExecutable(): Promise<api.ExecutableTarget|null> {
```
vector-of-bool: Lowercase `prepare`

src/cmake-tools.ts
```diff
@@ -793,19 +829,9 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
-        log.debug('Configuration of project failed.');
-        return null;
-      }
-    }
-    const target = await this.getOrSelectLaunchTarget();
-    if (!target) {
-      // The user has nothing selected and cancelled the prompt to select a target.
-      log.debug('No target selected.');
+
+    const targetExecutable = await this.PrepareLaunchTargetExecutable();
```
vector-of-bool: Compile error because this var was renamed.

src/cmake-tools.ts
KoeMai: Done

src/cmake-tools.ts
KoeMai: Done

PR #444, opened 2018-06-02
Bugfix/variants from settings

KoeMai:

- Default variants are used from configuartion settings
- Configuration changes for `defaultVariants` automatic refresh variants

- [x] Update of package.json needed. The style of schema is changed, the package default use the old one.


vector-of-bool:
Something's up with the CI branching stuff. I'll merge this since it is working in the one case.

Issue #407, opened 2018-04-28
Feature Request: Option to copy compile_commands.json to static path


DWagenk:
### Brief Summary

Hello! It would be great, to have an additional entry in `settings.json`, to define a static path where the current `compile_commands.json` should be copied/symlinked to. 

### Expected

The current `compile_commands.json` file is always accessible via a static path.

### Background

I'm using a per-configuration build directory like proposed here: https://github.com/vector-of-bool/vscode-cmake-tools/issues/195#issuecomment-311292876 
This leads to multiple `compile_commands.json` files. Since vscode-cmake-tools knows the active configuration it knows the currently valid `compile_commands.json` file. Other extensions, e.g. ms-vscode.cpptools or cquery-project.cquery don't and rely on a static path to the valid file. Having vscode-cmake-tools copy or symlink the currently valid file to a configurable location, for other extensions to pick it up, would be great.

Please consider adding this functionality! Thanks!

vector-of-bool:
Useful, and easy enough to make it in 1.0.0

vector-of-bool:
Ready in 1.0.1

PR #453, opened 2018-06-07
Support copying of compile_commands.json, and fix successful-build tests

vector-of-bool:
Implements #407


test/extension-tests/successful-build/test/configure-and-build.test.ts
```diff
@@ -107,103 +113,125 @@ suite('Build', async () => {
...
+           linux: [
+             {kitLabel: /^GCC/, compiler: 'GNU GCC/G++'},
+             {kitLabel: /^Clang/, compiler: 'Clang/LLVM'},
+           ],
+           win32: [
+             {kitLabel: /^GCC/, compiler: 'GNU GCC/G++'},
+             {kitLabel: /^VisualStudio/, compiler: 'Microsoft Visual Studio'}
+           ]
+         };
+         this.skip();  // XXX: Fix race between setting kit and building
```
KoeMai: should the test be disabled?

vector-of-bool: This test seems unreliable, at least on my system.

vector-of-bool:
@KoeMai it sems that me changing the kit label pattern has caused test failures on WIndows. I'm not sure what's going on here. Do you have any ideas?

test/extension-tests/successful-build/test/configure-and-build.test.ts
vector-of-bool: I am fine merging these disabled tests, although I'd prefer not to. I want to address the way kits are set in my other PR for 1.x.

PR #468, opened 2018-06-27
Project Outline + UI Tweaks

vector-of-bool:
## Awesome News!

This PR adds a long-desired feature: A project outline.

This adds a viewContainer to explorers that shows the outline of a CMake project. It lists all targets, and has several buttons and context menus to perform quick actions on individual targets.

In addition, progress notifications are now used in a variety of places. The ad-hoc progress bar in the status bar is no longer present: instead using a real progress notification.


vector-of-bool:
This _will_ require several tweaks and cleanups before finally releasing. I just want to get this groundwork laid down before proceeding on some other 1.1 changes.

PR #474, opened 2018-07-07
vscode-cpptools CustomConfigurationProvider

vector-of-bool:
## At long last!

This PR introduces the long-`await`ed feature to automatically configure vscode-cpptools with information from CMake.

This change uses the new cpptools extension API to hook into that extension and give it information about the project's source files.

@bobbrow would you take a look and remind me if I've forgotten anything important?

### DIFF
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -4,6 +4,7 @@
 
 import {CMakeCache} from '@cmt/cache';
 import * as cms from '@cmt/cms-client';
+import * as util from '@cmt/util';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import * as cpt from 'vscode-cpptools';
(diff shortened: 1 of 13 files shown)


src/cpptools.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * The actual class that provides information to the cpptools extension. See
+ * the `CustomConfigurationProvider` interface for information on how this class
+ * should be used.
+ */
+export class CppConfigurationProvider implements cpt.CustomConfigurationProvider {
+  /** Our name visible to cpptools */
+  readonly name = 'CMake Tools';
+  /** Out extension ID, visible to cpptools */
```
bobbrow: typo: `Out` -> `Our`

src/cpptools.ts
```diff
@@ -0,0 +1,139 @@
...
+    const lang = grp.language || 'CXX';
+    // Try the group's language's compiler, then the C++ compiler, then the C compiler.
+    const comp_cache = opts.cache.get(`CMAKE_${lang}_COMPILER`) || opts.cache.get('CMAKE_CXX_COMPILER')
+        || opts.cache.get('CMAKE_C_COMPILER');
+    // Try to get the path to the compiler we want to use
+    const comp_path = comp_cache ? comp_cache.as<string>() : opts.clCompilerPath;
+    if (!comp_path) {
+      rollbar.error('Unable to automatically determine compiler', {lang, fileGroup: grp});
+    }
+    const is_msvc = comp_path && (path.basename(comp_path).toLocaleLowerCase() === 'cl.exe');
```
bobbrow: **note:** clang on Windows will prefer msvc mode if a Windows SDK is installed.  In cpptools, we fix this for you automatically if `c_cpp_properties.json` is misconfigured, but I didn't test this for custom configs.  I probably should...

src/cpptools.ts
```diff
@@ -0,0 +1,139 @@
...
+    return {
+      defines: grp.defines || [],
+      includePath: (grp.includePath || []).map(p => p.path),
+      intelliSenseMode: is_msvc ? 'msvc-x64' : 'clang-x64',
+      standard: 'c++17',  // TODO: Switch on correct standard
```
bobbrow: Do you plan to set this for your upcoming release?  In my fork I believe I wrote some code that parsed through the command line to figure this out.  Or does CMake make this information available to you in some other way?

src/cpptools.ts
```diff
@@ -0,0 +1,139 @@
...
+      compilerPath: comp_path || undefined,
+    };
+  }
+
+  /**
+   * Update the configuration index for the files in the given file group
+   * @param sourceDir The source directory where the file group was defined. Used to resvolve
```
bobbrow: typo: `resvolve` -> `resolve`

src/extension.ts
```diff
@@ -120,6 +119,10 @@ class ExtensionManager implements vscode.Disposable {
   private readonly _projectOutlineDisposer
       = vscode.window.registerTreeDataProvider('cmake.outline', this._projectOutlineProvider);
 
+  private readonly _configProvider = new CppConfigurationProvider();
+  private _cppToolsAPI?: cpt.CppToolsApi;
```
bobbrow: This object is disposable but I don't see it being disposed anywhere or added to the extension context's "subscriptions".

src/cpptools.ts
vector-of-bool: Should be in the final release. Probably in this PR when I do some work to finish it.

src/extension.ts
vector-of-bool: I missed that one.

src/cpptools.ts
vector-of-bool: Done. Check `cpptools.ts`. I think I've covered 95% of use cases.

PR #485, opened 2018-07-22
Feature/kit variable

apples:
### This changes expansion variables, documentation

- Add a new expansion variable `${buildKit}`, which is the name of the current kit.
- If switching kits changes the binary directory, do not clean when switching kits.

### DIFF
--- a/src/driver.ts
+++ b/src/driver.ts
@@ -192,7 +192,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
       workspaceRootFolderName: path.basename(ws_root),
       generator: this.generatorName || 'null',
       userHome: user_dir,
-      buildKit: this._kit!.name,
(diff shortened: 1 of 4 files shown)


src/driver.ts
```diff
@@ -192,6 +192,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
       workspaceRootFolderName: path.basename(ws_root),
       generator: this.generatorName || 'null',
       userHome: user_dir,
+      buildKit: this._kit!.name,
```
vector-of-bool: Will have to look closer here that the non-`null` assertion won't cause any trouble.

apples:
I don't think it's possible for `_kit` to ever be null or undefined, but I changed it to a conditional just in case.

vector-of-bool:
Sorry to go dark. I was busy with another project for a few weeks. I'll get this in 1.1.2.

PR #486, opened 2018-07-23
Close mingw kit problems

KoeMai:

- Add generation of MinGw preferred generator for windows MinGw
- Add missing environment variables for windows MinGw
- Handle MinGW like a cross compiler for linux


src/kit.ts
```diff
@@ -627,26 +651,8 @@ export interface KitScanOptions {
...
-    let env_elems = env_path.split(sep);
-    if (env_elems.length === 1 && env_elems[0] === '') {
-      env_elems = [];
-    }
-    scan_dirs = env_elems;
-  }
-  if (opt.minGWSearchDirs) {
-    scan_dirs = scan_dirs.concat(convertMingwDirsToSearchPaths(opt.minGWSearchDirs));
-  }
+  const kit_options = opt ? opt : {};
```
vector-of-bool: Can simplify to `opt = opt || {}`, without a new local var.

src/kit.ts
```diff
@@ -663,13 +669,17 @@ export async function scanForKits(opt?: KitScanOptions) {
       scanPaths = scanPaths.concat(pathvar.split(sep));
     }
 
-    if (scanPaths) {
+    if (kit_options) {
```
vector-of-bool: I don't think this branch will even be `false`, since `kit_options` is always a valid object

KoeMai:
The modifikation is done from my side. I have updated the sourcecode according to your review comments.

src/kit.ts
```diff
@@ -174,22 +176,44 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
...
+          let make_version_output = execMake.stdout;
+          if (make_version_output.length == 0)
+            make_version_output = execMake.stderr;
+          const output_line_sep = make_version_output.trim().split('\n');
+          const isMake = output_line_sep[0].includes('Make');
+          const isMingwTool = output_line_sep[1].includes('mingw32');
+
+          if (isMake && isMingwTool) {
+            gccKit.preferredGenerator = {name: 'MinGW Makefiles'};
+            gccKit.environmentVariables = {PATH: '${env.PATH};' + binParentPath};
```
vector-of-bool: I don't think env var expansion works in this context.

KoeMai: done

src/kit.ts
```diff
@@ -174,22 +176,44 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
...
+    const isWin32 = process.platform === 'win32';
+    if (isWin32 && bin.toLowerCase().includes('mingw')) {
+      const binParentPath = path.dirname(bin);
+      const mingwMakePath = path.join(binParentPath, 'mingw32-make.exe');
+      if (await fs.exists(mingwMakePath)) {
+
+        // Check for working mingw32-make
+        const execMake = await proc.execute(mingwMakePath, ['-v'], null, {environment: {PATH: binParentPath}}).result;
+        if (execMake.retc != 0) {
+          log.debug('Bad mingw32-make binary ("-v" returns non-zero)', bin);
```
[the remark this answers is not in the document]
Randshot: Hm, I don't know how the `log.debug()` command exactly works, but I would imagine that it either inserts the arguments after each other with or without inserting a space between. If it does the former one, it doesn't really matter, but if it is the second one, then it makes the log entry pretty much unreadable.

Randshot:
Is this one ready to merge, or still WIP?

src/kit.ts
```diff
@@ -174,22 +176,44 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
...
+          let make_version_output = execMake.stdout;
+          if (make_version_output.length == 0)
+            make_version_output = execMake.stderr;
+          const output_line_sep = make_version_output.trim().split('\n');
+          const isMake = output_line_sep[0].includes('Make');
+          const isMingwTool = output_line_sep[1].includes('mingw32');
+
+          if (isMake && isMingwTool) {
+            gccKit.preferredGenerator = {name: 'MinGW Makefiles'};
+            gccKit.environmentVariables = {PATH: binParentPath};
```
vector-of-bool: We'll still want to append to the regular `%PATH%`. This will just replace it with the single MinGW path and create a lot of trouble.

vector-of-bool:
Sorry, I let this one sit a bit. Another comment.

lygstate:
Can be closed  now

vector-of-bool:
This was superseded in #596

PR #496, opened 2018-07-26
More general variant substitution

Randshot:
### This changes visible behaviour and documentation

- Variant option substitution only accepts `${variant:VARIANTNAME}` syntax.
- The name of the currently active choice of the provided `VARIANTNAME` variant option will be replaced.
- If the variant option is undefined, an empty string will be
expanded instead.

## The purpose of this change

The previous implementation only added a special case for `buildType` collision. With that I mean, that if the user defined a variant also named `buildType`, the driver would add the replacement key as `buildLabel`, since `buildType` was already taken.
While this behaviour is correct, the implementation didn't account for other possible name collision that could occur, leading to possible overriding of the replacement value for the colliding replacement key.

The implementation in this PR tries to fix that, by providing a syntax similar to the `${env:VAR}` and `${command:COMMAND}` one.

**Feedback for improvements are welcome!**


vector-of-bool:
This gets me thinking about maybe just dropping the short-form completely, and maybe using `variant:` as the replacement prefix instead of `variant_`. That would match the special-syntax for `command:` and `env:`. What do you think? It wouldn't be a ton of work for users to update their configs with the new syntax, and there would be minimal risk of naming collision.

Randshot:
That is a viable option to consider.

Randshot:
@vector-of-bool I have implemented your suggested syntax.

Issue #518, opened 2018-08-22
output window code page 


longlongwaytogo:
When i used vscode-cmake-tools to build project, the output text  defualt with utf8 code ; but how i change it to GBK(chinese charset)

![error](https://user-images.githubusercontent.com/16498273/44440127-63868480-a5f9-11e8-94ef-72907cc0371e.png)

andreeis:
Closing this issue. Please reopen if the merge that is referenced did not help with your case.

Issue #267, opened 2017-11-29
Output and terminal windows charset problem


xzzh999:
Hi, bigcow! Great work! 

When i build msvc c++ project, the output window always print garbled, 
i tried follow issue's suggestion but they were complicated and didn't work.

https://github.com/Microsoft/vscode/issues/3550

https://github.com/OmniSharp/omnisharp-vscode/issues/1775

can you add some charset encoding settings to make this easy?

![](https://github.com/xzzh999/markdown-pictures/blob/master/vscode-wrong-charset.png?raw=true)

xzzh999:
vscode version: 1.18.1, shell: 1.7.9, node: 7.9.0

vector-of-bool:
I've had a few troubles with character encoding. This is on the backlog.

feng-ye:
I am using the following work around:
1. create a bat file contains
`chcp 65001 && cmake %*`
2. set cmake.cmakePath to the bat file

I hope this will help.

xzzh999:
For now i wrote some python scripts to generate and build the project myself, and every charset encoding is under control in vscode, but i hope this extension will improved later, then  i will change back.

EternalPhane:
Maybe you can temporary patch it by adding an option in `settings.json` to let user manually set their preferred output encoding, and provide a permanent solution later.

pengbins:
> I am using the following work around:
> 
> 1. create a bat file containing
>    `chcp 65001 && cmake %*`
> 2. set cmake.cmakePath to the bat file
> 
> I hope this will help.
@feng-ye  I tried your work around, but got following error. It seems .bat can't run.
[rollbar] Unhandled exception: Unhandled Promise rejection: build Error: Impossible: CMake driver died immediately after successful configure {}

bobbrow:
This looks like a duplicate of #153 which may have been fixed by PR #550. Is this still an issue for you?

Issue #245, opened 2017-10-31
Chinese is not displayed correctly


tim37021:
I use visual studio 2017. The build output does not display correctly.

![image](https://user-images.githubusercontent.com/5507113/32241500-40fb519c-beab-11e7-8e7e-b63b7e6154a2.png)

However, if I use task.json to build it works perfectly
![image](https://user-images.githubusercontent.com/5507113/32241517-55239d1e-beab-11e7-9e31-f3ec855c9f64.png)

xdray:
I get same question

vector-of-bool:
@Randshot Look familiar? I bet its the same output code page issue. I want to move the build logs to an integrated terminal, and then maybe this issue would be easy to fix.

If not that, it may be enough to use the same technique for launching `vswhere` as here when running `--build`?

Randshot:
It certainly looks familiar. Moving to the integrated terminal may fix this issue, assuming that it doesn't use code pages.

bobbrow:
This looks like a duplicate of #153 which may have been fixed by PR #550.  Is this still an issue for you?

Issue #153, opened 2017-05-20
output message is garbled


qhuman:
when I build a project, output message like this. I use vs2012 compiler
[vscode] Executing command: cmake --build d:/vscode_test/hello/build --target ALL_BUILD --config Debug -- /m /property:GenerateFullPaths=true
Microsoft(R) ��������汾 4.7.2046.0
[Microsoft .NET Framework �汾 4.0.30319.42000]
��Ȩ���� (C) Microsoft Corporation����������Ȩ����

������ʱ��Ϊ 2017/5/20 16:59:57��
     1>�ڵ� 1 �ϵ���Ŀ��D:\vscode_test\hello\build\ALL_BUILD.vcxproj��(Ĭ��Ŀ��)��
     1>��Ŀ��D:\vscode_test\hello\build\ALL_BUILD.vcxproj��(1)���ڽڵ� 1 �����ɡ�D:\vscode_test\hello\build\ZERO_CHECK.vcxproj��(2) (Ĭ��Ŀ��)��
     2>InitializeBuildStatus:
         ���ڴ�����Win32\Debug\ZERO_CHECK\ZERO_CHECK.unsuccessfulbuild������Ϊ��ָ����AlwaysCreate����
       CustomBuild:
         ���������Ϊ���¡�
       FinalizeBuildStatus:
         ����ɾ���ļ���Win32\Debug\ZERO_CHECK\ZERO_CHECK.unsuccessfulbuild����
         ���ڶԡ�Win32\Debug\ZERO_CHECK\ZERO_CHECK.lastbuildstate��ִ�� Touch ����
     2>�����������Ŀ��D:\vscode_test\hello\build\ZERO_CHECK.vcxproj��(Ĭ��Ŀ��)�Ĳ�����
     1>��Ŀ��D:\vscode_test\hello\build\ALL_BUILD.vcxproj��(1)���ڽڵ� 2 �����ɡ�D:\vscode_test\hello\build\hello.vcxproj��(3) (Ĭ��Ŀ��)��
     3>InitializeBuildStatus:
         ���ڴ�����hello.dir\Debug\hello.unsuccessfulbuild������Ϊ��ָ����AlwaysCreate����
       CustomBuild:
         ���������Ϊ���¡�
       ClCompile:
         ���������Ϊ���¡�
       Link:
         ���������Ϊ���¡�
         hello.vcxproj -> D:\vscode_test\hello\build\Debug\hello.exe
       FinalizeBuildStatus:
[...]

PR #550, opened 2018-09-27
Add config "outputLogEncoding"

pengbins:
This change try to fix a codepage problem mentioned in some issues, such as  #518 , #267 ,  #245 and #153. 
The output messages of cmake build command are garbled when  the default system codepage is not utf8.

It is fixed by adding a user config item "outputEncoding" to set the character encoding of messages in the output console.  

- How to use
    * Set "outputEncoding" to your system default codepage. (Ctrl+,  and search outputEncoding)
image
    * The default value is "utf8".  For Chinese user, set it to "GBK".  
    * Supported value list can be found [here](https://github.com/ashtuchkin/iconv-lite#supported-encodings) .
- Results
    * Output console used to be unreadable when the default system codepage is not utf8.
image
    * Output console is readable after set "outputEncoding" to  GBK.
image

- How it works
This change introduces a new dependency   [iconv-lite](https://github.com/ashtuchkin/iconv-lite).
The console programs emit text encoded with the default system codepage. This change uses iconv-lite to decode those text to utf8 , so they can be  displayed correctly on output console. To decode,  iconv-lite
needs to know the default system codepage which is given by  user's config "outputEncoding".


vector-of-bool:
This change is excellent!!

The output codepage issues has been persistently troublesome almost from the very beginning. I've been unable to test and was never sure how to approach this problem.

One think I am wondering is if there is the possibility to auto-detect the code page. I found that Windows exposes a `_CODEPAGE` environment variable for the current codepage number. I've found a table of these numbers to their names, but I don't think the names that microsoft provides will work with `iconv-lite`.

Instead of trying to get them all down immediately, could you add a lookup that uses the `_CODEPAGE` value to default to your `GBK` when `_CODEPAGE` is the value you have on your system? As users report this issue I can add entries to this table so that people won't have to manually change their encoding.

vector-of-bool:
Alternatively, you could try using this table I made from codepage number to names. Maybe iconv just supports it out-of-the box? That would be awesome, and changing the output encoding setting wouldn't ever be required.

[code-pages.ts.txt](https://github.com/vector-of-bool/vscode-cmake-tools/files/2453670/code-pages.ts.txt)

pengbins:
> **Last question:** Can you verify that the auto-detection works on your system?
Yes, it works on my system. But I didn't test on other language system except mine.
My code page number is 936, encoding name is gb2312.

nim:
Problem is still here

Randshot:
@nim Can you please open a separate issue for this. It's easier to track that way :)

nim:
https://github.com/Microsoft/vscode/issues/72270

PR #590, opened 2018-11-23
Added option to show the log when building.

Som1Lse:

- The default has been changed to showing the log, but **not** focusing on it.
- Focusing on the log is still an option.
- It is also possible to do nothing.
- A potentially nice feature, but as of now not implemented, would be to only show the log when a warning/error/etc. occurs.

## Other Notes/Information

The option naming is not ideal, as this was hacked together in 5 minutes. Feel free to change it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -891,6 +891,15 @@
             "fatal"
           ],
           "scope": "window"
+        },
+        "cmake.revealLog": {
+          "type": "string",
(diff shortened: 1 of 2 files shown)


vector-of-bool:
Looks good! I may tweak some naming, but it's otherwise a fine addition.

Thanks!

Issue #528, opened 2018-09-01
cmake.copyCompileCommands didn't work correctly;


skatebill:
### Brief Issue Summary

cmake.copyCompileCommands works incorrectly;

### Expected:

1. Set "cmake.copyCompileCommands" to "${workspaceRoot}/compile_commands.json"
2. compile project
3. compile_commands.json copied to the WorkspaceRootFolder 

### Apparent Behavior:

the compile_commands.json didn't appear at WorkspaceRootFolder 
instead it is copied to 
[[AppData\Local\Programs\Microsoft VS Code\\${workspaceRoot}\compile_commands.json]]
yes the Foldername is "${workspaceRoot}"
![image](https://user-images.githubusercontent.com/4061585/44944857-b38ef380-ae10-11e8-9d4d-47f06e21c109.png)

### CMake Tools Log

[cmake] Configuring done
[cmake] Generating done

### Platform and Versions

- **Operating System**: [[Windows]]
- **CMake Version**: [[3.12.1]]
- **VSCode Version**: [[1.26.1]]
- **CMake Tools Extension Version**: [[1.1.1]]
- **Compiler/Toolchain**: [[Clang 8.0.0 for MSVC]]

### Other Notes/Information

I Checked the sourcecode of cmake-tools

I believe the problem is the copyCompileCommands did'n expand
[cmake-tools.ts](https://github.com/vector-of-bool/vscode-cmake-tools/blob/develop/src/cmake-tools.ts)
at line 468
```javascript
const copy_dest = this.workspaceContext.config.copyCompileCommands;
```
without calling the expand.expandString function like
![image](https://user-images.githubusercontent.com/4061585/44944865-d5887600-ae10-11e8-8cca-036208a72cf8.png)

vector-of-bool:
Sorry for the delay.

That's embarrassing. I'll take a look at it.

zorg93:
This can be closed now

PR #591, opened 2018-11-24
bugfix copyCompileCommands

zorg93:
## This change addresses item #528


vector-of-bool:
Sorry. That lint failure is my fault. Give me a moment.

PR #594, opened 2018-11-25
Upgrade CppToolsApi to use V2.

lygstate:

Implement canProvideBrowseConfiguration and provideBrowseConfiguration
And also use _cachedSourceFileConfiguration(the latest retrieved configuration) when _getConfiguration failed for specifc 
file URI.

## The purpose of this change
Currently,  code complete can not browse external include paths such as Qt and so on.

### DIFF
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -132,6 +132,8 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
    */
   private _lastUpdateSucceeded = true;
 
+  private _workspaceBrowseConfiguration:cpt.WorkspaceBrowseConfiguration = {browsePath:[]};
+
   /**
    * Get the SourceFileConfigurationItem from the index for the given URI
    * @param uri The configuration to get from the index
@@ -153,6 +155,24 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
    * @param uris The file URIs to look up
    */
   async provideConfigurations(uris: vscode.Uri[]) { return util.dropNulls(uris.map(u => this._getConfiguration(u))); }
+
+  /**
+   * A request to determine whether this provider can provide a code browsing configuration for the workspace folder.
+   * @param token (optional) The cancellation token.
+   * @returns 'true' if this provider can provider a code browsing configuration for the workspace folder.
+   */
+  async canProvideBrowseConfiguration() {
+    return true;
+  }
+
+  /**
+   * A request to get the code browsing configuration for the workspace folder.
+   * @returns A [WorkspaceBrowseConfiguration](#WorkspaceBrowseConfiguration) with the information required to
+   * construct the equivalent of `browse.path` from `c_cpp_properties.json`.
+   */
+  async provideBrowseConfiguration() {
+    return this._workspaceBrowseConfiguration;
+  }
 
   /** No-op */
   dispose() {}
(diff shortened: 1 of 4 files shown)


src/cpptools.ts
```diff
@@ -132,13 +132,23 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
    */
   private _lastUpdateSucceeded = true;
 
+  private _workspaceBrowseConfiguration:cpt.WorkspaceBrowseConfiguration = {browsePath:[]};
+
+  private _cachedSourceFileConfiguration: cpt.SourceFileConfigurationItem|undefined;
```
vector-of-bool: You can use a shorthand for `T | undefined`:

```suggestion
  private _cachedSourceFileConfiguration?: cpt.SourceFileConfigurationItem;
```

src/cpptools.ts
```diff
@@ -132,13 +132,23 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
...
   private _getConfiguration(uri: vscode.Uri): cpt.SourceFileConfigurationItem|undefined {
     const norm_path = util.platformNormalizePath(uri.fsPath);
-    return this._fileIndex.get(norm_path);
+    let currrentConfig = this._fileIndex.get(norm_path);
+    if (!currrentConfig) {
+      currrentConfig = this._cachedSourceFileConfiguration;
+    } else {
+      this._cachedSourceFileConfiguration = currrentConfig;
+    }
+    return currrentConfig;
```
vector-of-bool: I'm not feeling confident in this. I feel like returning whatever was the previous configuration will cause a lot of "CMake Tools is doing the configuration wrong" reports.

I'd feel better about this if this showed a notification that the configuration from a different file was being used.

lygstate:
@vector-of-bool This pull request would make vscode-cmake-tools' experience be much better.

src/cpptools.ts
lygstate: How to show a notification about this:) Not familer with notification things.

src/cpptools.ts
lygstate: Nice idea

src/cpptools.ts
lygstate: How about 
```
      vscode.window.showErrorMessage('The path to the compiler for one or more source files was not found in ' +
                                     'the CMake cache. If you are using a toolchain file, this probably means ' +
                                     'that you need to specify the CACHE option when you set your C and/or C++ ' +
                                     'compiler path');
```

vector-of-bool: Look at `showInformationMessage` [in the VSCode API docs](https://code.visualstudio.com/docs/extensionAPI/vscode-api). It's just like `showErrorMessage` but isn't for errors.

You can also see `showInformationMessage` being used in a few places around the codebase.

Here's my suggested wording:

```
The current file is not attached to a CMake target, so CMake Tools is using the configuration information for "`${otherFile}`". Add this source file to a project target for more accurate configuration information.
```

I'd also only show this message in the case that the file they are opening is contained within the current workspace folder. For example, if they open a header file from an external library, this message will not be useful because they will not have the external header in their CMake project.

lygstate: Updated, have a look

lygstate:
@vector-of-bool Have a look, ready to merge

src/cpptools.ts
vector-of-bool: The message should only appear if the file is within their workspace directory, otherwise it will be shown for files that aren't part of their project and for which the message is not helpful.

Thinking about this, I may want to ping some of the cpptools folks on this subject: They have their own logic for inferring the options that should be used with files not controlled by CMake, and adding this may break that.

vector-of-bool: For now, remove this logic and maybe put it in a separate PR. I have another idea for "best guess" on files that are missing that I'd like to try out.

PR #597, opened 2018-11-27
Cache config for file URI.

lygstate:

- 
- 
-

### DIFF
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -143,13 +143,31 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
 
   private _workspaceBrowseConfiguration: cpt.WorkspaceBrowseConfiguration = {browsePath: []};
 
+  /**  Used for cache the source file configuration */
+  private _cachedSourceFileConfiguration?: cpt.SourceFileConfigurationItem;
+
   /**
    * Get the SourceFileConfigurationItem from the index for the given URI
    * @param uri The configuration to get from the index
    */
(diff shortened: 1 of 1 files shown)


lygstate:
@vector-of-bool Can you share your idea about how to deal with the condition when config can not found for the specific file.

vector-of-bool:
@lygstate sorry for the delay.

I had an earlier rendition of a "best guess" on finding the compilation information from `compile_commands.json`. [Here is the important function](https://github.com/vector-of-bool/vscode-cmake-tools/blob/1.0.0/src/compdb.ts#L94).

With that function I could "stem" the file path such that the headers and sources ended up with the same "stem" key.

My new idea involves generating a secondary map of configuration information based on this stem key. If lookup in the primary map fails, it will fall back to this "best guess" map. Since header files will map to the same stem as the corresponding source file, it is very likely that the "best guess" will be correct (if the user has a source file corresponding to the header.) An appropriate diagnostic can be shown recommending that the user add the header file to the source list.

If the secondary map has no configuration for a stemmed file path, then there is no way to reliably determine the configuration information.

lygstate:
@vector-of-bool ping for fixes

lygstate:
@bobbrow That's mainly comes from .inc file and .h files, you propose a good idea, can you suggest a configuration sample? We may create a link in warning notification for automatically create such a configuration file, YAML or JSON

bobbrow:
I'm going to close this PR and create a new feature request for the behavior described. We can continue the discussion there if necessary.

Issue: #1479

Issue #611, opened 2018-12-17
Unable to scan toolkits.


pokowaka:
### Brief Issue Summary

Unable to use under windows 10. Scanning for toolkit fails

### Expected:

Be able to scan for toolkit

### CMake Tools Log

[kit] Error: ENOENT: no such file or directory, open 'C:\Users\XX\AppData\Local\Temp\vs-cmt-0.2646040883366907.bat.env'
[kit] Error: ENOENT: no such file or directory, open 'C:\Users\XX\AppData\Local\Temp\vs-cmt-0.5676718415170368.bat.env'
[kit] Error: ENOENT: no such file or directory, open 'C:\Users\XX\AppData\Local\Temp\vs-cmt-0.7996598566719535.bat.env'
[kit] Error: ENOENT: no such file or directory, open 'C:\Users\XX\AppData\Local\Temp\vs-cmt-0.5296463357321997.bat.env'
[kit] Error: ENOENT: no such file or directory, open 'C:\Users\XX\AppData\Local\Temp\vs-cmt-0.5546743853638219.bat.env'
[kit] Error: ENOENT: no such file or directory, open 'C:\Users\XX\AppData\Local\Temp\vs-cmt-0.058505082376307804.bat.env'
[kit] Error: ENOENT: no such file or directory, open 'C:\Users\XX\AppData\Local\Temp\vs-cmt-0.3845789596045024.bat.env'

### Developer Tools Log

```
<!-- Paste the log contents HERE -->
```

### Platform and Versions

- **Operating System**: Windows
- **CMake Version**: cmake version 3.12.18081601-MSVC_2
- **VSCode Version**: 1.29.1
- **CMake Tools Extension Version**: 1.1.3
- **Compiler/Toolchain**: Visual C++ 2017

### Other Notes/Information

Completely removing cmake-tools does not fix the issue.

pokowaka:
I've done some debugging, and can now provide you with something useful:

The generation of the batch file in:


async function collectDevBatVars(devbat: string, args: string[], major_version:number, common_dir:string): Promise<Map<string, string>|undefined>

depends on the existence of the file ${devbat}.  However it is possible for that file not to exists, causing no .env file to be written.

In my case: 

devbat="C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvarsall.bat"

Which does not exist, causing the following failure in the batch script:

[...]

PR #613, opened 2018-12-18
Discover all Visual Studio Installs.

pokowaka:
It is possible to have multiple versions of Visual Studio. For example
Professional and Enterprise. Previously only one version could be
registered, potentially causing problems.

Closes: #611
Test: Scanning of kits succeed, with reasonable (but longer) display name.

Now shows the complete name of Visual Studio in discovered kits

## The purpose of this change

Fix a bug when there are multiple installations of visual studio, but making only one available.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -374,7 +374,7 @@ export async function vsInstallations(): Promise<VSInstallation[]> {
   for (const inst of vs_installs) {
     const majorVersion = parseInt(inst.installationVersion);
       if (majorVersion >= 15) {
-      inst.instanceId = `VisualStudio.${majorVersion}.0`;
+      inst.instanceId = `${inst.displayName}`;
     }
     if (inst_ids.indexOf(inst.instanceId) < 0) {
       installs.push(inst);


lygstate:
Gocha have a try:)

lygstate:
@pokowaka I am confused why only one version could be registered。
I prefer
      inst.instanceId = `VisualStudio.${majorVersion}.0.Commnity`;
      inst.instanceId = `VisualStudio.${majorVersion}.0.Professional`;
and so on.

pokowaka:
The reason only one version gets registered is due to line 379. Only the first instance id gets registered.

I don't think it is worth it to have a display name translator that saves 5 chars for the following reasons.

- The code will be more complicated, as we will have to determine which type is installed (Enterprise/Community/Prof/etc). 
- The proposed scheme will cause complications if Microsoft drops or adds a new "release type". For example community editions where named express at a certain point.
- Not clear what the added value of .0. in the string is.

lygstate:
@pokowaka For consistence with older version of visual Studio.

blakehurd:
Hello, per the wording of this issue, would this potentially help with CMake Tools not discovering msvc_x64 / x64 kit when it scans? It picks up amd64, x86, amd64_arm, etc., but does not clearly pick up x64 separately even though the requisite exe/etc. are located right next to x86 per the Visual Studio Installer. Ended up needing to define a kit anyway to specify a vcpkg toolchain, but would've helped to have this at the start.

pokowaka:
@vector-of-bool  how can we move forward addressing this issue?

lygstate:
@vector-of-bool How about give me admin permission to merge request and submit pathches?

bobbrow:
I'm happy to merge this, but is it still an issue?  I see both "Build Tools" and "Enterprise" for VS 2017 on my machine in the latest version of CMake Tools.

image

pokowaka:
> I'm happy to merge this, but is it still an issue? I see both "Build Tools" and "Enterprise" for VS 2017 on my machine in the latest version of CMake Tools.
> 
> image

The problem only arises when the same major version is the same. The case would be installing 2019 Enterprise next to 2019 Build Tools. Only one of the two will show up.

bobbrow:
Is it not a problem with Preview builds then?  I have Build Tools 2017 and Enterprise 2017 (Preview).

bobbrow:
I've been testing this on my PC and there are still some issues.
* My VS 2019 and 2019 Preview builds are not both showing up because they have the same display name (`isPrerelease` differentiates these)
* The description string for choosing a kit ("using compilers from ...") is a bit too long with this change. It would be more informational if the major.minor version appeared here since the displayName is already visible.

I'm trying to think of a way to have all this information available with the current Kits schema, but this may just require a forced full rescan of kits to get right.

bobbrow:
I have a new PR at #719 that should address all of these pieces.

pokowaka:
Newer change address all concerns. PR no longer relevant.

PR #619, opened 2018-12-26
MSVC double warnings and wrong paths

KoeMai:
### This changes visible behavior of MSVC problems

- Update MSVC message RegEx, to identify problems from MSVC
- Add diagnostic test case

### DIFF
--- a/src/driver.ts
+++ b/src/driver.ts
@@ -731,6 +731,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
         return [
           '/m',
           '/property:GenerateFullPaths=true',
+          '/consoleloggerparameters:NoSummary'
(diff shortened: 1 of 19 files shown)


jclay:
I was experience this problem as well and tried your branch out, it works great!

src/driver.ts
```diff
@@ -731,6 +731,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
         return [
           '/m',
           '/property:GenerateFullPaths=true',
+          '/consoleloggerparameters:NoSummary'
```
Randshot: Hard coded build arguments were removed in a recent PR, since it can cause issues.

KoeMai: Done

PR #626, opened 2019-01-05
Support usage of environment variables for toolchainFile in kits

blakehurd:
### This changes kit handling to support environment variables (${SOME_ENV_VAR} pattern) for toolchainFile setting

- Add a processKit method to post-process kits from the .json spec 
- Leverage processKit method to resolve environment variables used for the toolchainFile setting.
- Add unit test to ensure it works.

## The purpose of this change
Scratching one of my itches using this extension. I wanted to use an env var to specify where my vcpkg toolchain is installed, rather than hardcode it. This does what I want, but feel free to suggest a better way if you can think of one.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -741,7 +741,7 @@ export function descriptionForKit(kit: Kit) {
 
 /**
  * Process parameters of the kit and perform any desired transformations.
- * 
+ *
  * Right now, this resolves environment variables used in the tool chain file. This logic could be expanded elsewhere to enable environment variable resolution elsewhere.
  *
  * May no longer be necessary if we can leverage some part of https://github.com/Microsoft/vscode/issues/2809 should it be completed and shipped.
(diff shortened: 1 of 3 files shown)


rcdailey:
Why has this not been reviewed yet? I think this is a major issue. Android's NDK includes a CMake toolchain file with it, but all developers install the NDK to different locations. Hence it's important to be able to do something like this:

```
"toolchainFile": "${env:ANDROID_NDK}/cmake/android.toolchain.cmake",
```

@vector-of-bool Can you provide some feedback?

src/kit.ts
```diff
@@ -739,6 +739,26 @@ export function descriptionForKit(kit: Kit) {
...
+ * Right now, this resolves environment variables used in the tool chain file. This logic could be expanded elsewhere to enable environment variable resolution elsewhere.
+ *
+ * May no longer be necessary if we can leverage some part of https://github.com/Microsoft/vscode/issues/2809 should it be completed and shipped.
+ *
+ * @param kit A kit json structure
+ */
+export function processKit(kit: Kit) {
+  if (kit.toolchainFile) {
+    // map ${foo} -> process.env[foo] to resolve corresponding environment variables
+    kit.toolchainFile = kit.toolchainFile.replace(/\$\{(\w+)\}/, (_str: string, p1: string, ..._args) => {
```
Randshot: Have you checked whether case sensitivity is an issue here?
This here might be useful: https://github.com/vector-of-bool/vscode-cmake-tools/blob/develop/src/expand.ts

bobbrow:
This PR was replaced by #991.

Issue #604, opened 2018-11-30
Compatibility issues with Espressif ESP-IDF


huming2207:
### Brief Issue Summary

First of all, thanks for your great plugin to allow CMake projects works with VSCode. But I've found some issues with ESP-IDF CMake project. When an ESP-IDF CMake project (especially with C++ code and/or header-only libraries) is opened, the plugin cannot handle the included headers correctly.

### Expected:

1. Open an ESP-IDF project
2. Everything works fine 

### Apparent Behavior:

1. Open an ESP-IDF project
2. It may returns this:

![image](https://user-images.githubusercontent.com/4463497/49266046-c5804100-f4a8-11e8-96c0-9bef41d37858.png)

But actually the "missing" headers do exist with correct filesystem permission applied. If I right click the header file and click "Go to definition", it will even jump to that header file correctly, although there are warning exists with underlines.

Also, sometimes if I open some header files, a warning like this will also pops up:

![image](https://user-images.githubusercontent.com/4463497/49266359-00cf3f80-f4aa-11e8-93ca-3fd42bd57f13.png)

### CMake Tools Log

There is no significant logs that indicates a warning/error so far. I will provide that later on my own desktop PC at home, if necessary.

### Developer Tools Log

(Same as above)

### Platform and Versions

- **Operating System**: Ubuntu 18.10, XFCE
- **CMake Version**: 3.12.1
- **VSCode Version**: 1.29.1
- **CMake Tools Extension Version**:  1.13
- **Compiler/Toolchain**: xtensa-esp32-elf-gcc 5.2.0

huming2207:
Hello @vector-of-bool, 

I've done another test with one of my incompleted ESP32 project.

Here is my project: https://github.com/huming2207/QueSera

I haven't got time to continue working on this project these days as I was busy dealing with some other business for work.

Here are the warning screenshots:

![image](https://user-images.githubusercontent.com/4463497/49327644-7117b700-f5b7-11e8-8e6d-182c58d508f9.png)


![image](https://user-images.githubusercontent.com/4463497/49327516-1a11e200-f5b7-11e8-80bc-20fcf946880d.png)

Here is the devtool's log, you can find there are some warnings and errors here: [devtool.log](https://github.com/vector-of-bool/vscode-cmake-tools/files/2635820/devtool.log)


Here is the CMake Tools Log:

```
[cms-client] Configuring using the "Unix Makefiles" CMake generator
[...]

Issue #599, opened 2018-11-27
cpptools Integration - trying to to understand


enelson1001:
### Brief Issue Summary
I have just started using VSCode and CMake Tools (CMT).  I came across some problems with include statements in header files.  I have looked at issue #481 as it seems close to what I am seeing.
1. I used CMake: Edit usr-local CMake kits to add my toolchain otherwise my toolchain did not show even though it is in my environmental PATH.  
**Question - Shouldn't CMT automatically detect my compiler if it was in my PATH?**
2. Started new project and selected my toolchain kit and is listed in bottom menu. When popup came up "CMake would like to configure intelliSense for this folder" I clicked allowed.
3. If I opened a header file in my project 1st I would receive a bunch of green squiggle lines on includes.
 
![image](https://user-images.githubusercontent.com/43481343/49104415-24408300-f23b-11e8-9be1-b2427fd87639.png)

But if I open Greetings.cpp 1st and then Greetings.h most green squiggles go away.
![image](https://user-images.githubusercontent.com/43481343/49104493-594cd580-f23b-11e8-9f1b-1841f412d853.png)

Reading issue #481 states you need to open the cpp file first before the header file which the OP thought was fixed.  
**Question - Do we need to open up the cpp file first before the header file?**

4. I want to include header file from external library (esp-idf) whose components are part of the build directory but their CMakeLists.txt files do not contain the *.h files.  In picture below, the build folder shows the driver folder and the freetos folder but I still get the green squiggly lines trying to include headers from their folders.
 
![image](https://user-images.githubusercontent.com/43481343/49105013-b8f7b080-f23c-11e8-8263-c7e7b0953487.png)

If I take the suggestion and add these to IncludePath a new c_cpp_properties file is created and then all the green squiggly lines come back.
![image](https://user-images.githubusercontent.com/43481343/49105355-97e38f80-f23d-11e8-9a5d-c16076770f67.png)

**Question - How do I add these includes and not get the green squiggly lines**

5. If I look at the newly created c_cpp_properties.json file is lists the wrong compiler path.
![image](https://user-images.githubusercontent.com/43481343/49105476-d9743a80-f23d-11e8-835d-fdf8378736d7.png)

**Question - Why is the wrong compiler path listed?**

### Platform and Versions

- **Operating System**:  Linux Mint 18
- **CMake Version**:  3.12.4
- **VSCode Version**:  1.29
- **CMake Tools Extension Version**:  1.1.2
[...]

PR #636, opened 2019-01-19
Cross compile support for cpptools integration.

bjosa:
# Cross Compile support for cpptools integration

Added compile flags and sysroot to the compiler path for cpptools.

## This change addresses [#637]

Generally all cross compiling toolchains that depend on special compiler flags and a sysroot. These flags have to be published to the cpptools compilePath option. The cpptools extension then parses the compiler with These flags for the right include paths and defines.

### This changes cpptools integration

- Add compile flags to cpptools compilePath
- Add sysroot to cpptools compilePath

### DIFF
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -184,7 +184,7 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
    * @param fileGroup The file group from the code model to create config data for
    * @param opts Index update options
    */
-  private _buildConfigurationData(fileGroup: cms.CodeModelFileGroup, opts: CodeModelParams, target: TargetDefaults):
+  private _buildConfigurationData(fileGroup: cms.CodeModelFileGroup, opts: CodeModelParams, target: TargetDefaults, sysroot: string):
       cpt.SourceFileConfiguration {
     // If the file didn't have a language, default to C++
     const lang = fileGroup.language || 'CXX';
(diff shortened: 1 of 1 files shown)


bjosa:
Maybe related to #604, #599

src/cpptools.ts
```diff
@@ -208,17 +208,28 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
         newBrowsePath.push(includePathItem);
       }
     }
+
+    let cpptoolsCompilerPath = (flags.length === 0) ? comp_path : `${comp_path} ${flags.join(' ')}`;
+    if (sysroot && cpptoolsCompilerPath.indexOf("--sysroot") < 0) {
+      if (sysroot.match(/\s/)) {
+        cpptoolsCompilerPath += ` "--sysroot=${sysroot}"`;
```
bobbrow: The opening `"` seems to be in the wrong place.  Shouldn't it be after `--sysroot=`?  And is `sysroot` guaranteed not to have additional `"`'s in it?

bjosa: Thank's for looking over this. Think this should be ok. The `sysroot` option points to an path. The path will be quoted if it contains a space. The double quote surrounds `--sysroot` so the whole term, option and path, is interpreted as an command line argument.

icegood:
I don't get how to use it. Please, update official documentation. At least I don't see any way to set sysroot somewhere in options.

bjosa:
Just provide a `cmake-kits.json` file with the specific settings for your toolchain. See the [cmake-kits](https://github.com/microsoft/vscode-cmake-tools/blob/develop/docs/kits.md) section of the documentation for more information.

Example cmake-kits entry with an CMake toolchain file. The CMake toolchain file of the sdk then defines the sysroot path and all neccessary compiler flags:

```json
{
    "name": "arm-linux-gnueabihf",
    "toolchainFile": "/home/user/sdk/toolchain.cmake",
    "cmakeSettings": {
        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
        "CMAKE_STAGING_PREFIX": "${workspaceFolder}/build/${buildKit}/out",
        "CMAKE_PREFIX_PATH": [
            "${workspaceFolder}/build/${buildKit}/out"
        ]
    },
    "preferredGenerator": {
        "name": "Ninja"
    }
}
```

PR #640, opened 2019-01-22
configure standard by language

xVanTuring:
### This changes 
1. src/cpptools.ts

* pass lang parameter to help configure the `standard`
## The purpose of this change
configure standard by language

### DIFF
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -32,10 +32,10 @@ interface TargetDefaults {
   defines: string[];
 }
 
-export function parseCompileFlags(args: string[]): CompileFlagInformation {
+export function parseCompileFlags(args: string[], lang: string = 'CXX'): CompileFlagInformation {
   const iter = args[Symbol.iterator]();
   const extraDefinitions: string[] = [];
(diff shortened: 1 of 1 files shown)


src/cpptools.ts
```diff
@@ -55,19 +55,21 @@ export function parseCompileFlags(args: string[]): CompileFlagInformation {
...
+          standard = 'c++17';
+        } else if (std.endsWith('++11') || std.endsWith('++0x')) {
+          standard = 'c++11';
+        } else if (std.endsWith('++2a')) {
+          // Not yet supported...
+        } else if (std.endsWith('++98')) {
+          standard = 'c++98';
+        } else if (std.endsWith('++03')) {
+          standard = 'c++03';
+        }
```
bobbrow: There is some code a few lines down that deals with "unknown standard control flags".  That code is now incorrect since it will set c++17 for a C file.  It seems this function already sets a default value for `standard` so this is unnecessary, but the warning may still be useful.  Can you replicate the warning for the CXX block as well?

bobbrow:
Replaced by #730

PR #644, opened 2019-01-27
Explicitly setting VS compiler cores and BuildInParallel to true.

giladreich:
CMake build command Explicitly setting VS compiler cores and BuildInParallel property to true.

The build command would look something like this:
```
cmake --build . --config Debug --target ALL_BUILD -- /m:10 /p:BuildInParallel=true /p:GenerateFullPaths=true
```

And would be more consistent with the way we set cores(i.e `-j 10`) when using any other generators that are not using the VS compiler.


Randshot:
@ca-mtg the hard coded build flags were removed, since the user can manually add them via `buildArgs`.
Going to close this one for now.

PR #647, opened 2019-02-05
Perform Build as Task (among other minor updates)

emanspeaks:
### Proposed behavior changes

- Perform build in an integrated terminal
- Provide more robust checking of Windows paths stored in environment variables
- Allow skipping the definition of kit environment when doing Visual Studio builds
- Load CMake Cache as text file (previewHtml is deprecated anyways)
- Remove hard-coded flags for Visual Studio builds

## The purpose of this change

While attempting to build a project with Visual Studio 2017, I encountered issues caused by the presence of the hard-coded flags.  Since the user can add `buildArgs` in the settings, these flags should be manually added by the user, or set as the default value of `buildArgs`.  Once I removed the flags, the build at least started, but encountered errors with using Intel compilers in Visual Studio.  Suspecting environment problems, I added the capability to ignore the VS environment.  I was still unable to verify the cause of the issue, but noticed that if I copy and pasted the command generated by CMake Tools into an integrated terminal, the build would run successfully.  I subsequently changed the build command to provide a command string for a task.  I also wanted to add custom `configureSettings` that are loaded from environment variables, so I needed to enhance this capability to eliminate a number of errors.


src/driver.ts
```diff
@@ -213,8 +213,9 @@ export abstract class CMakeDriver implements vscode.Disposable {
 
   getEffectiveSubprocessEnvironment(opts?: proc.ExecutionOptions): proc.EnvironmentVariables {
     const cur_env = process.env as proc.EnvironmentVariables;
+    const kit_env = (this.ws.config.ignoreVsEnv) ? {} : this.getKitEnvironmentVariablesObject();
```
Randshot: I think it would be better to check for `ignoreVsEnv` here:
https://github.com/vector-of-bool/vscode-cmake-tools/blob/b3a7aca1f86052ef94111163514906474c3d1f47/src/kit.ts#L638-L648
Or maybe rename the option to `ignoreKitEnv`?

emanspeaks: Renamed per your suggestion!

emanspeaks:
This should be ready to merge after someone reviews it...just waiting on a review.

PR #666, opened 2019-03-29
added a launchTargetDirectory command

dcourtois:
## Changes

- add a new `cmake.launchTargetDirectory` command.

This allows to use variable substitution to get the active launch target's directory in launch configurations. It works like `cmake.launchTargetPath` but instead of the path to the file, it's the path to the directory containing it.

## Why

In almost any complex project, lots of additional files are built and usually output next to the main executable (configuration files, data, etc.) When debugging this kind of projects, we have to set the launch working directory to the directory containing all those files. And if we configured CMake build directory using `${builtType}` we're screwed : we have no way of automatically knowing what working directory we should use.

Until this patch, I used the `${input:...}` feature, but it's error prone, and a bit tedious. With this command, we can easily access the directory and the launch configurations are a lot simpler. Also debugging no longer requires any user interaction (well, except for actually hitting F5 :) )

## Notes

I updated the doc and added a unit test. It basically just uses the existing `launchTargetPath` command and just does a `path.dirname` on the result, so the changes are pretty trivial.

I hope you'll accept this request, it would really make our life easier :)

Regards,
Damien.

### DIFF
--- a/docs/debugging.rst
+++ b/docs/debugging.rst
@@ -82,8 +82,8 @@ specifying a a command-based substitution in the appropriate field of
 ``launch.json``.
 
 Here is a minimal example of a ``launch.json`` that uses the
-``cmake.launchTargetPath`` to start a debugger on the active selected launch
-target:
+``cmake.launchTargetPath`` and ``cmake.launchTargetDirectory`` to start a debugger
+on the active selected launch target:
 
 .. code:: javascript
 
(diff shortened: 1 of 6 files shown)


dcourtois:
Pull request of hell, just noticed its ID number :p
As a side note: is there any contributor there ??

Randshot:
@dcourtois Hm, seems like something went wrong with travis. Can't merge the PR like this.
Maybe try pushing a new commit :man_shrugging: ?

Randshot:
@vector-of-bool Something seems to be wrong with the travis-ci check.

dcourtois:
Hi, thanks for checking this ! It seems to be a problem with the Travic infrastructure right ? Is there something I can do ?

Randshot:
No, I don't have the needed permissions for changing the CI stuff, so I (we) can't really do anything.
You could try opening a new PR though.

dcourtois:
Ok, will do that to see if it changes anything.

PR #669, opened 2019-04-04
Additional commands

dcourtois:
## Changes

- add a new `cmake.launchTargetDirectory` command.

This allows to use variable substitution to get the active launch target's directory in launch configurations. It works like `cmake.launchTargetPath` but instead of the path to the file, it's the path to the directory containing it.

## Why

In almost any complex project, lots of additional files are built and usually output next to the main executable (configuration files, data, etc.) When debugging this kind of projects, we have to set the launch working directory to the directory containing all those files. And if we configured CMake build directory using `${builtType}` we're screwed : we have no way of automatically knowing what working directory we should use.

Until this patch, I used the `${input:...}` feature, but it's error prone, and a bit tedious. With this command, we can easily access the directory and the launch configurations are a lot simpler. Also debugging no longer requires any user interaction (well, except for actually hitting F5 :) )

## Notes

I updated the doc and added a unit test. It basically just uses the existing `launchTargetPath` command and just does a `path.dirname` on the result, so the changes are pretty trivial.

I hope you'll accept this request, it would really make our life easier :)

Regards,
Damien.

### DIFF
--- a/docs/debugging.rst
+++ b/docs/debugging.rst
@@ -82,8 +82,8 @@ specifying a a command-based substitution in the appropriate field of
 ``launch.json``.
 
 Here is a minimal example of a ``launch.json`` that uses the
-``cmake.launchTargetPath`` to start a debugger on the active selected launch
-target:
+``cmake.launchTargetPath`` and ``cmake.launchTargetDirectory`` to start a debugger
+on the active selected launch target:
 
 .. code:: javascript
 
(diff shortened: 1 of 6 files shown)


dcourtois:
Seems travis-ci really is broken...

Randshot:
Yeah, nothing I can do here :/

Randshot:
Hm, I just checked on travis-ci and the build got triggered, but it doesn't seem to report back to GitHub?

dcourtois:
Unfortunately, I'm a complete noob in ci toolings..
Oh ! travis just passed ! but now its appveyor stalling :D

Randshot:
Ok, both passed, gonna merge this one.

ndotb:
Thank you for adding this, @dcourtois .  I was just looking for it.

Issue #682, opened 2019-05-11
Wrong cmake generator when selecting "Visual Studio Community 2017 - amd64" kit


atrantan:
### Brief Issue Summary

When I'm trying to switch to the `Visual Studio Community 2017 - amd64` kit the chosen generator is `Visual Studio 15 2017` but should be `Visual Studio 15 2017 Win64`

### Expected:

```
[driver] Switching to kit: Visual Studio Community 2017 - amd64
[cms-client] Configuring using the "Visual Studio 15 2017 Win64" CMake generator
```

### Apparent Behavior:

```
[driver] Switching to kit: Visual Studio Community 2017 - amd64
[cms-client] Configuring using the "Visual Studio 15 2017" CMake generator
```

### Developer Tools Log

```
<!-- Paste the log contents HERE -->
```

### Platform and Versions

- **Operating System**: Windows
- **CMake Version**: 3.12.1
- **VSCode Version**: 1.33.1
- **CMake Tools Extension Version**: 1.1.3
- **Compiler/Toolchain**: Visual Studio 15 2017 Win64

KoeMai:
I created a patchset. But after some later tests I saw that this is not the right solution ->closed it without merge. The cmake generator for Visual Studio supports architecture and toolset arguments, see at the [cmake documentation](https://cmake.org/cmake/help/v3.4/generator/Visual%20Studio%2012%202013.html). The extension already use them.

What do you mean with it nothing happens?
How does your selected kit from the kit-file looks like?

atrantan:
@KoeMai Sorry, I have updated the issue yo clarify thé behaviour

atrantan:
@KoeMai for the moment, and yo make it work, I need to 
- explicitely type the command before opening the source folder with vscode
```
cmake ../ -G "Visual Studio 15 2017 Win64
```
- ... then open the folder

KoeMai:
If you run the cmake configuration step inside vscode. How does your kit file looks like?

atrantan:
@KoeMai Do you mean the CMakeCache file?

andreeis:
We fixed this for CMake Tools 1.4, but slightly different. We are passing the platform attribute of the kit generator instead of passing the architecture in the generator name.
Closing as duplicate of https://github.com/microsoft/vscode-cmake-tools/issues/964.

PR #684, opened 2019-05-15
Fix wrong VS generator selection (#682)

KoeMai:
Fix the wrong selection of prefered generator on 64bit builds
with visual studio.

- Correct generation of preferred generators for Visual Studio Projects on Arch 64 Projects


KoeMai:
I created this patchset, after some later tests I saw that this is not the right solution.
The cmake generator for Visual Studio supports architecture and toolset arguments, see at the [cmake documentation](https://cmake.org/cmake/help/v3.4/generator/Visual%20Studio%2012%202013.html).
The extension already use them.

PR #686, opened 2019-05-16
conditionally add install target

### LINKED ISSUE #504: Target install missing
### Brief Issue Summary

Target "install" can no longer be selected from the bottom toolbar; only "all" is available.

### Expected:

One should be able to select target "install" the same way as target "all".

andy-held:
## This change addresses item #504
closes #504

Conditionally add an `install` meta target if there is a project that has an install rule.

## The purpose of this change

cmake-server does not give an install target in project targets.

### DIFF
--- a/src/cms-driver.ts
+++ b/src/cms-driver.ts
@@ -165,7 +165,7 @@ export class CMakeServerClientDriver extends CMakeDriver {
       filepath: 'A special target to build all available targets',
       targetType: 'META',
     }];
-    if(build_config.projects.some(project => project.hasInstallRule))
+    if(build_config.projects.some(project => (project.hasInstallRule)? project.hasInstallRule: false))
     {
       metaTargets.push({
         type: 'rich' as 'rich',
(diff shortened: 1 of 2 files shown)


andy-held:
Hm, this will not work with CMake versions prior to 3.10, I think, c.f. https://cmake.org/cmake/help/v3.10/manual/cmake-server.7.html

andy-held:
This seems to have passed the tests, but I actually expect it to not work with CMake 3.10 and lower due to the missing field.

src/cms-driver.ts
```diff
@@ -159,6 +159,21 @@ export class CMakeServerClientDriver extends CMakeDriver {
       log.error('Found no matching code model for the current build type. This shouldn\'t be possible');
       return [];
     }
+    const metaTargets = [{
```
KoeMai: It would be possible to check the version of the cmake executable (`this.cmake`).

yeganer:
This is awesome.

Looking at the response from the CMake server, individual targets also have the `hasInstallRule` property. One could use that to generate install rules for individual targets.
This would be very useful in a scenario where running the full install produces errors but the install of individual components (which are currently being developed) is possible.

src/cms-driver.ts
andy-held: What do you think about checking, whether the `hasInstallRule` field exists?

andy-held: I think this should do it, shouldn't it?

PR #719, opened 2019-07-26
Fix kits to allow for multiple VS SKUs+Preview in Kits

bobbrow:
Clean up naming of compilers, and discover all kits on PC

## Other Notes/Information
We may want another change that forces a rescan of kits so that we don't have to support the old schema indefinitely

Replaces PR #613

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -21,6 +21,16 @@ const log = logging.createLogger('kit');
 type ProgressReporter = vscode.Progress<{message?: string}>;
 
 /**
+ * Cache the results of invoking 'vswhere'
+ */
+interface VSInstallationCache {
+  installations: VSInstallation[];
+  queryTime: number;
+}
+
+let cachedVSInstallations: VSInstallationCache|null = null;
+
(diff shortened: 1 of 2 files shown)


pokowaka:
Awesome! Let me close my pull request.

src/kit.ts
```diff
@@ -628,7 +649,15 @@ export async function getVSKitEnvironment(kit: Kit): Promise<Map<string, string>
   console.assert(kit.visualStudio);
   console.assert(kit.visualStudioArchitecture);
   const installs = await vsInstallations();
-  const requested = installs.find(inst => inst.instanceId == kit.visualStudio);
+  const match = (inst: VSInstallation) =>
+      // old Kit format
+      (inst.instanceId == kit.visualStudio) ||
```
bobbrow: Actually, this is wrong.  `kit.visualStudio` will be in `VisualStudio.${majorVersion}.0` format in the old schema

lygstate:
ping for merge

PR #720, opened 2019-07-27
cmake-file-api(7)

KoeMai:
## Implementation of the CMake 3.15.0-rc2 `cmake-file-api(7)`
This modification addresses the new `cmake-file-api(7)`.
-	Setup of a new driver for `cmake-file-api(7)`
-	All needed `CodeModel` interfaces from CMake ServerApi Driver are extracted (CMake TreeView or vscode-CppTool)
-	Automatic extension features are implemented

## Still open issues:
- [x] Tests for CodeModel interfaces
- [x] Harmonization of test suite for other drivers missing
- [x] Implement sysroot support


src/drivers/cmfileapi-driver.ts
```diff
@@ -0,0 +1,218 @@
...
+
+      this._cache = await loadCacheContent(path.join(reply_path, cache_obj.jsonFile));
+
+      // load targets
+      const codemodel_obj = indexFile.objects.find((value: index_api.Index.ObjectKind) => value.kind === 'codemodel');
+      if (!codemodel_obj) {
+        throw Error('No code model object found');
+      }
+      this._target_map = await loadConfigurationTargetMap(reply_path, codemodel_obj.jsonFile);
+      this._codeModel = await loadExtCodeModelContent(reply_path, codemodel_obj.jsonFile);
```
bobbrow: I think there would be value in allowing users to use the codemodel that is persisted to disk in the event that they don't want to reconfigure frequently.  (e.g. #984)  Is there any other way to load the code model today, or do we need to invent this concept in the driver?

src/drivers/cmfileapi-driver.ts
```diff
@@ -0,0 +1,218 @@
...
+    const cmake = this.cmake.path;
+    log.debug('Invoking CMake', cmake, 'with arguments', JSON.stringify(args));
+    const env = await this.getConfigureEnvironment();
+    const res = await this.executeCommand(cmake, args, outputConsumer, {environment: env}).result;
+    log.trace(res.stderr);
+    log.trace(res.stdout);
+    if (res.retc == 0) {
+      this._needsReconfigure = false;
+    }
+    await this._reloadPostConfigure();
```
bobbrow: Does it make sense to call `_reloadPostConfigure` if the configure failed?  Should this be in the `if (res.retc == 0)` block?

src/drivers/cmakefileapi/api_helpers.ts
```diff
@@ -0,0 +1,244 @@
...
+export async function loadIndexFile(reply_path: string): Promise<index_api.Index.IndexFile|null> {
+  log.debug(`Read reply folder: ${reply_path}`);
+  if (!await fs.exists(reply_path)) {
+    return null;
+  }
+
+  const files = await fs.readdir(reply_path);
+  log.debug(`Found index files: ${JSON.stringify(files)}`);
+
+  const index_file = files.find(filename => filename.startsWith('index-'));
```
bobbrow: The spec says that it is possible that multiple index files can exist at the same time.  Can we check and make sure we get the latest?  They will be sorted lexicographically.

src/drivers/cmakefileapi/api_helpers.ts
```diff
@@ -0,0 +1,244 @@
...
+      STRING: api.CacheEntryType.String,
+      PATH: api.CacheEntryType.Path,
+      FILEPATH: api.CacheEntryType.FilePath,
+      INTERNAL: api.CacheEntryType.Internal,
+      UNINITIALIZED: api.CacheEntryType.Uninitialized,
+      STATIC: api.CacheEntryType.Static,
+    };
+    const type = entry_type_translation_map[el.type];
+    if (type === undefined) {
+      rollbar.error(`Unknown cache entry type ${el.type}`);
```
bobbrow: I see this use of rollbar here, but not around any of the file operations.  Is it intentional that we let those exceptions bubble up to the top?

src/drivers/cmakefileapi/api_helpers.ts
bobbrow: I believe this is a duplicate.  I wrote it again because github.com wasn't showing me the preview.

src/drivers/cmakefileapi/api_helpers.ts
```diff
@@ -0,0 +1,244 @@
...
+
+export async function loadConfig(paths: index_api.CodeModelKind.PathInfo,
+                                 reply_path: string,
+                                 configuration: index_api.CodeModelKind.Configuration) {
+  const projects = await Promise.all(
+      (configuration.projects).map((_, index) => loadProject(paths, reply_path, index, configuration)));
+  return {projects} as CodeModelConfiguration;
+}
+export async function loadExtCodeModelContent(reply_path: string, codeModel_filename: string) {
+  const codeModelContent = await loadCodeModelContent(path.join(reply_path, codeModel_filename));
```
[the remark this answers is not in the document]
KoeMai: WIP: There are a lot of version numbers in the CMake FileAPI. I need to look at this more in detail.

src/drivers/cmfileapi-driver.ts
KoeMai: The idea of this PR was implementing the possibilities of the ServerAPI for FileAPI.
Let us restructure the driver to allow a simple CMake configuration check (expectation between CmakeTools (UI) and CMake driver). The FileAPI gives the possibility to load the old code model inside of the CMake driver. The problem is that we have no way to test if all Kit/Variant/Settings are as expected by the UI. The UI can only force the driver into a configuration.

I´m not sure if #984 and the possibility to use a preconfigured build folder by a new CMake-Driver instance should be part of this PR.
I do not see the required changes to make it simple possible.
Could we move this into a new issue?

bobbrow: It's fine to move it to a new issue.  I just wanted to raise the question in case it was easy to do right now.

src/cmake-tools.ts
```diff
@@ -36,6 +36,8 @@ import {Property} from './prop';
 import rollbar from './rollbar';
 import {setContextValue} from './util';
 import {VariantManager} from './variant';
+import { CMakeFileApiDriver } from '@cmt/drivers/cmfileapi-driver';
+import * as driver_api from '@cmt/drivers/driver_api';
```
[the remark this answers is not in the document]
bobbrow: Or just leave it for now.  We can normalize all of this stuff later.  Let's just prioritize getting this PR finished.

src/cmake-tools.ts
```diff
@@ -265,26 +266,42 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+      }
+
+      if (communicationMode != 'fileAPI') {
         log.warning(
           localize('please.upgrade.cmake',
-            'CMake Server is not available with the current CMake executable. Please upgrade to CMake {0} or newer.',
-            versionToString(cmake.minimalServerModeVersion)));
-        drv = await LegacyCMakeDriver
-                  .create(cmake, this.workspaceContext.config, kit, workspace, preConditionHandler, preferredGenerators);
+            'CMake FileApi is not available with the current CMake executable. Please upgrade to CMake {0} or newer.',
```
bobbrow: I think the purpose of this message was related to the fact that "legacy" mode provides little actual CMake support.  You can still get decent support without file api, so people might complain about this message.  We could probably show a similar message if neither server mode nor file api are available.

"For the best experience, CMake server or file-api support is required.  Please upgrade CMake to {0} or newer." (serverModeVersion)

src/cmake-tools.ts
```diff
@@ -265,26 +266,42 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
     const preferredGenerators = this.getPreferredGenerators();
     const preConditionHandler = async (e: CMakePreconditionProblems) => this.cmakePreConditionProblemHandler(e);
-    if (this.workspaceContext.config.useCMakeServer) {
-      if (cmake.isServerModeSupported) {
-        drv = await CMakeServerClientDriver
-                  .create(cmake, this.workspaceContext.config, kit, workspace, preConditionHandler, preferredGenerators);
+    let communicationMode = this.workspaceContext.config.cmakeCommunicationMode;
+
+    if (communicationMode == 'automatic') {
+      if (cmake.isFileApiModSupported) {
```
bobbrow: Looks like there might be a typo in the name of this property. ("Mode")

src/config.ts
```diff
@@ -188,6 +198,17 @@ export class ConfigurationReader implements vscode.Disposable {
 
   get useCMakeServer(): boolean { return this.configData.useCMakeServer; }
 
+  get cmakeCommunicationMode(): CMakeCommunicationMode {
+    let communicationMode = this.configData.cmakeCommunicationMode;
+    if (this.useCMakeServer) {
+      log.warning(localize(
+          'please.upgrade.configuration',
+          'The setting \'useCMakeServer\' is replaced by \'cmakeCommunicationMode\'. Please upgrade your configuration.'));
```
bobbrow: Perfect.  Thanks!

src/config.ts
```diff
@@ -188,6 +198,17 @@ export class ConfigurationReader implements vscode.Disposable {
...
+      communicationMode = 'serverAPI';
```
bobbrow: If the settings conflict, I would rather take `cmakeCommunicationMode` over this one.  (e.g. if communication mode is fileAPI, we shouldn't downgrade it to server).  I think you should only allow setting it to server here if communicationMode is automatic (the default value).  I believe vscode has an API that would let us know if "automatic" was set in the user config versus being the default, unset value, but we might be able to ignore that for now.

src/drivers/cmakefileapi/api_helpers.ts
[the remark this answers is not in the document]
bobbrow: codemodel and cache are version 2.
cmakeFiles is version 1.

src/drivers/cmfileapi-driver.ts
KoeMai: I changed it. CMake only generates a code model if it runs successfully.
Additionally, I changed the behavior of the `doInit`-function. The `doInit`-function runs on driver creation. It is not necessary to configure if there is already a CodeModel is present. But this brings the driver into an inconsistent state between extension expected kit and the used kit/settings for the generation of CMake cache.

src/drivers/cmfileapi-driver.ts
KoeMai: I tested the new `doInit` behavior of the driver. I think it fixed issue #984, but there are so many dependencies.

src/drivers/cmakefileapi/api_helpers.ts
```diff
@@ -49,6 +48,13 @@ export async function loadCacheContent(filename: string): Promise<Map<string, ap
   const file_content = await fs.readFile(filename);
   const cache_from_cmake = JSON.parse(file_content.toString()) as index_api.Cache.CacheContent;
 
+  const expected_version = {major: 2, minor: 0};
+  const detected_version = cache_from_cmake.version;
+  if (detected_version.major != expected_version.major || detected_version.minor != expected_version.minor) {
+    log.warning(`Cache object version (${detected_version.major}.${detected_version.minor}) of FileAPI is not as ` +
```
bobbrow: All warnings should be localized.  Can you call the `localize` function for this and extract out the variables?
```
localize(
  'cache.object.version',
  'Cache object version ({0}.{1}) of cmake-file-api is unexpected. Expecting ({2}.{3}). IntelliSense configuration may be incorrect.',
  detected_version.major,
  detected_version.minor,
  expected_version.major,
  expected_version.minor)
```
```

File: src/drivers/cmakefileapi/api_helpers.ts [hunk 12]
```diff
@@ -83,7 +89,16 @@ function convertFileApiCacheToExtensionCache(cache_from_cmake: index_api.Cache.C
 
 export async function loadCodeModelContent(filename: string): Promise<index_api.CodeModelKind.Content> {
   const file_content = await fs.readFile(filename);
-  return JSON.parse(file_content.toString()) as index_api.CodeModelKind.Content;
+  const codemodel = JSON.parse(file_content.toString()) as index_api.CodeModelKind.Content;
+  const expected_version = {major: 2, minor: 0};
+  const detected_version = codemodel.version;
+
+  if (detected_version.major != expected_version.major || detected_version.minor != expected_version.minor) {
+    log.warning(`Code model version (${detected_version.major}.${detected_version.minor}) of FileAPI is not as ` +
```
bobbrow: Please localize this warning as well.

bobbrow:
Thank you so much for this!

src/drivers/cmfileapi-driver.ts
stefanweiser: Just to highlight how bad #984 is: This issue may decide over moving from previous IDEs to vscode in our company or not. And I guess this is also the point for many other guys. Other features of vscode are superior, but since we have very large code base, reconfiguring and therefore performing clean builds all the time is seriously hurting the performance of our developers. Moreover I don't see a way to work around this than excluding the plugin completely from our workflow at the moment, but doing so carries other disadvantages with it.

Issue #564, opened 2018-10-15
Variable substitution in launch.json for build directory path


nsubtil:
As a follow-on to the discussion in #559, it would be useful to enable variable substitution for `${cmake.buildDirectory}` in launch.json.

The intended use case would be applications that require cwd to be somewhere inside the build directory, or that require a path within the build directory to be passed in as a command-line argument (e.g., for loading data generated during the build).

HorstBaerbel:
This is very much needed for e.g. cwd

andreeis:
Closing as dupe of https://github.com/microsoft/vscode-cmake-tools/issues/1165. Reactivate with more details if the solutions presented there don't work for you.

Issue #334, opened 2018-03-06
${buildType} doesnt work in launch.json


zbendefy:
I'm trying to set up a launch.json in VSCode, where upon launching, the currently selected built type is selected. So if I set cmake to debug, the Build/Debug/MyApp will launch, but if I select release, then Build/Release/MyApp will launch.

I tried to accomplish this according to the documentation here (I'm not sure if I followed it correctly): https://vector-of-bool.github.io/docs/vscode-cmake-tools/configuring.html?highlight=

In my launch.json I have something like this:

> 
>       {
> 		"name": "MyApp",
> 		"type": "cppdbg",
> 		"request": "launch",
> 		"program": "${workspaceFolder}/../Build/${buildType}/MyApp",  // <----- Here is where I tried to select the buildtype
> 		"stopAtEntry": false,
> 		"cwd": "${workspaceFolder}/Binary",
> 		"externalConsole": true,
> 		"MIMode": "gdb",
> 		"setupCommands": [
> 				{
> 						"description": "Enable pretty-printing for gdb",
> 						"text": "-enable-pretty-printing",
> 						"ignoreFailures": true
> 				}
> 		] 
> 	}

### Platform and Versions

- **Operating System**: Linux Mint 18.3 64bit cinnamon
- **CMake Version**: 3.5.1
- **VSCode Version**: 1.20.1
- **CMake Tools Extension Version**: 0.10.4
- **Compiler/Toolchain**: GCC 6.3.0

Randshot:
That's not how this works. The variable substitution, at least to my knowledge, only works in the settings.json and maybe in the variant and kit json files (I can't remember right now).

vector-of-bool:
CMake Tools has no access to do those substitutions. You may want to look at `${command:cmake.launchTargetProgramPath}`, which will substitute the path to the selected launch target.

zbendefy:
@vector-of-bool Thanks, I actually found that, however I still need the buildType, because there are some other folders I need to set as Link directories. Is there a way to get that with a ${command:cmake.<something>}?

vector-of-bool:
Not at the moment, but it's a valid feature request. I'll log it

vincent-breysse:
Is there any progress on this feature ? I actually have pretty close needs.
I'd like to write something like this in my settings.json :
```
{

"catch2TestExplorer.executables": [
        {
            "name": "${filename}",
            "description": "${relDirpath}/",
[...]

PR #726, opened 2019-08-04
Add new commands, cmake.buildType and cmake.buildDirectory

notskm:
## This change addresses items #564 and #334

- `${command:cmake.buildType}` can be used to get the selected build type.
- `${command:cmake.buildDirectory}` can be used to get the current build directory.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1069,6 +1069,10 @@ class ExtensionManager implements vscode.Disposable {
 
   launchTargetDirectory() { return this.withCMakeTools(null, cmt => cmt.launchTargetDirectory()); }
 
+  buildType() { return this.withCMakeTools(null, cmt => cmt.currentBuildType()); }
+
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -1131,12 +1135,33 @@ async function setup(context: vscode.ExtensionContext, progress: ProgressHandle)
...
+    'launchTargetPath',
+    'launchTargetDirectory',
+    'buildType',
+    'buildDirectory',
+    'debugTarget',
+    'launchTarget',
+    'selectLaunchTarget',
+    'resetState',
+    'viewLog',
+    'compileFile',
```
bobbrow: It looks like `'tasksBuildCommand'` was lost in this reformatting.  Can you add it back?

notskm: Sorry about that. Should be fixed.

bobbrow:
Thank you!

PR #737, opened 2019-08-14
Recalculate total for ctest for every run

yokuyuki:
### This changes ctest pass rate

- Total for ctest shouldn't be based off total of tests parsed but rather how many tests are run as you can use ctest args to limit how many tests are fun

## The purpose of this change

For example, if I have 100 tests, but I run my ctest for only one test, it shouldn't show the status as failing and that 1/100 tests passing, but rather that it is passing and 1/1 tests passing.


bobbrow:
Thank you for your contribution!

PR #744, opened 2019-08-20
Adds user configured kit location

mjvankampen:

- Add a kit file to the config section
- Process this kit file in line with the workspace kit file

## Other Notes/Information

The current PR does not allow variable expansion as the driver (which is the only place which has the expansion options) is started after the kits have been identified. I felt that refactoring this would leave me a bit out of my depth and out of this scope.

In addition there are some formatting changes (automatically done by clang-format). I guess these are ok as they are in line with the clang-format file that is provided.


src/config.ts
```diff
@@ -8,7 +8,7 @@ import * as util from '@cmt/util';
 import * as os from 'os';
 import * as vscode from 'vscode';
 
-export type LogLevelKey = 'trace'|'debug'|'info'|'note'|'warning'|'error'|'fatal';
+export type LogLevelKey = 'trace' | 'debug' | 'info' | 'note' | 'warning' | 'error' | 'fatal';
```
bobbrow: We would like for formatting changes to be done as a separate PR.  I was planning to do a full re-formatting pass on the extension once we clear out the outstanding PR's so that we don't have to deal with lots of merge conflicts.  It also makes it really hard to see what "real" changes you've actually made.

mjvk: Understandable, will try to get those back in original state again! I had format on save active so it happened automatically :-)

mjvk: Formatting should be ok now.

src/kit.ts
```diff
@@ -865,6 +865,19 @@ function convertMingwDirsToSearchPaths(mingwDirs: string[]): string[] {
   return mingwDirs.map(mingwDir => path.join(mingwDir, 'bin'));
 }
 
+export function configKitsPathForWorkspaceDirectoryPath(wsPath: string, kitFile: string): string {
+  return path.join(wsPath, kitFile);
+}
+
+export function configKitsPathForWorkspaceFolder(ws: vscode.WorkspaceFolder, kitFile: string): string {
+  return configKitsPathForWorkspaceDirectoryPath(ws.uri.fsPath, kitFile);
```
bobbrow: Instead of hard coding this property to be relative to the workspace folder, we should enable variable substitution.  A quick and dirty way to support this is to import the `expandString` method from `expand.ts` and create a partial `ExpansionOptions` object to manage the substitutions.

```
const vars: expand.ExpansionVars = {
  workspaceRoot: ...,
  workspaceFolder: ...,
  buildType: "${buildType}",  // invalid substitution, just replace with itself
  workspaceRootFolderName: ...,
  generator: "${generator}", // invalid substitution
  userHome: ...,
  buildKit: "${buildKit}" // invalid substitution
};

return expand.expandString(kitFile, vars);

Comment (mjvk) [reply src/kit.ts]: Sounds like a nice idea. I did not see how to do this at first (as you need a driver for expansion options, which only exists after you have selected a kit). But I guess this should work. If you have a cleaner suggestion let me know as well!

File: src/extension.ts [hunk 3]
```diff
@@ -469,20 +472,45 @@ class ExtensionManager implements vscode.Disposable {
...
+      kitsWatcher.push(ws_kits_path);
+    }
+
+    // Determine wether we need to watch the config kits file:
+    const kit_file_path = await this._configKitsPath;
+    if (kit_file_path) {
+      kitsWatcher.push(kit_file_path);
+    }
+
+    this._kitsWatcher = new MultiWatcher(...kitsWatcher);
```
bobbrow: Thinking about this some more... The MultiWatcher requires paths that use glob patterns.  The config path could be an absolute path and would need to create a file watcher using `fs.watch()` instead of the vscode API.  It seems we would either need to restrict the config path to glob patterns, or add a new class in `watcher.ts` that wraps the `fs.FSWatcher` object (similar to `IndividualWatcher`).

mjvk: I'm not sure what you mean here. How are the other paths glob patterns? I think I am missing something :-).

bobbrow: I tried running your code and I'm not seeing any of the file watchers working.  The `vscode.workspace.createFileSystemWatcher` API only works for files inside the workspace folder and asks for a glob pattern as input (`MultiWatcher` creates file watchers of this type).  The user-local kits are outside the workspace too, so file watching doesn't work on that either, but that's not your bug. 😃

mjvk: Ah I see, I thought if the others work this should also work. This explains the ```_editorWatcher```  a bit better as well. Is this something you want to fix in this PR or do as something separate?

bobbrow: It seems like the path of least resistance is to just change the IndividualWatcher to use [`fs.watch`](https://nodejs.org/api/fs.html#fs_fs_watch_filename_options_listener) which would fix both problems at the same time.  Is that something you'd be willing to do?

mjvk: Willing yes, I need to check on time and knowledge (this is my first piece of typescript). I'll let you know.

bobbrow: Thanks for updating the PR. I was running into some problems with it on Windows.  I meant to respond back, but was thinking I might just update the branch when I'm done.  Sorry for the delay, I've been working on 0.26.0 of the cpptools extension as well and we've had a lot of stuff to coordinate with that release.

mjvk: I can imagine :-). I need to spend a bit more time on this as I am a bit unsure why tests are failing. In my eyes nothing should have changed yet as using the new watcher is an optional argument. Not 100% sure when I can make time again though.

mjvk: @bobbrow In the latest release of vs code I noticed:

> Filewatcher update (Chokidar)
During this milestone, we updated our file watcher for macOS and Linux to version 3.x, which brings improved performance, reduced memory consumption, and a smaller package size. You can learn more from the Chokidar repo and Chokidar 3 blog post. <

Will this impact this issue? I.e. am I trying to fix something that is/will be fixed?

bobbrow: VS Code file watchers are still based on glob patterns which won't let you look outside of the `${workspaceFolder}`.  It would be interesting to investigate using Chokidar instead of `fs.watch` for this, but I don't think it is required.

bobbrow: The multi-root PR is fixing up some of the file watcher stuff to use Chokidar.  After that goes in, I'll help merge this with those changes.

mjvk: Hi still up for this one?

bobbrow: Thanks for the reminder. We do still plan to get this in.  I'm sorry it's been so long.

mjvk: Is it wise to continue on this PR or has so much changed in the meantime that it is better to start over? Some of the functions I modified have disappeared entirely and such.

andreeis: We will analyze what is best after releasing 1.4. We will finalize on this, on way or another, thank you very much for your contribution and patience.

bobbrow:
This PR was replaced by #1701.  Thank you for the feature suggestion.  We'll get it into the extension soon.

PR #707, opened 2019-07-13
Restructuring of the cmake driver

KoeMai:
## Restructuring of the driver backend

In the last months, I changed parts of the driver backend. The idea was to separate the driver from the UI parts. The UI, configuration management and CMake interaction are now separated mostly.
Now, the CMake interaction of the driver can be tested independently from UI.
Additionally, small inconsistencies are removed.

Issue #512, opened 2018-08-06
Selecting a Kit without a compatible cmake generator breaks all builds until restart.


phi1010:
### Brief Issue Summary

The error message "[rollbar] Unhandled exception: Unhandled Promise rejection: build Error: Unable to determine CMake Generator to use {}" prevents me from building after selecting the MinGW toolchain. It also breaks Visual Studio kits in my case -- I've used the Visual Studio kits on Windows with before successfully, but after adding C:\MinGW\bin to the Path variable, the scan recognized the GCC and now refuses to build with any kit. 

### Expected:

Starting any cmake addin command (Configure, Build, Debug Target) will start cmake / the selected compiler / the target.

### Apparent Behavior:

Output tab opens in the CMake/Build view mode, and shows a single-line error message.

In the developer log, it seems as if the change kit command actually fails to switch away from GCC back to the good kits. The bottom toolbar however also doesn't update, and always shows "VisualStudio 14.0 - x86" (the last good selection) after choosing the GCC kit -- no matter what kit I choose afterwards.

Restarting VisualStudio Code fixes the issue with MSVC temporarily until I select GCC as toolchain again.

### CMake Tools Log

```
[rollbar] Unhandled exception: Unhandled Promise rejection: build Error: Unable to determine CMake Generator to use {}

```

### Developer Tools Log

```
/C:/Program Files/Microsoft VS Code/resources/app/out/vs/workbench/workbench.main.js:1375  INFO no standard startup: not a new window
/C:/Program Files/Microsoft VS Code/resources/app/out/vs/workbench/workbench.main.js:4103 Overwriting grammar scope name to file mapping for scope source.cs.
Old grammar file: file:///c%3A/Program%20Files/Microsoft%20VS%20Code/resources/app/extensions/csharp/syntaxes/csharp.tmLanguage.json.
New grammar file: file:///c%3A/Users/********/.vscode/extensions/maximetinu.identical-sublime-monokai-csharp-theme-colorizer-0.10.0/syntaxes/csharp.tmLanguage.json
e.register @ /C:/Program Files/Microsoft VS Code/resources/app/out/vs/workbench/workbench.main.js:4103
/C:/Program Files/Microsoft VS Code/resources/app/out/vs/workbench/workbench.main.js:270 [Extension Host] vscode-icons is active!
/C:/Program Files/Microsoft VS Code/resources/app/out/vs/workbench/workbench.main.js:270 [Extension Host] activating extension
[...]

Issue #641, opened 2019-01-25
Multiple configure runs in parallel


gotschmarcel:
### Brief Issue Summary

It’s possible to run CMake configure multiple times in parallel which in the end leads to an error. 

### Expected:

It should not be possible to run cmake multiple times in parallel. 

### Platform and Versions

- **Operating System**: macOS Mojave
- **CMake Version**: 3.12.1
- **VSCode Version**: 1.30.2
- **CMake Tools Extension Version**: 1.1.3

bobbrow:
This issue has been addressed in CMake Tools 1.2.0 which is available in the VS Code Marketplace or on the [Releases Page](https://github.com/microsoft/vscode-cmake-tools/releases/tag/v1.2.0)

Because the publisher id has changed, VS Code does not automatically upgrade you yet. The Marketplace team will enable this in the next few days.  If you want to install 1.2.0 now, please make sure you uninstall any earlier versions of CMake Tools manually to avoid any conflicts.

PR #747, opened 2019-08-24
Improve stability of CMake generator selection

KoeMai:
### This changes behavior and stability of the extension

- Replaces the CMake generator check from callback by previously CMake generator check and provide it as variable to the driver

## The purpose of this change

Issue #512 reports that the extension crashes completely if there was a generator selection problem. 
The cmake generator is checked at [start of the cmake client](https://github.com/microsoft/vscode-cmake-tools/blob/10cee0f78041fb74cd73c1543d2eb180cabeeba9/src/cms-client.ts#L623) when the cmake server protocol requires the information. If there is no useable build system (generator), then there cmake-server client throws an error. But there is no way to return this error into the cmake-server-driver instance. The driver instance is broken, and cmake-tools does not know it and will not create a new instance. 

The simplest solution was to move the generator selection to the cmake-tools context so that the driver does not have to select the generator. It only uses it like other build information.

This kind of error is only possible in cmake-server-api driver, legacy driver and file-api driver need the generator before executing the cmake configure step.

## Additional information

- This PR extracts some commits from my big PR #707 and #720  to improve stability of generator selection.

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -344,6 +344,7 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         try {
           this._statusMessage.set('Reloading...');
           await drv.setKit(kit, this.getPreferredGenerators());
+          this.workspaceContext.state.activeKitName = kit.name;
           this._statusMessage.set('Ready');
         } catch (error) {
           vscode.window.showErrorMessage(`Unable to set kit "${error}".`);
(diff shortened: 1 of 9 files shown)


bobbrow:
#707 is about ready to go, if you can respond to my feedback, we can get that one committed so there will be less churn to review in this PR.

bobbrow:
Should also address #641

Issue #759, opened 2019-09-08
getCTestPath() behavior in Windows is not correct


emanspeaks:
### Brief Issue Summary

If the user does not specify a path for CTest, the extension is supposed to use the CMake path to find CTest.  The logic for this search in Windows does not correctly located CTest, and thus always simply returns `ctest`, thus incorrectly assuming that CTest is in `PATH`.  Update the logic for getCTestPath() to account for the fact that the file name is `ctest.exe` and has different stat mode flags than Linux.

### Platform and Versions

- **Operating System**: Windows 10 build 1903
- **CMake Version**: 3.15.3
- **VSCode Version**: 1.38.0
- **CMake Tools Extension Version**: develop branch 2670472e461e5f6e2b0022c0531e11a349809de8

PR #760, opened 2019-09-08
Fixed PATH clobbering by bad kit file env

emanspeaks:
### This changes kit file caching of PATH variable on MinGW kits

- rename `PATH` to `CMT_MINGW_PATH` in kits file
- correct usage of `stat.isFile` to function call (`stat.isFile()`) in `getCTestPath()`

## Other Notes/Information

In the issues above, the problem occurs because the `ctest` command is not found in the `PATH` variable as defined in the output of `getEffectiveSubprocessEnvironment()`.  This may be caused by the actual absence of CTest in `PATH`, but it is also possible that the true environment PATH can be clobbered by the values in the kit file if using the kit environment is enabled.  MinGW kits, in their original behavior, would record the current value of `PATH` at the time of the scan as a kit environment variable. In my case, I scanned for kits before I had added the CMake directory to my environment path, and thus CTest could not be found.  Because the kit environment takes precedence when enabled, it was using the old cached value and hiding CTest from the `executeCommand()` call even after I updated my system PATH to include the path to CMake and CTest.

See issues #759 for more issues about why CTest is downmoding to searching on the `PATH` in the first place.


package-lock.json
```diff
@@ -0,0 +1,1935 @@
+{
+  "name": "cmake-tools",
```
bobbrow: Please delete `package.lock`.  This project uses yarn.lock instead.  Maybe it would be worth adding `package.lock` to `.gitignore` to avoid accidentally submitting this file in the future.

src/kit.ts
```diff
@@ -208,7 +208,7 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
       const binParentPath = path.dirname(bin);
       const mingwMakePath = path.join(binParentPath, 'mingw32-make.exe');
       if (await fs.exists(mingwMakePath)) {
-        const ENV_PATH = `${binParentPath};${process.env['PATH']}`;
+        const ENV_PATH = `${binParentPath}`;//;${process.env['PATH']}`;
```
bobbrow: Can you replace the commented code with a comment that explains why we don't need to store the %PATH% here?

src/kit.ts
```diff
@@ -223,7 +223,7 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
 
           if (isMake && isMingwTool) {
             gccKit.preferredGenerator = {name: 'MinGW Makefiles'};
-            gccKit.environmentVariables = {PATH: ENV_PATH};
+            gccKit.environmentVariables = {CMT_MINGW_PATH: ENV_PATH};
```
bobbrow: Do we need to force a rescan of kits?  (i.e. will developers get this fix without a rescan?)

emanspeaks: If we can force a rescan the first time the extension is activated after updating to this version of CMT, great and yes!  But I don't want to otherwise force scans...though a scan will be necessary to fix the problem.

emanspeaks:
Can someone help me with addressing the failed Travis CI builds?  I'm not sure I understand what is broken there or how to fix it.

bobbrow:
Can you revert the changes to `package.json` and see if that addresses the issue?  I wasn't sure why it was changed.  I don't see anything else in this change that could potentially fail the build.

emanspeaks:
Yup the package revert fixed it!  I think this is good to go now.

bobbrow:
Thank you for the bug fix!

PR #768, opened 2019-09-13
Change rollbar to a logger, and add telemetries for configuring and building

xisui-MSFT:
1. Changed rollbar to a logger. It doesn't send any data anymore, but still has all other previous functions such as logging error and triggering debugger.
1. Added telemetries for
    - CMake exe not found.
    - Configure: CMake exe version, generator, config type, duration, warning count, and error count
    - Build: Config type, duration, warning count, and error count

### DIFF
--- a/src/drivers/driver.ts
+++ b/src/drivers/driver.ts
@@ -612,16 +612,12 @@ export abstract class CMakeDriver implements vscode.Disposable {
       const timeEnd: number = new Date().getTime();
 
       const cmakeVersion = this.cmake.version;
-      const telemetryProperties: {[key: string]: string} = {
+      const telemetryProperties: telemetry.Properties = {
         CMakeExecutableVersion: cmakeVersion ? `${cmakeVersion.major}.${cmakeVersion.minor}.${cmakeVersion.patch}` : '',
         CMakeGenerator: this.generatorName || '',
         ConfigType: this.isMultiConf ? 'MultiConf' : this.currentBuildType || '',
       };
-      const azureSphereTargetApiSet = this.cmakeCacheEntries.get('AZURE_SPHERE_TARGET_API_SET');
-      if (azureSphereTargetApiSet) {
-        telemetryProperties['AzureSphereTargetApiSet'] = azureSphereTargetApiSet.value;
-      }
-      const telemetryMeasures: {[key: string]: number} = {
+      const telemetryMeasures: telemetry.Measures = {
         Duration: timeEnd - timeStart,
       };
       if (consumer instanceof CMakeOutputConsumer) {
@@ -641,7 +637,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
         rollbar.error('Wrong build result type.');
         telemetryMeasures['ErrorCount'] = retc ? 1 : 0;
       }
-      telemetry.logEvent('parse', telemetryProperties, telemetryMeasures);
+      telemetry.logEvent('configure', telemetryProperties, telemetryMeasures);
 
       return retc;
     } finally { this.configRunning = false; }
@@ -743,10 +739,10 @@ export abstract class CMakeDriver implements vscode.Disposable {
     const timeStart: number = new Date().getTime();
     const child = await this._doCMakeBuild(target, consumer);
     const timeEnd: number = new Date().getTime();
-    const telemetryProperties: {[key: string]: string} = {
+    const telemetryProperties: telemetry.Properties = {
       ConfigType: this.isMultiConf ? 'MultiConf' : this.currentBuildType || '',
     };
-    const telemetryMeasures: {[key: string]: number} = {
+    const telemetryMeasures: telemetry.Measures = {
       Duration: timeEnd - timeStart,
     };
     if (child) {
--- a/src/rollbar.ts
+++ b/src/rollbar.ts
@@ -1,98 +1,15 @@
 /**
  * Wrapper around Rollbar, for error reporting.
- */ /** */
-
-import * as path from 'path';
-import * as vscode from 'vscode';
-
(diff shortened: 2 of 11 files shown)


src/drivers/driver.ts
```diff
@@ -603,7 +607,42 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
       const retc = await this.doConfigure(expanded_flags, consumer);
+      const timeEnd: number = new Date().getTime();
+
+      const cmakeVersion = this.cmake.version;
+      const telemetryProperties: {[key: string]: string} = {
+        CMakeExecutableVersion: cmakeVersion ? `${cmakeVersion.major}.${cmakeVersion.minor}.${cmakeVersion.patch}` : '',
+        CMakeGenerator: this.generatorName || '',
+        ConfigType: this.isMultiConf ? 'MultiConf' : this.currentBuildType || '',
+      };
+      const azureSphereTargetApiSet = this.cmakeCacheEntries.get('AZURE_SPHERE_TARGET_API_SET');
```
bobbrow: Hmm... I don't think AzureSphere telemetry belongs in here.  Shouldn't your extension log this instead?

src/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+    return null;
+}
+
+function getPackageInfo(): IPackageInfo {
+    const packageJSON: util.PackageJSON = util.thisExtensionPackage();
+    return {
+        name: `${packageJSON.publisher}.${packageJSON.name}`,
+        version: packageJSON.version,
+        aiKey: "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217"
```
bobbrow: I don't know if we should be reusing this key for another extension or not.  Can you ask @pieandcakes and @sean-mcmanus?

src/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+    aiKey: string;
+}
+
+let telemetryReporter: TelemetryReporter | null;
+
+export function activate(context: vscode.ExtensionContext): void {
+    try {
+        telemetryReporter = createReporter();
+        if (telemetryReporter) {
+            context.subscriptions.push(telemetryReporter);
```
bobbrow: If you add this to the `context.subscriptions` you shouldn't need to manually dispose it later, right?

src/telemetry.ts
```diff
@@ -0,0 +1,57 @@
...
+import * as vscode from 'vscode';
+import * as util from './util';
+import TelemetryReporter from 'vscode-extension-telemetry';
+
+interface IPackageInfo {
+    name: string;
+    version: string;
+    aiKey: string;
+}
+
```
bobbrow: consider adding:

```ts
export type Properties = { [key: string]: string };
export type Measures = { [key: string]: number };
```

and using them as the types for the function below.

src/rollbar.ts
```diff
@@ -1,114 +1,27 @@
 /**
  * Wrapper around Rollbar, for error reporting.
- */ /** */
-
-import * as path from 'path';
-import * as vscode from 'vscode';
-
-import Rollbar = require('rollbar');
+ */
```
bobbrow: Do you plan to rename the rollbar in a subsequent PR?

src/drivers/driver.ts
```diff
@@ -603,7 +607,42 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+          }
+        });
+        telemetryMeasures['ErrorCount'] = errorCount;
+        telemetryMeasures['WarningCount'] = warningCount;
+      } else {
+        // Wrong type: shouldn't get here, just in case
+        rollbar.error('Wrong build result type.');
+        telemetryMeasures['ErrorCount'] = retc ? 1 : 0;
+      }
+      telemetry.logEvent('parse', telemetryProperties, telemetryMeasures);
```
bobbrow: This event should probably be called "configure"

src/telemetry.ts
sean-mcmanus: I think it's fine -- the C# and another extension uses the same key -- you're on an email thread about this from last year.

xisui-MSFT: Pierson mentioned in an earlier thread that this is a general key and we can use it.

src/drivers/driver.ts
xisui-MSFT: Sure. That's the named used in VS but we can't really be consistent with them anyway...

src/rollbar.ts
xisui-MSFT: I didn't have such plan. I can do it in this PR if you think renaming is necessary.

src/telemetry.ts
xisui-MSFT: Sure that's a good idea.

src/drivers/driver.ts
xisui-MSFT: This is written before that session id discussion. Since we can use session id to query the cmake telemetries for azure sphere projects, I guess we can remove this.

xisui-MSFT: Discussed with Oleg, actually this is still the best way for us since this is the only way we can get build telemetry with TargetApiSet.

src/rollbar.ts
bobbrow: Don't do it in this PR since the diff noise will be too large

src/telemetry.ts
xisui-MSFT: I was following [this usage instruction](https://github.com/Microsoft/vscode-extension-telemetry). Probably isn't necessary?

sean-mcmanus:
When is this going to check in? The usage of rollbar is causing problems with component governance.

Issue #637, opened 2019-01-20
Cpptools integration does not respect sysroot and compiler flags when crosscompiling


bjosa:
### Brief Issue Summary

Using a cross compile toolchain e.g. `cortexa9t2hf-neon-linux-gnueabi` with a sysroot folder for the target filesystem.
The CMake toolchain file sets architecture specific compile flags and the CMAKE_SYSROOT variable.

When using the cpptools integration, to autoconfigure the Microsoft C/C++ cpptools Extension, these flags and the sysroot folder are not respected from the cpptools extension.  Seems cmake-tools does not hand these settings over to the cpptools extension.

We get error squiggles for not existing header files and incorrect intellisens due to wrong defines that describe the architectur e.g. hard-float vs. soft-float.

### Wrong:

cpptools extension output:

```
Custom configurations received:
  uri: file:///c%3A/.../test.cpp
  config: {
  "defines": [],
  "standard": "c++17",
  "includePath": [
    "C:/...test/src",
    "C:/.../test/build/src",
    "C:/.../test/include",
    "C:/sdk/sysroots/cortexa9t2hf-neon-linux-gnueabi/usr/include/testlib"
  ],
  "intelliSenseMode": "clang-x64",
  "compilerPath": "C:/sdk/sysroots/x86_64-pokysdk-mingw32/usr/bin/arm-linux-gnueabi/arm-linux-gnueabi-g++.exe"
}
Custom browse configuration received: {
  "browsePath": [
    "C:/.../test/include",
    "C:/sdk/sysroots/cortexa9t2hf-neon-linux-gnueabi/usr/include/testlib",
    "C:/.../test/src",
    "C:/.../test/build/src"
  ],
  "standard": "c++17",
  "compilerPath": "C:/sdk/sysroots/x86_64-pokysdk-mingw32/usr/bin/arm-linux-gnueabi/arm-linux-gnueabi-g++.exe"
}
...
sending compilation args for C:\...\TEST\SRC\HEWO\TEST.CPP
  include: C:\...\TEST\SRC
  include: C:\...\TEST\BUILD\SRC
  include: C:\...\TEST\INCLUDE
  include: C:\SDK\SYSROOTS\cortexa9t2hf-neon-linux-GNUEABI\USR\INCLUDE\TESTLIB
[...]

PR #777, opened 2019-09-16
Adapt cpptools integration to API verison 3

bjosa:
## This change addresses item #637 

Changed cpptools integration to use api version 3.

## The purpose of this change

Cpptools changed in version 3 the api for the `WorkspaceBrowseConfiguration` and `SourceFileConfiguration`. The compiler flags are now included in the `compilerArgs` property.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -64,6 +64,13 @@
   dependencies:
     "@types/chai" "*"
 
+"@types/chai-string@^1.4.1":
+  version "1.4.2"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -64,6 +64,13 @@
   dependencies:
     "@types/chai" "*"
 
+"@types/chai-string@^1.4.1":
```
bobbrow: I don't see anything in this PR that would require an update to `yarn.lock`.  Can you revert this?

bobbrow: I take that back.  It looks like the lock file is out of sync.

bobbrow:
Thank you @bjosa!

Issue #710, opened 2019-07-17
The 'cmake-server(7)' is deprecated.


Colengms:
Perhaps obvious, but just wanted to point out that with CMake 3.15 installed, I get the following warning when configuring:

CMake Deprecation Warning:
  The 'cmake-server(7)' is deprecated.  Please port clients to use the
  'cmake-file-api(7)' instead.

jnickborys:
Bumping this. I hit it too.

rustedwizard:
Same here.

Ian-Re:
same here.

JesvinJoseph:
same here

tkit1994:
Same here.

cestlascorpion:
same here. And how can I use the  "cmake-file-api(7)" instead?

Zingam:
You can't. That's a task for the extension to implement. I have looked up the sources and I am not sure if that would be a trivial task. So I don't expect that to change soon but for now I don't think that this warning message should bother you.

RTS007:
Same here.

Toxe:
In the meantime, is there any way to disable this warning?

Randshot:
From what I know, this requires significant changes in core parts of the extension, so I think that we won't see any update on this soon, depending on how many people are actively involved in maintaining this extension.

bobbrow:
I believe @KoeMai's PR implements this https://github.com/microsoft/vscode-cmake-tools/pull/720.  I haven't had a good chunk of time to review it yet, but it's on my short list.

Randshot:
Oh, didn't see that one. Looks promising, but like I expected a rather large effort.

nmoinvaz:
Uninstalling cmake 3.15 and installing the previous version 3.14.6 appears to work for me. Obviously its not a long term solution.

eddyekofo94:
Any fix thus far?

eddyekofo94:
Make has [this](https://cmake.org/cmake/help/v3.15/release/3.15.html#deprecated-and-removed-features) one their CMake 3.15 Release Notes.

roy-tian:
I put this line in settings.json and the warning disppears:
`"cmake.useCMakeServer": false`

Zingam:
Off-topic: Is anybody running VSCode + CMake 3.15 (particularly CMake form the snap package) on Ubuntu?

timxian:
[...]

PR #784, opened 2019-09-21
Separate code model driver interface definition from CMS driver

KoeMai:
## This change addresses item #710 CMake File API

This changes introduce a code model driver interface, based on CMake Server API. This change allows to make the driver exchangeable by the CMake File API driver. The CMake Server driver use internal the existing code model types from cms-client. This type definitions contains more information than needed by cmake tools extension. The new code model driver interface only contains extension required information. This allows to extend the code model interface dependent on the frontend needs and transform the information out of the cmake server or file API specific code model representation.
Implementation or exchange of drivers does not require modifications in the front end parts. And could be tested by driver unit tests.

- Add code model driver interface
- Use code model driver interface in frontend extension parts (tree, cpptool integration)
- Replace identification of code model providing CMake driver via CodeModelDriver abstract class.


KoeMai:
I prefer to merge and update this PR before PR #720. This PR adds tests and separates the Code Model from the CMake server implementation. The PR 720 is then only for the integration of the nee CMake FileAPI used. The FileAPI Driver should later run tests of this PR successfully.
Is this ok? Or should I merge these two PRs?

bobbrow:
Thanks for the clarification.  We'll review this one before #720 then.

src/drivers/codemodel-driver-interface.ts
```diff
@@ -0,0 +1,118 @@
...
+   * (Source CMake Documentation cmake-file-api(7))
+   *
+   * \todo clarify need of INTERFACE_LIBRARY type
+   */
+  type: TargetTypeString;
+
+  /** A string specifying the absolute path to the target’s source directory. */
+  sourceDirectory?: string;
+
+  /** Name of the target artefact on disk (library or executable file name). */
```
bobbrow: \[sp\] "artefact" -> "artifact"

src/drivers/codemodel-driver-interface.ts
```diff
@@ -0,0 +1,118 @@
...
+  fullName?: string;
+
+  /** List of absolute path of a target build artifact. */
```
bobbrow: "List of absolute paths of a target's build artifacts" ?

src/drivers/codemodel-driver-interface.ts
```diff
@@ -0,0 +1,118 @@
...
+  artifacts?: string[];
+
+  /**
+   * List of compilation information for artifacts of this target.
+   * It contains groups of source files which there compilation information.
```
bobbrow: "It contains groups of source files for which there is compilation information."

src/drivers/codemodel-driver-interface.ts
```diff
@@ -0,0 +1,118 @@
...
+   */
+  fileGroups?: CodeModelFileGroup[];
+
+  /**
+   * Represents the CMAKE_SYSROOT
+   * \todo Implementation may depend on CMAKE_SYSROOT_LINK or CMAKE_SYSROOT_COMPILE, I think CMAKE_SYSROOT_COMPILE
```
bobbrow: Do you know the answer to this question?

src/drivers/codemodel-driver-interface.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * Describes a file group to describe the build settings.
+ */
+export interface CodeModelFileGroup {
+  /** List of source files with the same compilation information */
+  sources: string[];
+
+  /** Specific's the language (C, C++, ...) for the toolchain */
```
bobbrow: \[sp\] "Specific's" -> "Specifies"

src/drivers/codemodel-driver-interface.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  /** Defines */
+  defines?: string[];
+
+  /** CMake generated file group */
+  isGenerated: boolean;
+}
+
+/**
+ * Describes cmake project and all it related targets
```
bobbrow: "it" -> "its"

src/drivers/codemodel-driver-interface.ts
```diff
@@ -0,0 +1,118 @@
+import {CMakeDriver} from '@cmt/drivers/driver';
+import * as vscode from 'vscode';
+
+/**
+ * This file contains the API description between IDE parts and the cmake model driver.
+ * This API CodeModel contains only current required cmake code model parts.
+ * There are more information provided by CMake as mapped.
```
bobbrow: "There is more information provided by CMake than is mapped."

src/drivers/cms-client.ts
```diff
@@ -224,6 +224,7 @@ export interface CodeModelFileGroup {
   includePath?: {path: string; isSystem?: boolean;}[];
   defines?: string[];
   sources: string[];
+  isGenerated: boolean;
```
bobbrow: Why do we need to keep these interfaces in `cms-client.ts`?  Shouldn't they be replaced with the new ones in `codemodel-driver-interface.ts`?

src/drivers/driver.ts
```diff
@@ -647,10 +647,12 @@ export abstract class CMakeDriver implements vscode.Disposable {
         });
         telemetryMeasures['ErrorCount'] = errorCount;
         telemetryMeasures['WarningCount'] = warningCount;
-      } else {
+      } else if (consumer) {
         // Wrong type: shouldn't get here, just in case
         rollbar.error('Wrong build result type.');
         telemetryMeasures['ErrorCount'] = retc ? 1 : 0;
+      } else {
+        // Driver tests do not provide a CMakeOutputConsumer.
```
bobbrow: Can we assert that we are in test mode?

test/unit-tests/cms-driver/cms-driver-codemodel.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    if (driver instanceof codemodel_api.CodeModelDriver) {
+      driver.onCodeModelChanged(cm => { codemodel_data = cm; });
+    }
+    await driver.configure([]);
+    expect(codemodel_data).to.be.not.null;
+    expect(codemodel_data!.configurations.length).to.be.eql(4);
+  }).timeout(90000);
+
+  test('Test project information', async () => {
+    const config = ConfigurationReader.createForDirectory(root);
```
bobbrow: All of these tests start out with the exact same code.  Can you refactor it into a helper method that returns the code model so we don't have so many copies of the same code?

test/unit-tests/cms-driver/cms-driver-codemodel.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      driver.onCodeModelChanged(cm => { codemodel_data = cm; });
+    }
+    await driver.configure([]);
+    expect(codemodel_data).to.be.not.null;
+
+    const target = codemodel_data!.configurations[0].projects[0].targets.find(t => t.type == 'UTILITY'
+                                                                                  && t.name == 'runTestTarget');
+    expect(target).to.be.not.undefined;
+
+    // may could be used to exclude file list from utility targets
```
bobbrow: "may" -> "maybe"?

test/unit-tests/cms-driver/cms-driver-codemodel.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    expect(target!.fileGroups![0].isGenerated).to.be.true;
+  }).timeout(90000);
+  // sysroot for file groups
+
+  test('Test generation of code model with one configuration like make on linux', async () => {
```
bobbrow: Logically, should this test be right after the multi-config one for VS (at the beginning of the suite)?

src/drivers/cms-client.ts
KoeMai: The code model interface contains only types that are needed by the extension. The interface types are very similar to the CMake ServerAPI types but are not identical. In the case of `isGenterated`, it was necessary to extend the cms client types to get the types easy converted into the code model types. 

I understand the driver classes as a mapper so that the ServerAPI types have a different meaning than the code model types. The FileAPI types have a different schema then the ServerAPI types. The code model types make it consistent for the CMake tools.

src/drivers/codemodel-driver-interface.ts
KoeMai: I tested it seams to forward only CMAKE_SYSROOT to the FileAPI.
CMAKE_SYSROOT_LINK and CMAKE_SYSROOT_COMPILE do not effect.

src/drivers/driver.ts
KoeMai: This modification was not necessary.

PR #827, opened 2019-10-08
Apple Clang 11.0 not detected

bobbrow:
## The purpose of this change

Fix compiler detection issues on macOS


thomthom:
I was looking at trying this before the released version. I see the extension is compiled using npm, but I'm not sure how to use the locally built extension.

bobbrow:
If you have cloned the repository and checked out this branch, you should be able to test it out by doing the following:
1. Open the repository in VS Code.
2. Open the Terminal window. (<kbd>CTL</kbd>+<kbd>`</kbd>)
3. Run `yarn install` in the Terminal (make sure the current working dir is the root of the repo)
4. Press <kbd>F5</kbd>.

This will launch a new [Extension Host] window.  It works the same as a normal VS Code instance and will let you preview the change (as well as other changes that have not yet been released).

PR #853, opened 2019-10-31
SetKit fix (kit selection is not persisted properly when triggered at…

andreeis:
setKit may be called also very early, before a CMake driver instance is created. That can happen, for example, when a project folder is just opening and it doesn't have an active kit saved, triggering the kits quick pick to show up.
Not having a CMake driver instance at the time of selecting the kit causes the selection to not be persisted properly. 
The fix is to create a new instance of the CMake driver in setKit, if none exists yet.

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -362,7 +362,8 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
     this._activeKit = kit;
     if (kit) {
       log.debug(localize('injecting.new.kit', 'Injecting new Kit into CMake driver'));
-      const drv = await this._cmakeDriver;  // Use only an existing driver, do not create one
+      //const drv = await this._cmakeDriver;  // Use only an existing driver, do not create one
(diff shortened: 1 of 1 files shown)


src/cmake-tools.ts
```diff
@@ -362,7 +362,8 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
     this._activeKit = kit;
     if (kit) {
       log.debug(localize('injecting.new.kit', 'Injecting new Kit into CMake driver'));
-      const drv = await this._cmakeDriver;  // Use only an existing driver, do not create one
+      //const drv = await this._cmakeDriver;  // Use only an existing driver, do not create one
```
bobbrow: There's a reason for this, and we don't understand it yet.  The tests fail when you call `getCMakeDriverInstance()` here.  I pushed a new branch with an alternate fix and we'll see how tests like it.

andreeis:
Abandoning this PR, better fix with https://github.com/microsoft/vscode-cmake-tools/pull/854

PR #888, opened 2019-11-08
update flow for missing CMakeLists.txt

bobbrow:

When CMakeLists.txt is not discovered in the root of the folder, give the user the open to open settings and update the cmake.sourceDirectory setting


Zingam:
@bobbrow This a very useful feature:
> When CMakeLists.txt is not discovered in the root of the folder, give the user the open to open settings and update the cmake.sourceDirectory setting
Yet I had to close and re-open the folder for the newly set path to CMakeLists.txt to be detected by the extension. Is this expected? Clicking on the Configure on the left CMake pane would pop up the error message again and again before reloading the project.

bobbrow:
No, that's not expected.  It worked for me, but I just tried another case today and I saw the behavior your described.  I'm not sure why.

Issue #886, opened 2019-11-08
 The "path" argument must be of type string - issue after update 1.2.1


paulelsner:
### Brief Issue Summary
Hi, after the update today I get the following issue:

 ```[kit] Successfully loaded 5 kits from /home/xx/repos/xxx/.vscode/cmake-kits.json
[rollbar] Unhandled exception: Unhandled Promise rejection: Post-folder-open TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined {"folder":{"uri":{"$mid":1,"fsPath":"/home/xx/repos/xxx","external":"file:///home/xx/repos/xxx","path":"/home/xx/repos/xxx","scheme":"file"},"name":"xxx","index":0}}
[rollbar] Unhandled exception: Unhandled Promise rejection: build TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined {}
```
I cannot configure and cannot build anymore :(
### Platform and Versions

- **Operating System**: remote on Ubuntu with Windows as host
- **CMake Version**: 3.15.1
- **VSCode Version**: 1.40.0
- **CMake Tools Extension Version**: 1.2.1
- **Compiler/Toolchain**: Clang 7

paulelsner:
Furhter testing shows that it is not connected to the vscode update 1.40.0 but the the extension update. And the issue happens on Mac as well on Linux. Windows works fine.

I found a workaround:
open `/home/xx/.vscode/extensions/ms-vscode.cmake-tools-1.2.1/dist/main.js`

and remove all lines from the function `getVSInstallForKit`

You do not need to search for VisualStudio installations on Linux anyhow ;)

bobbrow:
That function won't be called on linux unless there is a Visual Studio Kit in your cmake-kits.json.  Is that the case?  Do you have Visual Studio Kits committed to your repo?  Can you share your Kits file(s)?

paulelsner:
Yes there are Kits for Windows, Linux and Mac defined, since we do cross platform development.

This is my shortend Kit file
```[
    {
        "name": "My Clang 7 C++17",
        "compilers": {
            "C": "/usr/bin/clang-7",
            "CXX": "/usr/bin/clang++-7"
        },
        "cmakeSettings": {
            "CMAKE_CXX_STANDARD": "17",
            "CMAKE_CXX_STANDARD_REQUIRED": "ON",
[...]

PR #890, opened 2019-11-08
Fix for issue 886: ignore vs kits processing when not on windows

andreeis:
Fix for GitHub issue #886 . Visual Studio installations may be defined (together with other non VS tools) in the user kits json file that is parsed on linux and/or mac, in which case getVSInstallForKit should be a no-op to avoid other invalid issues down the road.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -725,20 +725,23 @@ export async function scanForClangCLKits(searchPaths: string[]): Promise<Promise
 }
 
 async function getVSInstallForKit(kit: Kit): Promise<VSInstallation|undefined> {
-    if (process.platform === "win32") {
-        console.assert(kit.visualStudio);
-        console.assert(kit.visualStudioArchitecture);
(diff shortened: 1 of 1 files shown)


src/kit.ts
```diff
@@ -725,18 +725,20 @@ export async function scanForClangCLKits(searchPaths: string[]): Promise<Promise
...
+        const match = (inst: VSInstallation) =>
+            // old Kit format
+            (legacyKitVSName(inst) == kit.visualStudio) ||
+            // new Kit format
+            (kitVSName(inst) === kit.visualStudio) ||
+            // Clang for VS kit format
+            (!!kit.compilers && kit.name.indexOf("Clang") >= 0 && kit.name.indexOf(vsDisplayName(inst)) >= 0);
+
+        return installs.find(inst => match(inst));
+    }
```
bobbrow: can you add an explicit `return undefined` for the non-windows case?

Issue #864, opened 2019-11-01
Add the ability to rescan for kits in the Kits QuickPick


Zingam:
I tried the current extension on a machine that also has BlackBerry QNX Momentics installed which contains a GCC toolchain. VSCode finds the the QNX GCC toolchain as a kit but fails to find MSVC (I have Visual Studio 2019 Preview installed).
On another machine which contains only Visual Studio 2019, VS2019 Preview and Clang LLVM 9.0. For some reason the only toolkit that the CMake Tools find is called Visual Studio Enterprise:
![Untitled](https://user-images.githubusercontent.com/3337419/68024559-1bf69080-fcb3-11e9-8802-3e1de11c62e1.png)
There is no Visual Studio 2019 Community Version and no Visual Studio 2019 Community Preview version.
Maybe I used to have a preview version of VS Enterprise once installed (long time ago before the release of VS2019).
The extension also doesn't find Clang. It is not in the path but maybe it should look for a variable LLVM_ROOT or something.

bobbrow:
1.2.0 should have a fix for this.  We didn't get it ready in time to release this week.  We're targeting Monday now.

bobbrow:
CMake Tools 1.2.0 is now available in the VS Code Marketplace or on the [Releases Page](https://github.com/microsoft/vscode-cmake-tools/releases/tag/v1.2.0).

Because the publisher id has changed, VS Code does not automatically upgrade you yet. The Marketplace team will enable this in the next few days.  If you want to install 1.2.0 now, please make sure you uninstall any earlier versions of CMake Tools manually to avoid any conflicts.

Zingam:
This is not fixed. I still see VS Enterprise Preview instead of any of the VS 2019 Community versions that I have installed.

bobbrow:
Did you rerun the "Scan for Kits" command?

Zingam:
You are right! I have completely forgotten about that command:

> User-Local Kits can be updated by running Scan for Kits in the command palette. The following process occurs to find available kits:

I'd click on the status bar to select a kit and I'd think that's all I needed to do. Perhaps the select menu kit drop-down list could feature a Re-Scan option to remind to noobs like me that there is such an option.

bobbrow:
That's a good idea.  I'll change the issue title and turn this into a feature request.

PR #893, opened 2019-11-09
Add 'Scan for Kits' option to Kits' QuickPick

Zingam:
Adds the ability to rescan for kits in the Kits QuickPick

## This change addresses item

Add the ability to rescan for kits in the Kits QuickPick #864

Adds an option to rescan the kits as a special kit item to the QuickPick.
Moves all special kits to the top of the list in the QuickPick.


Zingam:
This is a disaster. I'll have to redo this PR.

PR #894, opened 2019-11-09
Clang: Use environment variable LLVM_ROOT when scanning for kits

Zingam:
LLVM_ROOT is a possible, common environment variable that points to a LLVM/Clang installation
This change adds the ability to use LLVM_ROOT to locate clang-cl on Windows in case LLVM is not installed in the default location.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -826,8 +826,16 @@ export async function scanForKits(opt?: KitScanOptions) {
     prs = prs.concat(compiler_kits);
     if (isWin32) {
       const vs_kits = scanForVSKits(pr);
-
-      const clang_cl_path = ['C:\\Program Files (x86)\\LLVM\\bin', 'C:\\Program Files\\LLVM\\bin', ...scanPaths];
+      const clang_cl_path: string[] = [];
+      const defaultPaths = ['C:\\Program Files (x86)\\LLVM\\bin', 'C:\\Program Files\\LLVM\\bin', ...scanPaths];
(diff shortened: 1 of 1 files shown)


src/kit.ts
```diff
@@ -826,8 +826,16 @@ export async function scanForKits(opt?: KitScanOptions) {
     prs = prs.concat(compiler_kits);
     if (isWin32) {
       const vs_kits = scanForVSKits(pr);
-
-      const clang_cl_path = ['C:\\Program Files (x86)\\LLVM\\bin', 'C:\\Program Files\\LLVM\\bin', ...scanPaths];
+      const clang_cl_path: string[] = [];
+      const defaultPaths = ['C:\\Program Files (x86)\\LLVM\\bin', 'C:\\Program Files\\LLVM\\bin', ...scanPaths];
+      if (typeof(process.env.LLVM_ROOT) !== 'undefined') {
```
bobbrow: Can you just do a "falsy" test here instead?
```
if (process.env.LLVM_ROOT) {
  ...
}
```

src/kit.ts
```diff
@@ -826,8 +826,16 @@ export async function scanForKits(opt?: KitScanOptions) {
...
+        const llvmRoot = path.normalize(process.env.LLVM_ROOT + "\\bin");
+        const llvmPath = defaultPaths.find(p => p === llvmRoot );
```
bobbrow: style: remove the space before the closing `)`

bobbrow:
Replaced by #930

Issue #895, opened 2019-11-10
Scan for kits fails if there is a non-executable file with base name "clang-cl" in PATH


Lightirius:
### Brief Issue Summary
Apparently cmake-tools extension tries to execute any file with base name "clang-cl" found in PATH and fails on non-executable files.

### Apparent Behavior:

1. Add any folder with "clang-cl.txt" file to PATH
2. Run cmake-tools scan for kits command
3. Scan for kits fails with:
[rollbar] Unhandled exception: Unhandled Promise rejection: scanForKits Error: spawn UNKNOWN {}

### CMake Tools Log

[extension] [9178] cmake.scanForKits started
[extension] Rescanning for kits
[kit] Scanning for Kits on system
[kit] Checking for kit: Visual Studio Build Tools�2019 Release - x86
[kit] Checking for kit: Visual Studio Build Tools�2019 Release - amd64
[kit] Checking for kit: Visual Studio Build Tools�2019 Release - x86_amd64
[kit] Checking for kit: Visual Studio Build Tools�2019 Release - x86_arm
[kit] Checking for kit: Visual Studio Build Tools�2019 Release - amd64_arm
[kit] Checking for kit: Visual Studio Build Tools�2019 Release - amd64_x86
[kit] Scanning directory C:\Windows\system32 for compilers
[kit] Scanning directory C:\Program Files (x86)\Common Files\Oracle\Java\javapath for compilers
[kit] Scanning directory C:\Windows for compilers
[kit] Scanning directory C:\Windows\System32\WindowsPowerShell\v1.0\ for compilers
[kit] Scanning directory C:\Program Files\dotnet\ for compilers
[kit] Scanning directory C:\Windows\System32\OpenSSH\ for compilers
[kit] Scanning directory C:\Windows\System32\Wbem for compilers
[kit] Scanning directory C:\Program Files\7-Zip for compilers
[kit] Scanning directory D:\apps\scoop\apps\yarn\current\Yarn\bin for compilers
[kit] Skipping scan of not existing path D:\apps\scoop\apps\yarn\current\global\node_modules\.bin
[kit] Found 0 kits in directory D:\apps\scoop\apps\yarn\current\global\node_modules\.bin
[kit] Scanning directory D:\apps\scoop\apps\python\current for compilers
[kit] Scanning directory D:\apps\scoop\apps\nodejs\current for compilers
[kit] Scanning directory D:\apps\scoop\apps\nodejs\current\bin for compilers
[...]

PR #896, opened 2019-11-10
Handle exceptions on testing non-executable 'clang-cl*' files

Zingam:
This fixes: https://github.com/microsoft/vscode-cmake-tools/issues/895

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -116,7 +116,13 @@ interface ClangVersion {
 
 async function getClangVersion(binPath: string): Promise<ClangVersion|null> {
   log.debug(localize('testing.clang.binary', 'Testing Clang binary: {0}', binPath));
-  const exec = await proc.execute(binPath, ['-v']).result;
+  let exec : proc.ExecutionResult;
+  try {
+    exec = await proc.execute(binPath, ['-v']).result;
+  } catch (e) {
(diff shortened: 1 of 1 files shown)


bobbrow:
Something like this: https://github.com/microsoft/vscode-cmake-tools/commit/4c1c3f5de377370976f63edc878d64af2ffae824

Zingam:
@bobbrow Thanx! I'll look into this deeper later and will also make the necessary changes to the call sites.

bobbrow:
I can just submit my fix if you want.  You shouldn't need to update any call sites.  This will take care of all of them

Zingam:
I think this error message will be confusing and needs to be updated:
`log.debug(localize('bad.clang.binary', 'Bad Clang binary ("-v" returns non-zero): {0}', binPath));`
and we could also log a debug message in  getCMakeExecutableInformation(path: string)

Zingam:
If you think that's not necessary. Then please submit your fix and I'll delete mine.

bobbrow:
I thought the message was ok.  It says that clang-cl.txt is a bad clang binary.

bobbrow:
It's part of the debug log, so it has more details about what went wrong. I'm ok with changing it though. Maybe showing the full command line would be better?

Zingam:
@bobbrow That might be a good idea actually.

Issue #885, opened 2019-11-08
Can't choose MSYS Generator after update


xubury:
Hi, I have this issue that I can't use MSYS Makefiles after an update. I am using MSYS2+mingw64 and I have to use MSYS Makefiles to configure normally.So I set "cmake.generator" to "MSYS Makefiles", but it keep switching to Mingw Makefiles for some reason. See the screenshot below.
Really appreciate any help, Thanks!
![20191108162136](https://user-images.githubusercontent.com/46261364/68461381-1534c400-0245-11ea-9bc8-0e9ff87f1b2c.png)

xubury:
So I thought editing cmake-tools-kits.json would work, but after setting generator variable to MSYS makefiles it switch to Unix Makefiles.
Manually using `cmake -G  "MSYS Makefiles"` works which means my cmake should be able to work with MSYS Makefiles.
![20191108164643](https://user-images.githubusercontent.com/46261364/68462417-73fb3d00-0247-11ea-8ab6-bf6189fa65e4.png)
![20191108164713](https://user-images.githubusercontent.com/46261364/68462591-cb011200-0247-11ea-97d8-e579967a2e0a.png)

xubury:
This problem may be related to #880 
If I don't use the mingw64 shell, i got the same error.
![20191108204428](https://user-images.githubusercontent.com/46261364/68477821-5a6aed00-0269-11ea-8b7a-7dfbc099aa35.png)

bobbrow:
1.2.2 was released today.  Please give that a try and let us know if you continue to have issues.

Issue #880, opened 2019-11-07
VSCode cannot find CMake generator


tnakaicode:
### Brief Issue Summary

I update VSCode 1.40.0. And, CMake cannot find CMake generator.
In previous VSCode version, I can use CMake/Build.

### CMake Tools Log

```
Unable to determine what CMake generator to use. Please install or configure a preferred generator, or update settings.json, your Kit configuration or PATH variable. Error: No usable generator found.
```

### Platform and Versions

- **Operating System**: Windows 10
- **CMake Version**: 3.14.0
- **VSCode Version**: 1.40.0
- **CMake Tools Extension Version**: 1.2.1
- **Compiler/Toolchain**: mingw-w64 gcc, g++

mischelebuha:
The same problem with "settings.json" and "cmake-kits.json...

chillophil:
I had the same issue after upgrading VSCode to version 1.40.0.

Previously I had set my PATH environment via settings.json `cmake.environment`. After the update it seems CMake-Tools were unable to find my toolset (generator: Unix Makefiles).

I fixed this issue with setting `environmentVariables` in cmake-kits.json for my toolchain.

bobbrow:
Can you share any of your settings?  I'm not able to reproduce this problem.

**settings.json**
```
{
    "cmake.mingwSearchDirs": [
        "C:\\mingw-w64\\mingw64\\bin"
    ],
    "cmake.generator": "MinGW Makefiles"
}
```

**cmake-tools-kits.json**
```
  ...
  {
    "name": "GCC 7.2.0",
    "compilers": {
      "C": "C:\\mingw-w64\\mingw64\\bin\\gcc.exe",
      "CXX": "C:\\mingw-w64\\mingw64\\bin\\g++.exe"
    },
    "preferredGenerator": {
      "name": "MinGW Makefiles"
    },
    "environmentVariables": {
      "CMT_MINGW_PATH": "C:\\mingw-w64\\mingw64\\bin"
    }
  },
  ...
```

bobbrow:
1.2.2 was released today.  Please give that a try and let us know if you continue to have issues.

tnakaicode:
I solved! Thank you.

VSCode: v1.40.1
CMake Tolls: v1.2.2

renebarto:
I'm running 1.41.1 and I still have problems. Maybe this is specifically due to using a toolchain file, but I still get the same error:

[...]

PR #906, opened 2019-11-12
User defined generator has priority over any other selection criteria

andreeis:
Issues like https://github.com/microsoft/vscode-cmake-tools/issues/880 and https://github.com/microsoft/vscode-cmake-tools/issues/885 started to occur after the 1.2 update because the following logic changed, starting with https://github.com/microsoft/vscode-cmake-tools/commit/6d04ebd7d1e86c11d17f28296a0d9d22e2fabee0:
- before the above commit, the extension would trigger the 'best generator' search only if the user did not define in settings (via "cmake.generator") any specific generator
- after the above commit, the 'best generator' search is triggered regardless of the cmake.generator setting. Even if it includes its value, it can happen that some other generator be picked instead, in case any criterias are not met (like tools not being on the path).

Some other aspect of issue 885 is still fixed by PR https://github.com/microsoft/vscode-cmake-tools/pull/903. That change is needed for the scenario when the user would not define any "cmake.generator" in settings, but would define a preferred generator in the kits json. Even if make.exe would be on the path, making MSYS Makefiles generation possible, the extension would still pick something else instead (without that fix), depending what will be found in the path (Ninja, Unix or MingW...).

Issue 880 can be worked around if mingw32-make.exe is on the path, or pointed to via settings. But with this fix, configuring would still work properly, as it used to up to 1.1.3 release.

### DIFF
--- a/src/drivers/driver.ts
+++ b/src/drivers/driver.ts
@@ -320,7 +320,17 @@ export abstract class CMakeDriver implements vscode.Disposable {
     if (kit.preferredGenerator)
       preferredGenerators.push(kit.preferredGenerator);
 
-    this._generator = await this.findBestGenerator(preferredGenerators);
+    // Use the "best generator" selection logic only if the user did not define already
+    // in settings (via "cmake.generator") a particular generator to be used.
+    if (this.config.generator) {
+      this._generator = {
+        name: this.config.generator,
+        platform: this.config.platform || undefined,
+        toolset: this.config.toolset || undefined,
+      };
+    } else {
+      this._generator = await this.findBestGenerator(preferredGenerators);
+    }
   }
 
   protected abstract doSetKit(needsClean: boolean, cb: () => Promise<void>): Promise<void>;


src/drivers/driver.ts
```diff
@@ -320,7 +320,17 @@ export abstract class CMakeDriver implements vscode.Disposable {
     if (kit.preferredGenerator)
       preferredGenerators.push(kit.preferredGenerator);
 
-    this._generator = await this.findBestGenerator(preferredGenerators);
+    // Use the "best generator" selection logic only if the user did not define already
+    // in settings (via "cmake.generator") a particular generator to be used.
+    if (this.config.generator) {
```
bobbrow: Would it make sense to put this logic in findBestGenerator instead?

src/drivers/driver.ts
```diff
@@ -320,7 +320,17 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+      this._generator = {
```
bobbrow: I believe this needs to be a `Promise.resolve( {...} )`

PR #907, opened 2019-11-12
Changelog and version update for CMake Tools Extension Release #1.2.2

andreeis:
Changelog and version update for CMake Tools Extension Release #1.2.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # What's New?
 
+## 1.2.2
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # What's New?
 
+## 1.2.2
+
+
```
bobbrow: Can you list the bugs we fixed here?

PR #937, opened 2019-11-21
'Scan for Kits' option to Kits' QuickPick

Zingam:
# This change addresses item #864

- Adds an option to re-scan the kits as a special kit item to the QuickPick.
- Moves all special kits to the top of the list in the QuickPick.

# The purpose of this change

To enable/remind the user to easily invoke the command  Rescan for Kits from the Kits QuickPick.

# Other Notes/Information

I had to resubmit this #893 due to a failed merge.


bobbrow:
I'm going to postpone this one temporarily until multi-root is committed since much of this code has moved around and I don't want it to get lost.

bobbrow:
Multi-root is committed.  Can you merge and resolve conflicts?

Zingam:
I guess this will need more fixes.

src/kit.ts
```diff
@@ -22,6 +22,14 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 
 const log = logging.createLogger('kit');
 
+/**
```
Zingam: @bobbrow I think it makes sense to replace these special kit variables with a enum. Also do we really need to have the type `export type scanforkitsKitType = '__scanforkits__';`

Zingam:
I think this is OK now.

src/kit.ts
```diff
@@ -901,6 +909,9 @@ export async function descriptionForKit(kit: Kit): Promise<string> {
     const compilers = Object.keys(kit.compilers).map(k => `${k} = ${kit.compilers![k]}`);
     return localize('using.compilers', 'Using compilers: {0}', compilers.join(', '));
   }
+  if (kit.name === scanforkitsKitName) {
+    return localize('rescanning.for.kits', 'Rescanning for kits');
```
bobbrow: "Search for compilers on this computer"

src/kit.ts
bobbrow: I don't have a strong opinion here.

Zingam: Well I did convert these to an enum in my latest commit.

Issue #610, opened 2018-12-14
CMake not found but is part of Visual Studio installation. Could automatically check


fedepad:
I start by saying that I don't know TypeScript but can follow somehow code.

In a Windows system, CMake can be installed as part of Visual Studio and not necessarily as a separate standalone program downloaded from the CMake website.
For example, in my case, I have Windows 10 and Visual Studio Community 2017 version 15.9.4 .
I installed CMake from within Visual Studio because it is part of one of the workload component.
In such a situation (my case), CMake and the related tools, get installed in the following path

C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin

in which one finds: cmake.exe, cpack.exe, ctest.exe and cmcldeps.exe .

So in this case there is a CMake in the system, but it's not found by vscode-cmake-tools because it's neither in the `PATH` or in one of the two hard-coded guesses 
https://github.com/vector-of-bool/vscode-cmake-tools/blob/358d8dc533168b94ecec73c32af81659e3c82975/src/paths.ts#L137
I was thinking that maybe one could add a field to
https://github.com/vector-of-bool/vscode-cmake-tools/blob/358d8dc533168b94ecec73c32af81659e3c82975/src/kit.ts#L342
like `hasCMake` and then check for the existence by building the path similarly to
https://github.com/vector-of-bool/vscode-cmake-tools/blob/358d8dc533168b94ecec73c32af81659e3c82975/src/kit.ts#L501
having something like this:
`const hasCMakeVSBuiltin = fs.exists(path.join(inst.installationPath, 'Common7', 'IDE', 'CommonExtensions', 'Microsoft', 'CMake', 'CMake', 'bin', 'cmake.exe')) ;`

and set to true or false accordingly the `hasCmake` field.
Then for example one could add this candidate to
https://github.com/vector-of-bool/vscode-cmake-tools/blob/358d8dc533168b94ecec73c32af81659e3c82975/src/paths.ts#L137

or similar.
By automatically finding CMake among the extensions by checking that path for VS installation then I think it would ease the setup and avoid some manual tuning (I know, it's minor if you know where CMake was installed as part of VS).
Sure one can write that path in the settings.json, that is what I did.

andreeis:
Closing since this is fixed in CMake Tools Beta2. Please reactivate if you encounter any other symptoms of this issue.

PR #942, opened 2019-11-23
Detection of bundled CMake in Visual Studio

Zingam:
## This change addresses item #610

- Moved some *common* functions into a new package 'installs/visual-studio'.

## The purpose of this change

Enables the detection of bundled CMake packages in Visual Studio.

## Other Notes/Information

A future improvement could be to scan the system once or on demand for all CMake versions and let the user pick one via the UI. When 'cmake.cmakePath' is not set  the extension appears to query the system for CMake executables continuously.


src/paths.ts
```diff
@@ -128,19 +130,51 @@ class Paths {
...
+            for (const tool_path of bundled_tool_paths) {
+              if (await fs.exists(tool_path.cmake)) {
+                // Append the bundled Ninja build system path to the VSCode's process' environment variable 'PATH'
+                if (await fs.exists(tool_path.ninja)) {
+                  if (process.env.hasOwnProperty('PATH')) {
+                    const env_paths = (process.env.PATH as string).split(';');
+                    const ninja_path = path.dirname(tool_path.ninja);
+                    const ninja_base_path = env_paths.find(path_el => path_el === ninja_path);
+                    if (ninja_base_path === undefined) {
+                      (process.env.PATH as string) = (process.env.PATH as string).concat(';' + ninja_path);
```
bobbrow: Changing the process.env only works inside the node process and I wouldn't expect it to work automatically for spawned processes like CMake.  Can you split this out into a new function to get the Ninja path that we can call from `getVSKitEnvironment` instead?

Zingam:
@bobbrow I'll wait for Multi-Root, File API, etc to be merged before I go over my current PRs again.

bobbrow:
Multi-root is in.  I don't think this will conflict with file-api.

src/installs/visual-studio.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * Module for querying MS Visual Studio
```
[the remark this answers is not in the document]
Zingam: Are these concerns still valid?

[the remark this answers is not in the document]
bobbrow: It looks like it merged cleanly, but I don't think you should bring as much stuff in here.  Really, this file would just be about querying the installation information.  Some of the other stuff in here is more related to Kits.  I'll mark the functions that I think should move back to `kit.ts`.

src/installs/visual-studio.ts
```diff
@@ -0,0 +1,118 @@
...
+import * as logging from '../logging';
+import * as proc from '../proc';
+import {thisExtensionPath} from '../util';
+import * as nls from 'vscode-nls';
+
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+const log = logging.createLogger('kit');
```
bobbrow: I think this logger should have a different name.

src/installs/visual-studio.ts
```diff
@@ -0,0 +1,118 @@
...
+let cachedVSInstallations: VSInstallationCache|null = null;
+
+
+/**
+ * Construct the display name (this will be paired with an
+ * arch later to construct the Kit.name property).
+ *
+ * @param inst The VSInstallation to use
+ */
+export function vsDisplayName(inst: VSInstallation): string {
```
bobbrow: This should go back to `kit.ts`

src/installs/visual-studio.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+  cachedVSInstallations = {
+    installations: installs,
+    queryTime: now
+  };
+  return installs;
+}
+
+export function vsVersionName(inst: VSInstallation): string {
```
bobbrow: This should go back to `kit.ts`

Zingam:
I hope this initial version is OK for now. I plan to implement a more sophisticated CMake detection later -> https://github.com/microsoft/vscode-cmake-tools/issues/943#issuecomment-564397901

src/paths.ts
```diff
@@ -128,28 +142,70 @@ class Paths {
...
+      }
+
+      for (const tool_path of bundled_tool_paths) {
+        if (await fs.exists(tool_path.cmake)) {
+          // Append the bundled Ninja build system path to the VSCode's process' environment variable 'PATH'
+          if (await fs.exists(tool_path.ninja)) {
+            vsCMakePaths.ninja = tool_path.ninja;
+          }
+          // CMake can be still used without Ninja
+          vsCMakePaths.cmake = tool_path.cmake;
```
bobbrow: Should this break out of the loop once you find one that works?

Zingam: Yes, that was the intention. Thank you.

PR #948, opened 2019-12-01
Fix cancel build

KoeMai:
### This changes cancel process for CMake Server driver

- Stops CMake-Server on cancel, by ending of cmake server pipe
- The cancel process waiting until return of build process

## The purpose of this change

- Build commands which are aborted can then be reused
- No stuck on start of cmake driver by multiply cmake server on same build directory


src/cmake-tools.ts
```diff
@@ -784,6 +783,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
       consumer.dispose();
     }
   }
+  /**
+   * Implementation of `cmake.build`
+   */
+  async build(target_?: string): Promise<number> {
+    this.m_promise_build = this.RunBuild(target_);
+    const run_build = this.m_promise_build;
```
bobbrow: Do we need these two intermediate variables or can you just return `await run_build` here?

src/drivers/driver.ts
```diff
@@ -899,20 +908,26 @@ export abstract class CMakeDriver implements vscode.Disposable {
       return null;
   }
 
+  /**
+   * Is called then the current process should be stopped.
```
bobbrow: \[sp\] `Is` -> `If`

test/unit-tests/cms-driver/cms-driver.test.ts
```diff
@@ -62,19 +62,24 @@ suite('CMake-Server-Driver tests', () => {
...
-
-    if (!cleanupBuildDir(path.join(defaultWorkspaceFolder, 'build'))) {
+    let isDone = false;
+    if (!isDone && !cleanupBuildDir(path.join(defaultWorkspaceFolder, 'build'))) {
       done('Default build folder still exists');
+      isDone = true;
     }
 
-    if (!cleanupBuildDir(path.join(emptyWorkspaceFolder, 'build'))) {
+    if (!isDone && !cleanupBuildDir(path.join(emptyWorkspaceFolder, 'build'))) {
```
bobbrow: I don't understand the logic change here.  Shouldn't we attempt to clean up all build folders?

src/cmake-tools.ts
KoeMai: It is only possible to have one waiting function on a promise. I had the problem that the build method and the stop build method locked each other. The m_promise_build synchronizes different tasks. The copy of the m_promise ensures that the build function can wait.

bobbrow: Promises aren't copied, right?  I was just suggesting trimming some of the code that looked unnecessary.
```ts
  async build(target_?: string): Promise<number> {
    this.m_promise_build = this.RunBuild(target_);
    return await this.m_promise_build;
  }
```
You're saying that that doesn't work?

KoeMai: I had some trouble with the `m_promise_build` Promise, but I don´t remember the exact problem. This helper variable does not make sense.

bobbrow:
Thank you for fixing this!

Issue #863, opened 2019-11-01
Duplicated launch targets


Zingam:
Maybe this is fixed already but I am unable to find such issue.
Whenever I need to select a launch target I see duplicated entries:

![Screenshot from 2019-11-01 13-41-51](https://user-images.githubusercontent.com/3337419/68022823-16e31280-fcae-11e9-8d3b-06c1da08f62f.png)

Maybe whenever there is just one target in the project it should be selected by default and no select popup menu should be displayed.

Zingam:
I can still see the duplicated entries in 1.2.1.

KoeMai:
I tried to [reconstruct ](https://github.com/KoeMai/vscode-cmake-tools/commit/d18925d3c07b28d8338a4447044858602d616972) the dublicate entries. But I don´t see it. Are this duplicate launch configurations or execution targets? When is the dialog shown? How does the launch configuration looks like? And which tool versions are used?

Zingam:
When I click on "Debug" on the status bar. I have been observing this for a while now with this project of mine:
https://github.com/Zingam/GunBox
I tried to quickly debug it but couldn't get any idea what was going on. The above screenshot was made on Ubuntu. Here a a couple of screenshot on Windows.

Here I clicked: [all]  "Set the active target to build" on the status bar.

![Untitled](https://user-images.githubusercontent.com/3337419/69029201-f53b9800-09dc-11ea-83f1-95bb875bc64d.png)


Here I clicked: [Debug] "Select target to launch" on the status bar.

![Untitled2](https://user-images.githubusercontent.com/3337419/69029226-02588700-09dd-11ea-9a07-11ce5619c94e.png)

KoeMai:
The project contains to executable targets from different projects (see cmake: project outline). 
May be it is helps to add the project into the target description.

![twoExecuableTargets](https://user-images.githubusercontent.com/21360085/69124841-df91a580-0aa4-11ea-858e-7c902a79ad88.PNG)

Zingam:
@KoeMai This is puzzling. I have just one add_executable in the source. Also that other library ThirdParty/glad also appears twice.
I use Visual Studio 2019 mostly and I don't observe that kind of duplicated executable targets there.

![image](https://user-images.githubusercontent.com/3337419/69191847-ffb87780-0b2b-11ea-8ff5-2a733fe51719.png)

Zingam:
These are the debug targets as seen in VS2019 and exactly as expected:

[...]

PR #958, opened 2019-12-06
Multi-root support

xisui-MSFT:
The major UI differences comparing to the original version of CMake Tools are:
1. Status bar (blue bar on the bottom of the screen) should show the “active folder”, and all status after are corresponding to the current active folder. By default, active folder changes based on the file being viewed/edited in vscode. This behavior can be changed by user/workspace setting “Auto select active folder” to false. If so, clicking on the “active folder” on status bar will show a menu for picking the desired active folder. This change does not affect single root case.
1. A lot of commands with “all”/”for all projects”, such as “configure all”, “build all” and “debug all”.
1. For project outline view, top level buttons are changed to “-all” version of the commands.


package.json
```diff
@@ -256,19 +365,32 @@
...
+        "title": "%cmake-tools.command.cmake.folders.selectActiveFolder.title%"
       }
     ],
     "menus": {
       "commandPalette": [
         {
           "command": "cmake.configure"
         },
+        {
+          "command": "cmake.configureAll"
```
bobbrow: Does it make sense to have these "All" commands available when doing single root?  Can we add a "when" clause that enables these only when multi-root is active?  (use `util.setContextValue` to define a variable that can be checked here)

package.nls.json
```diff
@@ -4,32 +4,45 @@
     "cmake-tools.command.cmake.scanForKits.title": "Scan for Kits",
     "cmake-tools.command.cmake.selectKit.title": "Select a Kit",
     "cmake-tools.command.cmake.setVariant.title": "Select Variant",
+    "cmake-tools.command.cmake.setVariantAll.title": "Select Variant for All Projects",
     "cmake-tools.command.cmake.configure.title": "Configure",
+    "cmake-tools.command.cmake.configureAll.title": "Configure All",
```
bobbrow: use "All Projects" to be consistent with the other labels.

package.json
xisui-MSFT: I think somethingall = something under single root mode make more sense, so the user can avoid some extra work when writing tests for an extension depending on cmake tools.

bobbrow: This just determines whether the command appears in the command palette or not. I believe extensions can still invoke the command (though we should check).

Zingam:
@xisui-MSFT Does this PR address this issue (which I think is an issue) #863
Basically if you have multiple `project()` commands in your CMakeLists.txt hierarchy a new entry is added for each `project` with its subsequent sub-`projects`, resulting in multiple entries of the same projects and targets, etc. Also clicking on the duplicated Target items in the Outline causes this error message:
image

xisui-MSFT:
@Zingam I don't think so. But since looks like you have already found the problem, you can always create a pull request yourself :).

Zingam:
@xisui-MSFT Thank you for the reply. I think that any fixes of that "issue" have to be discussed with MSFT and also all the multi-root and file-api stuff will have to be merged first. I'll create a simple multi-root test sample.

src/tree.ts
```diff
@@ -427,30 +428,81 @@ class ProjectNode extends BaseNode {
...
+  private readonly _folders = new Map<string, WorkspaceFolderNode>();
+
+  addAllCurrentFolders() {
+    for (const wsf of vscode.workspace.workspaceFolders || []) {
+      this._folders.set(wsf.name, new WorkspaceFolderNode(wsf));
+    }
+  }
+
+  addFolder(folder: vscode.WorkspaceFolder) {
+    this._folders.set(folder.name, new WorkspaceFolderNode(folder));
```
bobbrow: Does this work with multiple folders having the same name?

src/tree.ts
```diff
@@ -427,30 +428,81 @@ class ProjectNode extends BaseNode {
...
-export class ProjectOutlineProvider implements vscode.TreeDataProvider<BaseNode> {
-  private readonly _changeEvent = new vscode.EventEmitter<BaseNode|null>();
-  get onDidChangeTreeData() { return this._changeEvent.event; }
+interface ExternalUpdateContext {
+  launchTargetName: string|null;
+  defaultTarget?: string;
+}
 
+class WorkspaceFolderNode extends BaseNode {
+  constructor(readonly wsFolder: vscode.WorkspaceFolder) { super(`wsf/${wsFolder.name}`); }
```
bobbrow: Does this work with multiple folders having the same name?

src/tree.ts
```diff
@@ -460,14 +512,18 @@ export class ProjectOutlineProvider implements vscode.TreeDataProvider<BaseNode>
...
-      if (!node) {
-        // Request for root node
-        return this._children;
-      } else {
+      if (node) {
         return node.getChildren();
       }
+      // Request for root nodes
+      if (this._folders.size <= 1) {
+        for (const folder of this._folders.values()) {
```
bobbrow: I'm a little confused about the scenario that this covers.  I'll drop by and ask about it.

src/util.ts
```diff
@@ -543,4 +534,26 @@ export function checkFileExists(filePath: string): Promise<boolean> {
...
+  }
+}
+
+export function chokidarOnAnyChange(watcher: chokidar.FSWatcher, listener: (path: string, stats?: fs.Stats | undefined) => void) {
+  return watcher.on('add', listener)
+                .on('change', listener)
+                .on('unlink', listener);
+}
+
+export function isString(x: any) {
```
bobbrow: This is how we do it in cpptools:
```
export function isString(input: any): input is string {
    return typeof(input) === "string";
}
```
This has the benefit of also promoting `input` to `string` for anything inside the block after `if (isString(x))`

src/tree.ts
```diff
@@ -460,14 +512,18 @@ export class ProjectOutlineProvider implements vscode.TreeDataProvider<BaseNode>
...
+          return folder.getChildren();
+        }
+      }
+      return [...this._folders.values()];
     } catch (e) {
-      rollbar.error(localize('error.rendering.children.nodes', 'Error while rendering children nodes'));
+      rollbar.error('Error while rendering children nodes');
```
bobbrow: Can you restore the localized string?

src/tree.ts
```diff
@@ -427,30 +428,81 @@ class ProjectNode extends BaseNode {
...
+
+  removeFolder(folder: vscode.WorkspaceFolder) {
+    this._folders.delete(folder.name);
+    this._changeEvent.fire(null);
+  }
+
+  updateCodeModel(folder: vscode.WorkspaceFolder, model: cms.CodeModelContent|null, ctx: ExternalUpdateContext) {
+    let existing = this._folders.get(folder.name);
+    if (!existing) {
+      rollbar.error('Updating code model on folder that does not yet exist?');
```
bobbrow: Should this be localized?

src/variant.ts
```diff
@@ -234,18 +231,14 @@ export class VariantManager implements vscode.Disposable {
   private async _reloadVariantsFile(filepath?: string) {
     const validate = await loadSchema('schemas/variants-schema.json');
 
-    const workdir = util.getPrimaryWorkspaceFolder();
-    if (!workdir) {
-      // Can't read, we don't have a dir open
-      return;
-    }
+    const workdir = this.folder.uri.fsPath;
```
bobbrow: Is `this.folder` always an object?

src/workspace.ts
```diff
@@ -13,9 +15,9 @@ import {StateManager} from '@cmt/state';
 export class DirectoryContext {
   constructor(
       /**
-       * The directory path that this context corresponds to
+       * Absolute path to the directory associated with this context
```
bobbrow: I wouldn't say that this is an "absolute path".  This is a folder object.

src/workspace.ts
```diff
@@ -31,9 +33,9 @@ export class DirectoryContext {
    * @param dir The directory for which to create a context
```
bobbrow: `@param` was renamed.

src/folders.ts
```diff
@@ -0,0 +1,190 @@
...
+    this.cmakeTools.dispose();
+    this.kitsController.dispose();
+  }
+}
+
+export class CMakeToolsFolderController implements vscode.Disposable {
+  private readonly _instances = new Map<string, CMakeToolsFolder>();
+
+  private readonly _beforeAddFolderEmitter = new vscode.EventEmitter<vscode.WorkspaceFolder>();
+  private readonly _afterAddFolderEmitter = new vscode.EventEmitter<CMakeToolsFolder>();
```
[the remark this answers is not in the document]
bobbrow: Hmm... nevermind.  It looks like you can access the cmakeTools instance and the kitsController too.

---
In reply to: [361532229](https://github.com/Microsoft/vscode-cmake-tools/pull/958#discussion_r361532229) [](ancestors = 361532229)

test/extension-tests/successful-build/test/toolchain.test.ts
```diff
@@ -17,7 +17,9 @@ suite('[Toolchain Substitution]', async () => {
     testEnv = new DefaultEnvironment('test/extension-tests/successful-build/project-folder', 'build', 'output.txt');
     cmt = await CMakeTools.create(testEnv.vsContext, testEnv.wsContext);
 
-    const kits = await kitsAvailableInWorkspaceDirectory(testEnv.projectFolder.location);
+    const user_kits = await readKitsFile(USER_KITS_FILEPATH);
```
bobbrow: Why does this read the User kits now? I don't think it's supposed to.

src/tree.ts
xisui-MSFT: Restore the localized string?

src/tree.ts
xisui-MSFT: I don't think it would be helpful. But we can if you feel strongly.

src/extension.ts
```diff
@@ -1105,25 +643,125 @@ class ExtensionManager implements vscode.Disposable {
...
 
+  private _cleanOutputChannel(cmt: CMakeTools) {
+    if (cmt.workspaceContext.config.clearOutputBeforeBuild) {
+      log.clearOutputChannel();
+    }
+  }
+
   // The below functions are all wrappers around the backend.
+  async mapCMakeTools(fn: CMakeToolsMapFn): Promise<any>;
+  async mapCMakeTools(cmt: CMakeTools|undefined, fn: CMakeToolsMapFn): Promise<any>;
```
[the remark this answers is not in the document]
xisui-MSFT: It's vector-of-bool's code. I guess it feels more natural to specify CMakeTools then what to do with it, just like 'sth => ...'. But sure I can change this.

src/tree.ts
bobbrow: Yes, the code used to be: `localize('error.rendering.children.nodes', 'Error while rendering children nodes')`, but now it is just: `'Error while rendering children nodes'`.  The string should be localized.

src/variant.ts
xisui-MSFT: Yes. There's one variant instance for each folder, and folder is passed in the constructor. It is only constructed in cmake-tools.ts line 153. Since there must be a folder for CMakeTools, this.folder is always an object.

src/tree.ts
xisui-MSFT: Ah read it wrong. Sure I'll change it back. Not sure why I changed it thought...

src/workspace.ts
xisui-MSFT: vector-of-bool's code and comment. Will fix.

src/tree.ts
bobbrow: If it is an error that developers should see, then it needs to be localized.  If it is not an error and/or doesn't need to be seen, then use `log.debug`.  Either way, we've been localizing everything except for "trace" messages.

src/extension.ts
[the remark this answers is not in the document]
bobbrow: There's so much code moved around that the diff doesn't tell me what was already there and what you added.

src/extension.ts
```diff
@@ -58,46 +54,115 @@ function reportProgress(progress: ProgressHandle|undefined, message: string) {
...
+      rollbar.takePromise('Post-folder-open', {folder: info.folder}, this._postWorkspaceOpen(info));
+    });
+    this._folders.onAfterRemoveFolder (async info => {
+      console.assert(this._folders.size === vscode.workspace.workspaceFolders?.length);
+      this._codeModelUpdateSubs.delete(info.uri.fsPath);
+      if (!vscode.workspace.workspaceFolders?.length) {
+        await this._setActiveFolder(undefined);
+      } else {
+        if (this._folders.activeFolder?.folder.uri.fsPath === info.uri.fsPath) {
+          await this._setActiveFolder(vscode.workspace.workspaceFolders[0]);
```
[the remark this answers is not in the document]
xisui-MSFT: This is the case where the active folder gets removed.

src/variant.ts
bobbrow: What I meant was whether this can ever be `null` or `undefined`.  (i.e. if the user activates CMake Tools without opening a folder)

src/extension.ts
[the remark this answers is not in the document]
xisui-MSFT: This one isn't in the original code but in his pr.

src/variant.ts
xisui-MSFT: Unlike before, each CMakeTools is bonded to a folder now. So it won't be null or undefined.

src/extension.ts
```diff
@@ -1105,25 +643,125 @@ class ExtensionManager implements vscode.Disposable {
...
+        this._cleanOutputChannel(activeFolder.cmakeTools);
+        if (await this._ensureActiveKit(activeFolder.cmakeTools)) {
+          return fn!(activeFolder.cmakeTools);
+        }
+        return -1;
+      }
+      rollbar.error(localize('no.active.folder', 'No active foler.'));
+      return 0;
+    } else if (cmt instanceof CMakeTools) {
+      this._cleanOutputChannel(cmt);
```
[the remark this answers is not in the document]
xisui-MSFT: It's not only used for build. It's also used for configure, launch, debug, etc. Before my change, I think it's in driver.ts, but when, let's say configuring two project, the output of the former one will get cleared. And that's why I put it here. We can put it in multiple functions (configure, build, launch, etc.) as well, but I feel it better to put it here.

src/variant.ts
bobbrow: Thanks.  That's what I wanted to know.

src/extension.ts
```diff
@@ -1105,25 +643,125 @@ class ExtensionManager implements vscode.Disposable {
...
+      if (await this._ensureActiveKit(cmt)) {
+        return fn!(cmt);
+      }
+      return -1;
+    } else {
+      fn = cmt;
+      log.clearOutputChannel();
+      for (const folder of this._folders) {
```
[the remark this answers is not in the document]
xisui-MSFT: I feel it cleaner this way. If we have different named functions, we'll likely need a "distribute" function to test multiroot mode and route the operations to each of them. We can change the order of parameters though, which should make it look better.

src/extension.ts
```diff
@@ -1141,42 +779,69 @@ class ExtensionManager implements vscode.Disposable {
...
+  executableTargets(folder?: vscode.WorkspaceFolder | string) { return this.mapQueryCMakeTools(folder, async cmt => (await cmt.executableTargets).map(target => target.name)); }
 
-  launchTarget(name?: string) { return this.withCMakeTools(null, cmt => cmt.launchTarget(name)); }
+  tasksBuildCommand(folder?: vscode.WorkspaceFolder | string) { return this.mapQueryCMakeTools(folder, cmt => cmt.tasksBuildCommand()); }
 
-  selectLaunchTarget(name?: string) { return this.withCMakeTools(null, cmt => cmt.selectLaunchTarget(name)); }
+  async debugTarget(folder?: vscode.WorkspaceFolder, name?: string): Promise<vscode.DebugSession | null> { return this.mapCMakeToolsFolder(folder, cmt => cmt.debugTarget(name)); }
 
-  resetState() { return this.withCMakeTools(null, cmt => cmt.resetState()); }
+  async debugTargetAll(name?: string): Promise<(vscode.DebugSession | null)[]> {
```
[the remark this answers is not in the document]
xisui-MSFT: "name" is mostly for testing. The user will only see the command. If there's no preset target, I believe there will be quick pick menu for each folder.

.vscode/launch.json
```diff
@@ -126,6 +126,92 @@
...
+                }
+            },
+            "osx": {
+                "env": {
+                    "CMT_TESTING": "1",
+                    "CMT_QUIET_CONSOLE": "1",
+                    "HasVs": "false"
+                }
+            }
+            // "smartStep": true
```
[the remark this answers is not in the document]
xisui-MSFT: The description is "Automatically step through generated code that cannot be mapped back to the original source." Not sure how useful it is. Probably should ask Colen?

src/cmake-tools.ts
```diff
@@ -920,9 +912,10 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
-  async setVariant() {
-    const ret = await this._variantManager.selectVariant();
-    if (ret) {
+  async setVariant(name?: string) {
+    // ! for compatibility with return code style...
+    const ret = ! await this._variantManager.selectVariant(name);
+    if (!ret) {
       await this.configure();
     }
     return ret;
```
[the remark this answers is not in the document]
xisui-MSFT: This return now becomes correct because I inverted the meaning of ret :). SelectVariant returns true when succeeding, but succeeding code for cmake commands is 0.

src/folders.ts
```diff
@@ -0,0 +1,190 @@
+/**
+ * Class for managing workspace folders
+ */ /** */
```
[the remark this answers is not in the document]
xisui-MSFT: This exists in a lot of files. I'd suggests to investigate why, and if there's no reason then remove all in a separate PR.

src/cmake-tools.ts
```diff
@@ -252,15 +265,7 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
-    let workspace = null;
-    const rootPath = getPrimaryWorkspaceFolder();
-    if (!rootPath) {
-      throw new Error(localize('no.workspace.error', 'CMake Tools is not available without an open workspace'));
-    }
-    if (vscode.workspace.workspaceFolders) {
-      workspace = lightNormalizePath(rootPath.fsPath);
-    }
-
+    const workspace = this.folder.uri.fsPath;
```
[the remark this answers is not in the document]
xisui-MSFT: Because previously CMakeTools can exist without a folder. But now a folder is required.

src/cmake-tools.ts
```diff
@@ -920,9 +912,10 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
   /**
    * Implementation of `cmake.setVariant`
    */
-  async setVariant() {
-    const ret = await this._variantManager.selectVariant();
-    if (ret) {
+  async setVariant(name?: string) {
+    // ! for compatibility with return code style...
```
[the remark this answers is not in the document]
xisui-MSFT: I changed it from success (true) / failure (false) pattern to return code (0 for success) pattern. So the previous naming was inappropriate, and the current name should be fine. Can you see a better name?

src/folders.ts
```diff
@@ -0,0 +1,190 @@
...
+export class CMakeToolsFolder {
+  private constructor(readonly cmakeTools: CMakeTools, readonly kitsController: KitsController) { }
+
+  static async init(cmakeTools: CMakeTools) {
+    return new CMakeToolsFolder(cmakeTools, await KitsController.init(cmakeTools));
+  }
+
+  get folder() { return this.cmakeTools.folder; }
+
+  private readonly _subscriptions: vscode.Disposable[] = [];
```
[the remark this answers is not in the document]
xisui-MSFT: No. Was used during my changes, now we can remove it.

src/logging.ts
```diff
@@ -18,7 +18,6 @@ export enum LogLevel {
   Fatal,
 }
 
-export type LogLevelKey = 'trace'|'debug'|'info'|'note'|'warning'|'error'|'fatal';
```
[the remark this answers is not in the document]
xisui-MSFT: It's imported at line 237. Seems like there are something changed in ts 3.8, and keeping this line would make this file fail to compile.

src/paths.ts
```diff
@@ -117,15 +117,16 @@ class Paths {
   }
 
   async getCMakePath(wsc: DirectoryContext): Promise<string|null> {
+    // TODO?
     const raw = await expandString(wsc.config.raw_cmakePath, {
       vars: {
-        workspaceRoot: wsc.dirPath,
-        workspaceFolder: wsc.dirPath,
+        workspaceRoot: wsc.folder.uri.fsPath,
```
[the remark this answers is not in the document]
xisui-MSFT: DirectoryContext is initialized via CMakeTools. So it should always contain a valid WorkspaceFolder.

src/state.ts
```diff
@@ -11,37 +11,45 @@ import * as vscode from 'vscode';
  * invalid states.
  */
 export class StateManager {
-  constructor(readonly extensionContext: vscode.ExtensionContext) {}
+  constructor(readonly extensionContext: vscode.ExtensionContext, readonly folder: vscode.WorkspaceFolder) {}
+
+  private _get<T>(key: string): T | undefined {
+    return this.extensionContext.globalState.get<T>(this.folder.uri.fsPath + key);
```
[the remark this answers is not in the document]
xisui-MSFT: The main issue is when there's an existing single root project and a new folder is added. Let's say user has been working on project A, and it has kits set. Then folder B is added. Now since the workspace is changed, the user will be prompted to select kit again.

src/status.ts
```diff
@@ -44,6 +46,9 @@ export class StatusBar implements vscode.Disposable {
   }
 
   constructor() {
+    this._activeFolderButton.command = 'cmake.selectActiveFolder';
+    this._activeFolderButton.tooltip = localize('active.folder.tooltip', 'Active folder');
```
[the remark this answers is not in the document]
xisui-MSFT: It can't be used to select active folder if autoSelectActiveFolder is true.

src/util.ts
xisui-MSFT: This version in cpptools doesn't work for string objects, i.e., typeof(new String('whatever')) is object instead of string.

src/cmake-tools.ts
```diff
@@ -651,9 +648,6 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         return -1;
       }
     }
-    if (this.workspaceContext.config.clearOutputBeforeBuild) {
```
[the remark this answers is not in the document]
xisui-MSFT: Yes, _cleanOutputChannel in extension.ts. I just saw that it's not used in the "else" part of mapCMakeTools. I can see if I can change it to check the workspace setting.

src/folders.ts
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Handle workspace change event.
+   * @param e Workspace change event
+   */
+  private async _onChange(e: vscode.WorkspaceFoldersChangeEvent) {
+    // Un-register each CMake Tools we have loaded for each removed workspace
+    for (const folder of e.removed) {
+      const cmtf = this.get(folder);
+      if (cmtf) {
+        this._beforeRemoveFolderEmitter.fire(cmtf);
```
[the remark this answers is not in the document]
xisui-MSFT: Yes. I should remove addFolder function though. _addFolder and _removeFolder shouldn't be called directly.

src/extension.ts
```diff
@@ -275,124 +349,71 @@ class ExtensionManager implements vscode.Disposable {
...
-    this._activeWorkspaceFolder = ws;
-    // Drop the old kit watcher on the floor
-    this._resetKitsWatcher();
-    // Re-read kits for the new workspace:
-    await this._rereadKits(progress);
+    this._folders.setActiveFolder(ws);
+    this._statusBar.setActiveFolderName(ws?.name || '');
+    this._statusBar.setActiveKitName(this._folders.activeFolder?.cmakeTools.activeKit?.name || '');
     this._setupSubscriptions();
+    telemetry.logEvent('setActiveFolder');
```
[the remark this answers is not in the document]
xisui-MSFT: Erika mentioned this in that thread earlier. I don't have a preference.

src/extension.ts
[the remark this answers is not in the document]
xisui-MSFT: But yeah we probably don't really need this. Do you want me to remove it?

test/extension-tests/successful-build/test/toolchain.test.ts
xisui-MSFT: The original function (kitsAvailableInWorkspaceDirectory) reads user kits as well.

src/extension.ts
```diff
@@ -1105,25 +643,125 @@ class ExtensionManager implements vscode.Disposable {
...
+  private _cleanOutputChannel(cmt: CMakeTools) {
+    if (cmt.workspaceContext.config.clearOutputBeforeBuild) {
+      log.clearOutputChannel();
+    }
+  }
+
   // The below functions are all wrappers around the backend.
+  async mapCMakeTools(fn: CMakeToolsMapFn): Promise<any>;
+  async mapCMakeTools(cmt: CMakeTools|undefined, fn: CMakeToolsMapFn): Promise<any>;
+  async mapCMakeTools(cmt: CMakeTools|undefined|CMakeToolsMapFn, fn?: CMakeToolsMapFn): Promise<any> {
```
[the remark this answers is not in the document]
xisui-MSFT: No, for example DebugTarget returns a DebugSession.

src/state.ts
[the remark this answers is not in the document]
bobbrow: That makes sense. But I think that is the exception rather than the rule. You could have some global state that tracks the most recently active kit which is used when there is no value for the workspaceState yet (because it is a new workspace).  I think it is reasonable to require the user to pick the target and maybe the variant again when the workspace adds a second folder, but you could follow this fallback pattern for all properties if you like.

src/status.ts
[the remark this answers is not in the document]
bobbrow: I didn't see any logic that changes the tooltip based on the value of autoSelectActiveFolder.  You'll need to update it when the setting changes then.

src/util.ts
bobbrow: But String objects should never be used and are condemned as bad practice.  I didn't see `new String` anywhere in the code. Is VS Code sending us String objects?  If we want to be safe against String objects, the code should be:

```ts
export function isString(input: any): input is string {
    return typeof(input) === "string" || input instanceof String;
}
```

test/extension-tests/successful-build/test/toolchain.test.ts
bobbrow: you're right. sorry I missed that.

src/extension.ts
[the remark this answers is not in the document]
bobbrow: ok.

src/extension.ts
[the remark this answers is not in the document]
bobbrow: Can you fix it please?

terrelln:
Just wanted to drop in and say that I'm really looking forwards to this feature!

src/cmake-tools.ts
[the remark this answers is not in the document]
xisui-MSFT: Interesting... I thought that wasn't the case as the root folder is always used.... But anyway we don't need to worry about this in the new code. cmaketools can only be accessed via folder. So even if there's an invalid cmaketools instance somehow, it can't be accessed. If there's nothing open. Then nothing would happen for commands related to cmake tools.

src/cmake-tools.ts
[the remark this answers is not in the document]
xisui-MSFT: Yeah I did notice this. I did it this way because there was some timing issue with
```
    const ret = await this._variantManager.selectVariant(name);
    if (ret) {
      await this.configure();
    }
    return !ret;
```
Seems like that return line can bypass the awaits... I can try your ways though.

src/cmake-tools.ts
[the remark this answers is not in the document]
xisui-MSFT: Resolving this comment as duplicate.

src/extension.ts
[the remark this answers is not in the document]
xisui-MSFT: Actually you are right. The original mapCMakeTools provides three conditions but would become a default/optional parameter can only provide two. So I'll create another function.

src/kitsController.ts
```diff
@@ -0,0 +1,530 @@
...
+          isCloseAffordance: true,
+        },
+        {
+          title: localize('no.button', 'No'),
+          isCloseAffordance: true,
+        });
+
+      if (chosen !== undefined && (chosen.title === yesButtonTitle)) {
+        KitsController.userKits = new_definition_user_kits;
+        duplicateRemoved = true;
```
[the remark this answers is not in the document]
xisui-MSFT: Merged master and didn't see conflict on kits related things...

src/util.ts
xisui-MSFT: Not sure if this argument worth it but I was following the Array.IsArray implementation in the standard... The second version you provided doesn't work for String.prototype...

src/extension.ts
```diff
@@ -113,51 +178,53 @@ class ExtensionManager implements vscode.Disposable {
...
-   * Get the CMakeTools instance associated with the given workspace folder, or `null`
-   * @param ws The workspace folder to search
-   */
-  private _cmakeToolsForWorkspaceFolder(ws: vscode.WorkspaceFolder): CMakeTools|null {
-    return this._cmakeToolsInstances.get(ws.name) || null;
+  private async _pickFolder() {
+    const selection = await vscode.window.showWorkspaceFolderPick();
+    if (selection) {
+      const cmtFolder = this._folders.get(selection);
+      console.assert(cmtFolder, 'Folder not found in folder controller.');
```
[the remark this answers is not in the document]
xisui-MSFT: Discussed offline, was a confusion.

src/extension.ts
```diff
@@ -275,124 +349,71 @@ class ExtensionManager implements vscode.Disposable {
...
-    if (!kit_name) {
-      // No prior kit. Done.
-      return new_cmt;
+  async selectActiveFolder() {
+    if (vscode.workspace.workspaceFolders?.length && !this._workspaceConfig.autoSelectActiveFolder) {
+      const lastActiveFolderPath = this._folders.activeFolder?.folder.uri.fsPath;
+      const selection = await vscode.window.showWorkspaceFolderPick();
+      if (selection) {
+        // Ingore if user cancelled
+        await this._setActiveFolder(selection);
```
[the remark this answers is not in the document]
xisui-MSFT: Discussed offline, was a confusion.

src/cmake-tools.ts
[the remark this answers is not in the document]
xisui-MSFT: Had another check on the old code, clearOutputBeforeBuild setting hasn't been used very well. So we'll likely change the behavior anyway. Any thought?

src/util.ts
bobbrow: We should never have String objects and certainly should not be manipulating the prototype, so I don't think we need this level of protection, but if you feel strongly about it, you can leave it.  Just add the `x is string` part after the signature so that the type will be promoted to string by IntelliSense.

src/cmake-tools.ts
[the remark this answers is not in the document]
bobbrow: I don't want to start clearing for every operation.  Build and configure is fine to keep, but the others are not.

src/cmake-tools.ts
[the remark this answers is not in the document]
bobbrow: ok

src/extension.ts
```diff
@@ -659,478 +567,217 @@ class ExtensionManager implements vscode.Disposable {
...
+      rollbar.error(localize('no.active.folder', 'No active folder.'));
+      return -2;
     }
-    newKit = this._allKits.find(kit => kit.name === kitName);
-    await this._setCurrentKit(newKit || null);
-    // if we are showing a quickpick menu...
-    this._pickKitCancellationTokenSource.cancel();
+
+    if (await this._ensureActiveKit(cmt)) {
+      return fn!(cmt);
```
bobbrow: This shouldn't need a `!` anymore.

src/folders.ts
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * Create a new instance of the backend to support the given workspace folder.
+   * The given folder *must not* already be loaded.
+   * @param folder The workspace folder to load for
+   * @returns The newly created CMakeTools backend for the given folder
+   */
+  private async _addFolder(folder: vscode.WorkspaceFolder) {
+    const existing = this.get(folder);
+    if (existing) {
+      rollbar.error(localize('double.loaded.folder','Double-loaded workspace folder'), { wsUri: folder.uri.toString() });
```
bobbrow: how about "The same workspace folder was loaded twice"

src/state.ts
```diff
@@ -11,37 +11,56 @@ import * as vscode from 'vscode';
...
-  constructor(readonly extensionContext: vscode.ExtensionContext) {}
+  private readonly latestPrefix: string = 'latest.';
+
+  constructor(readonly extensionContext: vscode.ExtensionContext, readonly folder: vscode.WorkspaceFolder) {}
+
+  private _get<T>(key: string): T | undefined {
+    const actualKey = this.folder.uri.fsPath + key;
+    let value = this.extensionContext.workspaceState.get<T>(actualKey);
+    if (value) {
+      this.extensionContext.globalState.update(this.latestPrefix + actualKey, value);
```
bobbrow: globalState should only look at "key" or else it will be unique for all folders which is not what we want.

src/state.ts
```diff
@@ -11,37 +11,56 @@ import * as vscode from 'vscode';
...
+    } else {
+      value = this.extensionContext.globalState.get<T>(this.latestPrefix + actualKey);
```
bobbrow: globalState should use "key"

src/state.ts
```diff
@@ -11,37 +11,56 @@ import * as vscode from 'vscode';
...
+    }
+    return value;
+  }
+
+  private _update(key: string, value: any): Thenable<void> {
+    const actualKey = this.folder.uri.fsPath + key;
+    this.extensionContext.globalState.update(this.latestPrefix + actualKey, value);
```
bobbrow: globalState should use "key"

src/state.ts
xisui-MSFT: Got it, had some misunderstanding.

xisui-MSFT: Wait but if we make this key, then the kit selection menu will never pop up again automatically when the project is firstly opened...

bobbrow: ☹️ I guess we'll have to revert it back to use globalState.  Sorry for dragging you through that.

xisui-MSFT: Isn't this way better though as it allows the user to use different settings in different workspace for the same folder?

xisui-MSFT: Discussed offline. Go back to the original way.

Zingam:
I get a crash with the attached sample project configured as is.
[CMakeMultiProject.zip](https://github.com/microsoft/vscode-cmake-tools/files/4080207/CMakeMultiProject.zip)

bobbrow:
@Zingam it looks like your `sourceDirectory` setting is incorrectly configured.  You're pointing to a file instead of a directory.

```
    "cmake.sourceDirectory": "${workspaceFolder}/CMakeProject01/CMakeLists.txt"
```

I think the extension could be more tolerant of this case.  Can you open a bug to address it?

bobbrow:
For those following, we just posted a pre-release of multi-root support here: https://github.com/microsoft/vscode-cmake-tools/releases/tag/1.3.0-beta

We would appreciate it if you would install it and provide feedback.

Thank you!

PR #963, opened 2019-12-09
Consistently normalize absolute and relative paths

GeorchW:
Normalize even absolute paths, not just relative ones. Otherwise, absolute paths containing '.' and '..' may not be normalized out, 

The personal itch I was tackling was that, when compiling with `zapcc`, the diagnostics had paths containing '..' . This made VSCode open a separate editor when clicking on the diagnostic, displaying the same file using a different path, while not displaying the squiggles in the original editor.


GeorchW:
Uhm, is there something wrong with this PR?

bobbrow:
@GeorchW, I'm sorry, we haven't had a chance to review it yet.  December is a month where most MS employees take vacation, and we're spread pretty thin right now.  My developer assigned to this project also had to take some emergency time off this month so things are running more slowly than expected.  I apologize for that.  We will take a look as soon as we can.

bobbrow:
Thank you for your contribution.  Again, I'm sorry it has taken so long to commit it.

PR #973, opened 2019-12-17
Fixed public API

Deerenaros:
### This adds public API back again

- Added pretty small `ExtAPI` interface to manage with `ExtentionManager` outside
- Repaired `targetChangedEvent()`
- Implemented `Kit` management outside too
- A little there and here

## The purpose of this change
2k20 is coming, happy new year!

## Other Notes/Information
Yeah, TypeScript is a great language, but this is my first encounter with one. Also, I haven't read any code of conduct, my apologies. Anyway, there are really little changes so I hope you can manage them much better than me.

If I am anywhere wrong with architecture or your vision for the project please write to me.

Thank you for distribution that great addon!

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -8,6 +8,7 @@
  */ /** */
 
 import {DebugSession, Disposable, Event, Terminal} from 'vscode';
+import { Kit } from './kit';
 
 
 /**
@@ -143,7 +144,8 @@ export interface RichTarget {
 export type Target = NamedTarget|RichTarget;
 
 /**
- * The CMake Tools extension API obtained via `getExtension().exports`
+ * The CMake Tools extension API obtained via
+ * `getExtension("ms-vscode.cmake-tools").exports`
  */
 export interface CMakeToolsAPI extends Disposable {
 
@@ -303,4 +305,13 @@ export interface CMakeToolsAPI extends Disposable {
    * Get the build command string for the active target
    */
   tasksBuildCommand(): Thenable<string|null>;
+
+
+  readonly kitChangedEvent: Event<void>;
+  readonly currentKit: Kit|null;
+}
+
+export interface ExtAPI extends Disposable {
+  active(): CMakeToolsAPI|null;
+  readonly acitveChangedEvent: Event<void>;
 }
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -294,13 +294,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
   /**
    * Event fired after CMake configure runs
    */
-  get onReconfigured() { return this._onReconfiguredEmitter.event; }
   private readonly _onReconfiguredEmitter = new vscode.EventEmitter<void>();
-
+  get onReconfigured() { return this._onReconfiguredEmitter.event; }
   get reconfigured() { return this.onReconfigured; }
 
   private readonly _onTargetChangedEmitter = new vscode.EventEmitter<void>();
   get targetChangedEvent() { return this._onTargetChangedEmitter.event; }
+
+  private readonly _onKitChangedEmmiter = new vscode.EventEmitter<void>();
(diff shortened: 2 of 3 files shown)


bobbrow:
We apologize for letting this PR sit for so long without attention.  We feel that the API needs to be redesigned before we go about updating and re-enabling it - including making the API available as an npm module (similar to [cpptools' API](https://www.npmjs.com/package/vscode-cpptools)) instead of an unversioned TypeScript file you copy from here into your extension.  We haven't put this on the roadmap yet but we continue to track the 👍 reactions to the feature request.  I am going to close this PR for now.

yurivesnin:
Thank you anyway for the response. This is rly necessary feature while vscode is slightly unusable without respectation of legacy projects, which could have very variable and customized make manager, even cmake can be far away from standarts. In another hand, extensions can be managed by lowcost along of using eclipse or CLI the hard way. And, this will be much more good to use one project builder instead maintain twice of them (make and vsproj for example).

As I said, thank you anyway. Let me know if something will be changed.

bobbrow:
What functionality do you need?  Many of the APIs can be emulated by invoking CMake Tools' exported commands directly by your extension (e.g. `vscode.commands.executeCommand('cmake.configure')`).  If there is something missing in the set of exported commands, it might be faster for us to just add a command for that and you wouldn't need the API at all.

Deerenaros:
We had an issue to build embedded application with set of various
libraries. Some of them had cmake support, some not. Some of projects had
git submodules to manage dependencies, some just linking to statically
defined directories. Also, embedded compilers for our platforms (some of
them) support only posix environment, so we used patched cygwin. All of it
was painfully and hard to automate in devops inventing. The most painfull
point was necessity to maintain vsproj and make/cmake as well so we tried
to find userfriendly IDE and vscode was one of the most pleasant choice due
to very flexible design. Anything works well but intellisense and platform
choosing, which one does not catch our paths to used libraries sources. So
I explored a way to handle it byself in selfwroten extension, but managed
innability to do it due to hidden API. The main goal was not to building
application (not only), but realtime errors reporting and autocomplete
feature. As I said, various of libraries written with make management and
soureces placed somethere statically, while some of them with cmake and
submodules with recoursive building.

bobbrow:
Sorry, my question could have been more specific.  Which functions from the API were you hoping to use?  We can perhaps map some of these back to commands that exist in the extension now, or potentially add new commands to provide similar functionality.

Deerenaros:
In my previous task I had to manage `Kit`s dynamically due to the platform
chosen and file opened. Also, there were some issues with files' stack (on
walking headers through F12). Currently, I can't remember closer. I think
it is better to repair intellisense linking with providing better mapping
to the sources, but in projects with many deep dependencies it would be a
little harder. Anyway, the project is closed, so do not target me with such
requests. It would be much more flexible to provide a clean and useful API
instead of growing functionality.

ср, 24 мар. 2021 г. в 23:01, Bob Brown ***@***.***>:

bobbrow:
I'm sorry if my question bothered you.  I only asked because I was hoping to provide you with the support you needed.  I view the commands and the API as essentially the same thing -- it's just a different syntax and I would rather improve the commands than add this API.  An API of this form is only needed in my opinion if one extension (CMake Tools) needs to request information from another extension (yours, for example).

yurivesnin:
Don’t worry, it would be great if my answers will make this tool more flexible and useful!

As I see today, my trying to make extension was asymmetric in communications. Just as you said – registered methods used to choose right `Kit`, which used to provide right parameters to intellisense. Of course, extension wasn’t and won’t be posted, my apologize here. Any commands will be useful, but here we built kits dynamically providing sources paths and bindings in case of using platform, operations system, their content, environment variables, git state and much more. This already was done by set of wrappers and we failed on add-on this features into vscode. Anyway, there are around year past, so maybe something changed, or I take a lot of mistakes and it could be done much easier.

PR #975, opened 2019-12-18
Honor parallelsJobs value when the generator is a Visual Studio one (multicore support)

tonka3000:
## This changes
the perfomance of a build when the cmake generator is Visual Studio one f.e. `Visual Studio 15 2017 Win64` because cmake build single-core by default. Adding `/maxcpucount` as additional flag (it's the same flag as `-j` in gcc) will activate multicore build (when the Visual Studio solution is generated by Visual Studio ´/maxcpucount´ is the default).

This PR always add the number of cores calculate by this extension (as defined with `cmake.parallelJobs` setting), so the flag always looks like f.e `/maxcpucount:8`. msbuild would have also the option to detect the cores by itself with `/maxcpucount`, but the implementation for that in this extension would be much more complicated.

## The purpose of this change

Multicore-build per default => better speed in most cases


bobbrow:
Thank you!

PR #978, opened 2019-12-19
Add support for Unity Testing Framework output parsing:

Morozov-5F:
- I've implemented support for Unity Fixture dialect only as it's more or less standardized and that's what I currently use.
- There are some whitespace changes -- I used clang-format as the "contributing" page suggests.

### This changes CTest support

- Add support for [Unity Test Framework](https://github.com/ThrowTheSwitch/Unity) output parsing. Right now it parses the output produced by built-in extension [Unity Fixture](https://github.com/ThrowTheSwitch/Unity/tree/master/extras/fixture).  

## The purpose of this change
I'm currently working with CTest and Unity and I would like to have those fancy decorators in my source code if the tests are failing. I decided to implement this support.


Morozov-5F:
I'm sorry -- I had some TSLint issues on my side and some warnings slipped in. I've addressed them in a separate commit.

src/ctest.ts
```diff
@@ -188,9 +190,62 @@ export function parseCatchTestOutput(output: string): FailingTestDecoration[] {
...
+export function parseUnityFixtureOutput(output: string): FailingTestDecoration[] {
+  const lines_with_ws = output.split('\n');
+  const lines = lines_with_ws.map(l => l.trim());
+  const decorations: FailingTestDecoration[] = [];
+  const failedTestRegex = /^([\.!]*)(\.\..*):(\d+):(.*):FAIL: (.*)/;
+
+  for (const line of lines) {
+    const parseRes = failedTestRegex.exec(line);
+    if (parseRes) {
+      const [_all, testsRanStr, file, lineNumberStr, testName, message] = parseRes;
```
bobbrow: I'm pretty sure you can do:
```
const [, , file, lineNumberStr, , message] = parseRes;
```
or better:
```
const [/*all*/, /*testsRanStr*/, file, lineNumberStr, /*testName*/, message] = parseRes;
```
instead.  (I do see that the code you're modelling this on doesn't do this, but ownership of the code has changed since that was written 😄)

Morozov-5F: Thanks, that'll simplify the code! I prefer the second option -- that'll make it easy to spot those parameters if they're ever required.

src/ctest.ts
```diff
@@ -188,9 +190,55 @@ export function parseCatchTestOutput(output: string): FailingTestDecoration[] {
...
+ * * `FAIL`: failure indication. It could also be `PASS` or `IGNORE`.
+ * * `Expected 1 was 0`: assertion message. Should always be on one line expect the cases when user
+ *                       explicitly puts the `\n` in the message. That shouldn't typically happen
+ *                       and looks like a known limitation to me.
+ */
+export function parseUnityFixtureOutput(output: string): FailingTestDecoration[] {
+  const lines_with_ws = output.split('\n');
+  const lines = lines_with_ws.map(l => l.trim());
+  const decorations: FailingTestDecoration[] = [];
+  const failedTestRegex = /^([\.!]*)(\.\..*):(\d+):(.*):FAIL: (.*)/;
```
bobbrow: I loaded this regex and your sample output into a regex tester and it didn't match.  There seems to be an extra required `:` before the `FAIL`.

This framework's output is troublesome though and I expect there to be bugs in any regex that attempts to parse it.  Can you suggest to the Unity folks that they add a delimiter before the file name?  Maybe an extra `:` or something?  This will allow you to get more reliable parsing.

Morozov-5F: Oh, my bad. The sample output is missing an important part, I'm very sorry about it. I was using Unity Fixture in my private project and I changed the actual output a little bit and it turns out I removed some extra characters by accident. Here's the correct sample output that I've made on an example file: 
```
.../Users/morozov/git/unity-fixture-output-test/main.c:24:TEST(first, third):FAIL: Test failed
```
I've made a simple utility called [unity-fixture-output-generator](https://github.com/Morozov-5F/unity-fixture-output-generator) which produces correct output. 

I think you're right and I've mentioned that in my comment to the function -- there will always be bugs, whichever regex to use. I've attempted to handle the case that I think is common -- tests are just ran without any parameters at all. However, I think there are some ways around that could alter framework output and make the fail message search much more reliable: 
1. Use -v flag when running executable containing test. That way framework produces more output but it's more predictable: 
```
TEST(first, first) PASS
TEST(first, second) PASS
TEST(first, third)/path/to/unity-fixture-output-test/main.c:24::FAIL: Test failed
```
Even though there are no delimiters between path and test name, it's still possible to parse it using, for example, this regex: `^(TEST\(.*\))(.*):(\d+):FAIL: (.*)`.
2. It's possible to get rid of dots at all by using `-s` flag when running tests. The looks like that:
```
/path/to/unity-fixture-output-test/main.c:24:TEST(first, third):FAIL: Test failed
```
In that case we can use such regex: `^(.*):(\d+):(.*):FAIL: (.*)`. 
3. Propose Unity guys to add a configurable separator between dots and filename as you've mentioned. 
4. Add two extension options: first configures testing framework (possible values: `catch` and `user`) and second defines user regex for testing framework. For example: 
``` json
{
   "cmake.ctestFramework": "catch"
}
```
or 
``` json
{
    "cmake.ctestFramework": "user",
    "cmake.ctestFrameworkRegex": "^([\\.!]*)(\\.\\..*):(\\d+):(.*):FAIL: (.*)"
}
```
I think this option is the most flexible one and I could implement that as part of separate PR. 

Sorry for a long-read I hope I don't overcomplicate things here.

bobbrow: If Unity doesn't have a consistent way to parse test results, then I'm hesitant to approve adding direct support for it in this extension.

In that case, I would want to go with something like your option 4 and allow developers to set their own customizable regex based on the test framework they use and the way they configure it to run.  I would suggest the name `"cmake.ctest.failurePatterns"` and make it an array of strings (which are interpreted as Regex's)

Morozov-5F: Sounds great. I can implement the option in this PR or as a separate PR, whichever you prefer. I’m inclined to make a separate PR, what do you think?

bobbrow: It's probably easiest to start with a new branch/PR, but I'll leave that up to you.  Go ahead and close this PR if you plan to replace it with a new one.  Thanks!

Morozov-5F:
After a discussion this PR will be closed and another one will be created.

PR #980, opened 2019-12-22
Mark all tests as 'notrun' in case of build failure when running CTest

Morozov-5F:
### This changes CTest visible behaviour

- Mark all tests as 'notrun' in case of build failure during a CTest run.

## The purpose of this change
It can be confusing sometimes that tests are marked as passed when I add new test case into the test code which exercises not yet implemented behaviour. This thing can occur when doing a TDD -- most of the time you end up adding a non-exisiting (yet) function call to the test code and automatically running tests to make sure everything fails as expected. Even though it fails VSCode UI shows that tests are "all green" and it may be confusing.


Morozov-5F:
Good idea! Propose to make this change in a separate PR.

Issue #965, opened 2019-12-10
Can't debug in Visual Studio Code


zhouhao27:
### Brief Issue Summary

Can't make the Visual Studio Code debugger work.

### Expected:

1. Run `CMake Configuration`

<img width="585" alt="Screenshot 2019-12-10 at 4 01 19 PM" src="https://user-images.githubusercontent.com/8099731/70507693-d35eac80-1b67-11ea-848e-92d4b72309da.png">

The result:
```
[cmake] Configuring done
[cmake] CMake Deprecation Warning:
[cmake]   The 'cmake-server(7)' is deprecated.  Please port clients to use the
[cmake]   'cmake-file-api(7)' instead.
[cmake] 
[cmake] 
[cmake] Generating done
```

2. Select Clang 11.0

<img width="538" alt="Screenshot 2019-12-10 at 4 15 43 PM" src="https://user-images.githubusercontent.com/8099731/70507966-67c90f00-1b68-11ea-9a3a-edcf4d41725a.png">

3. Allow `IntelliSense for this folder`

<img width="503" alt="Screenshot 2019-12-10 at 4 15 53 PM" src="https://user-images.githubusercontent.com/8099731/70508058-98a94400-1b68-11ea-877a-6fb36e1a8209.png">

4. Build

<img width="358" alt="Screenshot 2019-12-10 at 4 18 06 PM" src="https://user-images.githubusercontent.com/8099731/70508124-b9719980-1b68-11ea-808d-be5007cb2df2.png">

The result:
```
[build] Starting build
[proc] Executing command: /usr/local/bin/cmake --build /Users/xxx/Desktop/cmake/hello/build --config Debug --target all -- -j 6
[build] Scanning dependencies of target say-hello
[build] [ 25%] Building CXX object CMakeFiles/say-hello.dir/hello.cpp.o
[build] [ 50%] Linking CXX static library libsay-hello.a
[build] [ 50%] Built target say-hello
[build] Scanning dependencies of target hello
[build] [ 75%] Building CXX object CMakeFiles/hello.dir/main.cpp.o
[build] [100%] Linking CXX executable hello
[build] [100%] Built target hello
[build] Build finished with exit code 0
```

5. Set the break point and run Debug

<img width="397" alt="Screenshot 2019-12-10 at 4 08 23 PM" src="https://user-images.githubusercontent.com/8099731/70508208-f178dc80-1b68-11ea-8a41-2e0fe129d56b.png">

[...]

PR #988, opened 2019-12-27
Workaround for MacOS lldb #965

kaidokert:
### This changes default fallback lldb search apth

- On MacOS, the system lldb is not a MI enabled debugger, and will always fail.
- Xcode dev tools always includes a working MI debugger at `/Applications/Xcode.app/Contents/Developer/usr/bin/lldb-mi`. 
- This change will address the issue when /usr/bin/clang is the selected compiler, and debugging without `launch.json` always fails when hitting Ctrl+f5

Also note, `/usr/bin/lldb` cannot be normally replaced by user, so it's always going to be Apple supplied binary there. There is no reason to try launching a debug session with it.

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -161,6 +161,10 @@ export async function getDebugConfigurationFromCache(cache: CMakeCache, target: 
     } else {
       // Look for lldb
       clang_debugger_path = compiler_path.replace(clang_compiler_regex, 'lldb');
+      if(clang_debugger_path == '/usr/bin/lldb' && platform == 'darwin') {
+        // Default Apple lldb in /usr/bin will definitely not work, try lldb-mi instead
+        clang_debugger_path = '/Applications/Xcode.app/Contents/Developer/usr/bin/lldb-mi';
+      }
       if ((clang_debugger_path.search(new RegExp('lldb')) != -1) && await checkDebugger(clang_debugger_path)) {
         return createLLDBDebugConfiguration(clang_debugger_path, target);
       }


bobbrow:
I believe latest versions of XCode actually stopped distributing lldb-mi since LLVM removed it from their distribution.  This is a problem the cpptools team is working on.  I don't think hard-coding a path is going to be the right solution.

In the meantime, you can use the `cmake.debugConfig` setting to override the miDebuggerPath.

kaidokert:
You are correct, it was removed in Xcode 11 apparently: https://travis-ci.org/kaidokert/travis-explore/builds/630198899 . 

Adding this in settings.json does override the location correctly
```
{
    "cmake.debugConfig": {
        "miDebuggerPath": "/tmp/lldb"
    }
}
```

However, [documentation is not helpful here](https://vector-of-bool.github.io/docs/vscode-cmake-tools/debugging.html?highlight=hotkey#quick-debugging) so for first time users, they'll still get the default `/usr/bin/lldb` which will fail in a confusing manner. I ran into this the second time on a new machine, and forgot what the tweak was.

Also, `checkDebugger()` function here should still be doing a better job, if it knows `lldb` will be launched with `--interpreter=mi`, it should probably test for that and fail if it's not supported.

coppersoft:
I'm sorry, but this workaround didn't work for me.

OS: MacOS X Catalina: 10.15.4
Visual Studio Code: 1.44.2
ms-vscode.cpptools: 0.27.0
ms-vscode.cmake-tools: 1.3.1

I downloaded Xcode 10.3, extracted llvm-mi from it and put it inside

`/Applications/Xcode.app/Contents/Developer/usr/bin/`

Then I added this to settings.json

```
"cmake.debugConfig": {
        "miDebuggerPath": "/Applications/Xcode.app/Contents/Developer/usr/bin/lldb-mi"
}
```

But nothing changes. I launch the debug, but everything hangs, no breakpoint hit and it seems that nothing is executed.

I tried also setting "miDebuggerPath" to the llvm-mi bundled in the cpptools

`~/.vscode/extensions/ms-vscode.cpptools-0.27.0/debugAdapters/llvm-mi`

But nothing changes...

Am I doing something wrong?

Thank you

PR #992, opened 2019-12-31
Filter out duplicate targets from the target selector

bobbrow:
## This change addresses item #863

We remove duplicate targets from the quick pick menu.


Zingam:
I looked at the PR briefly. I don't think this is the correct approach, which only masks the underlying problem, which I tried to summarize https://github.com/microsoft/vscode-cmake-tools/issues/863#issuecomment-559561523
I think the issue is in tree.js Is it really necessary to add each project the way it is returned by the server? What purpose does this serve? If there is a deeper hierarchy of projects, including the multi-project feature there will be an explosion of items in the outliner and then there will be the same project/target in two or more places.
So the structure shouldn't contain the targets/projects multiple times and the outliner should just display the folder hierarchy.
Also please note that if you expand a target/project in one of the branches as listed in the outliner then if you try to click on the same target/project on another one there will be an error message and it won't expand. This maybe a VSCode issue but I am not quite sure.

bobbrow:
We haven't had time to properly address #863, and this seemed like a low cost/sufficient workaround until we can get to the other issue.

Issue #288, opened 2017-12-26
Allow setting environment variables via shell/cmd script


emptyVoid:
It would be great to have support for setting environment variables via a script in addition to the current name-value lists (i.e. `cmake.environment`). This would help building against third-party libraries relying on shell/batch scripts to prepare the build environment (e.g., _Intel Performance Libraries_ - each library exposes a .sh/.bat script which sets relevant environment variables).

vector-of-bool:
Sounds useful.

What sort of design/UX would you like to see?

emptyVoid:
I'm not really sure how it's usually done, though I was thinking about something like this:
```js
    "cmake.environmentScript": "source ${workspaceRoot}/path/to/script.sh"
```
and on Windows:
```js
    "cmake.environmentScript": "${workspaceRoot}\\path\\to\\script.bat"
```

KoeMai:
Do you need the script only in the build step or in the configure step of Cmake?

I would prefer a preConfigure and preBuild Script step. 
Is this script step multiproject specific, project specific, variant specific or kit specific?

vector-of-bool:
I almost feel like this is getting close to the idea of another Kit type. The VS kits are generated in a similar manner (by invoking the vcvars batch files and collecting environment variables). Perhaps it belongs there?

emptyVoid:
> Do you need the script only in the build step or in the configure step of Cmake?

Actually all of them:
- in the configure step: to detect the presence of a third-party library (e.g., _Intel Performance Libraries_ set `<LIB>ROOT` environment variables)
- in the build step: to pick up correct paths and/or compiler flags (e.g., _Intel Performance Libraries_ set `CPATH`/`INCLUDE` and `LIBRARY_PATH`/`LIB` environment variables)
- in the test step and debug: to launch an application built against a third-party library  (e.g., _Intel Performance Libraries_ set `LD_LIBRARY_PATH`/`PATH` environment variables)

KoeMai:
I think the ```environmentPreparationScript``` belongs to the kits.  It is a additional preparation step before cmake is running. May be it needs to be added to each kit,  because it is not a separate compiler more a extended compiler configuration.

mika-fischer:
I would also be interested in something like this. We already have a batch file that sets all necessary variables for our version of VS2017 build tools as well as some others we need internally. It would be great if we could just make cmake-tools use that as a kit.

Has there been any progress in that direction? What would be the direction forward?

[...]

PR #995, opened 2020-01-06
Use a shell script to set environment variables for a kit

pisker:
### This adds a feature

- Added the possibility to use a shell / batch script that sets the environment variables for a kit.
- Added a unit test for this feature

## Other Notes/Information
Some frameworks like the esp-idf / ESP8266_RTOS_SDK (ESP32 / ESP8266) use shell scripts to set the environment variables (especially update the PATH) for a build.

Uses the same approach as the function that gets the environment variables correspnding to a VS dev batch file (kit.ts collectDevBatVars)


not7cd:
@pisker, how easily do you think, one could source files? 
I also found more related issues:
https://github.com/microsoft/vscode-remote-release/issues/141
https://github.com/microsoft/vscode-cmake-tools/issues/288

pisker:
@not7cd the feature works by setting the filename of a shell script which is run during initialization. No source / call command is required. 

One constrained is that the shell script is (currently?) not run in bash. The source command is bash specific so it is not available within the shell script. The "." command can be used instead.

pkohout:
Any plans on when this will be released? Would really improve the vs code experience for us.

schemas/kits-schema.json
```diff
@@ -35,6 +35,22 @@
                 "type": "string",
                 "description": "Architecture to target"
             },
+            "environmentVariablesShellScript": {
+                "type": "string",
+                "description": "Filename of a shell script which sets environment variables for the kit"
+            },
+            "environmentVariablesShellScriptDirectory": {
```
bobbrow: I prefer to add as few properties as possible to make this work.  Can you achieve your goal without this?  Ideally the scripts are able to run regardless of the current working directory (e.g. they patch them up themselves).

schemas/kits-schema.json
```diff
@@ -35,6 +35,22 @@
                 "type": "string",
                 "description": "Architecture to target"
             },
+            "environmentVariablesShellScript": {
+                "type": "string",
+                "description": "Filename of a shell script which sets environment variables for the kit"
```
bobbrow: I would recommend making this always be treated as an absolute path instead of just a file name.  Can you update the description to clarify this?

I would also like to name this something else.  Perhaps `environmentSetupScript` or `environmentConfigurationScript`?  Then the description can say "The absolute path to a script that modifies the environment for the Kit"

schemas/kits-schema.json
```diff
@@ -35,6 +35,22 @@
...
+            },
+            "environmentVariablesShellScriptDirectory": {
+                "type": "string",
+                "description": "Directory in which to change before running the shell script"
+            },
+            "environmentVariablesShellScriptArgs": {
```
bobbrow: There are a lot of rules around quotes and escaping of arguments for the different shells which don't appear to be handled in this PR.  Do we really want to try to emulate those behaviors, or can we get by with just supporting a script path?

test/unit-tests/kitmanager.test.ts
```diff
@@ -23,4 +25,28 @@ suite('Kits test', async () => {
...
+
+  test('Test load env vars from shell script', async() => {
+    const fname_extension = process.platform == 'win32' ? 'bat' : 'sh';
+    const fname = `cmake-kit-test-${Math.random().toString()}.${fname_extension}`;
+    const script_path = path.join(paths.tmpDir, fname);
+    // generate a file with test batch / shell script that sets two env vars
+    if (process.platform == 'win32')
+      await fs.writeFile(script_path, `set "TESTVAR12=abc"\r\nset "TESTVAR13=cde"`);
+    else
+      await fs.writeFile(script_path, `export "TESTVAR12=abc"\r\nexport "TESTVAR13=cde"`);
```
bobbrow: for non-Windows, don't add a `\r` with the newline

test/unit-tests/kitmanager.test.ts
```diff
@@ -23,4 +25,28 @@ suite('Kits test', async () => {
       'VSCode Kit 2',
     ]);
   });
+
+  test('Test load env vars from shell script', async() => {
+    const fname_extension = process.platform == 'win32' ? 'bat' : 'sh';
+    const fname = `cmake-kit-test-${Math.random().toString()}.${fname_extension}`;
+    const script_path = path.join(paths.tmpDir, fname);
+    // generate a file with test batch / shell script that sets two env vars
+    if (process.platform == 'win32')
```
bobbrow: There should be curly braces around the `if` and `else` cases

src/kit.ts
```diff
@@ -769,6 +848,13 @@ export async function effectiveKitEnvironment(kit: Kit, opts?: expand.ExpansionO
           util.map(util.chain(host_env, kit_env, vs_vars), ([k, v]): [string, string] => [k.toLocaleUpperCase(), v]));
     }
   }
+  if (kit.environmentVariablesShellScript) {
+    const shell_vars = await getShellScriptEnvironment(kit);
+    if (shell_vars) {
+      // replace hostvars with variables set by shell script
```
bobbrow: This means you can't add your own custom variables script if you are using the Visual Studio compiler.  Is that intentional?

src/kit.ts
```diff
@@ -580,6 +595,70 @@ async function collectDevBatVars(devbat: string, args: string[], major_version:
...
+  } catch (error) { log.error(error); }
+  if (!env || env === '') {
+    console.log(`Error running ${kit.environmentVariablesShellScript} with:`, output);
+    return;
+  }
+
+  // split and trim env vars
+  const vars
+      = env.split('\n').map(l => l.trim()).filter(l => l.length !== 0).reduce<Map<string, string>>((acc, line) => {
+          const mat = /(\w+)=?(.*)/.exec(line);
```
bobbrow: We prefer it if you don't abbreviate variable names.  I think you mean `match` here.

src/kit.ts
```diff
@@ -580,6 +595,70 @@ async function collectDevBatVars(devbat: string, args: string[], major_version:
   return vars;
 }
 
+/**
+ * Gets the environment variables set by a shell script.
+ * @param kit The kit to get the environment variables for
+ */
+export async function getShellScriptEnvironment(kit: Kit): Promise<Map<string, string>|undefined> {
+  console.assert(kit.environmentVariablesShellScript);
+  const fname = Math.random().toString() + (process.platform == 'win32' ? '.bat' : '.sh');
```
bobbrow: We prefer it if you don't abbreviate variable names.  I think you mean `filename` here.

src/kit.ts
```diff
@@ -580,6 +595,70 @@ async function collectDevBatVars(devbat: string, args: string[], major_version:
...
+
+  const script = [];
+  if (process.platform == 'win32') { // windows
+    if (kit.environmentVariablesShellScriptDirectory) {
+      script.push(`cd /d "${kit.environmentVariablesShellScriptDirectory}"`); // switch to directory if set
+    }
+    script.push(`call "${kit.environmentVariablesShellScript}" ${args}`); // call the user batch script
+    script.push(`set >> ${env_path}`); // write env vars to temp file
+    }
+  else { // linux based
```
bobbrow: `else` goes on the previous line after the `}`

src/kit.ts
```diff
@@ -755,7 +834,7 @@ export async function getVSKitEnvironment(kit: Kit): Promise<Map<string, string>
 }
 
 export async function effectiveKitEnvironment(kit: Kit, opts?: expand.ExpansionOptions): Promise<Map<string, string>> {
-  const host_env = objectPairs(process.env) as [string, string][];
+  let host_env = objectPairs(process.env) as [string, string][];
```
bobbrow: `host_env` shouldn't be populated here if you're going to throw the result away when an environment script runs.

src/kit.ts
pisker: Okay, I fixed this

schemas/kits-schema.json
pisker: Yes, environmentSetupScript and "The absolute path to a script that modifies the environment for the Kit" sounds better.

schemas/kits-schema.json
pisker: A script path should be fine.

test/unit-tests/kitmanager.test.ts
pisker: Thats right

test/unit-tests/kitmanager.test.ts
pisker: Okay, I added them

src/kit.ts
pisker: Okay

src/kit.ts
pisker: Thats true

src/kit.ts
pisker: No, that was not intentional. I moved the if (environmentSetupScript) statement (which sets the host_env variable) before the if (visualStudio) statement.

src/kit.ts
pisker: Okay, I changed that so host_env only becomes populated if the shell / setup script was not run successfully.

schemas/kits-schema.json
pisker: Yes, no problem

pisker:
> @pisker, how easily do you think, one could source files?
The setup script is now run from /bin/bash. Because of that bash commands like source are available now

PR #1001, opened 2020-01-10
Update yarn lock.  Fix issue with Buffer causing a deprecation warning

Colengms:
Addressed warning: "Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead."

Also ran "yarn upgrade" to pick up updated dependencies.


package.json
```diff
@@ -1101,15 +1101,16 @@
     "ajv": "^4.7.5",
     "es6-promisify": "~5.0.0",
     "follow-redirects": "^1.5.10",
+    "handlebars": "^4.5.3",
```
bobbrow: Is this needed? I see it in the `resolutions` section already and we don't actively use this in the extension.

bobbrow: You can also remove `follow-redirects`. We forgot to remove that in a previous commit.

Issue #960, opened 2019-12-07
CMake Tools is sending x64 intelliSenseMode for x86 compiler


Colengms:
When using CMake Tools as a configuration provider for vscode-cpptools, I'm seeing that it returns x64 intelliSenseMode for x86 compiler.  'msvc-x86' is expected:

```
  "intelliSenseMode": "msvc-x64",
  "compilerPath": "C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe",
```

It looks like the only 2 values it returns are msvc-x64 and clang-x64.  The following is from cpptools.ts, line ~270.

```
intelliSenseMode: is_msvc ? 'msvc-x64' : 'clang-x64',
```

The x86 intelliSenseModes are relatively recent additions in cpptools, so this probably predates them.  Though, if the compiler is gcc, gcc intelliSenseMode's should be used.

andreeis:
Closing since this is fixed in CMake Tools Beta2. Please reactivate if you encounter any other symptoms of this issue.

PR #1002, opened 2020-01-10
Support additional IntelliSenseModes

bobbrow:
## This change addresses item #960

This could be improved in the future, but is a good start for now.


src/cpptools.ts
```diff
@@ -126,6 +126,31 @@ export function parseCompileFlags(args: string[], lang?: string): CompileFlagInf
   return {extraDefinitions, standard};
 }
 
+/**
+ * Determine the IntelliSenseMode.
+ */
+function getIntelliSenseMode(compiler_path: string) {
+  const compiler_name = path.basename(compiler_path || "").toLocaleLowerCase();
+  if (compiler_name.endsWith('cl.exe')) {
```
Colengc: The compiler might be clang-cl.exe , which should use clang intellisense modes.  Perhaps put this check after the check for clang?

bobbrow: I thought that clang-cl defaulted to msvc-style command switches.  Is my understanding wrong?

Colengms: (With the correct user account this time).   Command line switches, yes (a subset, anyway), but I believe the intelliSenseMode should still be clang.

Colengms: As discussed on another thread; since clang-cl supports MSVC language extensions, as well as clang language extensions not supported by MSVC, it's unclear to me whether clang or msvc intelliSense mode would be more correct.  (Does VS use clang or clang-cl in their clang support?  If clang-cl, perhaps this should be built into clang mode?).

Right now, I think cpptools will always(?) switch it to msvc mode.  There is a check that will switch from clang to msvc mode IF "Visual Studio" or "Windows Kits" is seen in an include path returned from probing the compiler. I'm seeing these paths returned, so it's unclear to me in what condition they would not be, or why that would be a reliable indicator that a different intelliSense mode needs to be used.

Also, I'm seeing CMake Tools detect clang-cl.exe on Windows LLVM installs, but not clang.exe .  So, someone using CMake Tools will never get the option to use clang mode intelliSense, as it will always get switched to msvc mode.

PR #1019, opened 2020-01-23
feat: add cmake gui

nieroger:
### This changes

Adds a new option "Open Configuration" to the 3-dotted menu in the treeview section. Opens a WebView with the cached cmake options (only boolean so far). Cmake options can be changed and saved. After save, cmake is configured.


Zingam:
@rogern01 So this is a GUI editor for the  CMake cache? I suggest to name it appropriately and more specifically something like "CMake Cache Editor", respectively the commands "Edit Cache Variables", etc.
IMHO CMake Configuration (Open Configuration) is too generic and could also mean a set of parameters passed to CMake via the command line like those configured via "CMakeSettings.json" file in VS2019 or other methods, which I hope will be supported at some point.

i18n/chs/package.i18n.json
```diff
@@ -21,6 +21,7 @@
 	"cmake-tools.command.cmake.outline.cleanConfigure.title": "清除重新配置",
 	"cmake-tools.command.cmake.clean.title": "清理",
 	"cmake-tools.command.cmake.cleanRebuild.title": "清除重新生成",
+	"cmake-tools.command.cmake.openConfiguration.title": "清除重新生成",
```
bobbrow: Just FYI, you don't need to edit these i18n json files.  Our automated process will take care of it.  You are welcome to leave the changes (perhaps you wanted them for your fork), but they will be overwritten when we merge this in.

Zingam: @bobbrow 
I was wondering if I should removed these from my PR. I am not quite sure what to do if I need to add a new localized string or change one. What is the proper procedure?

bobbrow: You don't need to do anything for localization besides call the `localize` function in your code with a new string+id.  If you have feedback on an existing translation, we can take that back to our localizers and have them update it, but you would do that by opening a new issue rather than creating a pull request to edit the json directly.

rogern01:
@Zingam thanks for your feedback, I've renamed it to "CMake Cache Editor" and "Edit CMake Cache" respectively.

gocarlos:
@rogern01 can you have a look on the pipeline?

bobbrow:
Sorry for the delay in looking at this PR.  We do plan to evaluate it for 1.4.0, but the team has been tied up with other issues and we will get back to this as soon as we can.

gocarlos:
@bobbrow any updates?

src/cache.ts
```diff
@@ -180,6 +180,67 @@ export class CMakeCache {
...
+        const currentVal = found[1];
+        const newValueLine = line.replace(currentVal, value ? 'TRUE' : 'FALSE');
+        resolve(content.replace(line, newValueLine));
+      } else {
+        resolve('');
+      }
+    });
+  }
+
+  async replaceOption(key: string, value: boolean) {
```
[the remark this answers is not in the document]
bobbrow: Please also add a function comment.

src/cache.ts
```diff
@@ -180,6 +180,67 @@ export class CMakeCache {
...
+      } else {
+        resolve('');
+      }
+    });
+  }
+
+  async save(key: string, value: boolean) {
+    const content = await this.replaceOption(key, value);
+    if (await fs.exists(this.path)) {
+      if (content) {
```
bobbrow, line 229: nit: It is faster to check the `content` than it is to stat a file.  Check `content` first.

src/cache.ts
```diff
@@ -180,6 +180,67 @@ export class CMakeCache {
...
+          content = await this.replace(content, option.key, option.value);
+        }
+        resolve(content);
+      } else {
+        resolve('');
+      }
+    });
+  }
+
+  async save(key: string, value: boolean) {
```
bobbrow, line 226: Please add a function comment

src/cache.ts
```diff
@@ -180,6 +180,67 @@ export class CMakeCache {
...
+    const content = await this.replaceOption(key, value);
+    if (await fs.exists(this.path)) {
+      if (content) {
+        await fs.writeFile(this.path, content);
+      }
+    }
+  }
+
+  async saveAll(options: Array<{key: string, value: boolean}>) {
```
bobbrow, line 235: Please add a function comment.

src/cache.ts
```diff
@@ -180,6 +180,67 @@ export class CMakeCache {
...
+        resolve(content.replace(line, newValueLine));
+      } else {
+        resolve('');
+      }
+    });
+  }
+
+  async replaceOption(key: string, value: boolean) {
+    return new Promise(async resolve => {
+      const exists = await fs.exists(this.path);
```
bobbrow, line 201: It is unclear to me why you need to check the existence of the file and read it again.  I would have expected the cache entries to have already been parsed and placed in the `this._entries` member variable.

bobbrow: (this feedback applies to all new functions added to this class)

src/cmake-tools.ts
```diff
@@ -167,6 +168,12 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
    */
   private _cmakeDriver: Promise<CMakeDriver|null> = Promise.resolve(null);
 
+  /**
+   * webview object
+   * handles everything regarding webview
```
bobbrow, line 173: This comment could be better.  This object manages the CMake GUI.  I recommend referencing that explicitly.

src/cmake-tools.ts
```diff
@@ -167,6 +168,12 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+   */
+  private _webview: ConfigurationWebview | undefined;
```
bobbrow, line 175: I recommend giving this a better name so that everyone knows this manages the Cmake GUI.

src/cmake-tools.ts
```diff
@@ -599,17 +606,64 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+            }
+          });
+        },
+    );
+  }
+
+  /**
+   * Configure without cmake settings
+   */
+  rawConfigure(): Thenable<number> {
```
bobbrow, line 642: I think this could be factored better.  I don't see any significant difference between this function and `configure` besides that this one passes in `[]` and an extra `true` to `drv.configure`.  We should avoid copying large portions of the code around because it is cumbersome and error-prone to keep both functions up to date.

Instead, you should create a new `ConfigureType` (maybe `ConfigureType.Cache`) and add a case label to the existing `configure` function.  Then rename `rawConfigure` to `cacheConfigure` and just call `configure([], ConfigureType.Cache)` here.

src/cmake-tools.ts
```diff
@@ -909,6 +963,32 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
   async clean(): Promise<number> { return this.build('clean'); }
 
+  /**
+   * Implementation of `cmake.openConfiguration`
+   */
+  async openConfiguration(): Promise<number> {
+    if (!this._webview) {
+      const drv = await this.getCMakeDriverInstance();
+      if (!drv) {
+        vscode.window.showErrorMessage('No CMakeCache.txt file has been found. Please configure project first!');
```
bobbrow, line 973: User-facing messages need to be translated into the appropriate language.  Please call `localize` on your message here and give it an id.  Our translation team will take care of the rest.

src/cmake-tools.ts
```diff
@@ -909,6 +963,32 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+        return new Promise(resolve => resolve(1));
```
bobbrow, line 974: Do not return a Promise from an `async` function.  `async` does this auto-magically.  Just `return 1;` here.

src/cmake-tools.ts
```diff
@@ -909,6 +963,32 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+      await this._webview.initPanel();
+
+      this._webview.panel.onDidDispose(() => {
+        this._webview = undefined;
+      });
+    } else {
+      this._webview.panel.reveal();
+    }
+
+    return new Promise(resolve => resolve(0));
```
bobbrow, line 989: `return 0;`

src/extension.ts
```diff
@@ -644,7 +644,7 @@ class ExtensionManager implements vscode.Disposable {
    */
   async setKitByName(kitName: string, folder?: vscode.WorkspaceFolder) {
     if (folder) {
-      await this._folders.get(folder)?.kitsController.setKitByName(kitName);
+      const result = this._folders.get(folder)?.kitsController.setKitByName(kitName);
```
bobbrow, line 647: Nothing is returned here.  Please revert this.  It should be awaited.

src/telemetry.ts
```diff
@@ -28,7 +28,7 @@ export function activate(): void {
 
 export function deactivate(): void {
     if (telemetryReporter) {
-        telemetryReporter.dispose();
+        telemetryReporter.dispose().then(() => {}, () => {});
```
bobbrow, line 31: Please revert.  This looks like something we may want to address in the future, but it wouldn't be part of this PR.

tslint.json
```diff
@@ -2,7 +2,6 @@
 	"rules": {
 		"no-unused-expression": true,
 		"no-duplicate-variable": true,
-		"no-unused-variable": true,
```
bobbrow, line 5: Please revert.  lint rules should not be changed as part of a feature PR.

src/webview.ts
```diff
@@ -0,0 +1,226 @@
+import * as vscode from 'vscode';
+import { CMakeCache } from './cache';
+import * as api from './api';
+import { CMakeDriver } from './drivers/driver';
+
+interface IOption {
```
bobbrow, line 6: I will take a look at this next.  I just wanted to get initial feedback out so I have not reviewed this file yet.

src/cache.ts
rogern01: will any changes done to the `this._entries` variable be persisted? @bobbrow

rogern01:
any updates @bobbrow ?

andreeis:
> any updates @bobbrow ?

@rogern01, we are looking at this now. Thank you very much for your contribution and patience.

gocarlos:
@bobbrow can you tell me whats the status here?

bobbrow:
We've had to spend some time working on another project, so we've fallen behind on PR's and issues in CMake Tools.  We're hoping to wrap that up in a few weeks and then start an effort to finish up the open PR's in this repository.  We're sorry for the delay and appreciate your patience as we've had to split our engineering time across projects.

gocarlos:
starts to be a bit depressing that this is open since more than half a year...

bobbrow:
Yes, I'm sorry about that.  Unfortunately, it has taken longer than expected to wrap up our other project.  I still intend to take this for 1.5.

src/cache-view.ts
```diff
@@ -0,0 +1,262 @@
...
+              } else {
+                tr.classList.remove('invisible');
+              }
+            }
+          }
+        </script>
+    </head>
+    <body>
+      <div class="container">
+        <button id="save" onclick="save()">Save</button>
```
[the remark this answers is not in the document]
andreeis: I added the localization of the web view strings as a new sub-item of https://github.com/microsoft/vscode-cmake-tools/issues/1467. Resolving this conversation and will not block merging this PR on implementing this.

bobbrow:
@andris I added some theming changes as a starting point for supporting light vs dark.  Please reference cpptools as well.  Not all of the elements have the same border colors or font colors and that still needs to be fixed. If you want to split this out into a follow up task so we can merge this, that is fine, but we need to complete it before releasing 1.5.0

andreeis:
> @andris I added some theming changes as a starting point for supporting light vs dark. Please reference cpptools as well. Not all of the elements have the same border colors or font colors and that still needs to be fixed. If you want to split this out into a follow up task so we can merge this, that is fine, but we need to complete it before releasing 1.5.0

Yes, I added this as a new entry in work item https://github.com/microsoft/vscode-cmake-tools/issues/1467. 
As for merging this PR, I can remove the "#" and we can always add it back if @gocarlos replies with the reason behind it.

gocarlos:
Yes, i’m ok with removing it

bobbrow:
I replaced this PR with #1490.  Something is wrong with the branch and the extension won't activate in production mode.  I reconstructed the changes in a new branch off develop and everything appears to work as expected.

bobbrow:
@gocarlos With this PR having been released for a few months now, do you need to keep your fork of the CMake Tools extension in the Marketplace, or can it be removed?

gocarlos:
I removed it now

bobbrow:
Thanks!

Issue #1025, opened 2020-01-27
"sourceDirectory" is not a directory exception


DexterMagnific:
### Cannot configure when toplevel CMakeLists.txt not in wkspc folder

I know this bug seems to be fixed in #533, but I still have it.

Here is our dir structure:

```
projectRoot/
     ide/VSCode/ourProject.code-workspace
     src/CMakeLists.txt
```

the workspace file contains the following:

```
{
	"folders": [
		{
			"path": "../../src"
		},
	],
	"settings": {
		"cmake.sourceDirectory": "{workspaceFolder}/../../src",
		"cmake.buildDirectory": "${workspaceFolder}/../../build"
	}
}
```

### Expected:

1. CMake -> Clean Reconfigure
2. CMake configure is OK

### Apparent Behavior:

Exception:

[rollbar] Unhandled exception: Unhandled Promise rejection: cleanConfigure [cmake-server] Failed to activate protocol version: "sourceDirectory" is not a directory. {}

### Platform and Versions

- **Operating System**: Ubuntu
- **CMake Version**: 3.13.4
- **VSCode Version**: 1.41.1
- **CMake Tools Extension Version**: 1.2.3
- **Compiler/Toolchain**: GCC 8.3

bobbrow:
It looks like you have a typo here:
```
"cmake.sourceDirectory": "{workspaceFolder}/../../src"
```

Can you add a `$` before the `{workspaceFolder}` and see if that fixes it?

DexterMagnific:
Yes this was a typo. I get the same error with `$`. I even removed the variable and left only `../../src` with no success.

What is surprising is that I used to get the same exception but with the directory in the message. Now it is empty.

bobbrow:
Sorry, I gravitated toward that instead of the layout of your code.  Your ${workspaceFolder} is actually `../../src` based on what I see in the code-workspace file.  Can you try just:
```
"cmake.sourceDirectory": "${workspaceFolder}"
```
I created a test with a similar layout and that works for me.  I'm not sure why the error doesn't show the path.  I'll make sure there is something in the log that tells you what the failing folder is.

DexterMagnific:
I tried your suggestion, now I don't have the exception, but a popup saying:
"CMakeLists.txt was not found in the root of the workspace folder, "quickstart a new ...", "edit ...."".

And I don't understand why the workspace folder is src. I opened the workspace file from ide/VSCode so for me this is the workspace folder.

[...]

PR #1028, opened 2020-01-28
Print the value of sourceDirectory when it is invalid

bobbrow:
## This change addresses item #1025

Log an error if we detect that sourceDirectory is invalid.


src/drivers/driver.ts
```diff
@@ -398,6 +398,9 @@ export abstract class CMakeDriver implements vscode.Disposable {
         // Don't fail if CMakeLists.txt was accidentally appended to the sourceDirectory.
         this._sourceDirectory = path.dirname(this._sourceDirectory);
       }
+      if (!(await util.checkFileExists(this._sourceDirectory))) {
```
sean-mcmanus: Should this warning go above the "cmakelists.txt" correction above?

bobbrow: no, but this should be checking the folder, not a file.  I'm not sure why I didn't see that earlier.

PR #1035, opened 2020-01-30
Added command to hide launch and debug commands and debug button

xisui-MSFT:
I don't really like this way, but didn't find a better one...

The reason to hide those commands is that the program may not be a desktop program. So commands to set/get launch/debug targets are not hidden since they could still be useful even if the program is not targeting desktop.

### DIFF
--- a/package.json
+++ b/package.json
@@ -366,6 +366,14 @@
       {
         "command": "cmake.outline.revealInCMakeLists",
         "title": "%cmake-tools.command.cmake.outline.revealInCMakeLists.title%"
+      },
+      {
+        "command": "cmake.hideLaunchCommand",
+        "title": "%cmake-tools.command.cmake.hideLaunchCommand.title%"
+      },
+      {
+        "command": "cmake.hideDebugCommand",
+        "title": "%cmake-tools.command.cmake.hideDebugCommand.title%"
       }
     ],
     "menus": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -366,6 +366,14 @@
       {
         "command": "cmake.outline.revealInCMakeLists",
         "title": "%cmake-tools.command.cmake.outline.revealInCMakeLists.title%"
+      },
+      {
+        "command": "cmake.hideLaunchCommand",
```
bobbrow: You don't need to add this to `package.json`.  Just register it in `extension.ts`.  You extension can call it without it being in package.json.

src/extension.ts
```diff
@@ -854,6 +856,17 @@ class ExtensionManager implements vscode.Disposable {
   resetState(folder?: vscode.WorkspaceFolder) { return this.mapCMakeToolsFolder(cmt => cmt.resetState(), folder); }
 
   async viewLog() { await logging.showLogFile(); }
+
+  async hideLaunchCommand(shouldHide: boolean = true) {
+    // Don't hide command to launch/debug target here since the target can still be useful, one example is in launch.json
```
bobbrow: I don't understand this comment.  You are hiding the launch target command with the same "context" state.

src/status.ts
```diff
@@ -84,7 +84,7 @@ export class StatusBar implements vscode.Disposable {
       setVisible(item, this._visible && !!item.text);
     }
     setVisible(this._debugButton,
-               this._visible && vscode.extensions.getExtension('ms-vscode.cpptools') !== undefined
+               this._visible && !this._hideDebugButton && vscode.extensions.getExtension('ms-vscode.cpptools') !== undefined
```
bobbrow: I think you need to hide `this._launchTargetNameButton` as well.  What about the ctest button?

package.json
xisui-MSFT: I got an error message when debugging the extension... Maybe I don't need to add anything to package.json?

src/extension.ts
xisui-MSFT: Typo... Should be get/set launch/debug target

package.json
bobbrow: >Maybe I don't need to add anything to package.json?

That was my understanding.  For cpptools, commands that we don't want to show up in the command palette are not added to package.json.

Example:
```
    disposables.push(vscode.commands.registerCommand('cpptools.activeConfigName', onGetActiveConfigName));
```

src/status.ts
xisui-MSFT: As explained in the description, launchTargetName could still be useful, because it provides the wanted binary name. So it could be used in launch.json.

ctest isn't relevant since it has nothing to do with the desktop program assumption.

PR #1066, opened 2020-02-12
Show "Collapse all" command on project outline view

dirondin:
### This change adds a feature

- Increase version of dependency `vscode` and `engines.vscode` to ^1.30.0 (needs to using new createTreeView API)
- Use `window.createTrewView` instead `window.registerTreeDataProvider` for additional option `showCollapseAll` (its API used in vscode for adding Collapse all action on tree views)

image


bobbrow:
Thank you for your help with this!

PR #1082, opened 2020-02-17
Spanish translation is wrong. Update cmake-tools.i18n.json

pablo-benito:
The spanish translation for 'building.status'  is wrong. 

- update the i18n/esn/src/cmake-tools.i18n.json

## The purpose of this change

The spanish translation for 'building.status' in this context is absolutely wrong. 'Edificio' is a noun, and means house, or construction.. 
The correct translation is 'Compilando' as in building.target


bobbrow:
Thanks for spotting this!  I'll let the translators know.  They'll update the translation source on their side which will auto generate a new pull request with the changes.

lukaschal:
Fixed as suggested.

Src\vscode-extensions\vscode-cmake-tools.xlf
ID: 0; "./src/cmake-tools.building.status"
SRC: Building
new tgt: Compilando

bobbrow:
Thanks @lukaschal. Replacing this PR with #1086

bobbrow:
@lukaschal it looks like there's a similar issue with Portuguese.  I'll open another bug.

Issue #1099, opened 2020-02-27
Custom browse configuration path irregularities


RobRuckOB:
(I initially mistakenly reported this bug to the C/C++ extension team)

Whilst looking at another unrelated issue with the C/C++ extension I turned on debug logging and saw something unusual in the paths listed in the "Custom browse configuration received:" section. Specifically, there are repeated lower-cased versions of paths (I can just about believe that's intentional due to platform differences) and/or paths with the workspace folder appended twice (I can't think of any reason for this). If this is a somewhat scattergun approach to finding things, that's not ideal, but it also doesn't seem to work quite right - I see some paths in the lower case or repeated style that don't appear correctly in the list of paths at all.

- OS and Version: OSX 10.14.6
- VS Code Version: 1.42.1
- C/C++ Extension Version: 0.26.3
- CMake Version: 3.16.2
- CMake Tools Extension Version: 1.3.1
- Compiler/Toolchain: clang 9.1.0

**To Reproduce**
My c_cpp_properties.json:
`{
    "configurations": [
        {
            "name": "Mac",
            "includePath": [
                "${workspaceFolder}/**",
                "~/.conan/data/**"
            ],
            "defines": [],
            "macFrameworkPath": [
                "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/System/Library/Frameworks"
            ],
            "compilerPath": "/usr/bin/clang",
            "cStandard": "c11",
            "cppStandard": "c++17",
            "configurationProvider": "vector-of-bool.cmake-tools"
        }
    ],
    "version": 4
}`
I simply start up VS Code, and the paths issue is visible in the start up debug output from the extension.

**Expected behavior**
Only valid paths are set in the browse configuration.

**Screenshots**
Lower casing of path (Notice /users):
<img width="442" alt="image" src="https://user-images.githubusercontent.com/42342894/75325319-83db7300-5870-11ea-9d8c-a089a64b6200.png">

Lower casing of path and repetition of the workspace folder:
<img width="857" alt="image" src="https://user-images.githubusercontent.com/42342894/75324793-9c975900-586f-11ea-8751-4c43d694eea9.png">

Subsequent errors:
<img width="1139" alt="image" src="https://user-images.githubusercontent.com/42342894/75325022-fdbf2c80-586f-11ea-92d5-f455b535c041.png">

bobbrow:
[...]

PR #1103, opened 2020-03-02
Normalize all paths sent to cpptools

bobbrow:
## This change addresses item #1099

Normalize relevant paths sent to cpptools.  This should remove the duplicate entries in the browse configuration seen on Windows and macOS.  It also makes the paths sent for SourceFileConfigurationItem objects consistent with the browse path.


bobbrow:
Stylistically we decided not to go that route.  Thanks for the review!

PR #1109, opened 2020-03-10
Add support for Objective-C and Objective-C++ languages

marksisson:
### This changes visible behavior of logging

- Add OBJC and OBJCXX as recognized languages in vscode-cmake-tools cpptools parser


bobbrow:
We'll take a look. Thank you!

marksisson:
I do use the cpptools extension, and this change to cmake-tools has not caused any problems for me.  As you have stated, I do not get IntelliSense support for the Objective-C/C++, but this update does cleanup the cmake-tools output.

bobbrow:
Ok, so this just cleans up output from CMake Tools?  I can approve it then.

Issue #1045, opened 2020-02-04
Change the configuration provider id to "ms-vscode.cmake-tools"


bobbrow:

andreeis:
CMake Tools 1.4.0 Beta is available on GitHub.
https://github.com/microsoft/vscode-cmake-tools/releases/tag/1.4.0-beta

Please try it out and let us know if you encounter any issues.

andreeis:
This fix is available in CMake Tools #1.4.0 which has been released.

Issue #870, opened 2019-11-03
can't detect gcc 9.2.0 on linux with non-English output


xspeed1989:
$gcc --version
gcc (GCC) 9.2.0
Copyright © 2019 Free Software Foundation, Inc.

CMakeTools ver:0.0.17

$cmake --version
cmake version 3.16.0-rc3

CMake suite maintained and supported by Kitware (kitware.com/cmake).

log:
[Extension Host] [CMakeTools] 2019-11-03T19:46:26.486Z [debug] [extension] [2484] cmake.scanForKits started
workbench.desktop.main.js:sourcemap:243 [Extension Host] [CMakeTools] 2019-11-03T19:46:26.487Z [debug] [extension] Rescanning for kits
workbench.desktop.main.js:sourcemap:243 [Extension Host] [CMakeTools] 2019-11-03T19:46:26.487Z [debug] [kit] Scanning for Kits on system
workbench.desktop.main.js:sourcemap:243 [Extension Host] [CMakeTools] 2019-11-03T19:46:26.488Z [debug] [kit] Scanning directory /usr/bin for compilers
workbench.desktop.main.js:sourcemap:243 [Extension Host] [CMakeTools] 2019-11-03T19:46:26.488Z [debug] [kit] Skipping scan of not existing path /home/john/.local/bin
workbench.desktop.main.js:sourcemap:243 [Extension Host] [CMakeTools] 2019-11-03T19:46:26.488Z [debug] [kit] Found 0 kits in directory /home/john/.local/bin
workbench.desktop.main.js:sourcemap:243 [Extension Host] [CMakeTools] 2019-11-03T19:46:26.488Z [debug] [kit] Skipping scan of not existing path /home/john/.local/bin
workbench.desktop.main.js:sourcemap:243 [Extension Host] [CMakeTools] 2019-11-03T19:46:26.488Z [debug] [kit] Found 0 kits in directory /home/john/.local/bin
workbench.desktop.main.js:sourcemap:243 [Extension Host] [CMakeTools] 2019-11-03T19:46:26.488Z [debug] [kit] Scanning directory /bin for compilers
workbench.desktop.main.js:sourcemap:243 [Extension Host] [CMakeTools] 2019-11-03T19:46:26.489Z [debug] [kit] Scanning directory /usr/bin/site_perl for compilers
[...]

PR #1116, opened 2020-03-11
Update cpptools.ts

ChristianS99:
change old extensionId to new

I had problem, that .vscode/settings.json was created with "C_Cpp.default.configurationProvider": "vector-of-bool.cmake-tools", changing this to ms-vscode.cmake-tools fixed Intellisense for me.

I took a look at the source and I *think*, this is the relevant part, but since I didn't want to dive in extension development, maybe someone can confirm, if this is the right place...

### DIFF
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -181,7 +181,7 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
   /** Our name visible to cpptools */
   readonly name = 'CMake Tools';
   /** Our extension ID, visible to cpptools */
-  readonly extensionId = 'vector-of-bool.cmake-tools';
+  readonly extensionId = 'ms-vscode.cmake-tools';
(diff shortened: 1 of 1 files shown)


bobbrow:
cpptools has special code to support both id's for us due to the ownership transition, so I'd be surprised if the rename was what fixed it for you.  We were planning on doing this in 1.4.0 once a critical mass had updated to the latest cpptools.  Thanks for doing it for us. 😃 

Tagging issue: #1045

ChristianS99:
I retested by changing back to vector-of-bool.cmake-tools, and you're right. that's not the fix.
it seems intellisense works with clang, but not with gcc.
but funnily, when i select clang kit and push F12 on cout, it opens the gcc header...

ChristianS99:
Further digging in my original problem, revealed, that it is actually #870.

PR #1130, opened 2020-03-18
Update and sort settings documentation

zalava:

- Sort the settings in alphabetical order
- Add ``cmake.parallelJobs`` to settings documentation


docs/settings.rst
```diff
@@ -238,6 +204,48 @@ produces in its output channel. Set to one of "trace", "debug", "info", "note",
...
+need to be on your ``$PATH`` for it to be found via Kit scanning.
+
+- Default: ``["C:\\MinGW"]`` (Search in C:\\MinGW for a MinGW installation)
+
+.. _conf-cmake.parallelJobs:
+
+``cmake.parallelJobs``
+**********************
+
+By specifying a number, you can define how many jobs are run in parallel during the build.
```
bobbrow, line 222: I would probably say that this controls how many CPU cores are used during the build.  `-j` is passed to build tools that support it, but not all do.  I believe msbuild uses `/maxcpucount:`.  I think if you just say that this controls the number of CPU cores used during the build, that should be sufficient.

zalava: You're right. I changed it.

bobbrow:
Thank you!

Issue #1124, opened 2020-03-17
CMake Cache Entries with double quotes produce error with CMake Tools


tmaslach:
### Brief Issue Summary
We have set statements like this in our CMakeLists.txt files:
`set("PROJECT_COMPONENT:${PROJECT_NAME}" controller CACHE STRING)
`

This causes the CMakeCache entry to have double quotes in it: 
`"PROJECT_COMPONENT:controller":STRING=controller
`

I believe the double quotes cause the following warning to get generated in the CMake output:
`[rollbar] Cache entry '"PROJECT_COMPONENT' has unknown type: 'controller":STRING' {}
`

If I had only a few of these, it wouldn't be an issue.  But, we have hundreds of entries that are getting double quotes, meaning the output is overwhelming.  I assume these messages are being printed by VSCode while it is extracting the values from CMakeCache.txt

Is there any way to fix this? Or at a minimum turn the warnings off?  The project builds fine, but many get confused because of the long list of messages.

### Expected:
If we have a set statement like this in our CMakeLists.txt file:
`set("PROJECT_COMPONENT:${PROJECT_NAME}" controller CACHE STRING)
`
I would expect no errors to be produced in the CMake/build output terminal.

### Platform and Versions
- **Operating System**: Linux and Windows 
- **CMake Version**: cmake-3.16.4
- **VSCode Version**: 1.43
- **CMake Tools Extension Version**: 1.3.1
- **Compiler/Toolchain**: GCC & VSProfessional

tmaslach:
I took a look at the code and this looked pretty easy to fix.  So, I will initiate a pull request.  Please let me know if you have issues with the code.  Tests all passed on Travis for linux, I'm waiting on Mac, but I don't expect any issues.

bobbrow:
Thanks for the PR.  We'll take a look. 😃

andreeis:
CMake Tools 1.4.0 Beta is available on GitHub.
https://github.com/microsoft/vscode-cmake-tools/releases/tag/1.4.0-beta

Please try it out and let us know if you encounter any issues.

tmaslach:
It's looking good so far!  

It is wonderful to not see 1000+ cache entry errors after building! :)

andreeis:
This fix is available in CMake Tools #1.4.0 which has been released.

PR #1133, opened 2020-03-19
Fix for: CMake Cache Entries with double quotes produce error with CMake Tools #1124

tmaslach:
## This change addresses item 
#1124 

## This change will properly:
parse CMake cache entries that follow this pattern:
"MY:VARIABLE":STRING=value

these can be generated by a user by doing the following in a CMakeLists.txt file:
set ("MY:VARIABLE" value CACHE STRING)

## Other Notes/Information
Test has been added as well


lygstate:
Can it squash and rebase to master?

tmaslach:
I got my 3 commits squashed into 1 and rebased with master.  Please look it over, but I think I did it right.

src/cache.ts
```diff
@@ -139,7 +139,8 @@ export class CMakeCache {
       if (line.startsWith('//')) {
         docs_acc += /^\/\/(.*)/.exec(line)![1] + ' ';
       } else {
-        const match = /^(.*?):(.*?)=(.*)/.exec(line);
+        const entry_pattern = line.startsWith('\"') ? /^\"(.*?)\":(.*?)=(.*)/ : /^(.*?):(.*?)=(.*)/;
```
bobbrow: Usually, I would prefer to use regex syntax to differentiate between the two cases:
```
^("(.*?)"|(.*?)):([^:]*?)=(.*)
```
Then you can use just one pattern.  (you'd have to update the unpacking on line 148 though since there's an extra capture group in this one).

tmaslach: Hi @bobbrow - thanks for the feedback.  I like what you are suggesting, but I have not been able to  get it working.  I plugged it at https://regexr.com/ as well and can reproduce the issue.  For whatever reason, the extra parens you added for the either-or aren't liked.  

I tried this, and it works: 
`(^"(.*?)"|(.*?)):([^:]*?)=(.*)`
But, I don't want the carat included in grouping.

I got this to work as well:
`"(.*?)"|(.*?):([^:]*?)=(.*)`
But, this makes the capturing a little crazy (could be in the first group or second group)

Is there something I'm missing that you can see?  I'm pretty decent with regex, but this one has me a little surprised.

tmaslach: Ignore my previous comment - it was too late when I was looking at this.  For the record, I deleted the comment because I couldn't seem to add a new comment otherwise on this thread.

bobbrow: Nno worries.  I was just going to comment that I tested my regex on https://regex101.com/ and it worked there.  I'm not familiar with the site you shared so I don't know what's going on there.

bobbrow: With your current set of changes though, I think you need to update line 147 to unpack the values correctly.  I believe it would be (add a comma):
```ts
const [, , name, typename, valuestr] = match;
```

tmaslach: Yep, already done.. I used the following to unpack:
`const [, name, , , typename, valuestr] = match;`
But, I am realizing that there is one caveat - the name for the quote case that I want is in group 2, but group 3 (or 1) for the non-quoted case. I'll need extra code for that.
Or, conversely, I can use group 1 for both, but manually need to strip the quotes if they are there.

Do you have a preference on one of those two ways?  Or perhaps another idea? 

See: https://regex101.com/r/djovn6/1

tmaslach: Ok, I think this will work:
`^(?|"(.*?)"|(.*?)):([^:]*?)=(.*)`
Learned something new today! :)

tmaslach: Hmm.. Spoke to soon, I think. Seems that typescript doesn't support this?

tmaslach: Another alternative - I kind of like this:
```
        const match = /^("(.*?)"|(.*?)):([^:]*?)=(.*)/.exec(line);
        if (!match) {
          rollbar.error(localize('failed.to.read.line.from.cmake.cache.file', 'Failed to read a line from a CMake cache file {0}', line));
          continue;
        }
        const filtered = match.filter(element => element != undefined)
        const [, , name, typename, valuestr] = filtered;
```

I'm not very well versed with javascript or typescript (I'm a C++ & Python guy).  So the above may not not be very clear to someone with more experience in typescript :)

bobbrow: I think that's what I suggested earlier. 😉 But your PR currently has the commas in the wrong place for unpacking the result.  Update it to what you typed above and I think it's good to go.

I would also recommend adding another test:
const str = "\"QUOTED\":STRING=value";

bobbrow: And yes, I realize now that I missed that regex101 was filtering the undefined results automatically, so my original proposed solution was missing that.

bobbrow: Just update the PR with your proposed solution and we'll approve it.

src/cache.ts
```diff
@@ -139,12 +139,13 @@ export class CMakeCache {
...
         docs_acc += /^\/\/(.*)/.exec(line)![1] + ' ';
       } else {
-        const match = /^(.*?):(.*?)=(.*)/.exec(line);
+        const match = /^("(.*?)"|(.*?)):([^:]*?)=(.*)/.exec(line);
         if (!match) {
           rollbar.error(localize('failed.to.read.line.from.cmake.cache.file', 'Failed to read a line from a CMake cache file {0}', line));
           continue;
         }
-        const [, name, typename, valuestr] = match;
+        const filtered = match.filter(element => element != undefined);
```
bobbrow, line 147: tslint is complaining about this line.  I believe you can just do:
```
const filtered = match.filter(element => element);
```

src/cache.ts
tmaslach: I had to fiddle with the filter line a little to get it past the linter, but it's compiling and all tests are passing on linux (including the new one I added based on your suggestion).
I expect Mac will pass on travis as well, so I think this PR will be good to go!

src/cache.ts
tmaslach: Let me try your solution.  It's cleaner than what I came up with.

tmaslach: Looks like it works! Thanks for the suggestion.

bobbrow: Did you push it?  It looks like you have something else currently

src/cache.ts
tmaslach: Sorry, my fault... It should now be what you expect.

src/cache.ts
tmaslach: Sorry, my fault... It should now be what you expect.

But - it does look like something is failing, but I don't think it's related to my change?   Either that or something odd is up with Travis.  Mac failed, but linux is acting terribly slow.
See: https://travis-ci.org/github/tmaslach/vscode-cmake-tools

bobbrow: It looks related to the line that filters out undefined results.  I think we need to remove the filter and replace with:
```ts
        const [, , quoted_name, unquoted_name, typename, valuestr] = match;
        const name = quoted_name || unquoted_name;
```

tmaslach: Interesting.. Wasn't expecting that to be related.  

Alright, updated the change to include the lines you suggested.  Too bad, was trying hard not to have a quoted and unquoted variable! 

All is green!

bobbrow: > Too bad, was trying hard not to have a quoted and unquoted variable!

No worries! 🤷‍♂ 😄

bobbrow:
Thanks for your PR.  It is now merged!

PR #1144, opened 2020-03-27
Update vswhere to a version that supports utf-8

bobbrow:


bobbrow:
@lygstate I didn't write this code originally so I'm making the minimum change required to fix the scenario.  If we decide to refactor the way we spawn processes in the future, we can revisit this decision.

lygstate:
OK, LTGM

bobbrow:
Something's wrong with the webhook.  I see the build succeeded on Travis, but the status is not being reported here.

Issue #1073, opened 2020-02-13
Multi-root workspaces don't provide per-folder browse configuration info


sean-mcmanus:
With a multi-root workspace, provideBrowseConfiguration is only returning info on one workspace folder and canProvideBrowseConfigurationsPerFolder should return true...if it that's impossible it should return the browse data for all the workspace folders instead of one.

This isn't working for the existing cpptools or our pending 1.0 release that has a completely different multi-root implementation.

sean-mcmanus:
In cpptools.ts, CppConfigurationProvder._workspaceBrowseConfiguration data just gets overwritten by the 2nd workspace folder data.

@xisui-MSFT @andreeis Are any of you able to work on this? This is a high priority for our cpptools 1.0 release.

sean-mcmanus:
Looks like provideFolderBrowseConfiguration also needs to be implemented.

sean-mcmanus:
I figured out a potential hack/workaround via skipping the bogus browse configuration if none of the paths share the same workspace folder, which is usually the case when we're given browse config info for the wrong workspace folder.

andreeis:
CMake Tools 1.4.0 Beta is available on GitHub.
https://github.com/microsoft/vscode-cmake-tools/releases/tag/1.4.0-beta

Please try it out and let us know if you encounter any issues.

andreeis:
This fix is available in CMake Tools #1.4.0 which has been released.

PR #1156, opened 2020-04-02
Add support for per-folder browse.path.

sean-mcmanus:
Fix for https://github.com/microsoft/vscode-cmake-tools/issues/1073


sean-mcmanus:
This generally works, but I'm still investigating some potential issues...

sean-mcmanus:
Are there any multi-root CMake Tools automated tests? I didn't see any.

bobbrow:
They are under `test/extension-tests/multi-root-UI`. There is a config in launch.json that will run them.  The tests are not currently written to handle Windows very well.  You should run them on linux or macOS.

src/cpptools.ts
```diff
@@ -304,6 +309,8 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
       compilerArgs: flags || undefined
     };
 
+    this._workspaceBrowseConfigurations.set(util.platformNormalizePath(opts.folder), this._workspaceBrowseConfiguration);
```
bobbrow, line 312: This seems incorrect.  `this._workspaceBrowseConfiguration` is supposed to be the accumulation of all browse paths across all folders (it is returned by `provideBrowseConfiguration`).  I think you will need to create a local `cpt.WorkspaceBrowseConfiguration` in `updateConfigurationData` (instead of reinitializing this one) and pass it through to the functions it calls.

src/cpptools.ts
```diff
@@ -175,6 +175,10 @@ export interface CodeModelParams {
    * The active target
    */
   activeTarget: string|null;
+  /**
+   * Workspace folder.
+   */
+  folder: string
```
bobbrow, line 181: Should we make this a Uri so that we know it's expected to be an absolute path?  If not, please add that to the comment.

src/cpptools.ts
```diff
@@ -284,7 +289,7 @@ export class CppConfigurationProvider implements cpt.CustomConfigurationProvider
     const {standard, extraDefinitions} = parseCompileFlags(flags, lang);
     const defines = (fileGroup.defines || target.defines).concat(extraDefinitions);
     const includePath = fileGroup.includePath ? fileGroup.includePath.map(p => p.path) : target.includePath;
-    const normalizedIncludePath = includePath.map(p => util.platformNormalizePath(p));
+    const normalizedIncludePath = includePath.map(p => util.lightNormalizePath(p));
```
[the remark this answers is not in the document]
sean-mcmanus: It was buggy. Causing duplicates with different casing. What consistency are you referring to?

src/cpptools.ts
sean-mcmanus: this._workspaceBrowseConfiguration isn't actually be user-hittable by our C/C++ extension. Are you referring to possible other callers?

src/cpptools.ts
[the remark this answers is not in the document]
sean-mcmanus: Are you saying I should fix the paths that are not filtered via platformNormalizePath? e.g. previously it would have ../Folder/.. and ../folder/.. paths.

src/cpptools.ts
sean-mcmanus: A Uri doesn't seem necessary and a caller using a relative path seems unlikely, but I can add more comment text.

src/cpptools.ts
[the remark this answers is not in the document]
bobbrow: I don't think the shipping bits don't have the `platformNormalize` fix in them yet.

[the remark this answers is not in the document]
sean-mcmanus: Yeah, that's what I figured. However, it looks like we need to not change the cases in this particular case, so I'm adding more comments to explain why.

src/cpptools.ts
sean-mcmanus: I think the current behavior is okay unless we get complaints from consumers. The usage of the API is expected to be how our C/C++ handles it via check the per-workspace folder data first and ignore the other data.

sean-mcmanus:
> They are under test/extension-tests/multi-root-UI. There is a config in launch.json that will run them.  The tests are not currently written to handle Windows very well.  You should run them on linux or macOS.

Ah, thanks...not sure how I didn't see that.

sean-mcmanus:
> They are under test/extension-tests/multi-root-UI. There is a config in launch.json that will run them.  The tests are not currently written to handle Windows very well.  You should run them on linux or macOS.
> 
> Ah, thanks...not sure how I didn't see that.

I added some browsePath checks to the test.

sean-mcmanus:
@bobbrow Why is the continuous-integration/travis-ci stuck?

bobbrow:
The travis CI is actually running, but the results are not being reported back to GitHub for some reason.  I've been checking here for results: https://travis-ci.org/github/microsoft/vscode-cmake-tools/pull_requests.  I also added my feedback to a bug report, but haven't heard anything back from the Travis team.  https://travis-ci.community/t/status-checks-randomly-broken/7876

We need to switch this repo over to Azure pipelines anyway, so I haven't investigated much.

test/unit-tests/cpptools.test.ts
```diff
@@ -61,20 +61,58 @@ suite('CppTools tests', () => {
...
+                sources: [sourceFile2],
+                isGenerated: false,
+                compileFlags: '-DFLAG1',
+                language: 'CXX'
+              }]
+            }]
+        }]
+      }]
+    };
+    provider.updateConfigurationData({cache, codeModel: codeModel2, activeTarget: 'target1', folder: smokeFolder});
```
bobbrow, line 110: I think it would be better to pass in all of the code models up front before doing any validation.  This would help validate that the different folder states do not pollute the API responses.

src/cpptools.ts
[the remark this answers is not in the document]
bobbrow: > Oh, yeah, there's a problem with vscode.workspace.getWorkspaceFolder returning undefined when the path is lowercased on Windows

Wouldn't that be a bug for vscode?  I don't think we can guarantee that CMake will return correctly-cased paths.  And vscode shouldn't fail a path comparison based on incorrect casing on Windows.

test/unit-tests/cpptools.test.ts
sean-mcmanus: Ok

sean-mcmanus:
How did this PR get submitted? It was not intentional.

bobbrow:
It's ok.  I do want you to update the normalize cases to use `platformNormalize` instead though.

bobbrow:
And it would probably be better to update the test to have unique target values in `codeModel2` just so we don't have to worry about bugs sneaking through.

sean-mcmanus:
@bobbrow Sure, but I think I should delay that change until we ship our next C/C++ extension version, so 0.27.0 aren't broken...in case there's a reason to ship a CMake Tools update before then.

bobbrow:
Can you create the PR now and we'll just wait to submit it then?  If we don't, we're going to forget since this is a month or more away.

sean-mcmanus:
> Can you create the PR now and we'll just wait to submit it then?  If we don't, we're going to forget since this is a month or more away.

https://github.com/microsoft/vscode-cmake-tools/pull/1163

Issue #1084, opened 2020-02-20
Ninja generator is used instead of Visual Studio 16 2019


Celeborn2BeAlive:
### Brief Issue Summary

When I select the kit "Visual Studio Community 2019 Release - amd64", vscode-cmake-tools is passing '-G Ninja' to cmake instead of '-G "Visual Studio 16 2019" -A x64'.

My cmake-tools-kit.json file contains the entry:
```json
  {
    "name": "Visual Studio Community 2019 Release - amd64",
    "visualStudio": "da769ae6",
    "visualStudioArchitecture": "amd64",
    "preferredGenerator": {
      "name": "Visual Studio 16 2019",
      "platform": "x64"
    }
  },
```

What is strange is that I get the correct behavior when I select "Visual Studio Community 2017 Release - amd64" (in that case I can see '-G "Visual Studio 15 2017" -A x64' passed to cmake, which is correct). For this one the entry is:
```json
  {
    "name": "Visual Studio Community 2017 Release - amd64",
    "visualStudio": "05246a08",
    "visualStudioArchitecture": "amd64",
    "preferredGenerator": {
      "name": "Visual Studio 15 2017",
      "platform": "x64"
    }
  },
```
that seems pretty similar to the one for VS 2019. So I don't know how vscode-cmake-tools interprets the file cmake-tools-kits.json, but it seems that it does treat both cases in the same way.

### Expected:

1. "CMake: Select a Kit" command
2. Choose "Visual Studio Community 2019 Release - amd64"
3. "CMake: Configure" command
4. cmake should be called with option '-G "Visual Studio 16 2019" -A x64'

### Apparent Behavior:

1. "CMake: Select a Kit" command
2. Choose "Visual Studio Community 2019 Release - amd64"
3. "CMake: Configure" command
4. cmake is called with option '-G Ninja'

### CMake Tools Log

```
[main] Configuring folder: gltf-viewer-tutorial-git 
[proc] Executing command: C:\Users\laure\scoop\shims\cmake.EXE --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -Hd:/projects/opengl_tds/gltf-viewer/gltf-viewer-tutorial-git -Bd:/projects/opengl_tds/gltf-viewer/build-gltf-viewer-tutorial -G Ninja
[cmake] Not searching for unused variables given on the command line.
[cmake] -- The C compiler identification is MSVC 19.23.28106.4
[...]

Issue #964, opened 2019-12-09
Wrong compiler platform selected with Visual Studio 16 2019 kit


johannesstricker:
### Brief Issue Summary
The automatically detected kit for Visual Studio 16 2019 for the x86 platform incorrectly uses the x64 compiler. This can be fixed by adding `"platform": "Win32"` to the kit configuration in `cmake-tools-kits.json` and changing `"visualStudioArchitecture"` from `"x86"` to `"Win32"`.

### Expected:
1. Scan for kits
2. Select the `Visual Studio 16 2019 - x86` kit
3. Configure and build
4. Should generate and build with the x86 compiler.

### Apparent Behavior:
4. The compiler that's being used is the x64 compiler.

### Fix:
Open the `cmake-tools-kits.json` and change
```
  {
    "name": "Visual Studio Enterprise 2019 Release - x86",
    "visualStudio": "11e6a10a",
    "visualStudioArchitecture": "x86",
    "preferredGenerator": {
      "name": "Visual Studio 16 2019"
    }
  }
```
to
```
  {
    "name": "Visual Studio Enterprise 2019 Release - x86",
    "visualStudio": "11e6a10a",
    "visualStudioArchitecture": "Win32",
    "preferredGenerator": {
      "name": "Visual Studio 16 2019",
      "platform": "Win32"
    }
  }
```

### Platform and Versions
- **Operating System**: Windows 10
- **CMake Version**: 3.16
- **VSCode Version**: 1.40.2
- **CMake Tools Extension Version**: 1.2.3
- **Compiler/Toolchain**: Visual C++ 2019 Enterprise

H-G-Hristov:
Are you sure about this line: `"visualStudioArchitecture": "x86",` -> `"visualStudioArchitecture": "Win32"`,. 
I'll see what I can do later about that issue.

Zingam:
@johannesstricker  Could you, please, provide the Debug Console output.

andreeis:
I am not reproducing. Can you attach a standalone repro?

bobbrow:
@andreeis I don't think it will repro with Ninja, but it should with the Visual Studio generator. @johannesstricker, using Ninja as your generator might also be a workaround.

andreeis:
CMake Tools 1.4.0 Beta is available on GitHub.
https://github.com/microsoft/vscode-cmake-tools/releases/tag/1.4.0-beta

Please try it out and let us know if you encounter any issues.

[...]

PR #1162, opened 2020-04-07
Fixes for GitHub issues 964 and 1084

andreeis:
Fix for GitHub issue https://github.com/microsoft/vscode-cmake-tools/issues/1084:
Remove the default values for cmake.prefferedGenerators in package.json and use them later, only if the deduction process is not identifying any. Without this fix, Ninja and Unix Makefiles would always be preferred (when installed), when the user would not override these defaults in the settings and regardless of the preferred generator defined for the kit.

Fix for GitHub issue https://github.com/microsoft/vscode-cmake-tools/issues/964:
Always generate the VS kits with full information about the platforms, no defaults.
VS architecture becomes host architecture and is the arch before "_".
Target architecture is the arch after "_". 
If no "_" is present in the end of the kit name, then the one architecture applies for host and target.

I am investigating some test failures (probably only baseline changes) but I am submitting the PR before, to get results also from the CI.


src/kit.ts
```diff
@@ -526,11 +561,31 @@ async function collectDevBatVars(devbat: string, args: string[], major_version:
 
 /**
  * Platform arguments for VS Generators
+ * For example, x86 and amd64_x86 will generate -A win32
  */
-const VsArchitectures: {[key: string]: string} = {
+const genPlatformFromVsHostTargetArchs: {[key: string]: string} = {
+  x86: 'win32',
```
bobbrow, line 567: If the key is a string, it would be best to add the keys with string syntax:

```ts
  'x86': 'win32',
```

src/kit.ts
```diff
@@ -642,8 +703,11 @@ export async function scanForVSKits(pr?: ProgressReporter): Promise<Kit[]> {
   const installs = await vsInstallations();
   const prs = installs.map(async(inst): Promise<Kit[]> => {
     const ret = [] as Kit[];
-    const arches = ['x86', 'amd64', 'x86_amd64', 'x86_arm', 'amd64_arm', 'amd64_x86'];
-    const sub_prs = arches.map(arch => tryCreateNewVCEnvironment(inst, arch, pr));
+    // TODO: exclude toolsets that are not yet installed.
```
bobbrow, line 706: Doesn't `tryCreateNewVCEnvironment` handle your TODO?  I thought it would only succeed if the tools were there and skip it otherwise.

.vscode/launch.json
```diff
@@ -213,18 +213,18 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": [
-                "${workspaceFolder}/test/extension-tests/vs-preferred-gen/project-folder",
+                "${workspaceFolder}/test/extension-tests/successful-build/project-folder",
```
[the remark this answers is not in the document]
andreeis: Right. The original path was not working, there's no vs-preferred-gen in the CMake Tools test repo. But indeed the above change is redundant. So I'd rather delete the whole block instead than revert because it wasn't doing anything before.

src/kit.ts
andreeis: It appears that vcvarsall.bat, when invoked for arm/arm64 on an install that does not include these architectures, it still creates a valid environment for x86. To me this is confusing, but I can follow up separately with the VS team, in case there is a good reason for this. The environment difference between the two cases is that for arm(64) installed, the first path is [hostArch]\[targetArch] instead of [hostArch]\[hostArch]. I am fixing it now by checking on this detail.

src/kit.ts
andreeis: I tried but I get "Unnecessarily quoted property 'x86' found."

src/kit.ts
```diff
@@ -434,6 +434,41 @@ function kitName(inst: VSInstallation, arch: string): string {
...
+  return hostArch;
+}
+
+/**
+ * Extract the toolset host architecture from the kit name suffix.
+ * The rule is: [hostArch]_[targetArch] or both [arch] if no "_" is present.
+ * @param hostTargetArch
+ */
+function kitTargetArch(hostTargetArch: string): string {
+  const targetArch: string = hostTargetArch.split("_")[1] || hostTargetArch;
```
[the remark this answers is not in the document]
andreeis: Yes, doing this for the next commit/push in this PR.

src/kit.ts
```diff
@@ -461,12 +455,7 @@ function kitTargetArch(hostTargetArch: string): string {
  * @param targetArch The architecture of the target
  */
 function kitHostTargetArch(hostArch: string, targetArch: string): string {
-  let hostTargetArch: string = hostArch;
-  if (targetArch && vsHostTargetArchsFromGenPlatform[targetArch] !== hostArch) {
-    hostTargetArch = hostTargetArch.concat("_").concat(vsHostTargetArchsFromGenPlatform[targetArch]);
-  }
-
-  return hostTargetArch;
+  return hostArch === targetArch ? hostArch : hostArch.concat("_").concat(targetArch);
```
bobbrow, line 458: The second case is probably better as:
```ts
`${hostArch}_${targetArch}`
```
Using backtick quotes will evaluate the variables.

src/kit.ts
```diff
@@ -602,18 +571,38 @@ const VsGenerators: {[key: string]: string} = {
...
   }
-  const variables = await collectDevBatVars(devbat, [`${arch}`], majorVersion, common_dir);
+  const variables = await collectDevBatVars(devbat, [`${kitHostTargetArch(hostArch, targetArch)}`], majorVersion, common_dir);
   if (!variables) {
     return null;
   } else {
+    // When invoked for arm or arm64, the vcvarsall.bat script creates an x86 environment,
+    // when the arm(64) components are not installed.
+    // The way to differentiate between the two cases is to check that the first path
+    // in the PATH environment variable ends with the target architecture.
```
bobbrow, line 593: Are you 100% sure this will always be the case forever?  I would hate to be inadvertently broken because msvc decided to change something without us knowing.

src/kit.ts
```diff
@@ -602,18 +571,38 @@ const VsGenerators: {[key: string]: string} = {
   16: 'Visual Studio 16 2019'
 };
 
-async function varsForVSInstallation(inst: VSInstallation, arch: string): Promise<Map<string, string>|null> {
-  console.log(`varsForVSInstallation path:'${inst.installationPath}' version:${inst.installationVersion} arch:${arch}`);
+async function varsForVSInstallation(inst: VSInstallation, hostArch: string, targetArch: string): Promise<Map<string, string>|null> {
+  console.log(`varsForVSInstallation path:'${inst.installationPath}' version:${inst.installationVersion} host arch:${hostArch} - target arch:${targetArch}`);
+
+  // TODO: if needed, cover also the scenario of VS older than 2017,
```
bobbrow, line 577: I thought the older VS versions are handled on line 583.  It looks like it's searching the path you mention in this comment.  Can this TODO be removed?

andreeis: Oh I didn't realize that. It's true. I will remove the comment and also do another fix related to x64 and amd64, because the older vcvars scripts don't recognize x64.

src/kit.ts
andreeis: No, not 100% sure. But I just noticed that a VS with arm/arm64 installed has some arm/arm64 specific vcvars, which call into the main one. I think that is a better indication.

bobbrow:
Please double-check my merge.  There was a conflict with another feature that was recently submitted.

package.json
```diff
@@ -1272,21 +1269,21 @@
     "mocha": "^4.1.0",
     "module-alias": "^2.2.1",
     "sinon": "~5.0.7",
-    "ts-loader": "^6.0.4",
-    "ts-node": "^6.0.0",
+    "ts-loader": "^6.2.2",
+    "ts-node": "^8.8.1",
```
bobbrow, line 1273: This new version of `ts-node` seems to be causing the build to fail on linux.  Can you please fix it or revert the change?

bobbrow: I saw your update, but the build still fails.  Unfortunately, I don't think this was fixed.

andreeis: Yes I just saw it fails. I am trying reverting next.

Issue #587, opened 2018-11-22
Notification errors after re-configuration.


sialiuk:
- Extension Name: cmake-tools
- Extension Version: 1.1.2
- OS Version: Windows_NT x64 10.0.17134
- VSCode version: 1.29.1

Every time when I re-configure my project, I get the next error notifications from VS Code:

"Element with id ::.::Target1 is already registered"
"Element with id ::.::Target2 is already registered"

and other, when Target1 and Target2 are names of targets from Cmake.

vector-of-bool:
This looks to be an issue with the project outline tree view. Do you have a sample project I can look at?

lflserg:
Have same issue

sialiuk:
> This looks to be an issue with the project outline tree view.
Yes, after I had disabled the project outline tree view, error notifications haven't shown. 

> Do you have a sample project I can look at?
Sorry, but I can't provide it.

vector-of-bool:
@lflserg Do you have a minimal project that can reproduce the issue? I know the line and error that is being hit, but I can't mentally model how it executes. I'll need to debug into it.

lflserg:
@vector-of-bool project I am trying it on is huge closed source so i cant share it, sorry

rtbo:
Have the same issue with a project (declaring `project()` in CMakeLists.txt) used as submodule in another project with `add_subdirectory(... EXCLUDE_FROM_ALL)`

rtbo:
bug can be triggered with this repo:
https://github.com/rtbo/vscode-cmake-tools587

(don't forget to `git submodule update --init`)

wridhdhi:
Even I face the same issue with a big project, 
I hit _configure_ -> it configures part of the targets , shows them in the _right pane_ and shows this error.
but it shows **configuring done** , and build files have been written also, 

also, configuring it manually on bash by `cmake ..` does configure it properly, so the issue is with parsing the targets by the extension I guess. 
![issue_cmake](https://user-images.githubusercontent.com/40772263/60683257-740fb880-9eb4-11e9-98cd-84f732efa6d6.png)

MBetters:
I'm getting the same issue :/

ph03:
Same error here - I think this is related to similar named subfolders added via cmakes `add_directory(subdirname)` if `subdirname` is repeated for different but same named subdirectories in different places of the project tree. Perhaps these internal elements can be prefixed with their "full" project tree path inside the extension to make them unique again?

[...]

PR #1169, opened 2020-04-11
 Fix for Github issue 587: duplicate project structure causes error

SchweizS:
This PR resolves: #587

## Changes
Added project sourceDirectory to class ProjectNode.
_rootDir is now initialized with sourceDirectory as prefix.
Added projectDirectory to tooltip of ProjectNodes

## Notes
Minimal project to reproduce issue: 
https://github.com/SchweizS/vscode-cmake-tools587

### DIFF
--- a/src/tree.ts
+++ b/src/tree.ts
@@ -135,7 +135,7 @@ function sortStringForType(type: codemodel_api.TargetTypeString): string {
 
 export class DirectoryNode<Node extends BaseNode> extends BaseNode {
   constructor(readonly prefix: string, readonly parent: string, readonly pathPart: string) {
-    super(`${prefix}${path.sep}${path.normalize(pathPart)}`)
+    super(`${prefix}${path.sep}${path.normalize(pathPart)}`);
   }
 
   private _subdirs = new Map<string, DirectoryNode<Node>>();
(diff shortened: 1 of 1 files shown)


SchweizS:
i updated the code in the reproduction repository. project(base2) was supposed to be commented out.



Additional explanation:
1) reveal a directory inside a target which is inside a directory
2) cause an update to the tree view (e.g. select a build/debug target) 
3) items gone

But i found the issue.
It was caused by triggering an event on multiple nodes (last commit).
as far as i can see it doesn't change anything if this event is only triggered once.

src/tree.ts
```diff
@@ -513,9 +513,12 @@ export class ProjectOutlineProvider implements vscode.TreeDataProvider<BaseNode>
     existing.updateCodeModel(model, {...ctx, nodesToUpdate: updates, folder});
 
     this._changeEvent.fire(null);
-    for (const node of updates) {
-      this._changeEvent.fire(node);
-    }
+    // i have no idea why this was added.
```
bobbrow, line 516: Can you just delete this instead of commenting it out?  I looked at the interface and it says that firing the event with `null` causes a full tree update which is all we need to do.

SchweizS: deleted (https://github.com/microsoft/vscode-cmake-tools/pull/1169/commits/8e165cf4815358a421359be12633e3e4beedf58c , https://github.com/microsoft/vscode-cmake-tools/pull/1169/commits/3eaeac6f9e464034afcc199d95af712b0e81f849 )

bobbrow:
Thank you so much!

PR #1176, opened 2020-04-15
Update to new vscode types

bobbrow:


bobbrow:
This is a bigger work item than expected.

PR #1182, opened 2020-04-17
Fix for Github issue 741: Add context menu for CMakeLists

### LINKED ISSUE #741: Add context menu items like "configure" for right clicking on explorer items.
### Current Behavior

It's not intuitive to know you have to type into the command palette.

### Expected:

Intuitively, right clicking on a CMakeLists.txt (or maybe ANY file in the project folder) should manifest a CMakeTools submenu containing actions like Configure in it.

SchweizS:
## Changes
- Added context menu entries for CMakeLists.txt in Explorer and CMake project outline.
- Added context menu entries for source files in CMake project outline.
- Removed context menu for all other file types (they all had "compile file", which resulted usually in compiler errors)
- Compile file only works if filename contains .c (e.g. .c, .cpp, .cxx)
- Added configure all projects to CMakeLists.txt files as icon in CMake project outline.

This should resolve #741

### DIFF
--- a/src/tree.ts
+++ b/src/tree.ts
@@ -225,7 +225,10 @@ export class SourceFileNode extends BaseNode {
     const item = new vscode.TreeItem(path.basename(this.filePath));
     item.id = this.id;
     item.resourceUri = vscode.Uri.file(this.filePath);
-    item.contextValue = 'nodeType=file';
+    const name = this.name.toLowerCase();
+    const cml = name == "cmakelists.txt";
+    const compilable = name.search(/\.c/)!==-1;
+    item.contextValue = ['nodeType=file', `compilable=${compilable}`, `cmakelists=${cml}`].join(',');
     item.command = {
       title: localize('open.file', 'Open file'),
       command: 'vscode.open',
--- a/package.json
+++ b/package.json
@@ -620,8 +620,33 @@
         },
         {
           "command": "cmake.outline.configureAll",
-          "when": "view == cmake.outline && viewItem =~ /nodeType=target/",
-          "group": "3_targetActions@1"
+          "when": "view == cmake.outline && viewItem =~ /nodeType=file/ && viewItem =~ /cmakelists=true/",
+          "group": "1_fileActions@1"
+        },
+        {
+          "command": "cmake.outline.buildAll",
+          "when": "view == cmake.outline && viewItem =~ /nodeType=file/ && viewItem =~ /cmakelists=true/",
+          "group": "1_fileActions@2"
+        },
+        {
+          "command": "cmake.outline.cleanAll",
+          "when": "view == cmake.outline && viewItem =~ /nodeType=file/ && viewItem =~ /cmakelists=true/",
+          "group": "1_fileActions@3"
+        },
+        {
+          "command": "cmake.outline.cleanConfigureAll",
+          "when": "view == cmake.outline && viewItem =~ /nodeType=file/ && viewItem =~ /cmakelists=true/",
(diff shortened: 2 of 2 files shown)


src/tree.ts
```diff
@@ -225,7 +225,10 @@ export class SourceFileNode extends BaseNode {
     const item = new vscode.TreeItem(path.basename(this.filePath));
     item.id = this.id;
     item.resourceUri = vscode.Uri.file(this.filePath);
-    item.contextValue = 'nodeType=file';
+    const name = this.name.toLowerCase();
+    const cml = name == "cmakelists.txt";
+    const compilable = name.search(/\.c/)!==-1;
```
bobbrow, line 230: I think this should be more explicit. `.c` could exist in a part of the filename that is not the extension.  Consider using `path.extname(name)` to chop off the extension and compare it to the exact extensions we want to support.

SchweizS: Which extensions do we want to support?
So far there was no check and the 'compile file' command was shown even for txt files.
The explorer view supports .c and .cpp.

Common c/c++ source file extensions include
- cpp
- c
- cc
- cxx
- c++
according to [wiki](https://en.wikibooks.org/wiki/C%2B%2B_Programming/Programming_Languages/C%2B%2B/Code/File_Organization).

I'd go for the full list and probably also add the extensions to the explorer view.

bobbrow: I believe the `resourceLangId` is used for the explorer view so that we don't have to manage the extensions.

I'm also wondering if maybe you can just use the compilation information in the code model instead of file extensions.  This might also cover cases where custom extensions are used.  It looks like the filegroup that contains the source files with have "language" set to "C" or "CXX".  The filegroup with headers will not.  (I just did a quick check with cmake-file-api, I haven't checked with CMake Server to see if it's consistent)

SchweizS: Looks like CMake Server and file-api do the same thing.

I added a check for the language. i'll probably do another pull request in the future to clean up the tree.ts file.
i noticed some unused variables and codepaths.

src/tree.ts
```diff
@@ -210,7 +210,7 @@ export class DirectoryNode<Node extends BaseNode> extends BaseNode {
 }
 
 export class SourceFileNode extends BaseNode {
-  constructor(readonly prefix: string, readonly sourcePath: string, readonly filePath: string) {
+  constructor(readonly prefix: string, readonly sourcePath: string, readonly filePath: string, private _language?: string) {
```
bobbrow: This will sound nitpicky, and I apologize, but the member variables in these classes don't start with a leading underscore.  Can you remove it?

src/tree.ts
bobbrow: Awesome!  Thanks again for doing this!

src/tree.ts
SchweizS: i don't mind changing it, but shouldn't it have the underscore?

> Prefix private members/methods with an underscore and write them _withCamelCase ([coding style](https://vector-of-bool.github.io/docs/vscode-cmake-tools/development.html))

the private members of the other classes in that file start with the underscore.
 e.g. `private _subdirs = new Map<string, DirectoryNode<Node>>();`

bobbrow: Ah.  The other classes that declared member variables in the constructor signature didn't use underscores, and that's what threw me off.

It's fine though.  We will be changing the style guide when we migrate the docs off of the old website (which we don't have access to update).  We'll need to do a wholesale update when we do this, and we can update this at that time.

Issue #708, opened 2019-07-14
Improve coexistence with cmake-js


mmartell:
### Brief Issue Summary

Great extension, but doesn't play nicely with cmake-js.

### Expected:

We are using cmake-js to configure a node-addon-api. We run _cmake-js configure_ from the command line to generate our project. Then we build inside of VS Code by specifying "Do not use a kit".  It would be nice if your tool had a setting that avoided this popup every time we build, as we never want to run configure inside VS Code.

### Platform and Versions

- **Operating System**: All platforms
- **CMake Version**: 3.12
- **VSCode Version**: 1.36.1
- **CMake Tools Extension Version**: 1.1.3
- **Compiler/Toolchain**: Clang on Mac and Linux, VS 2017 on Windows

andreeis:
@mmartell, what popup are you referring to? The "Select a kit" QuickPick? We have another GitHub issue tracking the bug with that particular QuickPick being triggered too often.

Do you have the setting "cmake.configureOnOpen" set to false? That would help avoid configuring operations. Unfortunately, the build is always triggering a configure and we have another bug for adding a new setting to prevent that.

https://github.com/microsoft/vscode-cmake-tools/issues/1007 

https://github.com/microsoft/vscode-cmake-tools/issues/984

https://github.com/microsoft/vscode-cmake-tools/issues/588

andreeis:
CMake Tools 1.4.0 Beta is available on GitHub.
https://github.com/microsoft/vscode-cmake-tools/releases/tag/1.4.0-beta

Please try it out and let us know if you encounter any issues.

andreeis:
This fix is available in CMake Tools #1.4.0 which has been released.

PR #1192, opened 2020-04-21
Scan silently if there are no available kits

xisui-MSFT:
In the offline discussion, there was two decisions around this experience:
1. If opening a CMake project and there’s an toolchain file entry in settings.json, then don’t do anything related to Kits.
1. If opening a CMake project and there is NO toolchain file entry in settings.json and if there is no kits.json, then silently start scanning for kits.

We finally decided not to implement (1) because a kit is a super set of a toolchain file, and although it's possible to detect multiple settings in settings.json instead of only the toolchain file entry, this is unnecessarily complicated.


bobbrow:
This should also address #708

Issue #1187, opened 2020-04-18
VSCode C++ IntelliSense bug when adding new files


Yuri6037:
### Brief Issue Summary

VSCode C++ tools bugs when adding new files in a project manged by ms-vscode.cmake-tools.

When adding a new hpp/cpp file in the project workspace CMakeLists.txt, the VSCode C++ IntelliSense gets bugged and refuses to process correctly the new hpp/cpp file. It is needed to close VSCode, run cmake .. inside the build folder in a Powershell and then re-open VSCode in order to get it to work.

### Expected:

When saving the CMakeLists.txt of a CMake sub-directory, VSCode C++ tools should automatically be refreshed with the new files. That behaviour is already fine under Visual Studio Community 2019.

### Apparent Behavior:

VSCode C++ tools IntelliSense does not get refreshed with the new files and requires full editor reload with powershell console in order to rebuild the cmake project.

### Platform and Versions

- **Operating System**: Windows 10 x64
- **CMake Version**: cmake version 3.16.0-rc3
- **VSCode Version**: 1.42.1
- **CMake Tools Extension Version**: 1.3.1
- **Compiler/Toolchain**: Visual Studio Community 2019 Release - amd64

### Other Notes/Information

Tested on a projected first initialized with cmake Visual Studio 2019 generator.

andreeis:
Another workaround (besides reloading) is to configure after editing the CMakeLists.txt. Hopefully not that inconvenient before we get to fix this.

Yuri6037:
I can try to do it myself, at least if I can get a VSCode development environment working...

andreeis:
That would be great. Let us know if you have any questions about the code base or if you encounter any issues during developing the fix for this issue. We'll be happy to review a PR when you send one.

Yuri6037:
I've got something working in local where I can add new files, edit my CMake, and right on CMake save it runs CMake configure which then updates intellisense, a pretty quick fix

andreeis:
CMake Tools 1.4.0 Beta is available on GitHub.
https://github.com/microsoft/vscode-cmake-tools/releases/tag/1.4.0-beta

Please try it out and let us know if you encounter any issues.

andreeis:
This fix is available in CMake Tools #1.4.0 which has been released.

PR #1196, opened 2020-04-22
Added automatic CMake configure on CMakeLists save

Yuri6037:

- Add a file save subscription on extension load
- Run configure function when the file is named "CMakeLists.txt"

## Other Notes/Information

Not sure why but ALT+SHIFT+F to auto reformat changed the whole file. I've checked it and it still works fine on my own projects including a 3D Game Engine.
Also I'm new to TypeScript and Visual Studio Code Extension coding...

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -1,41 +1,41 @@
 /**
  * Root of the extension
  */
-import {CMakeCache} from '@cmt/cache';
-import {CMakeExecutable, getCMakeExecutableInformation} from '@cmt/cmake/cmake-executable';
-import {CompilationDatabase} from '@cmt/compdb';
+import { CMakeCache } from '@cmt/cache';
+import { CMakeExecutable, getCMakeExecutableInformation } from '@cmt/cmake/cmake-executable';
+import { CompilationDatabase } from '@cmt/compdb';
 import * as debugger_mod from '@cmt/debugger';
 import diagCollections from '@cmt/diagnostics/collections';
 import * as shlex from '@cmt/shlex';
-import {StateManager} from '@cmt/state';
-import {Strand} from '@cmt/strand';
-import {ProgressHandle, versionToString} from '@cmt/util';
-import {DirectoryContext} from '@cmt/workspace';
+import { StateManager } from '@cmt/state';
+import { Strand } from '@cmt/strand';
+import { ProgressHandle, versionToString } from '@cmt/util';
+import { DirectoryContext } from '@cmt/workspace';
 import * as path from 'path';
 import * as vscode from 'vscode';
 
 import * as api from './api';
-import {ExecutionOptions, ExecutionResult} from './api';
+import { ExecutionOptions, ExecutionResult } from './api';
 import * as codemodel_api from '@cmt/drivers/codemodel-driver-interface';
-import {BadHomeDirectoryError} from '@cmt/drivers/cms-client';
-import {CMakeServerClientDriver, NoGeneratorError} from '@cmt/drivers/cms-driver';
-import {CTestDriver} from './ctest';
-import {BasicTestResults} from './ctest';
-import {CMakeBuildConsumer} from './diagnostics/build';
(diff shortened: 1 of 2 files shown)


andreeis:
Thank you for investigating a fix for issue https://github.com/microsoft/vscode-cmake-tools/issues/1187. We will have a look. Can you make sure you include in the diffs only the fix and revert the formatting changes?

Yuri6037:
> Thank you for investigating a fix for issue #1187. We will have a look. Can you make sure you include in the diffs only the fix and revert the formatting changes?

I just reverted automatic format.

andreeis:
Your PR failed. Until we fix the way the PR validation is reported, you can find yours at this link: https://travis-ci.org/github/microsoft/vscode-cmake-tools/pull_requests and explore.

For example, the linter suggests these improvements:
ERROR: /home/travis/build/microsoft/vscode-cmake-tools/src/cmake-tools.ts:402:86 - Parentheses are prohibited around the parameter in this single parameter arrow function
ERROR: /home/travis/build/microsoft/vscode-cmake-tools/src/cmake-tools.ts:403:11 - Identifier 'str' is never reassigned; use 'const' instead of 'let'.

When you open the CMake Tools code base folder in VSCode, don't you see anything in the problems tab? You can ensure there are no new reports coming from your code.

Also, the new message you add needs to be localized and it is enough if you do this:
log.debug((localize('cmakelists.save.trigger.reconfigure', "We are saving a CMakeLists.txt file, attempting automatic reconfigure..."));
Where 'cmakelists.save.trigger.reconfigure' is just an arbitrary ID we write now, but soon the localization system will use it to link in all the translations done by the localization team.

Yuri6037:
> Your PR failed. Until we fix the way the PR validation is reported, you can find yours at this link: https://travis-ci.org/github/microsoft/vscode-cmake-tools/pull_requests and explore.
> 
> For example, the linter suggests these improvements:
> ERROR: /home/travis/build/microsoft/vscode-cmake-tools/src/cmake-tools.ts:402:86 - Parentheses are prohibited around the parameter in this single parameter arrow function
> ERROR: /home/travis/build/microsoft/vscode-cmake-tools/src/cmake-tools.ts:403:11 - Identifier 'str' is never reassigned; use 'const' instead of 'let'.
> 
> When you open the CMake Tools code base folder in VSCode, don't you see anything in the problems tab? You can ensure there are no new reports coming from your code.
> 
> Also, the new message you add needs to be localized and it is enough if you do this:
> log.debug((localize('cmakelists.save.trigger.reconfigure', "We are saving a CMakeLists.txt file, attempting automatic reconfigure..."));
> Where 'cmakelists.save.trigger.reconfigure' is just an arbitrary ID we write now, but soon the localization system will use it to link in all the translations done by the localization team.

That's strange I can run the code perfectly fine under VS Code. I don't know how you got those errors... I will try to fix from what you told me.

Yuri6037:
Also thank you for the Travis Link I was searching for it but couldn't find it. I will base myself on the Travis for checking build...

Yuri6037:
What should I do about the fakecontext ? It seam that Unit tests are not designed to support accessing subscriptions on the extension context.

andreeis:
> What should I do about the fakecontext ? It seam that Unit tests are not designed to support accessing subscriptions on the extension context.
> 
> In addition it seems that a lot of tests fails in parts of the code that I did not even change. Can you have a look at those other failing tests? I will try to allow the test tool to accept subscriptions.
> 
> EDIT2: It seems that fixing the first test fixed subsequent tests.
> 
> EDIT3: All fixed unit tests are now running.

I don't know very well, investigating. I wonder if the failing tests are a sign that maybe another place would be better for the subscription, which would not cause any test failure. Otherwise, the approach doesn't look too bad to me.

andreeis:
Looks good. I am going to approve. Ideally, this change could have been accompanied by a new test but it's ok now. We need to fix something else very similar which will require some new tests and we will cover this scenario among them.
Thank you very much for your contribution.

Issue #1183, opened 2020-04-17
No indication of active workspace in CMake project outline.


SchweizS:
### Summary
Currently there is no indication which workspace is active in multi root workspaces.

### Expected
![active_workspace](https://user-images.githubusercontent.com/18433748/79553646-12b08300-809d-11ea-8fbb-0ae9d5b55e4b.JPG)

### Suggestion
Add an active indication and a context menu/button to change workspace in the tree view.

### Required changes:
see: [Schweizs/vscode-cmake-tools/fix-issue-1183](https://github.com/SchweizS/vscode-cmake-tools/tree/fix-issue-1183)

i'm not quite sure how to localize this, so no pull request yet.
i guess new loca keys for "Workspace Folder" and "Workspace Folder (active)".
"Workspace Folder" is currently hardcoded.

### Platform and Versions
- **VSCode Version**:  1.44.0
- **CMake Tools Extension Version**: 1.3.1 / 2a56936a86fdd10ae1e197470b3dc1ccb0660095

andreeis:
@SchweizS, thank you for looking into this. Please start the PR and we will help with the changes needed for localization.

andreeis:
CMake Tools 1.4.0 Beta is available on GitHub.
https://github.com/microsoft/vscode-cmake-tools/releases/tag/1.4.0-beta

Please try it out and let us know if you encounter any issues.

andreeis:
This fix is available in CMake Tools #1.4.0 which has been released.

PR #1199, opened 2020-04-23
Fix issue 1183: add indication of active workspace to project outline

SchweizS:
Changes:
- Added indication of active workspace
- added icon to select active workspace in project outline

Todo:
- localization

#1183


andreeis:
The icon might not be the most relevant for the "change active workspace" operation. We would prefer 'gear' first, then if not possible the 'utility' icon. If you already tried other icons but got blocked in any way so you had to pick as a last resort the current (build like) icon then let us know and we will try to figure out. If you didn't try yet, please push another commit using one of the two recommended icons.

SchweizS:
i switched it to gear.
i just copy&pasted that part and forgot to change the icon to a better one.

src/extension.ts
```diff
@@ -812,7 +813,10 @@ class ExtensionManager implements vscode.Disposable {
     }
     vscode.window.showErrorMessage(localize('compilation information.not.found', 'Unable to find compilation information for this file'));
   }
-
+  async selectWorkspace(file?: vscode.WorkspaceFolder) {
```
bobbrow, line 816: Sorry, minor nitpick.  Can you call this `folder` instead of `file`?

SchweizS:
localize('workspace.active', 'Active Workspace');
localize('workspace', 'Workspace');

those to keys should probably be added to the localization  (tree.ts lines 461/463)

Issue #1189, opened 2020-04-19
Run in terminal in the bottom bar


xahon:
It would be very convenient to have 'Run' button bound to 'Run in terminal' action near 'Debug' one
![image](https://user-images.githubusercontent.com/18731681/79686693-9a58e600-8285-11ea-820c-d6bd30ba373c.png)

Issue #952, opened 2019-12-03
Status bar text should be theme-able


bobbrow:
The colours also seem to be at odds with the colour scheme.  Is this intentional?

Bright yellow in a low-contrast colour scheme (Solarized) is pretty unpleasant, and it seems impossible to remove.

<img width="713" alt="Screenshot 2019-11-27 at 09 30 35" src="https://user-images.githubusercontent.com/152388/69944148-deba3400-14e6-11ea-9331-b36cf9e08a1c.png">

_Originally posted by @willvousden in https://github.com/microsoft/vscode-cmake-tools/issues/911#issuecomment-560289866_

Issue #212, opened 2017-07-17
Configure buttons in the status bar


mental13:
I use other VSCode extensions that add buttons to the status bar and cmake tools makes it really cluttered.
I would appreciate options to edit this and hide buttons that are redundant for a specific project
- Run CTest
- Make build and debug target the same (one selection for both)
- Use icons for build and debug buttons
- Minimize the name of the current project (instead of `Cmake: ProjName: Debug: Ready` a simple `ProjName: Debug` would suffice in most cases)

leira:
Or we can configure to disable these status bar buttons all together.

vector-of-bool:
The buttons will have some clean up in 1.1, but I expect more clean up in 1.2, so I'm going to leave this open and move the milestone to 1.2

bobbrow:
VS Code recently added the capability to hide status bar icons from an extension. It's an all-or-nothing toggle, but it should address @leira's comment.

tanveerbadar:
Can I upvote this 1000x? Kit selection and build mode (debug/release) are the biggest offenders here, eating into already limited available space.

michaelkollmann:
It'd be very interesting for me as well to customize which buttons are being displayed in the status bar.

I'm using CMake Tools for embedded development and I have to use a different debugger (marus25.cortex-debug) then the native VSCode one. This makes the "Debug" button in the status bare useless for me as I can't use quick debug. So now there is this button down there in the menu bar and I have to tell all my colleagues that they must not click this button. Quiet annoying...

bobbrow:
@michaelkollmann, if you are using launch.json to configure debugging because the quick debug button doesn't work, you can actually move the launch.json config into the `cmake.debugConfig` setting.  This allows you to override any of the defaults (e.g. `"miDebuggerPath"`) that you may be needing to change.

bobbrow:
This issue was addressed in version 1.4 but was never closed.  New settings `cmake.statusbar.visibility` and `cmake.statusbar.advanced` were added to give you full control over the icons.

PR #1200, opened 2020-04-23
Statusbar update 212 952 1189

SchweizS:
### This adds:
- all buttons can be hidden (#212)
- added launch button (#1189)
- added option to disable coloring of CTest button (#952) 
- added option to truncate name of selected kit (#212)
- added option to only show build icon (#212)
- added option to reduce size of extension status (CMake: Debug: Ready => Debug: Ready) (#212)
- added tooltip to the build button

### Changes:
- CTest button is not colored by default.
- LaunchTarget is now in brackets `[target]` similar to build target
- Tooltip of debug, launch and build buttons now shows the target

### This does not address:
- build and debug target are still separate (#212)
- no custom coloring (#952)

## Notes:
- ~~as far as i can tell hideDebugButton is never called. maybe the entries related to that can be completely removed.~~
- there seems to be a bug which causes the workspace selection not to show up when a workspace is created by adding a second directory.
- visibility of buttons can be one of
  - default: current behaviour. Label is completely visible
  - compact: buttons only show minimal information or get truncated
  - icon: only icons are visible. buttons without icons are hidden
  - hidden: button is hidden


src/status.ts
```diff
@@ -2,258 +2,385 @@ import * as vscode from 'vscode';
...
+  setCTestEnabled = (v: boolean) => this._testButton.enabled = v;
+  setTestResults = (v: BasicTestResults|null) => this._testButton.results = v;
+  setIsBusy = (v:boolean) => this._buildButton.isBusy = v;
+  setActiveKitName = (v:string) => this._kitSelectionButton.text = v;
+
   hideDebugButton(shouldHide: boolean = true) {
-    this._hideDebugButton = shouldHide;
-    this.reloadVisibility();
+    if (false) return shouldHide;
+    // TODO: Find out if this is called...
```
bobbrow, line 384: This is a feature for other extensions that allows them to configure our behavior a bit. You can remove the comment.

package.json
```diff
@@ -1205,6 +1205,96 @@
...
+          "type": "string",
+          "default": "text",
+          "enum": [
+            "text",
+            "icon",
+            "hidden"
+          ],
+          "scope": "window"
+        },
+        "cmake.statusbar.launch_target.type": {
```
bobbrow, line 1261: Should be `launchTarget`

package.json
```diff
@@ -1205,6 +1205,96 @@
...
+        "cmake.statusbar.workspace.type": {
+          "type": "string",
+          "default": "text",
+          "enum": [
+            "text",
+            "hidden"
+          ],
+          "scope": "window"
+        },
+        "cmake.statusbar.build_target.type": {
```
bobbrow, line 1242: Should be `buildTarget`

package.json
```diff
@@ -1205,6 +1205,96 @@
...
+          "enum": [
+            "text",
+            "short",
+            "hidden"
+          ],
+          "scope": "window"
+        },
+        "cmake.statusbar.kit.length": {
+          "type": "integer",
+          "default": 10,
```
bobbrow, line 1220: 10 is really short and probably won't display most Kits in a manner that is helpful. I'm not convinced that this is a feature anyone will use. I would be happy with "text", "icon", and "hidden" though I don't see an "icon" option for this one. Since this aligns with the toolset that you'll be using to compile your code, I would vote for the $(tools) or $(briefcase) octicons to represent kits.

src/status.ts
```diff
@@ -2,258 +2,385 @@ import * as vscode from 'vscode';
...
+  isVisible():boolean {
+    return this._isBusy || true;
+  }
+  protected getType():ButtonType {
+    return this._config.statusbar.build.type;
+  }
+}
+
+export class StatusBar implements vscode.Disposable {
+  private readonly _kitSelectionButton = new KitSelection(this._config, 3.6);
```
bobbrow, line 336: It looks like the order of the status items was changed. Can you restore the original order?

SchweizS: changed.

package.json
SchweizS: increased to 20 and added option for icon (tools).

src/status.ts
```diff
@@ -2,258 +2,385 @@ import * as vscode from 'vscode';
...
     } else {
       testPassingText = localize('tests.passing', '{0}/{1} tests passing', passing, total);
     }
-    this._testButton.text = `$(${icon}) ${testPassingText}`;
-    this._testButton.color = good ? 'lightgreen' : 'yellow';
+    const icon = good ? 'check' : 'x';
+    return `$(${icon}) ${testPassingText}`;
   }
+}
+class BuildButton extends CheckCPPToolsButton {
```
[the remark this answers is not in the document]
SchweizS: just rechecked -> no.
removed dependency.

src/status.ts
```diff
@@ -2,258 +2,385 @@ import * as vscode from 'vscode';
...
-    this._testResults = v;
+class DebugButton extends CheckCPPToolsButton {
+  command = 'cmake.debugTarget';
+  text = '$(bug)';
+  tooltip = localize('launch.debugger.tooltip', 'Launch the debugger for the selected target');
+  protected getType():ButtonType {
+    return this._config.statusbar.debug.type;
+  }
+}
+class LaunchButton extends CheckCPPToolsButton {
```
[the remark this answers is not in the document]
SchweizS: just rechecked -> no.
removed dependency.

src/status.ts
SchweizS: ok.
i added this to the button
same with hideLaunchButton.

SchweizS:
> This is pretty cool and I'm glad that there is interest out there to address this. We've been debating here what we should do about the status bar because it does take up a lot of space. We've actually been thinking about moving this out of the status bar completely and into its own view (i.e. perhaps sharing space with the "outline view" panel as a separate section). We could then just leave a single icon on the status bar that could act as a shortcut to the panel. We haven't fleshed out the details, but that seemed like an option we wanted to explore.

I like that idea, but i think the statusbar should still remain an option.
The statusbar allows to see some settings at first glance, without having to change to a different view.
This is in my opinion a big win in comparison to other options.

If you add the new view, you could just set the default for the statusbar to hidden and add the new button.

> That said, I think this PR gives people a lot of control over how the status bar looks and is a good interim solution in the event we want to do the aforementioned work. However, my overall feedback for this PR is that this might give a little too much control and result in some settings bloat. It was somewhat cumbersome to change the settings for each button. I would expect that most people would be happy with a simple setting to control all of the buttons. "text"/"icon"/"hidden". "short" is mostly a problem for Kits with long names and I think it's not essential to try to solve that. Icon + Tooltip would be sufficient, I think. If we consolidate the settings, then in the event that we do move the controls out of the status bar, that's fewer settings for us to deprecate.

I had the same issue, but i think i have a solution for that.
settings
json
With that you have both: An easy way to change all buttons and (kind of) hidden settings for each individual button. It might be a good idea to add some descriptions if this gets approved.

package.json
```diff
@@ -1205,6 +1205,143 @@
           ],
           "scope": "window"
         },
+        "cmake.statusbar.type": {
```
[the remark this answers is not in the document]
bobbrow: I would also prefer we name these `visibility` instead of `type` as I think that conveys the meaning better.

I need to do more testing (sorry, I haven't finished yet - it's been a busy week) but I was also thinking about the states because there is some inconsistency with how they appear when you use this top level setting vs. what I expected.  I will flesh this out more when I get a chance, but I did want to submit this review feedback in the meantime.

src/extension.ts
```diff
@@ -872,6 +872,7 @@ class ExtensionManager implements vscode.Disposable {
 
   async hideLaunchCommand(shouldHide: boolean = true) {
     // Don't hide command selectLaunchTarget here since the target can still be useful, one example is ${command:cmake.launchTargetPath} in launch.json
+    this._statusBar.hideLaunchButton(shouldHide);
```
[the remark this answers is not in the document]
SchweizS: >  I don't remember why this wasn't added previously.

there was previously no status bar button for that.
i added that line because it looks like the same as with debug.

src/status.ts
```diff
@@ -2,258 +2,398 @@ import * as vscode from 'vscode';
...
-}
+abstract class Button {
+  constructor(protected readonly _config: ConfigurationReader, protected readonly priority: number) {
+    this._button.command = this._button.command;
+  }
+  protected readonly _button = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, this.priority);
+  private _forceHidden:boolean = false;
+  private _text: string = '';
+  private _tooltip: string|null = null;
+  readonly settings : string|null = null;
```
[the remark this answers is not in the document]
SchweizS: renamed to settingsName

src/status.ts
```diff
@@ -2,258 +2,398 @@ import * as vscode from 'vscode';
...
+  protected isVisible():boolean {
+    return true;
+  }
+  protected getType():ButtonType | null {
+    if (this.settings) {
+      return Object(this._config.statusbar.advanced)[this.settings]?.type || null;
+    }
+    return null;
+  }
+  protected getDefaultType():ButtonType {
```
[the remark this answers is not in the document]
SchweizS: getType and getDefaultType merged and renamed to _getType (private).

src/status.ts
```diff
@@ -2,258 +2,398 @@ import * as vscode from 'vscode';
...
+  getTooltipIcon():string {
+    return this._getCurrentText();
+  }
+  isVisible():boolean {
+    return this._isBusy || true;
+  }
+}
+
+export class StatusBar implements vscode.Disposable {
+  private readonly _kitSelectionButton = new KitSelection(this._config, 3.45); //3.6
```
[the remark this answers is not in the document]
SchweizS: fixed

src/status.ts
```diff
@@ -2,258 +2,398 @@ import * as vscode from 'vscode';
...
+    if (text.length === 0) {
+      return localize('no.kit.selected', 'No Kit Selected');
+    }
+    return text;
+  }
+  protected getTextIcon(): string {
+    return `$(tools)`;
+  }
+  protected getTextShort() {
+    const len = this._config.statusbar.advanced?.kit?.length || 20;
```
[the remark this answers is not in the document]
SchweizS: this should be fixed now.

src/status.ts
```diff
@@ -2,258 +2,398 @@ import * as vscode from 'vscode';
...
+class CheckCPPToolsButton extends Button {
+  protected isVisible() {
+    return vscode.extensions.getExtension('ms-vscode.cpptools') !== undefined;
+  }
+}
+class DebugButton extends CheckCPPToolsButton {
+  settings = 'debug';
+  command = 'cmake.debugTarget';
+  text = '$(bug)';
+  tooltip = localize('launch.debugger.tooltip', 'Launch the debugger for the selected target');
```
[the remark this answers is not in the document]
SchweizS: yes should be possible.
would you prefer:

> Launch` the debugger for the selected target
> *target*

or just
> *target*

package.json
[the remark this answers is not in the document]
bobbrow: Forgot to mention... I'm also considering alternate names for the enum values that work better with 'visibility' as the property name.  I'm open to feedback, but these are the names I'm thinking:
```
'default' or 'visible' or 'iconAndText'
'compact'
'icon' or 'iconOnly'
'hidden'
```

src/status.ts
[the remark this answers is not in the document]
bobbrow: I think either:

> Launch the debugger for the selected target: [target]

or

> Launch [target] in the debugger

would be fine.

(I think that the square brackets should be in the message around the target name - I'm not just using those to highlight the variable 😉)

src/extension.ts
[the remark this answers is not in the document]
xisui-MSFT: @bobbrow Yes this is a newly added button. From our perspective this looks good to me.

src/status.ts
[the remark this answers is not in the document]
bobbrow: I was thinking more along the lines of: if it's not a number, then use the default value `20`

src/status.ts
```diff
@@ -2,258 +2,396 @@ import * as vscode from 'vscode';
...
+    return `${this.getTooltipNormal()}\n${this.getTextNormal()}`;
+  }
+}
 
-  private _ctestEnabled: boolean = false;
-  public get ctestEnabled(): boolean { return this._ctestEnabled; }
-  public set ctestEnabled(v: boolean) {
-    this._ctestEnabled = v;
-    setVisible(this._testButton, v);
+class BuildTargetSelectionButton extends Button {
```
bobbrow, line 205: I think that the two target selection buttons should hide when "icon" mode is set. I consider them more as labels than buttons that do something (even though they do have a shortcut to the command that lets you change the target)

src/status.ts
```diff
@@ -2,258 +2,396 @@ import * as vscode from 'vscode';
...
 
-  setLaunchTargetName(v: string) {
-    this._launchTargetNameButton.text = v;
-    this._reloadDebugButton();
+class KitSelection extends Button {
+  settingsName = 'kit';
+  command = 'cmake.selectKit';
+  tooltip = localize('click.to.change.kit.tooltip', 'Click to change the active kit');
+
+  protected getTextNormal():string {
```
bobbrow, line 176: We should show the icon with the text in "normal" and "short" modes.

src/status.ts
```diff
@@ -2,258 +2,396 @@ import * as vscode from 'vscode';
...
-    this._targetName = v;
-    this._targetButton.text = `[${v}]`;
-    this.reloadVisibility();
+  protected getTextNormal() {
+    return `CMake: ${this._buildTypeLabel}: ${this._statusMessage}`;
+  }
+  protected getTextShort() {
+    return `${this._buildTypeLabel}: ${this._statusMessage}`;
   }
+}
```
bobbrow, line 169: I think the text for the cmake status label should disappear in "icon" mode

package.json
```diff
@@ -1205,6 +1205,143 @@
...
+              }
+            },
+            "status": {
+              "type": "object",
+              "properties": {
+                "type": {
+                  "type": "string",
+                  "default": "text",
+                  "enum": [
+                    "text",
```
bobbrow, line 1247: I'm wondering if the enum values should be the same for all of these settings (even if they don't have a visible effect). By having a single setting that configures all buttons, we are forced to interpret the values for each of the buttons already.

src/status.ts
SchweizS: now it is:
text -> CMake: type: status
short -> type: status
icon -> hidden

i'm thinking about adding an icon, but i'm not sure which one to use.

I also noticed, that the status message can be quite big. maybe it would be a good idea to move the status message into the tooltip.

src/status.ts
SchweizS: added with 1f93be77a94eddaf35e8f84bc6c2ce5a5a4ea8e8

src/status.ts
SchweizS: added with 798e21c6191ade40c56702bba03f0f1a512f7bdd

package.json
SchweizS: i added the missing values.

package.json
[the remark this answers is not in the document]
SchweizS: changed it to `visiblity`.
enum values are now
`default`, `compact`, `icon` and `hidden`

src/status.ts
```diff
@@ -2,258 +2,398 @@ import * as vscode from 'vscode';
...
+  }
+  isVisible() {
+    return super.isVisible() && !this._hidden;
+  }
+}
+class LaunchButton extends Button {
+  settings = 'launch';
+  command = 'cmake.launchTarget';
+  text = '$(play)';
+  tooltip = localize('launch.tooltip', 'Launch');
```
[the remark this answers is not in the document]
SchweizS: changed.

src/status.ts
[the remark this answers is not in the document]
SchweizS: added for debug, launch and build button
=>
```
Launch the debugger for the selected target:
[target]
```

SchweizS:
with the current changes i think we could also remove the label of the Build-button.
@bobbrow

i also updated the first post to reflect the changes

src/status.ts
bobbrow: `$(info)` might be an appropriate icon maybe?  I'm not sold on keeping the status the way it is right now, but that's not something you need to solve in this PR.  In my development, I only use it to select variants, not to look at status.

SchweizS: i added the icon and removed the status info from the label in compact mode.

SchweizS:
statusbar

the last commits contain minor improvements to compact mode and an overall cleanup.

macdems:
This is a great improvement to a serious cmake-tools drawback. The ability to turn-off selected buttons is something that was strongly missing. For example, I need to be able to select a variant, however I don't want to touch the kits (and accidentally clicking the 'select kit' statusbar button can result in half-an-hour recompilation of the whole project, so being able to hide it is important to me). Please accept this PR for the next release.

bobbrow:
We're almost done iterating on this.  It will be in the 1.4 release.  It's in the "fit and finish" phase right now.  I am going through all of the tooltips and settings combinations (there are a lot!) looking for consistency issues.  I have a few comments pending right now and plan to submit my feedback within the next day or so.

src/status.ts
```diff
@@ -1,259 +1,454 @@
...
   }
+}
+
+class BuildButton extends Button {
+  private static readonly _build = localize('build', 'Build');
+  private static readonly _stop = localize('stop', 'Stop');
+
+  settingsName = 'build';
+  command = 'cmake.build';
+  tooltip = localize('build.tooltip', 'Build the selected target');
```
bobbrow, line 361: While testing this, I thought the tooltip looked strange with all of that whitespace on the second line.  I would prefer that the [target] be on the same line as the description.

I'll add a comment to each of the classes where I think some changes should be made to the tooltips.

src/status.ts
```diff
@@ -2,258 +2,441 @@ import * as vscode from 'vscode';
...
+//---------------------------------------------
+//---------------- Button Class ---------------
+//---------------------------------------------
+class ActiveFolderButton extends Button {
+  private static readonly _autoSelectToolTip = localize('active.folder.auto.select.tooltip', 'Active folder');
+  private static readonly _toolTip = localize('active.folder.tooltip', 'Select Active folder');
+
+  settingsName = 'workspace';
+  command = "cmake.selectActiveFolder";
+  icon = 'folder-active';
```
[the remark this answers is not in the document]
bobbrow: This looks like it's fixed already.

src/status.ts
```diff
@@ -2,258 +2,441 @@ import * as vscode from 'vscode';
...
-   * of the button and the command that is executed when the button is pressed
-   */
-  private _isBusy: boolean = false;
-  setIsBusy(v: boolean) {
+  private _isBusy:boolean = false;
+  set isBusy(v: boolean) {
     this._isBusy = v;
-    this._reloadBuildButton();
+    this._button.command = v ? 'cmake.stop' : 'cmake.build';
+    this.text = this._isBusy? BuildButton._stop:BuildButton._build;
```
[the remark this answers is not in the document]
bobbrow: This looks like it's fixed already

src/status.ts
```diff
@@ -2,258 +2,441 @@ import * as vscode from 'vscode';
...
+    return `[${this.text}]`;
+  }
+}
+class LaunchTargetSelectionButton extends Button {
+  settingsName = 'launchTarget';
+  command = 'cmake.selectLaunchTarget';
+  tooltip = localize('select.target.tooltip', 'Select the target to launch');
+
+  protected getTextNormal():string {
+    return `[${this.text}]`;
```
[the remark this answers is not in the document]
SchweizS: this should show `[-]` if no target is selected.

i can hide this button if no target is selected, but i personally think it's good to have the button. If you just want to select the target without debugging or launching, you'd have to use the commands or the outline.
I thought about adding a text if no target is selected but that takes way too much space.

i think it's a good compromise to have this with as little space usage as possible there.

src/status.ts
```diff
@@ -2,258 +2,441 @@ import * as vscode from 'vscode';
 import {BasicTestResults} from '@cmt/ctest';
 import * as nls from 'vscode-nls';
 import {SpecialKits} from '@cmt/kit';
+import {StatusBarButtonVisibility as ButtonVisibility, ConfigurationReader } from '@cmt/config';
+
+// FIXME: Show workspace selection if a folder is added to workspace
```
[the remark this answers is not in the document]
SchweizS: it's an existing problem.

#1219

src/status.ts
SchweizS: removed the newline in Debug, Launch and Build.

src/status.ts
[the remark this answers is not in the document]
bobbrow: Please remove the comment. We'll track the issue you created instead.

src/status.ts
```diff
@@ -1,259 +1,454 @@
...
-    } else {
-      this._debugButton.text = `$(bug) ${localize('debug', 'Debug')}`;
-      if (this._visible) {
-        this._launchTargetNameButton.show();
-      }
+  protected getTextNormal(): string { return this._text; }
+  protected getTextShort(): string { return this.getTextNormal(); }
+  protected getTextIcon(): string { return ''; }
+
+  protected getTooltipNormal(): string|null { return this._tooltip; }
```
bobbrow, line 75: Tooltip issues:

**`default` mode**
Button|comment
---|---
**Status**| Remove the status from the tooltip. It should just say "Click to select the current build variant"

**`compact` mode**
Button|comment
---|---
**Workspace**| I don't think we need to show the folder name in the tooltip because we don't truncate the text.
**Status**| I think the status should come before the command text. Can you switch the order?
**Kit**| I think the active kit should come before the command text. Can you switch the order?
**Build**| The text should start with "CMake:"
**Build label**| I don't think we need to show the selected build in the tooltip because we don't truncate the text.
**Launch label**| The text should start with "CMake:"
**CTest**| I don't think we need to show the button text in the tooltip. (make it a one-liner)

**`icon` mode**
Button|comment
---|---
**Workspace**| I'm thinking this should look like: "CMake: Active Folder [${folder name}]\nClick to change"
**Status**| I'm thinking this should look like: "CMake: ${status text}\nClick to select the current build variant" (In the future, maybe status and variant will be split out so we can put the variant in square brackets, but I think that's out of scope for this PR)
**Kit**| I'm thinking this should look like: "CMake: Active Kit [${active kit}]\nClick to change the active kit"
**CTest**| I don't think we need to show the button text in the tooltip. (make it a one-liner)

src/status.ts
```diff
@@ -1,259 +1,454 @@
...
-    if (!this._launchTargetNameButton.text) {
-      this._debugButton.text = '$(bug)';
-      this._launchTargetNameButton.hide();
-    } else {
-      this._debugButton.text = `$(bug) ${localize('debug', 'Debug')}`;
-      if (this._visible) {
-        this._launchTargetNameButton.show();
-      }
+  protected getTextNormal(): string { return this._text; }
+  protected getTextShort(): string { return this.getTextNormal(); }
```
bobbrow, line 72: The CTest button has a '-' in the label in `compact` mode.  That should be removed.

src/status.ts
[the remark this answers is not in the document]
bobbrow: The first time you debug or launch a target it will ask you to select one, so I don't think we need to add this to the status bar when it's unset.  Perhaps you can change the tooltip based on whether a target is selected or not to explain that you will select and debug a target or select and launch a target, but I would rather not add `[-]` to the status bar as it provides no real value.

src/status.ts
SchweizS: changed

src/status.ts
SchweizS: ctest in icon mode should have the annotation of how many test did  pass.
=> second line is only visible if there are test results

i changed the labels (and tooltips) such that selected values (kit, folder, variant, targets) are in brackets and text is without bracket.

i didn't add Active XYZ to icon mode.
I don't think that it adds information because the second line of the tooltip already states what is meant.

I changed the tooltip of the workspace button to:
`Click to select the active workspace` and `Click to select the active workspace (auto)` to match the other tooltips.
i'm not sure if auto is enough or if we should use autoselect.
auto is visible if Cmake: Auto Select Active Folder is selected.

SchweizS:
current state

src/status.ts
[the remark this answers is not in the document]
SchweizS: i removed the text.

src/status.ts
```diff
@@ -138,12 +149,14 @@ abstract class Button {
...
-  private static readonly _autoSelectToolTip = localize('active.folder.auto.select.tooltip', 'Active folder');
-  private static readonly _toolTip = localize('active.folder.tooltip', 'Select Active folder');
+  // private static readonly _autoSelectToolTip = localize('active.folder.auto.select.tooltip', 'Active folder');
+  // private static readonly _toolTip = localize('active.folder.tooltip', 'Select Active folder');
+  private static readonly _autoSelectToolTip = localize('active.folder.auto.tooltip', 'auto');
 
   settingsName = 'workspace';
   command = 'cmake.selectActiveFolder';
   icon = 'folder-active';
+  tooltip = localize('click.to.select.workspace.tooltip', 'Click to select the active workspace');
```
bobbrow, line 159: It should say active "folder", not "workspace" to avoid confusing the terminology with what VS Code uses.

SchweizS: fixed.

src/status.ts
bobbrow: I think putting (auto) in the tooltip is confusing.  I would rather that be removed.

SchweizS:
i didn't want to touch everything which is related to that feature, so i just commented out he 
parts related to autoSelect in status.ts

bobbrow:
That's fine.  We will clean it up in a separate PR.

PR #1201, opened 2020-04-24
Add optional launch args

cogilvie:
### This changes [[visible behavior]]

- This allows specifying arguments to be used with Launch without Debug


andreeis:
Thank you for looking into fixing this.
1. We recommend is to not add a new array setting and use instead the args array in "cmake.debugConfig".

2. There might be various problems with quotes and argument passing in different terminals, which this PR does not cover and which would be better addressed centrally in VSCode by some API that would benefit all extensions.

3. We will not be ready to take this PR for the upcoming release but we want to investigate what options are there about passing an arguments array to a terminal in VSCode.

I opened an issue in the VSCode repo. https://github.com/microsoft/vscode/issues/96091. Let's see what this team will suggest.

Tudyx:
Any news on this PR?

andreeis:
We merged PR https://github.com/microsoft/vscode-cmake-tools/pull/1603 which overlapped with yours. That had simpler code and being more recent didn't have conflicts. It was easier to merge.
I actually merged some conflicts first, without realizing it's into your branch (I apologize) and that caused the typescript to fail (not where I merged, somewhere else, I wonder how it compiled before). Now there's one more conflict which you can solve if you want to continue with what this PR had extra than the other one, which is some nice refactoring around getDebugConfig. Not critical though (or let me know if I'm missing anything).

andreeis:
Closing this PR for now.

Issue #1167, opened 2020-04-09
Prevent CMake Tools activation when CMakeLists.txt is not in the workspace folder or at cmake.SourceDirectory


andreeis:
Since CMake Tools changed the activation condition from CMakeLists.txt found in the root to anywhere inside the workspace, we get into situations when it activates for non CMake repos. One example is when a non CMake repo contains some tests that have CMake files.

Preventing the activation or programmatically deactivating the extension might not be possible, but at least we would like to disable the features until the state of the workspace changes (it becomes a real valid CMake repo).

Integrate this functionality with the message box asking about QuickStart or cmake.sourceDirectory, providing a third option, to ignore this folder.

parched:
This looks great, thank you. It's not clear to me though if this is something I can set in setting.json. I would like to share share my multiroot workspace with my team and they not get the popup about missing CMakeLists at all. Will that be possible?

bobbrow:
Not in 1.4.0.  If you want to open a new issue to request we add a setting for it, we can consider doing that for a future release.

You can try out the experience today by installing the beta on our [releases page](https://github.com/microsoft/vscode-cmake-tools/releases/tag/1.4.0-beta)

andreeis:
This fix is available in CMake Tools #1.4.0 which has been released.

PR #1206, opened 2020-04-29
Support partial CMT activation for non cmake repos

andreeis:
There are repos that are not CMake based but which may have some CMake files further down in the folders structure (like tests suites) that will eventually trigger the CMake Tools activation. Having the buttons visible in status bar, the commands active in the pallette and also various popups that are trying to fix things are annoying. vscode-cmake-tools is a good example.
This PR implements a partial activation mode for CMake Tools when operating on repos that don't have a CMakeLists.txt in the root of the active folder or where cmake.sourceDirectory points at.
The partial activation mode is set once the user chooses the new "Ignore" button from the popup about missing CMakeLists.txt. The choice is saved in state. The second time the extension is activated the popup will not appear.
Full activation is given another chance with every change of the setting cmake.sourceDirectory and of the CMakeLists.txt file. Also a successful configure results in full activation (the cmake.sourceDirectory or the CMakeLists.txt files may have been changed outside VSCode and not go through the handling of the watchers).


package.json
```diff
@@ -80,51 +80,61 @@
       {
         "command": "cmake.viewLog",
         "title": "%cmake-tools.command.cmake.viewLog.title%",
+        "when": "!cmake:cmtPartiallyActive",
```
bobbrow, line 83: Maybe a context name with negated meaning would have saved you adding all of the `!`'s in this file. `cmake.cmtActive` or some such.

src/cmake-tools.ts
```diff
@@ -243,17 +250,38 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+        const quickStart = localize('quickstart.cmake.project', 'Create');
+        const changeSetting = localize('edit.setting', 'Locate');
+        const ignoreCMakeListsMissing = localize('ignore.activation', 'Ignore');
+        const result = await vscode.window.showErrorMessage(
+            localize('missing.cmakelists', 'CMakeLists.txt was not found in the root of the folder \'{0}\'. How would you like to proceed?', this.folderName),
+            quickStart, changeSetting, ignoreCMakeListsMissing);
+        if (result === quickStart) {
+          vscode.commands.executeCommand('cmake.quickStart');
+        } else if (result === changeSetting) {
+          vscode.commands.executeCommand('workbench.action.openSettings');
```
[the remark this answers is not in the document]
andreeis: Makes sense. I didn't figure out how to associate a tooltip to the buttons, to leave the initial text on them the same and add additional info as tooltip. 
Then, I tried adding small but relevant text to the buttons text which causes the UI to display the message partially (and the rest with a mouse over). Not ideal, but acceptable. 
Or you think we should make the original message ("CMakeLists.txt was not found ... ") bigger?
Or you're thinking of another separate message?

src/extension.ts
```diff
@@ -165,6 +181,18 @@ class ExtensionManager implements vscode.Disposable {
     telemetry.logEvent('open', telemetryProperties);
   }
 
+  // Partial activation means that the CMake Tools commands are hidden
+  // from the commands pallette and the status bar is not visible.
+  // The context variable "cmake:cmtPartiallyActive" is always equal
+  // to the state setting ignoreCMakeListsMissing.
+  // To have them both always in sync, cmake:cmtPartiallyActive is set
+  // by the getter and setter of ignoreCMakeListsMissing.
+  public async partialActivation(partiallyActive: boolean) {
```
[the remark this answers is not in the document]
andreeis: You're right, enableExtensionFeatures reads better. What do you think of a shorter one, enableCmtFeatures? 6 letters less.

src/extension.ts
```diff
@@ -179,6 +207,9 @@ class ExtensionManager implements vscode.Disposable {
    * The folder controller manages multiple instances. One per folder.
    */
   private readonly _folders = new CMakeToolsFolderController(this.extensionContext);
+  public getActiveFolder(): vscode.WorkspaceFolder {
+    return this._folders?.activeFolder?.cmakeTools.folder || vscode.workspace.workspaceFolders![0];
```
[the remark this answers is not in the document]
andreeis: I've seen this pattern in the code base but I can put a nice and safe guard, of course, it's always better.

src/extension.ts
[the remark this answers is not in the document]
andreeis: I know it can cause double negatives, but disableCmtFeatures doesn't sound bad to me either. enableCmtFeatures would confuse me because I would wonder why an extension would not have its features enabled.
Or, enableFullFeatureSet?

src/cmake-tools.ts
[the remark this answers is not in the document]
andreeis: Thanks for the idea with the open file dialog, it can be done. I'll send update soon.

src/cmake-tools.ts
```diff
@@ -243,17 +250,49 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+          vscode.commands.executeCommand('cmake.quickStart');
+        } else if (result === changeSetting) {
+          const openOpts: vscode.OpenDialogOptions = {
+            canSelectMany: false,
+            defaultUri: this.folder.uri,
+            filters: {"CMake files": ["txt"], "All files": ["*"]},
+            openLabel: "Load",
+          };
+          const cmakeListsFile = await vscode.window.showOpenDialog(openOpts);
+          if (cmakeListsFile) {
```
bobbrow, line 270: If no `cmakeListsFile` is selected, then we need to disable the extension features.

src/cmake-tools.ts
```diff
@@ -243,17 +250,49 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+        const quickStart = localize('quickstart.cmake.project', 'Create');
+        const changeSetting = localize('edit.setting', 'Locate');
+        const ignoreCMakeListsMissing = localize('ignore.activation', "Don't show again");
+        const result = await vscode.window.showErrorMessage(
+            localize('missing.cmakelists', 'CMakeLists.txt was not found in the root of the folder \'{0}\'. How would you like to proceed?', this.folderName),
+            quickStart, changeSetting, ignoreCMakeListsMissing);
+        if (result === quickStart) {
+          vscode.commands.executeCommand('cmake.quickStart');
+        } else if (result === changeSetting) {
+          const openOpts: vscode.OpenDialogOptions = {
```
bobbrow, line 263: This is pretty cool.  Is it possible to filter it to just show CMakeLists.txt instead of all .txt files?

package.json
```diff
@@ -80,51 +80,61 @@
...
       {
         "command": "cmake.setVariantAll",
         "title": "%cmake-tools.command.cmake.setVariantAll.title%",
+        "when": "cmake:enableFullFeatureSet",
         "category": "CMake"
       },
       {
         "command": "cmake.configure",
         "title": "%cmake-tools.command.cmake.configure.title%",
+        "when": "cmake:enableFullFeatureSet",
```
bobbrow, line 125: I'm thinking that maybe this command can always stay active so that if a user creates a CMakeLists.txt file in a folder that doesn't have one, they can use this command as an entry point to enable the extension. I've been doing this when testing previous PR's, so I noticed when I wasn't able to do it with this one.

src/cmake-tools.ts
andreeis: I tried but I wasn't able to unfortunately.

src/cmake-tools.ts
andreeis: I can do that. Originally I thought that we disable the features only if the user chooses "ignore".

package.json
andreeis: Yes good idea. I realize now that all configure commands and 'quick start' make sense to remain active.

bobbrow:
#1167

package.json
```diff
@@ -464,8 +453,7 @@
           "command": "cmake.resetState"
         },
         {
-          "command": "cmake.configureAll",
-          "when": "cmake:multiRoot && cmake:enableFullFeatureSet"
```
bobbrow: Don't remove the `cmake:multiRoot` criteria.

src/cmake-tools.ts
```diff
@@ -260,9 +260,14 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         if (result === quickStart) {
           vscode.commands.executeCommand('cmake.quickStart');
         } else if (result === changeSetting) {
+          // Open the search file dialog from the path set by cmake.sourceDirectory or from the current workspace folder
```
bobbrow, line 263: I thought the focus of this flow was that `cmake.sourceDirectory` is not set and we are going to set it now.  I don't think we should try to create the driver at this location in the code.  Just use the workspace folder for now.

src/cmake-tools.ts
```diff
@@ -1336,7 +1344,8 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
       return -2;
     }
 
-    const sourceDir = cmtFolder.folder.uri.fsPath;
+    const drv = await this._cmakeDriver;
```
bobbrow, line 1347: I don't think this should change. This should use the workspace folder.

src/extension.ts
```diff
@@ -181,20 +181,30 @@ class ExtensionManager implements vscode.Disposable {
...
-    if (activeFolder) {
-      const extensionState = new StateManager(this.extensionContext, activeFolder);
-      extensionState.ignoreCMakeListsMissing = !fullFeatureSet;
+  public enableFullFeatureSet(fullFeatureSet: boolean) {
+    const context = this.getActiveFolderContext();
+    if (context) {
+      context.ignoreCMakeListsMissing = !fullFeatureSet;
+      this._statusBar.setVisible(fullFeatureSet);
+    } else {
+      log.info(localize('', 'enableFullFeatureSet(\'{fullFeatureSet}\') called but not active folder found.'));
```
bobbrow: Did you mean to insert the value of `fullFeatureSet` here?  You need to use `{0}` and pass the parameter in afterwards:
```
localize('enableFullFeatureSet.no.active.folder', 'enableFullFeatureSet({0}) called but no active folder found.', fullFeatureSet);
```

Note: I also fixed a spelling mistake in the message.

bobbrow: And actually, I don't think this message belongs in the `info` stream.  I think `log.error` is more appropriate.

andreeis: Oh, I made a confusion with some other scenario. And I forgot to test that message.

package.json
andreeis: good catch

src/cmake-tools.ts
andreeis: As discussed offline, I reverted both changes regarding using sourceDir versus workspace folder (which require driver instantiation) to not risk any destabilization for the 1.4 release and I will open a new issue (low priority) describing the scenarios that are impacted by this to review later.

src/extension.ts
```diff
@@ -203,7 +203,7 @@ class ExtensionManager implements vscode.Disposable {
       context.ignoreCMakeListsMissing = !fullFeatureSet;
       this._statusBar.setVisible(fullFeatureSet);
     } else {
-      log.info(localize('enableFullFeatureSet.no.active.folder', 'enableFullFeatureSet(\'{fullFeatureSet}\') called but not active folder found.'));
+      log.error(localize('enableFullFeatureSet.no.active.folder', 'enableFullFeatureSet({0}) called but not active folder found.'), fullFeatureSet);
```
bobbrow, line 206: Sorry, I said `error` at first, but then realized it should be `debug` or `trace`

andreeis: Right.

Issue #1195, opened 2020-04-22
Latest develop branch broken


Yuri6037:
### Brief Issue Summary

I just tried to setup this extension in development mode, but everytime I run it on existing CMake C++ projects it fails with the following error in console:
"The path to the compiler for one or more source files was not found in the CMake cache. If you are using a toolchain file, this probably means that you need to specify the CACHE option when you set your C and/or C++ compiler path"

### Expected:

I suppose it should not have thrown an error.

### CMake Log

[variant] Loaded new set of variants
[kit] Successfully loaded 14 kits from C:\Users\Yuri Edward\AppData\Local\CMakeTools\cmake-tools-kits.json
[main] Configuring folder: Framework 
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.EXE" --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE "-Hc:/Users/Yuri Edward/BlockProject3D/Framework" "-Bc:/Users/Yuri Edward/BlockProject3D/Framework/build" -G "Visual Studio 16 2019"
[cmake] Not searching for unused variables given on the command line.
[cmake] -- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.18363.
[cmake] Writing module descriptor "BPF.cmake"...
[cmake] -- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.18363.
[cmake] -- Configuring done
[cmake] -- Generating done
[cmake] -- Build files have been written to: C:/Users/Yuri Edward/BlockProject3D/Framework/build/Tests/googletest-download
[cmake] Microsoft (R) Build Engine version 16.3.1+1def00d3d for .NET Framework
[cmake] Copyright (C) Microsoft Corporation. All rights reserved.
[cmake] 
[cmake]   Performing update step for 'googletest'
[cmake]   Current branch master is up to date.
[cmake]   No configure step for 'googletest'
[cmake]   No build step for 'googletest'
[cmake]   No install step for 'googletest'
[cmake]   No test step for 'googletest'
[cmake]   Completed 'googletest'
[cmake] -- Configuring done
[cmake] -- Generating done
[cmake] -- Build files have been written to: C:/Users/Yuri Edward/BlockProject3D/Framework/build

### CMake Tools Log (part of cause it's so large)

[...]

PR #1211, opened 2020-05-01
Silent scan for kits to avoid breaking changes in kits definition

andreeis:
This PR fixes the issue https://github.com/microsoft/vscode-cmake-tools/issues/1195. 
Silently scan for kits to ensure the users do not encounter kits breaking changes (that are solvable with scanning).
If this causes any inconvenience for the users, there is the setting cmake.scanForKitsOnOpen (defaulting to true) which can be set to false.
Many places available for hooking this. I chose at the end of workspace open and before configuring.
This happens also before the silent scan triggered by missing or empty kits json. That change will be in effect less than before this PR (for opening, never) but it can still happen when kits get deleted or corrupted between opening a workspace and any operations that triggers configuration.


package.json
```diff
@@ -80,51 +80,61 @@
       {
         "command": "cmake.viewLog",
         "title": "%cmake-tools.command.cmake.viewLog.title%",
+        "when": "!cmake:cmtPartiallyActive",
```
bobbrow, line 83: I don't think these changes belong in this PR

src/extension.ts
```diff
@@ -365,6 +365,10 @@ class ExtensionManager implements vscode.Disposable {
     if (should_configure) {
       // We've opened a new workspace folder, and the user wants us to
       // configure it now.
+      if (cmt.workspaceContext.config.scanForKitsOnOpen) {
+        await this.scanForKits();
```
bobbrow, line 369: Thinking about this... If we have to scan every time the folder opens, this seems excessive and will slow down IntelliSense for cpptools because the configurations won't be sent until after 'configure' happens.

We should only need to scan when we know something has changed with the Kits, right?  I think we could manage this with a piece of persistent state that tracks a "version number" or something. If we ever make a change that requires a rescan, then we bump the number. If the state doesn't exist, we do a scan (e.g. this will happen for everyone once they upgrade to 1.4, but then it won't happen again unless we change Kits again).  If we take this approach, we don't need a setting to disable it.

andreeis: True, good idea. 
I just pushed an update. Basically I reverted the whole thing. 
I found a better spot to hook the breaking changes scan at the beginning of general activation (it didn't have to be part of the configure process, I was following too closely the setting cmake.configureOnOpen).
I implemented some versioning functionality tied to the extension version defined in package.json.
To avoid yet another version to take care of and also to be able to rescan for kits appropriately when users go back and forth with various versions of CMake Tools I decided to do this rescan every time when the kits version doesn't match the extension version. It will be slightly more often than if we implement a new versioning for kits but I think it's acceptable and avoids too complicated implementation,.

package.json
andreeis: ups

src/extension.ts
```diff
@@ -1052,6 +1053,15 @@ export async function activate(context: vscode.ExtensionContext) {
         await vscode.window.showWarningMessage(localize('uninstall.old.cmaketools', 'Please uninstall any older versions of the CMake Tools extension. It is now published by Microsoft starting with version 1.2.0.'));
     }
 
+    // Rescan if the kits version (saved in the extension state) is not the same as the extension version,
+    // to avoid eventual breaking changes in the kits definitions.
+    const kitsVersion = context.workspaceState.get<string>('kitsVersion');
```
bobbrow, line 1058: Let's move this to an async function in kits.ts so that the related logic is all together.  Then call/await it here.

src/extension.ts
```diff
@@ -1052,6 +1053,15 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    const extVersion = vscode.extensions.getExtension('ms-vscode.cmake-tools')?.packageJSON.version;
```
bobbrow, line 1059: This can just be a const number that we increment whenever the kits version changes.

andreeis: Ok.

src/extension.ts
andreeis: Done

src/kit.ts
```diff
@@ -973,6 +973,18 @@ export async function scanForKits(opt?: KitScanOptions) {
   });
 }
 
+// Rescan if the kits versions (extension context state var versus value defined for this release) don't match.
+export async function scanForKitsIfNeeded(context: vscode.ExtensionContext) : Promise<void> {
+  const kitsVersionSaved = context.workspaceState.get<number>('kitsVersionSaved');
```
bobbrow, line 978: this should be `context.globalState`.

andreeis: good catch I just created new PR for this quick fix.

Issue #984, opened 2019-12-23
CMake tools should not configure the project if cache is available


WSUFan:
Hi guys.

I'm new to use CMake tools for VScode. Right now I'm developing a C++ project using the Docker and VScode remote SSH. Everything works fine except for one thing. The project is managed by CMake. So I have installed the CMake extension in VScode. When I reconnect to the remote container (like reload the VScode), the CMake tool will always configure the project first (it may think I did not configure the project before but I did!!). I believe all data is mounted through outside and can be persisted. So anyone knows how to solve this thing?

Thanks

bobbrow:
There is a setting `cmake.configureOnOpen` that controls whether a configure is run automatically when the folder opens.  By default it should be disabled, so I'm not sure why you're seeing what you're seeing.  But in general, the extension will do a configure before building if one hasn't been run yet in that session because it doesn't know the state of the project when it loads.

WSUFan:
Thanks for replying. Actually I disable that setting. And what do you mean by "current session"? The session state is not saved somewhere? Because when I reopen the VScode I will need to configure the project again even I have already configured it before.

bobbrow:
The "current session" is when you open the folder in VS Code.  If you close and reopen the folder that starts a new "session".

In any case, we recommend that CMake Tools configure your project on open and will probably change the default for the setting to `true` in a future release.  The C++ extension can't ask CMake Tools for project information so that IntelliSense will work better unless CMake Tools has run the configure step.

WSUFan:
Ok. That makes sense. Is that possible that the tool can cache the current session state so that it does not need to run the configuration again when each time I open the same folder?

bobbrow:
We'll have to look into it.  It seems like CMake should allow us to get project information if the cache file is present.  It will be up to the user to ensure the cache is up to date in this case though.

How long does your configure step take?

WSUFan:
Sounds great. My project is very big so it takes a couple of minutes to configure.

stefanweiser:
In my case vscode 1.43.2 seem to configure always after stopping the build. Even with cmake.configureOnOpen set to true. Then after opening vscode configures the workspace once. When I build then no configuration is done. When I let the build finish not configuration is done again, but if I interrupt the build and build it again, it will unnecessarily configure the workspace again.

andreeis:
CMake Tools 1.4.0 Beta is available on GitHub.
[...]

PR #1212, opened 2020-05-01
Skip configuring when cache is present and according setting is ON

andreeis:
This PR fixes issue 984: https://github.com/microsoft/vscode-cmake-tools/issues/984 

Allow the users to set when the extension should skip over configuring operations: new setting, plus verify if the cache file is present. 
Add info log message so that the user is aware of the need to keep the cache up to date.

Functionality around configureOnOpen remains the same. The PR causes configuring to be skipped (even if state is evaluated as dirty) before building, debugging or launching.

No easy way to fix the symptom of reconfiguring after stopping a build. It would require some rewrite. Tracking with issue https://github.com/microsoft/vscode-cmake-tools/issues/1214.


src/cmake-tools.ts
```diff
@@ -732,11 +732,22 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
     const drv = await this.getCMakeDriverInstance();
-    if (!drv || await drv.checkNeedsReconfigure()) {
+    if (!drv) {
       return true;
-    } else {
+    }
+
+    const needsReconfigure: boolean = await drv.checkNeedsReconfigure();
+
+    const skip_configure_if_cache_present = this.workspaceContext.config.skipConfigureIfCachePresent;
```
bobbrow, line 741: Inconsistent variable naming. Can you remove the underscores and use camelCase?

SandNet:
@bobbrow This change is missing documentation. Can't find this command on either https://vector-of-bool.github.io/docs/vscode-cmake-tools/settings.html or https://github.com/microsoft/vscode-cmake-tools

PR #1235, opened 2020-05-13
Add buildKitShort command, add shortName property to kits

janspisiak:

- add buildKitShort command
- add shortName property to kits

## Other Notes/Information

My first time with Typescript, please review carefully :)
Also I used TS 3.7 feature, don't know if it's possible.
I could not get tests running, I'm not sure why.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -62,6 +62,11 @@ export interface Kit {
   name: string;
 
   /**
+   * The short name of the kit
+   */
+  shortName?: string;
+
+  /**
    * The preferred CMake generator for this kit
    */
   preferredGenerator?: CMakeGenerator;
@@ -167,6 +172,15 @@ async function getClangVersion(binPath: string): Promise<ClangVersion|null> {
   };
 }
 
+function majorVersionSemver(semver: string) : string|null {
+  const major_version_re = /^(\d+)./;
+  const major_version_match = major_version_re.exec(semver);
+  if (major_version_match !== null) {
+    return major_version_match[1];
+  }
+  return null;
+}
+
 /**
  * Convert a binary (by path) to a CompilerKit. This checks if the named binary
  * is a GCC or Clang compiler and gets its version. If it is not a compiler,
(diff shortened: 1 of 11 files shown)


src/kit.ts
```diff
@@ -716,9 +737,13 @@ async function tryCreateNewVCEnvironment(inst: VSInstallation, hostArch: string,
   if (!variables) {
     return null;
   }
+  const major_version = majorVersionSemver(inst.installationVersion) ?? inst.installationVersion;
+  const arch = kitHostTargetArch(hostArch, targetArch);
+  const short_name = `msvc${major_version}_${arch}`;
```
janspisiak, line 742: I tried to make default shortNames sensible, but I am not 100% sure about them. For example clang/gcc only use major version a no arch, but I have no expierence with cross-compilation, so I am not sure if that's good enough.

bobbrow: We recently took a PR (#1200) to introduce compact and icon-only visibility modes for the status bar.  Maybe we don't need to do anything for this anymore?

janspisiak: My use case for this is that I need build directories separated by kits, but the default buildKit command produces really long kit names (and makes them un-openable through console in vscode). I know one can create custom kits and give them shorter custom name, but I thought that having sensible build folders out of the box would be nice. #1200 seems to only resolve status bar length problems. Does changing kits-schema introduce some kind of breaking change? I saw some recent PR that tries to solve the kits-needing-refresh problem.

bobbrow: > My use case for this is that I need build directories separated by kits, but the default buildKit command produces really long kit names

That makes sense.  We'll take a look at this and if we have more questions, we'll ask.  Just to set expectations though, we are testing the release build for 1.4 and plan to publish it in a day or so, so this won't make it into 1.4.  And since we won't release 1.5 for a few months after that and we have some other work we'll be starting that is not related to this extension, it may be a few weeks before we get to this PR.  Thank you in advance for understanding.

lygstate:
I suggest use buildKitName buildKitTarget buildKitMajorVersion buildKitFullVersion instead

lygstate:
Thanks for you wonderfull work

lygstate:
@janspisiak Hello, do you any objection about my idea?
If you don't have the time to update the pull-request, I'd like to update it:)

janspisiak:
@lygstate sorry for the delay, and thanks for suggestions :)
However I am not qualified enough to know, if there are any drawbacks to supporting that many commands (performance or other). I am now using cmake-kits.json to define my own name (thus also changing the build directory), and I think adding shortName property to kits is great as it makes possible to customize build dir, without having to overwrite full kit name.
But what you suggest could be possible, and there is some documentation https://code.visualstudio.com/api/extension-guides/command
which to me seems it should be possible to refactor buildKit commands into common namespace eg. _cmake.buildKit._{major, targetArch, targetOs}, or maybe parametrize buildKit? I don't know how would passing arguments from settings work however.
In summary, I would love to hear from the maintainers what do they think and if they would accept such big change :)

lygstate:
@janspisiak Gocha, so shortName can be customized in JSON file, and we can auto detect cmake.buildKit.{major, targetArch, targetOs}

lygstate:
At least, we need to make sure that the re-scan would not cause the shortName disappear, or it's would be no-lucky. And for shortName, can it be committed in workspace settings?, that's what I worried about shortName

lygstate:
I've create a updated version at https://github.com/microsoft/vscode-cmake-tools/pull/1337, that not using the shortName, but offer vendor hostOs targetOs targetArch versionMajor versionMinor separately,
so the user can choose what they want.

lygstate:
This can also be closed now

PR #1239, opened 2020-05-14
Fix scope for state variable that guards the silent scan for kits

andreeis:
Change the scope of the state variable that guards the silent scan for kits from workspace to global to avoid unnecessary scanning for each new folder.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -975,7 +975,7 @@ export async function scanForKits(opt?: KitScanOptions) {
 
 // Rescan if the kits versions (extension context state var versus value defined for this release) don't match.
 export async function scanForKitsIfNeeded(context: vscode.ExtensionContext) : Promise<void> {
-  const kitsVersionSaved = context.workspaceState.get<number>('kitsVersionSaved');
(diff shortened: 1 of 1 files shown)


src/kit.ts
```diff
@@ -981,7 +981,7 @@ export async function scanForKitsIfNeeded(context: vscode.ExtensionContext) : Pr
   // Scan also when there is no kits version saved in the state.
   if (!kitsVersionSaved || kitsVersionSaved !== kitsVersionCurrent) {
     await scanForKits();
-    context.workspaceState.update('kitsVersionSaved', kitsVersionCurrent);
+    context.globalState.update('kitsVersionSaved', kitsVersionCurrent);
```
bobbrow, line 984: You need to update the one on line 978 too.

andreeis: oh of course

bobbrow:
Do you know what is going on with the tests?

andreeis:
> Do you know what is going on with the tests?

I noticed just since this morning (or yesterday after noon) that they seldom fail and pass with rerun. I am not reproducing on my ubuntu.

Looking deeper into this now since it didn't happen just once.

PR #1247, opened 2020-05-18
Forgot one more scenario for when we want partial feature set enabled

andreeis:
Enable only the reduced features set of CMake Tools when there is no kit chosen from the "Select a kit" quick pick (which pops up immediately after opening a fresh folder that hasn't been successfully configured before).


andreeis:
Postponing, just to eliminate any risk for 1.4, especially since this is not a critical scenario at all.

andreeis:
Closing this PR, as this scenario was fixed differently in PR https://github.com/microsoft/vscode-cmake-tools/pull/1689.

Issue #1249, opened 2020-05-19
[1.4.0-beta] cpptools reports unknown architecture flag


mjvk:
### Brief Issue Summary

In 1.4.0-beta a warning is shown that cpptools does not know the target architecture (as set using -march=broadwell in cmake). This warning was not present in the previous version.

### Expected:

No warnings

### Apparent Behavior:

Warning:
```[cpptools] Unknown target architecture flag: broadwell```

### Platform and Versions

- **Operating System**: Windows
- **CMake Version**: 3.16.3
- **VSCode Version**: 1.45.1
- **CMake Tools Extension Version**: 1.40.0-beta
- **Compiler/Toolchain**: mingw

bobbrow:
Thanks for reporting this.  @michelleangela, perhaps we should remove the warning message, or change it to just log the target detected?  We're trying to detect the bitness, but it's valid to pass in chip-specific strings here too.

michelleangela:
Pull request created to remove warning logs. 
https://github.com/microsoft/vscode-cmake-tools/pull/1251

andreeis:
This fix is available in CMake Tools #1.4.0 which has been released.

PR #1251, opened 2020-05-19
Remove warning logs when parsing unknown target architectures

michelleangela:
### This changes logging of parsing target architectures

- Remove warning logs when parsing unknown target architectures

## The purpose of this change

Address issue #1249


bobbrow:
Thanks @michelleangela !

PR #1253, opened 2020-05-20
More kits and archs fixes

andreeis:
One more fix regarding the incompatibility between 'x86' architecture name used by VS and 'win32' platform name used by CMake preferred generator.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -434,9 +434,9 @@ export function vsDisplayName(inst: VSInstallation): string {
  *
  * @param inst The VSInstallation to use
  * @param hostArch The architecture of the toolset host (e.g. x86, x64|amd64)
- * @param targetArch The architecture of the toolset target (e.g. x86, x64|amd64, arm, arm64)
+ * @param targetArch The architecture of the toolset target (e.g. win32|x86, x64|amd64, arm, arm64)
  */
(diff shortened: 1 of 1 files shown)


src/kit.ts
```diff
@@ -467,10 +467,20 @@ function kitHostTargetArch(hostArch: string, targetArch?: string, amd64Alias: bo
...
-  return hostArch;
+
+  // CMake preferred generator platform requires 'win32', while vcvars are still using 'x86'.
+  // This function is called only for VS generators, so it is safe to overwrite
+  // targetArch with the vcvars naming.
+  // In case of any future new mismatches, use the vsArchFromGeneratorPlatform table
+  // instead of hard coding for win32 and x86.
+  // Currently, there is no need of a similar overwrite operation on hostArch,
+  // because CMake host target does not have the same name mismatch with VS.
+  targetArch = vsArchFromGeneratorPlatform[targetArch] || targetArch;
```
bobbrow, line 481: Should you also add x64 -> amd64 into `vsArchFromGeneratorPlatform` and remove lines 465-467?

bobbrow: Oh, maybe not.  That transformation is conditional on `amd64Alias`.

Issue #499, opened 2018-07-26
Add touch bar controls for cmake commands


rafmudaf:
The title pretty much covers it, but it would be really helpful to add vscode-cmake-tools commands to the touch bar. I already use the touch bar to run debug commands, so adding configure, build, and maybe ctest buttons would go pretty far for my workflow. I haven't actually seen a VS Code extension make use of this, but extensions can add [custom touch bar buttons](https://code.visualstudio.com/updates/v1_17#_touch-bar-support).

If it is decided that this would be worthwhile, I'm happy to take a stab at implementing it.

Randshot:
I did take a look at the vscode api, and it looks simple enough for that to be added. The only problem will be testing this feature, since I don't think that any of us (@vector-of-bool and the other contributors) have a MacBook with a touchbar. At least I only have an older model.

rafmudaf:
Got ya, let me know if there's any way I can help in testing as I do have a touch bar.

vector-of-bool:
If you are able to do VSCode development you could send PR. Otherwise, the turnaround time for iterating on the creating of touch bar items might be too high. Anyone else reading this who might be able to offer help: Your assistance would be greatly appreciated.

andreeis:
This should be fixed in CMake Tools 1.5.0 which was published today.
Let us know if you encounter any issues with this release.

PR #1263, opened 2020-05-25
MacBook Pro touchbar support

vptrbv:
## This change addresses item #499

Contributing Build, Launch and Debug commands to MacBook Pro Touch Bar.


bobbrow:
This looks pretty straightforward, but I don't have a way to validate this change since we don't have a mac with touchbar support to test on.  Are you able to share a photo of what the touchbar looks like when these are enabled? Uploading it to this PR would be fine.

package.json
```diff
@@ -810,6 +810,35 @@
...
           "when": "resourceFilename == CMakeLists.txt && cmake:enableFullFeatureSet"
         }
+      ],
+      "touchBar": [
+        {
+          "command": "cmake.configure",
+          "when": "!cmake:enableFullFeatureSet",
+          "group": "CMake:build"
+        },
+        {
```
andreeis, line 820: The condition negation is probably a typo here, right?  However, we want various commands (like quickStart, all configure operations) to be always on because they represent a way back from a partial to a full activation mode. I think you can remove the condition here.

vptrbv: Is quickStart needed when you already have configured (or even at least ready to be configured) project?

TouchBar is actually quite small, and placing custom commands there should have reasoned advantage of having it there at the moment: overextending commands and groups are just cut from Touch Bar, no slide gestures supported.

I can try to use images instead of default text title - but it [requires copies of already existing SVGs to be stored in PNG format](https://code.visualstudio.com/updates/v1_17#_touch-bar-support). So I need your advice here: should we do this?

There's another way to place more commands - `"alt"` property. It was supposed to switch command on Alt pressed - but I haven't noticed it's working. Just in case I'll check it again :)

vptrbv:
IMG_0348
IMG_0349
IMG_0350

vptrbv:
Since these commands use commands same as tasks, this PR needs (another fix)[https://github.com/microsoft/vscode-cmake-tools/pull/1268]

package.json
bobbrow: Quick start probably isn't needed for already configured projects, but it's possible someone deletes files and wants to start over.

"Always on" was referring to having the commands be available in VS Code, not the Touchbar, I think.  We're not asking for more commands in the Touchbar.

vptrbv:
> Are we waiting on PR #1268 before we commit this? Or is it ready to go now?

For now it's working only with no files open VSCode - in other cases it's providing currently open file as argument (if I'm not mistaken - checked some time ago). 
Fix for this was initial change in PR #1268 - but, as you know, it became something bigger.

andreeis (approved):

bobbrow:
We now build an installable extension with each commit.  You can test this out by downloading cmake-tools.vsix from https://github.com/microsoft/vscode-cmake-tools/actions/runs/310694927 and using the "install from vsix" command in VS Code.

Issue #1264, opened 2020-05-26
Is it possible to run commands like `cmake.build` from tasks.json?


ce3a:
### How to run the `cmake.build` command as a vscode task (from tasks.json)?

I would like to run the `cmake.build` command as a vscode task. My `task.json` looks like this:

```jsonc
    "version": "2.0.0",
    "tasks": [
        {
            "label": "Build",
            "command": "${command:cmake.build}",
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "problemMatcher": [ ... ],  //  ... for a proprietary compiler
        }
    ]
 ```

### Expected:

1. Tasks: Run Build Task
2. The `cmake-tools` cammand `cmake.build` shall be executed.

### Apparent Behavior:

Error message popup appers:

 > "Cannot read property 'fsPath' of undefined"

### CMake Tools Log

```
2020-05-26T08:05:28.964Z [debug] [extension] [6236] cmake.build gestartet
2020-05-26T08:05:28.965Z [fatal] [rollbar] Ausnahmefehler: Unbehandelte Promise-Ablehnung: build TypeError: Cannot read property 'fsPath' of undefined {}
```

### Developer Tools Log

```
[2020-05-26 10:05:28.977] [renderer1] [error] Cannot read property 'fsPath' of undefined: TypeError: Cannot read property 'fsPath' of undefined
	at CMakeToolsFolderController.get (c:\<HOME>\.vscode\extensions\ms-vscode.cmake-tools-1.4.0\dist\main.js:41014:48)
	at ExtensionManager.mapCMakeToolsFolder (c:\<HOME>\.vscode\extensions\ms-vscode.cmake-tools-1.4.0\dist\main.js:13673:59)
	at ExtensionManager.build (c:\<HOME>\.vscode\extensions\ms-vscode.cmake-tools-1.4.0\dist\main.js:13692:39)
	at c:\<HOME>\.vscode\extensions\ms-vscode.cmake-tools-1.4.0\dist\main.js:13833:35
	at c:\<HOME>\.vscode\extensions\ms-vscode.cmake-tools-1.4.0\dist\main.js:13844:15
	at e._executeContributedCommand (c:\opt\Microsoft VS Code\resources\app\out\vs\workbench\services\extensions\node\extensionHostProcess.js:627:894)
[...]

Issue #1234, opened 2020-05-13
Using command variables in task with CMake Tools


stlkr:
### Brief Issue Summary

Tried to use  this in tasks.json:

```json
{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "CMake Build",
            "command": "${command:cmake.build}",
            "type": "shell",
            "group": "build",
            "problemMatcher": []
        }
    ]
}
```
Or

```json
{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "CMake Build All",
            "command": "${command:cmake.buildAll}",
            "type": "shell",
            "group": "build",
            "problemMatcher": []
        }
    ]
}
```
### Expected:

CMake Tools to build project as when running "CMake: Build" or "CMake: BuildAll" command from command palette.

### Apparent Behavior:

Got error:
For "CMake Build": 
```
[rollbar] Unhandled exception: Unhandled Promise rejection: build TypeError: Cannot read property 'fsPath' of undefined {}
```
For "CMake Build All":
```
[rollbar] Unhandled exception: Unhandled Promise rejection: buildAll TypeError: str.replace is not a function {}
```
After "CMake Build All" task CMake tools becomes broken, and even from CMake side panel gives an error:

```
A CMake task is already running. Stop it before trying to build.
```

### Platform and Versions

- **Operating System**: Linux Mint 19
- **CMake Version**: 3.5.1
- **VSCode Version**: 1.38.1
- **CMake Tools Extension Version**: 1.3.1
- **Compiler/Toolchain**: arm-none-eabi-gcc 4.8

bobbrow:
It looks like the tasks pass in an unexpected argument to our commands.

stlkr:
Also, if(or when) command variable feature becomes working, it would be nice to publish the command IDs for CMake Tools somewhere in documentation. Because I wasn't able to find it anywhere except CMake Tools's sources.

andreeis:
This should be fixed in CMake Tools 1.5.0 which was published today.
Let us know if you encounter any issues with this release.

fjeremic:
@andreeis I just tried this via:

```
{
	"version": "2.0.0",
	"tasks": [
        {
            "label": "CMake Build",
            "command": "${command:cmake.build}",
            "type": "shell",
            "group": "build",
            "problemMatcher": []
        }
	]
} 
```

And it does not work. I get a notification saying:

[...]

PR #1268, opened 2020-05-27
Initial tasks support

vptrbv:
## This change addresses items #1264, #1234

In cases with file build command context, empty ws uri is possible. Additional check added.


src/folders.ts
```diff
@@ -82,7 +82,7 @@ export class CMakeToolsFolderController implements vscode.Disposable {
    * @param ws The workspace folder to search
    */
   get(ws: vscode.WorkspaceFolder | undefined): CMakeToolsFolder | undefined {
-    return ws ? this._instances.get(ws.uri.fsPath) : ws;
+    return (ws && ws.uri) ? this._instances.get(ws.uri.fsPath) : undefined;
```
bobbrow, line 85: I believe the path is passed in as element 2 in an array for the "run command from task" case.  This function should accept a new type for `ws` which is an array of two elements.

vptrbv: Thank you for suggestion - it's totally right, I've updated PR... And I guess we need test suite for tasks now.

src/folders.ts
```diff
@@ -79,10 +79,19 @@ export class CMakeToolsFolderController implements vscode.Disposable {
...
   /**
    * Get the CMakeTools instance associated with the given workspace folder, or undefined
-   * @param ws The workspace folder to search
+   * @param ws The workspace folder to search, or array of command and workspace path
    */
-  get(ws: vscode.WorkspaceFolder | undefined): CMakeToolsFolder | undefined {
-    return ws ? this._instances.get(ws.uri.fsPath) : ws;
+  get(ws: vscode.WorkspaceFolder | Array<string> | undefined): CMakeToolsFolder | undefined {
+    if (ws) {
+      if ((ws as vscode.WorkspaceFolder).uri) {
```
bobbrow, line 86: Casting in TypeScript is just asserting that a type is what you say it is.  This won't do what you want. 

You need to use `instanceof` to check object types and `typeof` to check intrinsic types.  For example, we do this in cpptools:
```ts
export function isString(input: any): input is string {
    return typeof(input) === "string";
}

export function isArray(input: any): input is any[] {
    return input instanceof Array;
}

export function isArrayOfString(input: any): input is string[] {
    return isArray(input) && input.every(isString);
}
```

vptrbv: Utilities extended, thank you!

src/folders.ts
```diff
@@ -79,10 +80,19 @@ export class CMakeToolsFolderController implements vscode.Disposable {
...
+   * @param ws The workspace folder to search, or array of command and workspace path
    */
-  get(ws: vscode.WorkspaceFolder | undefined): CMakeToolsFolder | undefined {
-    return ws ? this._instances.get(ws.uri.fsPath) : ws;
+  get(ws: vscode.WorkspaceFolder | Array<string> | undefined): CMakeToolsFolder | undefined {
+    if (ws) {
+      if (util.isArrayOfString(ws)) {
+        const command = ws as Array<string>;
+        return this._instances.get(command[command.length - 1]);
+      } else if ((ws as vscode.WorkspaceFolder).uri) {
```
bobbrow, line 90: You should use `ws instanceof vscode.WorkspaceFolder` to check the type or just test the existence of `ws.uri` directly. The `as` type assertion doesn't do what you want.

vptrbv: Since `vscode.WorkspaceFolder` is interface, instanceof is not working with it.

bobbrow:
We're still interested in taking this change.  Are you planning to resubmit it?

vptrbv:
Yes, but I wanted to get tests done. Unfortunately for me, as for a newbie, it's taking time to know TS and VSCode API as well. 
Is it ok that this PR will be open and Travis will be checking til the work is still going?

PS What are you thinking about migration to GitHub Actions?

vptrbv:
Long story short, command is working from task. But `${command:cmake.build}` var extension in task command returns `int` - and it leads to this error notification after build: 
```
Cannot substitute command variable 'cmake.build' because command did not return a result of type string.
```
I see two ways to fix this: change return value - or make CMake Tools `taskProvider` (able to support tasks of type `cmake`). Advice appreciated.

bobbrow:
I think that the task provider route is the more "correct" way to go, but it is a bit of work, so I'd understand if you didn't want to do it.

The piece I wasn't thinking about when you started work on this was that the result of `cmake.build` would need to be a command that vscode would then execute.  Another option could be to create a new command `cmake.buildCommand` or something like that that just returns the command line that we would have executed (the one you generally see in the output window) and then has vscode execute it.  There might be trouble with vscode interpreting the args though since those are supposed to be in the `args` array, so even this approach could be termed a "hack".

src/extension.ts
```diff
@@ -10,30 +10,31 @@ import * as vscode from 'vscode';
 import * as cpt from 'vscode-cpptools';
 import * as nls from 'vscode-nls';
 
-import {CMakeCache} from '@cmt/cache';
+import { CMakeCache } from '@cmt/cache';
```
bobbrow, line 13: If you could avoid reformatting code files while making changes, it helps make the review easier for us.  Formatting changes are generally supposed to be done as a separate PR with no code changes.

vptrbv: Got it, will revert. Just used recommended xaver.clang-format, as I format automatically when possible.

vptrbv:
Since we're providing own task type - do we need tasksBuildCommand as external command?

vptrbv:
Build task is working in Code itself - but test requires full extension activation to run tasks (as well as in Code, actually). `fetchTasks` isn't enough - and unfortunately I haven't found necessary Promise to wait. Angels are crying somewhere, when we remove tests - so I don't want to do it. Can you suggest the promise to wait?

vptrbv:
> PS What do you think about migration to GitHub Actions?

I'm actually tried vanilla CI/CD at my pet project - and it seems it's not that good for C++, as for TS.
Of course it could be done in separate branch - but I've asked because of (maybe) unknown for me reasons to keep up with Travis.

bobbrow:
Sorry I haven't had a chance to take a look at your updates yet.  I will do it soon.

If we make a change to CI, we will probably move to Azure Pipelines like the cpptools extension.  This project's CI is wrapped up in a rather involved Powershell script, so we haven't come up with a cost for migrating that to another system yet.

src/folders.ts
vptrbv: Sorry, just've noticed... `assertion`. I've seen practices doing exactly what I wanted this way: check type and field at the same time. Field check isn't actually as necessary here as type check - but it seems I need something differ from `instanceof` anyway.

bobbrow: You're right.  That's what we have `util.isArrayOfString` for. You'd need to write another type assertion function like that one that checks that the required properties exist and are of the correct type.  In this case, however, you are checking that the `uri` property exists explicitly, so I suppose that no type assertion is actually needed here.

vptrbv: Ended up with `util.isWorkspaceFolder`.

vptrbv:
There's another part I've forgot about: `buildAll`. 
As I don't actually understand how it should work for separate projects in workspace, when we have only one available target setter in status bar - I've just fixed the exception.

The point is: `buildAll` passed to `build` array of `${command:cmake.buildAll}` and first project path. Corresponding `expand.expandString` promise failed in `driver.ts`.

As I'm not using this in work (despite sometimes I'd wanted to) - I'm not really into this.

vptrbv:
@bobbrow @andreeis Please see comments with my eyes on it and comment, if it's possible.

vptrbv:
Ready to go (w/o test)

julianxhokaxhiu:
Can someone please take a look at this?

andreeis (approved):

Issue #1257, opened 2020-05-21
System (stl) headers are missed after updating to 1.4.0


sialiuk:
It looks like after updating to version 1.4.0 extension provides the wrong environment to cmake (Include Directories,), or something like that. 

- Operating System: Windows 10 x64
- CMake Version: 3.16.3
- VSCode Version: 1.45.1
- CMake Tools Extension Version: 1.4.0
- Compiler/Toolchain: MSVC 2019 x64
- Cmake generator: Ninja

I get compiler errors:
fatal error C1083: Cannot open include file: 'memory': No such file or directory
fatal error C1083: Cannot open include file: 'string': No such file or directory
.... and other system headers

The message just after Cmake: Configure command: 
![image](https://user-images.githubusercontent.com/1897738/82558900-5bb19680-9b77-11ea-9e67-abae561352d7.png)

Downgrading to 1.3.1 fixes the problem.

sialiuk:
Looks like related to the issue https://github.com/microsoft/vscode-cmake-tools/issues/1243

andreeis:
I am able to configure fine and the headers are found on my system for the kit and generator you mention (and for any other combinations). First, delete your build folder and try again. We have another bug that we're going to look at later regarding more robust cache and settings updates.

sialiuk:
Thank you, deleting build folder(with resetting build kits) has helped!
I guess it is a problem updating from 1.3.1 to 1.4.0. 
Most probably 1.4.0 has some new build environment settings that are missed in 1.3.1 and after updating initializes them incorrectly.

andreeis:
What do you mean resetting build kits? The kits should have been re-scanned automatically. You had to re-select the used kit?

andreeis:
@sialiuk, I understand now how resetting the kit was needed as a step for getting unblocked. 
What would help, if you have a chance, is reproduce again the migration from 1.3.1 to 1.4.0 problem on your side and share with us the contents of your user kits json file. Besides other issues at play here (which we're working on for 1.4.1), there might be a problem with the silent rescan for kits we implemented in 1.4.0 in the hopes to automatically solve the breaking change in kits definition. You can help us confirm this theory. 
If you manage to do this (if it's not handy to reproduce again don't worry), share also a 'debug' log (cmake.loggingLevel=debug in settings) of the first time you load your project in VSCode after updating to 1.4.0. If you happen to not configure on open, share a log of the first configure you do post migration.

[...]

Issue #1256, opened 2020-05-20
For vs2017, amd64_x86 target are not working.


lygstate:
It should use
C:\PROGRA~2\MICROS~1\2019\COMMUN~1\VC\Tools\MSVC\1425~1.286\bin\Hostx64\x86
not C:\PROGRA~2\MICROS~1\2019\COMMUN~1\VC\Tools\MSVC\1425~1.286\bin\Hostx64\x64\

The host and target may not be the same
```
d] [87/482  16% :: 8.542] Building C object aeron-client\src\main\c\CMakeFiles\aeron_static.dir\concurrent\aeron_mpsc_rb.c.obj
[build] [88/482  16% :: 8.872] Building C object aeron-client\src\main\c\CMakeFiles\aeron_static.dir\concurrent\aeron_counters_manager.c.obj
[build] [89/482  16% :: 8.885] Building C object aeron-client\src\main\c\CMakeFiles\aeron_static.dir\collections\aeron_int64_to_tagged_ptr_hash_map.c.obj
[build] [90/482  16% :: 9.001] Building C object aeron-client\src\main\c\CMakeFiles\aeron_static.dir\collections\aeron_str_to_ptr_hash_map.c.obj
[build] [91/482  17% :: 9.026] Building C object aeron-client\src\main\c\CMakeFiles\aeron_static.dir\concurrent\aeron_mpsc_concurrent_array_queue.c.obj
[build] [92/482  17% :: 9.039] Linking C shared library binaries\aeron.dll
[build] FAILED: binaries/aeron.dll lib/aeron.lib 
[...]

Issue #1243, opened 2020-05-15
[1.4.0-beta] VS environment not set correctly


bobbrow:
@andreeis Now it initializes a correct cross-compiler for MSVC CMake generator, but there is an issue. 

If you have `"cmake.preferredGenerators": ["Ninja"],` in global settings - CMake-Tools'll never initialize `vcvarsall.bat`. It's a bit unexpected. Shouldn't it initialize a correct environment anyway if both `visualStudio` and `visualStudioArchitecture` are present as per documentation?

_Originally posted by @Minimonium in https://github.com/microsoft/vscode-cmake-tools/issues/1084#issuecomment-629515919_

bobbrow:
Confirmed.  We'll fix this before releasing 1.4.0
```
varsForVSInstallation path:'C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise' version:16.5.30011.22 host arch:x86 - target arch:undefined
main.js:60823
Error running C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat x86_undefined, can not found INCLUDE
```

lygstate:
same issue
```
[main] 正在配置文件夹: aeron 
[driver] 删除 e:/CI-Cor-Ready/xemu-rpc/aeron/build-cmake/RelWithDebInfo/CMakeCache.txt
[driver] 删除 e:\CI-Cor-Ready\xemu-rpc\aeron\build-cmake\RelWithDebInfo\CMakeFiles
[proc] 执行命令: "C:\Program Files\CMake\bin\cmake.EXE" --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo -He:/CI-Cor-Ready/xemu-rpc/aeron -Be:/CI-Cor-Ready/xemu-rpc/aeron/build-cmake/RelWithDebInfo -G Ninja
[cmake] Not searching for unused variables given on the command line.
[cmake] -- The C compiler identification is unknown
[cmake] -- The CXX compiler identification is unknown
[cmake] CMake Error at CMakeLists.txt:51 (project):
[cmake]   No CMAKE_C_COMPILER could be found.
[cmake] 
[cmake]   Tell CMake where to find the compiler by setting either the environment
[...]

PR #1277, opened 2020-06-02
Several bug fixes for regressions caused by CMake Tools migration to 1.4.0

andreeis:
This fixes all issues related to kits, generators and VS environment caused by the migration to 1.4.0.
https://github.com/microsoft/vscode-cmake-tools/issues/1256
https://github.com/microsoft/vscode-cmake-tools/issues/1257
https://github.com/microsoft/vscode-cmake-tools/issues/1243

Most impactful of the fixes is that the silent scan for kits (since it was moved from extension.ts to kits.ts) was invoking a different helper with the same name (from a varied family of APIs with the same name) that is not writing into the kits file. Also, being called too early, it would cause various issues (which I didn't understand and solved yet but they don't reproduce anymore with the fix) with occasional exceptions caused by VS environment setup that would interrupt the scanning in the middle. In both cases, the old content of VS kits was left over and causing all sorts of incompatibilities.
The fix is to move the scanning for kits from earlier during the activation time to post workspace open, at configuration time.
Even if the kits are properly scanned, things don't all work well until the next configure (properties like architecture of VS kit or platform of generator are still the old ones from the cache). With configureOnOpen true, this is not a problem because the migration triggers a reload and a configuration is performed. I added a warning for configureOnOpen false to notify the user that a configuration is needed to be performed, since I didn't feel comfortable to override the meaning of configureOnOpen false for the user.
Bumped kits version to 2, even if there is no definition change between 1.4.0 and 1.4.1, to ensure the 1.4.1 upgrade will complete the rescanning started with 1.4.0.

Another fix is regarding any change in the generators space: either a new preferred generator of a kit, or a new cmake.generator or a change in cmake.preferredGenerators or (like it happens for 1.3.1-->1.4.0) a change in the defaulting/deducing mechanism. Such changes require deletion of the build folder and a reload. Only a reload or only a clean configure or only a delete build folder followed by configuring would not update the generator being used.
The way this impacts the migration to 1.4.0 is that 1.3.1 would wrongly select Ninja in some circumstances, while 1.4.0 fixes that, fact which is equivalent to a generator change. Things may not work according to the user expectations now since Ninja and VS generators are different in many aspects.
The fix in cmfileapi-driver.ts is making it possible now to reflect a generator change only by a reload + configure, no deletion of the build folder. Ideally, such generator change should be considered without a project reload, but that is for future (GitHub issue to be opened for future releases).
I did not investigate the same for server or legacy communication modes since fileApi is preferred.


src/extension.ts
```diff
@@ -421,6 +425,12 @@ class ExtensionManager implements vscode.Disposable {
         return;
       }
       await cmt.configure();
+    } else if (silentScanForKitsNeeded) {
+      // This warning will show up the first time after deciding not to configure, if a version change has been detected
+      // in the kits definition. This may happen during a CMake Tools extension upgrade.
+      // The warning is emitted only once because scanForKitsIfNeeded returns true only once after such change,
+      // being tied to global state variable.
+      log.warning(localize('configure.recommended', 'It is recommended to reconfigure after upgrading to a new kits definition.'));
```
bobbrow, line 433: Can we make this a popup message with a button to perform the scan immediately?

PR #1286, opened 2020-06-05
Fix #1285 - ninjaPath is undefined but tested against null

### LINKED ISSUE #1285: CMake build fails with 'The "path" argument must be of type string'.
### Brief Issue Summary
CMake build fails with 'The "path" argument must be of type string'.

### Apparent Behavior:
I either push Build button on the toolbar or start it via CMake: Build or via F7 and it fails with an exception.

### CMake Tools Log
`[rollbar] Unhandled exception: Unhandled exception Promise: build TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined {}`

### Platform and Versions
- **Operating System**: Windows 10 Home
- **CMake Version**: ? (didn't install separately, can't invoke via cmd)
- **VSCode Version**: 1.45.1 (user setup)
- **CMake Tools Extension Version**: 1.4.0
- **Compiler/Toolchain**: Visual Studio Community 2017 Release - amd64_x86

### Other Notes/Information
The problem is in `dist/main.js` file at the line 5620: it checks `null !== paths_1.default.ninjaPath` while `paths_1.default.ninjaPath` is `undefined`. 
I replaced
`if (null !== paths_1.default.ninjaPath) {`
with
`if (null !== paths_1.default.ninjaPath && undefined !== paths_1.default.ninjaPath) {`
and now it works fine.

bobbrow:
## This change addresses item #1285 

There is a chance that the ninja path is undefined, but the Kits logic expects a null. This change cleans up the types so that we only have to test for undefined.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -703,11 +703,11 @@ async function varsForVSInstallation(inst: VSInstallation, hostArch: string, tar
     variables.set('CC', 'cl.exe');
     variables.set('CXX', 'cl.exe');
 
-    if (null !== paths.ninjaPath) {
+    if (paths.ninjaPath) {
       let envPATH = variables.get('PATH');
       if (undefined !== envPATH) {
         const env_paths = envPATH.split(';');
-        const ninja_path = path.dirname(paths.ninjaPath);
+        const ninja_path = path.dirname(paths.ninjaPath!);
         const ninja_base_path = env_paths.find(path_el => path_el === ninja_path);
         if (undefined === ninja_base_path) {
           envPATH = envPATH.concat(';' + ninja_path);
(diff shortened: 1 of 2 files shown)


src/kit.ts
```diff
@@ -703,11 +703,11 @@ async function varsForVSInstallation(inst: VSInstallation, hostArch: string, tar
...
     variables.set('CXX', 'cl.exe');
 
-    if (null !== paths.ninjaPath) {
+    if (paths.ninjaPath) {
       let envPATH = variables.get('PATH');
       if (undefined !== envPATH) {
         const env_paths = envPATH.split(';');
-        const ninja_path = path.dirname(paths.ninjaPath);
+        const ninja_path = path.dirname(paths.ninjaPath!);
         const ninja_base_path = env_paths.find(path_el => path_el === ninja_path);
```
sean-mcmanus, line 711: What does that do?

bobbrow: It strips `undefined` from the type.

Issue #1269, opened 2020-05-28
Partial CMT activation configuration for non CMake workspace folder not persistent


kurthardin:
### Brief Issue Summary

When connecting to a remote multi-root workspace where one folder doesn't contain a CMakeLists.txt file, clicking the "Don't show again" button on the "CMakeLists.txt was not found in the root of the folder..." notification does not persist, i.e. the notification appears each time the workspace is opened.

### Expected:

1. Connect to remote development environment via remote-ssh extension.
2. Open multi-root workspace with one root folder that doesn't contain a CMakeLists.txt file.
3. Click "Don't show again" button on the "CMakeLists.txt was not found in the root of the folder..." notification.
4. Close and re-open workspace.
5. "CMakeLists.txt was not found..." notification is not shown again.

### Apparent Behavior:

1. Connect to remote development environment via remote-ssh extension.
2. Open multi-root workspace with one root folder that doesn't contain a CMakeLists.txt file.
3. Click "Don't show again" button on the "CMakeLists.txt was not found in the root of the folder..." notification.
4. Close and re-open workspace.
5. "CMakeLists.txt was not found..." notification is shown again.

### Platform and Versions

- **Operating System**: Windows 10 (local), Ubuntu 19.10 (remote)
- **CMake Version**: cmake version 3.14.5
- **VSCode Version**: 1.45.1
- **CMake Tools Extension Version**: 1.4.0
- **Compiler/Toolchain**: 8.0.0-3

andreeis:
This should be fixed in CMake Tools 1.4.1 which was published today.
Let us know if you encounter any issues with this release.

Issue #1259, opened 2020-05-22
[Suggestion] Allow disable automatic CMake configure when saving CMakeLists.txt


tt4g:
This pull request #1196 added new feature to automatic configure CMake when saving `CMakeLists.txt`.

This is a great feature that integrates CMake projects with IntelliSense.
However, there were things I wanted to turn off this feature.
Specifically, when working on a project that has multiple `CMakeLists.txt`, running CMake configure every time `CMakeLists.txt` is saved will hinder my work.

Imagine a project with the following structure:

```plaintext
.
+-- include
|   +-- CMakeLists.txt // Add `foo.hpp`, etc header files to project.
|   +-- foo.hpp
+-- src
|   +-- CMakeLists.txt // Add `cpp.hpp`, etc source files to project.
|   +-- foo.cpp
+-- CMakeLists.txt // Calls `add_subdirectory(include)` and `add_subdirectory(src)` as aggregators.
```

My project has many source and header files, I have divided the project source file declarations in this way.

This project is using Qt. The problem is that Qt must perform the task of scanning the source files and generating additional source files (Qt MOC sources) every time CMake configure time.
I have to update the `CMakeLists.txt` twice to add the header and source files.
CMake configure starts when the first `CMakeLists.txt` is saved, and it also starts scanning the source.
I can't work on it until I've finished scanning the source.

Could you add an option to turn off the this feature to get around the problem?
I'm sure there are others besides me who would like to turn this feature off for projects that perform time-consuming tasks in the CMake configure step.

GaskoCoding:
+1

wutiantong:
+1

fatfatson:
+1
I'm working on a workspace and there's only one folder has CMakeList.txt, everytime I modify&save it, vscode always shows error about something like "no CMakeList in other folder"

![image](https://user-images.githubusercontent.com/5842874/83507071-9cf06180-a4fa-11ea-9760-a85ed91c8ae8.png)

even though I click the "Don't show again", it still appear everytime.

bobbrow:
@fatfatson we should fix that, but did you try clicking the "Locate" button to tell the extension where the CMakeLists.txt file is?  Or you can also set the `cmake.sourceDirectory` setting manually.

bobbrow:
[...]

PR #1295, opened 2020-06-09
Partial/full activation mode and multi root. Disable automatic configures.

andreeis:
Introduce cmake.configureOnEdit to control the automatic configures triggered by editing CMakeLists.txt or cmake.sourceDirectory.

Refactored and fixed the whole partial/full activation modes in the multi root scenario.

This addresses GitHub issues:
https://github.com/microsoft/vscode-cmake-tools/issues/1259
https://github.com/microsoft/vscode-cmake-tools/issues/1269


package.json
```diff
@@ -1249,6 +1249,12 @@
           "description": "%cmake-tools.configuration.cmake.configureOnOpen.description%",
           "scope": "resource"
         },
+        "cmake.configureOnEdit": {
+          "type": "boolean",
+          "default": true,
+          "description": "%cmake-tools.configuration.cmake.configureOnOpen.description%",
```
bobbrow, line 1255: typo: `configureOnEdit`

src/cmake-tools.ts
```diff
@@ -271,8 +271,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
           const cmakeListsFile = await vscode.window.showOpenDialog(openOpts);
           if (cmakeListsFile) {
             const fullPathDir: string = path.parse(cmakeListsFile[0].fsPath).dir;
-            const relPathDir: string = path.relative(this.folder.uri.fsPath, fullPathDir);
-            vscode.workspace.getConfiguration('cmake').update("sourceDirectory", path.join("${workspaceFolder}", relPathDir));
+            const relPathDir: string = lightNormalizePath(path.relative(this.folder.uri.fsPath, fullPathDir));
```
bobbrow, line 274: Does this give you `../folder` for folders outside of `${workspaceFolder}`?

src/cmake-tools.ts
```diff
@@ -271,8 +271,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+            const drv = await this.getCMakeDriverInstance();
+            if (drv) {
+              const joinedPath = "${workspaceFolder}/".concat(relPathDir);
```
bobbrow, line 277: `path.join` is probably better to use than `.concat`

src/cmake-tools.ts
```diff
@@ -271,8 +271,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+              vscode.workspace.getConfiguration('cmake', this.folder.uri).update("sourceDirectory", joinedPath);
+              drv.config.updatePartial({sourceDirectory: joinedPath});
+            } else {
+              throw new Error(localize('unable.to.save.sourceDirectory', 'Unable to save "cmake.sourceDirectory". There is no valid cmake driver instance.'));
```
bobbrow, line 281: You don't need a `drv` to set `cmake.sourceDirectory`, so I think you should set it unconditionally. In that case you could just instruct the user to reload the window and the setting should take effect.  E.g. "Something went wrong while setting the cmake.sourceDirectory setting. Please run the "Reload Window" command for the change to take effect") Or give them a button to run the command automatically.

src/cmake-tools.ts
```diff
@@ -445,14 +452,22 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
       const str = td.uri.fsPath;
-      if (str.endsWith("CMakeLists.txt")) {
-        // Reset to a full activation mode.
-        // If required, the extension will switch back to partial activation mode
-        // after more analysis triggered by the below configure.
-        await enableFullFeatureSet(true);
-
-        log.debug(localize('cmakelists.save.trigger.reconfigure', "We are saving a CMakeLists.txt file, attempting automatic reconfigure..."));
-        await this.configure([], ConfigureType.Normal);
+      if (str === path.join(this.folder.uri.fsPath, "CMakeLists.txt")) {
```
bobbrow, line 455: Is this a multiroot problem that you're trying to fix?  If so, then I think you need to look at the `cmake.sourceDirectory` setting and look if the CMakeLists.txt is under that folder instead. 

Technically, I think that CMakeLists can reference folders anywhere in the filesystem, so that's not a 100% fix, but maybe addresses 99% of cases.

src/cmake-tools.ts
```diff
@@ -445,14 +452,22 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+        // should be full or partial, so there is no need to proactively enable full here,
+        // unless the automatic configure is disabled.
+        // If there is a configure or a build in progress, we should avoid setting full activation here,
+        // even if cmake.configureOnEdit is true, because this may overwrite a different decision
+        // that was done earlier by that ongoing configure process.
+        const drv = await this.getCMakeDriverInstance();
+        if (drv && !drv.configOrBuildInProgress()) {
+          if (drv.config.configureOnEdit) {
+            log.debug(localize('cmakelists.save.trigger.reconfigure', "Detected saving of CMakeLists.txt, attempting automatic reconfigure..."));
+            await this.configure([], ConfigureType.Normal);
```
bobbrow, line 466: Does `configure` enable the feature set?

src/cmake-tools.ts
```diff
@@ -445,14 +452,22 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+          } else {
+            await enableFullFeatureSet(true, this.folder);
+          }
+        }
```
bobbrow, line 470: I think for the else of `configOrBuildInProgress` you should log a message that the user needs to reconfigure and build again.

src/drivers/driver.ts
```diff
@@ -634,6 +639,15 @@ export abstract class CMakeDriver implements vscode.Disposable {
         return -1;
       }
 
+      // _beforeConfigureOrBuild contains the only scenario when we would want partial features set
```
bobbrow, line 642: If this is the case, then should the call to enable the full feature set go there instead? This would also give you the opportunity to turn it off if the conditions started to fail for some reason.

src/drivers/cmfileapi-driver.ts
```diff
@@ -97,6 +102,26 @@ export class CMakeFileApiDriver extends codemodel.CodeModelDriver {
       log.debug(`Reload CMake cache: ${this.cachePath} changed`);
       rollbar.invokeAsync('Reloading CMake Cache', () => this.updateCodeModel());
     });
+
+    this.config.onChange('sourceDirectory', async () => {
```
bobbrow, line 106: Should this config listener be enabled for all drivers (not just file api)?

src/drivers/cmfileapi-driver.ts
```diff
@@ -97,6 +102,26 @@ export class CMakeFileApiDriver extends codemodel.CodeModelDriver {
...
+      // unless the automatic configure is disabled.
+      // If there is a configure or a build in progress, we should avoid setting full activation here,
+      // even if cmake.configureOnEdit is true, because this may overwrite a different decision
+      // that was done earlier by that ongoing configure process.
+      if (!this.configOrBuildInProgress()) {
+        if (this.config.configureOnEdit) {
+          log.debug(localize('cmakelists.save.trigger.reconfigure', "Detected 'cmake.sourceDirectory' setting update, attempting automatic reconfigure..."));
+          await this.configure([]);
+        } else if (this.workspaceFolder) {
+          const folder = vscode.workspace.getWorkspaceFolder(vscode.Uri.file((this.workspaceFolder)));
```
bobbrow, line 118: extra parenthesis in the function call

src/extension.ts
```diff
@@ -182,30 +174,24 @@ class ExtensionManager implements vscode.Disposable {
...
+  // The context variable "cmake:enableFullFeatureSet" (which controls
+  // all the cmake commands and UI elements) is set to true,
+  // if there is at least one folder with full features set enabled.
+  // We need to add this private _enableFullFeaturesSetOnWorkspace here
+  // because currently there is no way of reading a context variable
+  // like cmake:enableFullFeatureSet, to apply the OR operation on it.
+  private _enableFullFeatureSetOnWorkspace = false;
+  public enableFullFeatureSet(fullFeatureSet: boolean, folder: vscode.WorkspaceFolder) {
+    this.getFolderContext(folder).ignoreCMakeListsMissing = !fullFeatureSet;
+    this._enableFullFeatureSetOnWorkspace = this._enableFullFeatureSetOnWorkspace || fullFeatureSet;
```
bobbrow, line 192: I find this pattern confusing. You need to set `_enableFullFeatureSetOnWorkspace` to `false` outside of this method because `fullFeatureSet` can't change the true to false. Is there a reason you can't just take the value of `fullFeatureSet`?

src/cmake-tools.ts
bobbrow: We also support `sourceDirectory` with `CMakeLists.txt` included in the path so you should account for that here.

src/cmake-tools.ts
```diff
@@ -445,14 +458,27 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+        if (drv && !drv.configOrBuildInProgress()) {
+          if (drv.config.configureOnEdit) {
+            log.debug(localize('cmakelists.save.trigger.reconfigure', "Detected saving of CMakeLists.txt, attempting automatic reconfigure..."));
+            await this.configure([], ConfigureType.Normal);
+          } else {
+            await enableFullFeatureSet(true, this.folder);
+          }
+        } else {
+          log.warning(localize('cmakelists.save.could.not.reconfigure',
+           'Detected saving of CMakeLists.txt but could not configure because there is no valid driver or the project is already building/configuring.'));
```
bobbrow, line 479: >Detected saving of CMakeLists.tx [](start = 12, length = 32)

I would reword it like this:

"Changes were detected in CMakeLists.txt but we could not reconfigure the project because another operation is already in progress."

src/cmake-tools.ts
andreeis: Good catch. It made me remember https://github.com/microsoft/vscode-cmake-tools/pull/1016/files and I actually debugged and I confirm that CMakeLists.txt gets removed from sourceDirectory before the CMakeLists path check. Do you think there is a different scenario where actually sourceDirectory would end with CMakeLists.txt once we hit the file content change listener?

andreeis:
I am looking at the failed tests. Again, "electron: Loading non context-aware native modules in the renderer process is deprecated...". I fixed this before with another PR a few weeks or months ago. I had to move a particular call somewhere else.

bobbrow:
>     return cache.get('CMAKE_GENERATOR')!.value;

can you use `?.` and make the return type optional?  I think you only compare this value to an optional.

(I like to avoid using `!.` where ever possible)

---
Refers to: src/drivers/cmfileapi-driver.ts:72 in 27bee85. [](commit_id = 27bee8554bc2ea229f5c31ab01b3e91252536d29, deletion_comment = False)

bobbrow:
>     return cache.get('CMAKE_GENERATOR')!.value;

Oh, it looks like this was merged in from another PR.  You can disregard.

---
In reply to: [642980937](https://github.com/Microsoft/vscode-cmake-tools/pull/1295#issuecomment-642980937) [](ancestors = 642980937)

---
Refers to: src/drivers/cmfileapi-driver.ts:72 in 159e0e6. [](commit_id = 159e0e6886bc9880c35715267ed656b78a623e6d, deletion_comment = False)

src/extension.ts
bobbrow: Do we ever need to turn support off?  I'm wondering if we should just leave the full feature set enabled once it's been enabled.

bobbrow: If you remove the state that would make CMake Tools enable the full feature set, then you would need to reload the window at that point to disable it.  I would be fine with that behavior.

andreeis:
The only fix needed for the Electron issue was to not call enableFullFeatureSet from driver.ts. To still achieve the same functionality (we want full activation when beforeConfigureOrBuild returns true and partial only for missing CMakeLists.txt), I introduced a -2 error code (which wasn't used already, only -1 was returned on any failure path) to recognize later that we have a missing CMakeLists.txt, thus calling enableFullFeatureSet in a place where Electron doesn't complain.

The reason for commenting out the sourceDirectory listeners in the driver (the previous commit) was that I had infrastructure issues on my Ubuntu and with no amount of cleanup (not even fresh cloning of CMake Tools) I wasn't able to get rid of the Electron test issue so I just made some experiments to leverage the GitHub CI.

PR #1297, opened 2020-06-09
Fix debugging on MacOS

andreeis:
When no miDebuggerPath is given via the cmake.debugConfig in settings, try to locate the lldb-mi that gets installed with the CppTools extension.
In case that is not found, continue with the logic that was already implemented: search for lldb-mi, then gdb, then lldb in the location of the compiler.


andreeis:
Ignore the changes in extension.ts. I don't know how those show up. That's why I closed my previous PR and started again, fresh and made sure that I synced to develop, made a new branch, edited the fix, git status showed only debugger.ts being modified. Then in the PR I see that change in extension.ts. I'll figure it out. To review it's only the change in debugger.ts.

src/debugger.ts
```diff
@@ -150,8 +152,12 @@ export async function getDebugConfigurationFromCache(cache: CMakeCache, target:
 
   if (!debuggerPathOverride) {
     const clang_compiler_regex = /(clang[\+]{0,2})+(?!-cl)/gi;
-    // Look for lldb-mi
-    let clang_debugger_path = compiler_path.replace(clang_compiler_regex, 'lldb-mi');
+    // Look for lldb-mi where the CppTools extension usually installs it
+    // and if not found, try also the compiler path.
+    const cpptoolsExtension = vscode.extensions.getExtension('ms-vscode.cpptools');
+    let clang_debugger_path = cpptoolsExtension ?
```
bobbrow, line 158: I think the test should be procedural.  Try to see if there is a lldb-mi that matches the compiler.  If not, then try to grab the one from cpptools as a last resort.

bobbrow:
It looks like you based this change off of the wrong branch instead of starting from develop.

> Forgot one more scenario for when we want partial feature set enabled  Commit: 
158799f

src/debugger.ts
andreeis: Ok, makes sense. What's your opinion on the searching for gdb and lldb (not -mi)? Anything you think doesn't make sense now in the search logic and I can remove it now since I'm at it?

Issue #1302, opened 2020-06-11
Multiple silent kits scanning in multi-root workspace


andreeis:
When opening a multi-root project, if there is a breaking change in the kits definition, the extension will trigger as many silent scans as the number of projects.

andreeis:
This should be fixed in CMake Tools 1.4.1 which was published today.
Let us know if you encounter any issues with this release.

PR #1303, opened 2020-06-11
Fix multiple silent kit scans in a multi-root workspace

andreeis:
Since the silent kits scanning is called at the end of opening a folder, a multi-root project will trigger unnecessarily more than one kits scanning.
The fix in this PR is to check if there is already a scan in progress but that was not enough because for a multi root project, the multiple scans are first triggered but the boolean gets set later and we still have more than one scanning processes, depending on various timings.
In addition to check for scans in progress, the call to silent scan is made only for the first folder.
Another alternative would be to move the silent scan earlier in the extension activation (or outside the scope of a folder) since this operation has nothing to do with a folder specifically. I remembered the issues with the first approach and thought that checking first folder only is not that bad.

https://github.com/microsoft/vscode-cmake-tools/issues/1302

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -365,8 +365,9 @@ class ExtensionManager implements vscode.Disposable {
 
     // Silent re-scan when detecting a breaking change in the kits definition.
     // Do this only for the first folder, to avoid multiple rescans taking place in a multi-root workspace.
-    const silentScanForKitsNeeded: boolean = vscode.workspace.workspaceFolders![0] === cmt.folder &&
-                                            await scanForKitsIfNeeded(cmt.extensionContext);
+    const silentScanForKitsNeeded: boolean = vscode.workspace.workspaceFolders !== undefined &&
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -364,7 +364,9 @@ class ExtensionManager implements vscode.Disposable {
     const cmt = info.cmakeTools;
 
     // Silent re-scan when detecting a breaking change in the kits definition.
-    const silentScanForKitsNeeded: boolean = await scanForKitsIfNeeded(cmt.extensionContext);
+    // Do this only for the first folder, to avoid multiple rescans taking place in a multi-root workspace.
+    const silentScanForKitsNeeded: boolean = vscode.workspace.workspaceFolders![0] === cmt.folder &&
```
bobbrow, line 368: You should always check whether the `workspaceFolders` is set or not before accessing the folders.  There are some cases where it can be undefined.

bobbrow:
#1302

PR #1315, opened 2020-06-17
Fix in how lldb-mi integrates with the debugger detection logic

andreeis:
The lldb-mi integration with the debugger detection logic was buggy: it was missing the regex check for the clang compiler.
Also fix the expectations of a unit test regarding gdb fallback, when CppTools is installed.

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -165,17 +165,17 @@ export async function getDebugConfigurationFromCache(cache: CMakeCache, target: 
         // 1b. lldb-mi installed by CppTools
         return createLLDBDebugConfiguration(cpptoolsDebuggerPath, target);
       }
+    }
+
+    // 2. gdb in the compiler path
+    clang_debugger_path = compiler_path.replace(clang_compiler_regex, 'gdb');
+    if ((clang_debugger_path.search(new RegExp('gdb')) != -1) && await checkDebugger(clang_debugger_path)) {
+      return createGDBDebugConfiguration(clang_debugger_path, target);
     } else {
-      // 2. gdb in the compiler path
-      clang_debugger_path = compiler_path.replace(clang_compiler_regex, 'gdb');
-      if ((clang_debugger_path.search(new RegExp('gdb')) != -1) && await checkDebugger(clang_debugger_path)) {
-        return createGDBDebugConfiguration(clang_debugger_path, target);
-      } else {
-        // 3. lldb in the compiler path
-        clang_debugger_path = compiler_path.replace(clang_compiler_regex, 'lldb');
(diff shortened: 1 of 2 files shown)


andreeis:
Something wrong in the debugger.ts fix causes the test to fail. Fixing it. Will send update.

src/debugger.ts
```diff
@@ -151,28 +151,31 @@ export async function getDebugConfigurationFromCache(cache: CMakeCache, target:
...
-    } else if (cpptoolsDebuggerPath && await checkDebugger(cpptoolsDebuggerPath)) {
-      // 2. lldb-mi installed by CppTools
-      return createLLDBDebugConfiguration(cpptoolsDebuggerPath, target);
+    if ((clang_debugger_path.search(new RegExp('lldb-mi')) != -1)) {
+      const cpptoolsExtension = vscode.extensions.getExtension('ms-vscode.cpptools');
+      const cpptoolsDebuggerPath = cpptoolsExtension ? path.join(cpptoolsExtension.extensionPath, "debugAdapters", "lldb-mi", "bin", "lldb-mi") : undefined;
+        // 1a. lldb-mi in the compiler path
+      if (await checkDebugger(clang_debugger_path)) {
+        return createLLDBDebugConfiguration(clang_debugger_path, target);
+      } else if (cpptoolsDebuggerPath && await checkDebugger(cpptoolsDebuggerPath)) {
```
bobbrow, line 164: It might improve code readability if we remove the `else` keywords in this function.  Each block returns early, so the `else` is not needed.  (not blocking, you can submit without this).

src/debugger.ts
```diff
@@ -151,28 +151,31 @@ export async function getDebugConfigurationFromCache(cache: CMakeCache, target:
   }
 
   if (!debuggerPathOverride) {
-    const clang_compiler_regex = /(clang[\+]{0,2})+(?!-cl)/gi;
     // Look for a debugger, in the following order:
-    const cpptoolsExtension = vscode.extensions.getExtension('ms-vscode.cpptools');
-    const cpptoolsDebuggerPath = cpptoolsExtension ? path.join(cpptoolsExtension.extensionPath, "debugAdapters", "lldb-mi", "bin", "lldb-mi") : undefined;
+    // 1. LLDB-MI
+    const clang_compiler_regex = /(clang[\+]{0,2})+(?!-cl)/gi;
     let clang_debugger_path = compiler_path.replace(clang_compiler_regex, 'lldb-mi');
```
bobbrow, line 157: You might want to rename this to `mi_debugger_path` since it covers `lldb-mi`, `gdb`, and `lldb`  (not blocking, you can submit without this)

src/debugger.ts
andreeis: Both good ideas and were quick for me to do now.

PR #1337, opened 2020-06-29
Add vendor hostOs targetOs targetArch versionMajor versionMinor attributes for kit.

### LINKED ISSUE #462: Question? How to use kits in     "cmake.buildDirectory"
I want to use something like this to building for different compiler & tartget : "${workspaceRoot}/build/${buildType}-$(buildKit}",

such as Visual Studio 2017 amd 64& Visual Studio 2015 x86 & LLVM Cross for Linux and so on.

lygstate:
Add vendor hostOs targetOs targetArch versionMajor versionMinor attributes for kit.
Fixes #462


bobbrow:
Thank you for your contribution.  One quick question.  Is it possible to query these attributes at the point where the kit is selected instead of modifying the definition of a Kit?  Changing the Kit definition on disk is disruptive for people and I haven't reviewed this PR extensively enough to determine whether it is actually required.

lygstate:
I am not sure if that possible, cause it's need the version info of the compiler, and that was detected at the scan procedure

lygstate:
Where does the travis result gone.

bobbrow:
The detection just invokes the compiler with `-v` and reads the stderr. I think that could be done when the kit is selected.

lygstate:
> The detection just invokes the compiler with `-v` and reads the stderr. I think that could be done when the kit is selected.

OK, I'll try it:)

lygstate:
@bobbrow done, wish you have time to do a review. I add a function getKitExpand to be called in _setKit

lygstate:
CI not triggered again.

lygstate:
> The detection just invokes the compiler with `-v` and reads the stderr. I think that could be done when the kit is selected.

I've modify the patch with you suggestion, need feedback, thanks

lygstate:
@bobbrow  ping for merge

andreeis:
We apologize for taking so long to finalize a review of this change. We are wrapping up another project. Thank you for your patience while we catch up with the PRs and high priority bugs.

andreeis (approved):

PR #1343, opened 2020-07-01
Allow minor version of File API protocol to be greater than expected

KyleFromKitware:
Minor versions of the File API protocol are backwards compatible -
they only add new fields, and don't remove or change the meaning of
existing fields. Therefore, it is safe for the minor version to be
greater than what we expect.


bobbrow:
Thank you @KyleFromKitware for the quick response!

PR #1348, opened 2020-07-01
Error objects are not automatically converted to strings

bobbrow:
The `localize` function does not automatically convert objects to strings.  If an object is passed in, the `{0}` string will remain in the final output.  This change identifies a few invocations of `localize` that are attempting to output error information and converts those errors to the correct format before passing them in as arguments.


src/cmake-tools.ts
```diff
@@ -650,14 +650,14 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+          compdb_path, expanded_dest, pardir, e.toString()));
         return;
       }
       try {
         await fs.copyFile(compdb_path, expanded_dest);
       } catch (e) {
         // Just display the error. It's the best we can do.
-        vscode.window.showErrorMessage(localize('failed.to.copy', 'Failed to copy "{0}" to "{1}": {2}', compdb_path, expanded_dest, e));
+        vscode.window.showErrorMessage(localize('failed.to.copy', 'Failed to copy "{0}" to "{1}": {2}', compdb_path, expanded_dest, e.toString()));
         return;
```
andreeis, line 661: any reason not to call errorToString for the 2 instances in this code fragment and only e.toString()?
All looks good otherwise.

bobbrow: `errorToString` is formatted to be useful for the Output window. These instances will show up in a popup window so we just want the error message here, not the full stack.

PR #1355, opened 2020-07-07
Implement cmake:hideBuildCommand context option.

tritao:

- New context option `cmake:hideBuildCommand` to hide the build command and keybindings.

This works like the already existing `cmake:hideDebugCommand`, and just does the same for the build actions.

## The purpose of this change

This change allows other extensions to provide a custom F7 keybinding while the CMake extension is loaded.


tritao:
Is anyone able to take a look at these changes? Thanks

andreeis (approved):

Issue #794, opened 2019-09-25
 [loc][Query][vscode-cmake-tools] Localizability of 0; "package.cmake-tools.configuration.cmake.enabledOutputParsers.description"


PDostalek:
Hello developers,
please help localization team to answer question from translator.
In case translator is not sure about context/meaning/localizability of string we raise a question.

Source:
Output parsers to use. Supported parsers 'cmake', 'gcc', 'gnuld' for GNULD-style inker output, 'msvc' for Microsoft Visual C++, and 'ghs' for the Green Hills compiler with --no_wrap_diagnostics --brief_diagnostics

File Name:
Src\vscode-extensions\vscode-cmake-tools.xlf

Query:
Please confirm to what "inker" refers to here.

bobbrow:
it is a typo.  It should be "linker"

PDostalek:
thx, closing

PR #1363, opened 2020-07-10
Fix i18n Typo (See #794)

SE2Dev:
### This changes a typo in the `cmake-tools.configuration.cmake.enabledOutputParsers.description` localization entry.

| i18n    | Original Term | Proposed Term |
| ------- | ------------- | ------------- |
| **nls** | inker         | linker        |
| **chs** | inker         | ~~链接器~~ (Handled by i18n team)       |
| **trk** | inker         | ~~bağlayıcı~~ (Handled by i18n team)    |

## The purpose of this change

This is a typo and can (and has) affected various i18n files.

## Other Notes/Information

I'm not 100% sure that the correct terms were used for the `chs` and `trk` localization files as I don't speak either of those languages; I was able to confirm that the terms were used by native speakers in various forums, etc., but it would probably be best to have someone double check them.

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -67,7 +67,7 @@
     "cmake-tools.configuration.cmake.ctestPath.description": "Path to CTest executable. If null, will be inferred from cmake.cmakePath (recommended to leave null).",
     "cmake-tools.configuration.cmake.ctest.parallelJobs.description": "The number of parallel test jobs. Use zero to use the value of cmake.parallelJobs.",
     "cmake-tools.configuration.cmake.parseBuildDiagnostics.description": "Parse compiler output for warnings and errors.",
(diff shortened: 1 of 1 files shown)


bobbrow:
Thank you!
If you just submit the typo fix to the .nls.json file, our i18n team will handle the rest.  Those language-specific files are auto-generated by an internal process that we have.

SE2Dev:
I've removed the changes to the other files, and only kept the change to the .nls one.

SE2Dev:
@bobbrow I'm not exactly sure why the CI build is failing on Mac now, given that the only chances I made since the original pull request were reverting 2 lines and force pushing. Other than that, I believe everything should be in order.

bobbrow:
It's not your fault.  There is a transient failure in the tests that we haven't root caused yet.

PR #1393, opened 2020-08-01
Add option to set CMAKE_BUILD_TYPE also on multi config generators

tonka3000:
### This changes behavior
This PR add the option to set `CMAKE_BUILD_TYPE` when a multi config generator like `Visual Studio` is set. I've add a new option named `cmake.setBuildTypeOnMultiConfig`.

* `true`: the `CMAKE_BUILD_TYPE` will be set in multi config generators
* `false` (default): the `CMAKE_BUILD_TYPE` wont be set in multi config generators

So the merge of that PR wont change anything for existing users!

## The purpose of this change
I know many projects which use `CMAKE_BUILD_TYPE` to force single config also on multi config generators, including myself 😄 . A example-code for that would be

```cmake
if(NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
endif()
if(CMAKE_BUILD_TYPE)
  set(CMAKE_CONFIGURATION_TYPES ${CMAKE_BUILD_TYPE})
endif()

set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)

foreach(_config_type ${CMAKE_CONFIGURATION_TYPES})
    string(TOUPPER ${_config_type} _config_type)
    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_${_config_type} "${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}")
    set(CMAKE_LIBRARY_OUTPUT_DIRECTORY_${_config_type} "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
    set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_${_config_type} "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")
endforeach()
```

I see that often when you have a project with multi platform support, were single config generators are the typical use. I used multi config generators like `Visual Studio` for years, but because it does only add complexity to a project without any real benefit (in CI it is much slower because we have to clone always all config and  such kind of problems) I completely get rid of it with the above technique after looking into many bigger open source projects which using cmake and using these technique. I know that this is not the "normal" case in multi config generators since cmake 3.0, so this is some kind of personal preference.

### The problem
The problem is that the build type choosed in the vscode menu is most of the time not in sync with the configured on when some project use the technique as above. This leads to some kind of problems when for example starting the `cmake build` command with f.e. use `Debug` during configure but in the vscode menu it is `Release` and so I get an error with the `Visual Studio` generator because the command line looks like `cmake --build . --config Release` but the config does not exists (thanks to setting CMAKE_CONFIGURATION_TYPES to a single value).

To be able to set the `CMAKE_BUILD_TYPE` is always have to set it in the `settings.json` file, which works, but it leads to the build type not in sync problem pretty fast.

## The solution
By just adding this new option, cmake tools can behave like it do until now, but if someone like me has projects with this type of `CMAKE_BUILD_TYPE` handling the user can just active support for it (in workspace or globally) and that without any kind of risk as far as I can tell. After activating this the user can change the build type normally via the vscode menu from cmake tools.

Greetings

Tonka


tonka3000:
Hey @bobbrow ,

any change to get this merged?

Greetings

Tonka

bobbrow:
I'm sorry that we're behind on PR's right now.  I'm hoping to get all caught up next week during CppCon.

tonka3000:
No problem :+1:

andreeis (approved):

PR #1397, opened 2020-08-05
Set application run directory to executable path

Shatur95:
### This changes run and debugging behavior

- Run and debug directory set to the application directory.

## The purpose of this change

Now, if the application writes files, then it will not write them to the workspace folder. It will also work correctly if it uses shared objects (plugins), which are copied to its folder. This is how it works in most IDE's.

## Other Notes/Information

I applied changed suggested by @alan-wr. But it works for run without debug. It looks like `cwd: '${workspaceFolder}',` not expanded.

[Testing project.zip](https://github.com/microsoft/vscode-cmake-tools/files/5030816/Test.zip)

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -1370,6 +1370,7 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
     }
     const termOptions: vscode.TerminalOptions = {
       name: 'CMake/Launch',
+      cwd: path.dirname(executable.path),
     };
     if (process.platform == 'win32') {
       // Use cmd.exe on Windows
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -60,7 +60,7 @@ async function createGDBDebugConfiguration(debuggerPath: string, target: Executa
     type: 'cppdbg',
     name: `Debug ${target.name}`,
     request: 'launch',
-    cwd: '${workspaceFolder}',
+    cwd: '${fileDirname}',
     args: [],
     MIMode: 'gdb',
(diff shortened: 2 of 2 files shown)


Shatur95:
I figured it out, now works as expected.

bobbrow:
Thanks @Shatur95!  We're finishing up another project this week and are hoping to get back to CMake Tools next week and review your PR.

Shatur95:
Thanks! I'm looking forward to it.

Shatur95:
@bobbrow, could you draft a new release with this change?

bobbrow:
If you want to build your own, the process is pretty quick.
1. Install VSCE: `npm install -g vsce`
2. In the root of the repository, run: `vsce package --yarn -o cmake-tools.vsix `
3. Run the "Extensions: Install From VSIX" command in VS Code and select the cmake-tools.vsix you built in step 2.

Shatur95:
Thanks! Yes, I can build it manually, but just need to do it on every machine.

bobbrow:
You shouldn't need to build it on every machine, but you will need to manually install it on each machine.  Is that what you meant?

Shatur95:
Yes, that's why I suggested draft release with recent fixes.
But I changed my mind, will built it manually until the next release. Thanks for quick response!

bobbrow:
Unfortunately we don't have an insiders program set up for this extension yet.  If we uploaded to the Marketplace it would go to everyone and we're not ready to do that yet.

alan-wr:
It is better to have daily or commit extension binary.

PR #1403, opened 2020-08-07
Do not care about the bat call result

lygstate:
Thirdparty software would cause
call "C:\Program Files (x86)\Microsoft Visual Studio 11.0\VC\vcvarsall.bat" x86
to be called but with error code
and the env are still set properly.
The following are the execute result:
```
configuration might not be installed.

C:\Users\大乖>call "C:\Program Files (x86)\Microsoft Visual Studio 11.0\VC\vcvarsall.bat" x86
'MySQL' 不是内部或外部命令，也不是可运行的程序
或批处理文件。
'MySQL' 不是内部或外部命令，也不是可运行的程序
或批处理文件。
```

- 
- 
-

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -528,18 +528,20 @@ async function collectDevBatVars(devbat: string, args: string[], major_version: 
     `@echo off`,
     `cd /d "%~dp0"`,
     `set "VS${major_version}0COMNTOOLS=${common_dir}"`,
-    `call "${devbat}" ${args.join(' ')} || exit`,
+    `set "INCLUDE="`,
+    `call "${devbat}" ${args.join(' ')}`,
     `cd /d "%~dp0"`, /* Switch back to original drive */
   ];
   for (const envvar of MSVC_ENVIRONMENT_VARIABLES) {
     bat.push(`echo ${envvar} := %${envvar}% >> ${envfname}`);
   }
+  const batContent = bat.join('\r\n');
   const batpath = path.join(paths.tmpDir, batfname);
   const envpath = path.join(paths.tmpDir, envfname);
   try {
     await fs.unlink(envpath);
   } catch (error) {}
-  await fs.writeFile(batpath, bat.join('\r\n'));
+  await fs.writeFile(batpath, batContent);
   const res = await proc.execute(batpath, [], null, {shell: true, silent: true}).result;
   await fs.unlink(batpath);
   const output = (res.stdout) ? res.stdout + (res.stderr || '') : res.stderr;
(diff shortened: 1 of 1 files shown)


src/kit.ts
```diff
@@ -567,7 +569,7 @@ async function collectDevBatVars(devbat: string, args: string[], major_version:
           return acc;
         }, new Map());
   if (vars.get('INCLUDE') === '') {
-    console.log(`Error running ${devbat} ${args.join(' ')}, cannot find INCLUDE`);
+    log.error(`Error running:\n${batContent}\n ${args.join(' ')}, cannot find INCLUDE with ${env}`);
```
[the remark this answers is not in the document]
lygstate: update with suggestion

lygstate:
ping

src/kit.ts
```diff
@@ -552,7 +554,7 @@ async function collectDevBatVars(devbat: string, args: string[], major_version:
   } catch (error) { log.error(error); }
 
   if (!env || env === '') {
-    console.log(`Error running ${devbat} ${args.join(' ')} with:`, output);
+    console.log(`Error running:\n${batContent}\n ${args.join(' ')} with:`, output);
     return;
```
andreeis, line 558: 2 more messages to localize

src/kit.ts
```diff
@@ -539,7 +540,8 @@ async function collectDevBatVars(devbat: string, args: string[], major_version:
   try {
     await fs.unlink(envpath);
   } catch (error) {}
-  await fs.writeFile(batpath, bat.join('\r\n'));
+  const batContent = bat.join('\r\n');
```
andreeis, line 543: Minor: batContent can be set one line earlier, then fs.writeFile can use batContent as second parameter.

src/kit.ts
```diff
@@ -567,7 +569,7 @@ async function collectDevBatVars(devbat: string, args: string[], major_version:
           return acc;
         }, new Map());
   if (vars.get('INCLUDE') === '') {
-    console.log(`Error running ${devbat} ${args.join(' ')}, cannot find INCLUDE`);
+    log.error(localize('script.run.error', 'Error running:\n{0}\nWith args:\n{1}, cannot find INCLUDE with {2}', batContent, args.join(' '), env));
     return;
   }
```
andreeis, line 574: Not sure how readable that message is, because it has very long parts.
I'm thinking that we still need devbat in the message, as it were before, then being followed by the args. Right now, the args are coming after the devbat content.
Then a print of the env is useful, but part of a different message phrase.

src/kit.ts
bobbrow: `console.log` output does not need to be localized.

src/kit.ts
lygstate: Don't understand, batContent  used latter

src/kit.ts
lygstate: Convert to log.error, easier for user to feedback.

src/kit.ts
andreeis: I don't see it now but I thought I saw twice the following: bat.join('\r\n').

andreeis (approved):

Issue #1280, opened 2020-06-03
Cmake build argument notation is wrong ( -- /maxcpucount:14)


VincentD21:
### Brief Issue Summary
It seems that the arguments for the cmake build call can be mixed between Windows and Unix notation. In the current case the 'maxcpucount' argument uses both.
### Expected:

Pressing the build butten will result either in:
`C:\msys64\mingw64\bin\cmake.EXE --build d:/Fortran_Projekte/windows/project/build --config Debug --target ALL_BUILD --maxcpucount 14`

or:

` C:\msys64\mingw64\bin\cmake.EXE --build d:/Fortran_Projekte/windows/project/build --config Debug --target ALL_BUILD /maxcpucount:14`

### Apparent Behavior:

`C:\msys64\mingw64\bin\cmake.EXE --build d:/Fortran_Projekte/windows/project/build --config Debug --target ALL_BUILD -- /maxcpucount:14`

### CMake Tools Log

```
[build] Build wird gestartet.
[proc] Befehl wird ausgeführt: C:\msys64\mingw64\bin\cmake.EXE --build d:/Fortran_Projekte/windows/project/build --config Debug --target ALL_BUILD -- /maxcpucount:14
[build] 
[build] Microsoft Visual Studio 2013, Version 12.0.40629.0.
[build] Copyright (C) Microsoft Corp. Alle Rechte vorbehalten.
[build] 
[build] Ungültige Befehlszeile. Unbekannter Schalter: maxcpucount:14.
[build] 
[build] Syntax:
[build] devenv  [solutionfile | projectfile | anyfile.ext]  [switches]
[build] 
[build] Das erste Argument für "devenv" ist i. d. R. eine Projektmappendatei oder eine Projektdatei.
[build] Sie können jedoch auch eine beliebige andere Datei als erstes Argument angeben, wenn die
[build] Datei automatisch in einem Editor geöffnet werden soll. Wenn Sie eine Projektdatei eingeben, sucht die IDE
[build] in dem Verzeichnis, das der Projektdatei übergeordnet ist,
[build] nach einer SLN-Datei mit dem gleichen Basisnamen wie die Projektdatei. Wenn keine entsprechende SLN-Datei vorhanden ist, sucht die
[build] IDE nach einer einzelnen SLN-Datei, die auf das Projekt verweist. Wenn keine entsprechende einzelne
[...]

PR #1421, opened 2020-08-26
Remove duplicate option start signs (#1280)

dnkbln:
The generator_args function deliver the option with platform depend
option signs.


bobbrow:
Thank you for your PR @dnkbln.  We will take a look as soon as we can.  We've been busy wrapping up some work in our other extensions and we'll be back soon to evaluate this.

andreeis (approved):

PR #1433, opened 2020-09-11
Message must bet set for vscode.Diagnostic in 1.49.0

xisui-MSFT:
Message must bet set for vscode.Diagnostic in 1.49.0. Otherwise there will be an exception with unreadable messages.


bobbrow (approved):

bigstream-valentyn:
Is there workaround for this?

Issue #1341, opened 2020-06-30
[cmakefileapi-parser] Code model version (2.1) of cmake-file-api is unexpected. Expecting (2.0). IntelliSense configuration may be incorrect.


pbaret:
I am trying to build a c++ project on windows with visual studio code and cmake-tools extension.

My versions are 
cmake:  3.18.0-rc1
cmake tools (vscode extension): 1.4.1

When I build, I get this annoying message about codemodel version.

```
[cmakefileapi-parser] Code model version (2.1) of cmake-file-api is unexpected. Expecting (2.0). IntelliSense configuration may be incorrect.
[cmakefileapi-parser] Code model version (2.1) of cmake-file-api is unexpected. Expecting (2.0). IntelliSense configuration may be incorrect.
```

This messages appear at configuration, and also when I build.

It is the exact same problem mentionned by someone else in this stackoverflow question : https://stackoverflow.com/questions/62445673/cmakefileapi-parser-warning-message-about-cmake-in-windows-vscode

One of the comments on stackoverflow suggests to edit the file 
`\build\.cmake\api\v1\reply\codemodel-v2-[???].json` and edit the minor version at the bottom of the file from '1' to '0'.

It does make the messages disappear for build after this changes, but as this file is generated at configuration, the warning comes back everytime you want to run cmake again... 

Also I don't know if this modification breaks anything or not.

**UPDATE: when writing this issue and checking my version of cmake, I figured out I was using a release candidate (for some reason it is the first displayed on cmake website and I did not pay attention when I downloaded it). I installed the latest cmake release (3.17.3) instead, and the problem "disappeared".**

bobbrow:
Thank you for reporting this issue.  I saw it on another developer's machine this week and forgot to open a ticket to fix this.  I looked for documentation from the CMake team about this revision but didn't find any.  I posted a question on their forum to ask about the changes and we'll see what they say.

https://discourse.cmake.org/t/cmake-file-api-code-model-version-2-1/1465

Please note that the log message is just a warning and if the code model hasn't changed substantially in revision 2.1, the message can be safely ignored.

KyleFromKitware:
> if the code model hasn't changed substantially in revision 2.1, the message can be safely ignored.

CMake upstream here, this is a minor update which only adds new fields to the File API output - it won't break older clients. The warning is safe to ignore.

I've opened #1343 to suppress the warning for newer versions.

bobbrow:
[...]

PR #1442, opened 2020-09-17
Allow code model minor version to be higher than expected

KyleFromKitware:
b6884ea5 allowed the cache object minor version to be higher than
expected, but did not do the same for the code model. Fix this.

Fixes: #1341


KyleFromKitware:
This time I tried it myself and verified that it worked :)

bobbrow (approved):

PR #1452, opened 2020-09-23
Preserve focus when executing "CMake: Run Without Debugging"

### LINKED ISSUE #1138: Terminal is focused after running "CMake: Run Without Debugging"
### Brief Issue Summary
When running the `CMake: Run Without Debugging` command, the terminal gets focused after the program has finished its execution.

### Expected:

The extension should not put focus on the terminal or an option should be provided to control this behavior.

### Apparent Behavior:

Focus is put on the terminal and no option is given to control this behavior.

### Platform and Versions

- **Operating System**: `Windows 10`
- **CMake Version**: `3.16.19112601-MSVC_2`
- **VSCode Version**: `1.43.1
fe22a9645b44368865c0ba92e2fb881ff1afce94
x64`
- **CMake Tools Extension Version**: `0.0.17`
- **Compiler/Toolchain**: `Visual Studio 2019 version 16.5`

estshorter:
Resolves #1138.
Prevent terminal from taking focus when executing "CMake: Run Without Debugging".

- Set `preserveFocus` argument  of `Terminal.show()` to `true`.


estshorter:
@bobbrow  Would you review this PR?
This is a one-liner fix and easy to review, I think.

andreeis (approved):

vid512:
Has this fix been merged into production build? I am using ms-vscode.cmake-tools version 1.7.0, and after executing "CMake: Run Without Debugging" (SHIFT+F5), the built-in terminal still grabs focus.

andreeis:
The code is in production. I just tried it and indeed you get focus if you happen to not have the terminal visible already. I don't know if that is a VSCode UI bug or not. But if you are already in the terminal tab window and you have "cmake.buildBeforeDebug" set to "false" (the default is "true") it will not take focus. If you set buildBeforeDebug to true or it's missing (which is also true) then before the launch command is sent to the terminal, a build is triggered which by design shows the "output channel" tab window. When the terminal is made visible again to show how the launch command runs, it captures focus.

PR #1470, opened 2020-10-07
Try to fix tests

andreeis:
Let's see what else fails in the workflow


andreeis:
Merging without review to be able to fix the regressed tests for all the PRs we are currently investigating, so that we are able to detect eventual new failures.

PR #1475, opened 2020-10-08
Upgrade all packages to the newest

lygstate:
Adjust webpack mainFields: to ['main', 'module'] for supporting newest json5.

Upgrade webpack to newest by using env variable BUILD_VSCODE_NLS instead unknown command line parameter `--vscode-nls`

After this patch, the cmake-tools installer size are srhink a lot,
from 4MB -> 1MB


andreeis:
@lygstate, we are catching up with PRs now. This PR looks good but since it's not a small one and also I am not familiar with some of the infrastructure areas I would like to clone and test it locally. I see source code changes besides the version upgrades that are mentioned in the title. I understand the gnu/c 18 to 17 standard names. What about the others? Are there opened reports that would be fixed by this? If not, help us with some repro steps to exercise the changes done here.

lygstate:
> @lygstate, we are catching up with PRs now. This PR looks good but since it's not a small one and also I am not familiar with some of the infrastructure areas I would like to clone and test it locally. I see source code changes besides the version upgrades that are mentioned in the title. I understand the gnu/c 18 to 17 standard names. What about the others? Are there opened reports that would be fixed by this? If not, help us with some repro steps to exercise the changes done here.

main issues comes from webpack/mocha/vscode-cpptools, we may upgrade one by one

andreeis:
No, it's ok, I'll merge this in and do more testing after the fact. If you have time, it would help if you apply the vsix that will be automatically built for this merge and do some testing on your side as well.

lygstate:
> No, it's ok, I'll merge this in and do more testing after the fact. If you have time, it would help if you apply the vsix that will be automatically built for this merge and do some testing on your side as well.

I've already test locally, didn't found issue yet:) I am using it daily life

andreeis (approved):

andreeis:
Wonderful, thank you.

PR #1478, opened 2020-10-09
support non-trivial command-lines for Compile Active File

### LINKED ISSUE #1227: C++ Compile Active File - doesn't work (Win 7)
VSCode Version: 1.44.2
C/C++ Version 0.27.1
OS Version: Windows_NT x64 6.1.7601

Steps to Reproduce:

1. Install Extensions: C/C++; CMake; CMake Tools
2. Select compiler: e.g. Visual Studio Professional 2019 - x86
3. Open and build any suitable CMake project. (This works no problems for me).
4. Open any C++ file in that project, and select "Compile Active File (Ctrl+F7)
5. Terminal window opens with a command prompt but nothing else happens. Active file is not compiled.

I think the issue is limited to Win7. My colleague has Win10 and Compile_Active_File works fine for him. 

I think the problem is that Compile_Active_File is trying to paste a long compile command into the terminal but this is not working. (I found a VSCode issue reported that just pasting long text into terminal does not work)

If I try pasting the command line in manually (1562 chars long) nothing happens. If I paste the command in 2 pieces, it works as expected.

I wondered if a possible fix/workaround would be to send the command to the terminal in smaller chunks.

Note: I initially reported this issue in VSCode repository ([https://github.com/microsoft/vscode/issues/96860]), but they suggested it was a C++ issue.

Trass3r:
In particular macro definitions containing quotes and spaces.
The code was trying to be too clever, splitting the commandline into arguments and re-merging (the split args weren't even used/modified).
Just let the shell do its job.
NB: this doesn't fix https://github.com/microsoft/vscode-cmake-tools/issues/1227#issuecomment-706111287

@bobbrow @andreeis

### DIFF
--- a/src/compdb.ts
+++ b/src/compdb.ts
@@ -1,5 +1,3 @@
-import * as shlex from '@cmt/shlex';
-
 import {createLogger} from './logging';
 import {fs} from './pr';
 import * as util from './util';
@@ -16,7 +14,7 @@ interface BaseCompileCommand {
   output?: string;
 }
 
-interface StringCompileCommand extends BaseCompileCommand {
+export interface StringCompileCommand extends BaseCompileCommand {
   command: string;
 }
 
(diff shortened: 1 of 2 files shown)


Trass3r:
Rebased.

elahehrashedi:
@Trass3r Thank you for your contribution. Some part of this PR has been considered in: https://github.com/microsoft/vscode-cmake-tools/pull/1648. In the mentioned PR, there is a fix for substrings that include ' ' space. I can see your comments that the arguments haven't been used, however, we needed to fix the parsing algorithm to send the correct set of arguments to IntelliSense.
Again, thank you for your contribution.

Trass3r:
This is not just about spaces but more so about arbitrarily complex shell expressions, one case was `-DMACROSTR="\"Text with spaces\""`.

elahehrashedi:
@Trass3r thank you for your explanation. The PR is updated  https://github.com/microsoft/vscode-cmake-tools/pull/1648. In case you are willing to test it, let us know about the test results under that PR.

PR #1488, opened 2020-10-14
Don't rebuild when doing command substitution

bobbrow:


bobbrow:
@dslijepcevic FYI

bobbrow:
Hmm... based on tests, it seems there is an expectation that this command ensures the binary is up to date.  We may need to dig into the history to find out why and/or who might be depending on this behavior.

dslijepcevic:
It's probably only `cmake.buildBeforeRun` that ensures the binary is up to date. When it is turned off, the expectation is gone. CMake Tools should build launch target exactly once when it is set to `true`, regardless of how many times command substitution is performed before launch.

andreeis (dismissed):

bobbrow:
From what I can tell, it seems that the current behavior for launchTargetPath is very much by design so that you can set the `program` in your launch.json to the executable target and have it be built for you automatically when debugging without adding a task dependency. I would argue that the other two commands were not added with this intent and can be modified. To get the behavior you want without breaking others who depend on this feature, we'll need to add a new command for launchTargetPath with a different name that does not ever do a build.

dslijepcevic:
I don't think that current `cmake.buildBeforeRun` behaviour is correct. If you manually specify an executable path for `program` in `launch.json`, CMake Tools will not build it before running even when it knows how to do it (if there is a CMake target with that output file). It seems that this functionality was wrongly implemented in command substitution so that only `${command:cmake.launchTargetPath}` can be built before running.

bobbrow:
I agree, but since the side effect has been around for as long as I can tell, there's no telling who may have taken a dependency on that behavior.  I don't want to take the risk of regressing existing users, so the only solution is to implement a new set of commands with the new behavior.

dslijepcevic:
Can we at least fix `cmake.buildBeforeRun` to build *any* program executable specified in the current launch configuration (i.e. not only those specified with a command substitution) before running it with F5, provided that we know how to build it from the CMake's target database?

bobbrow:
Unfortunately we don't get a notification from VS Code that a user is trying to launch the program unless there is a substitution, so there's no way to do that.

dslijepcevic:
Sure, I understand. While you're adding these new commands, could you please include one for getting the active build target? It's the only thing missing I think. It can be useful to have it in a build task when `cmake.buildTask` set to `true`. Something like `${command:cmake.buildTargetName}`. Thanks!

andreeis (approved):

bobbrow:
You can test out this fix with the VSIX here: https://github.com/microsoft/vscode-cmake-tools/suites/1370492497/artifacts/22440204

Use the "Install from VSIX" command in VS Code to install it.

PR #1490, opened 2020-10-15
Add CMake GUI

bobbrow:
Ported PR #1019 as that PR won't pass tests (and the extension won't activate in production mode) for some reason.  After re-creating the branch, all seems to be in order.

### DIFF
--- a/package.json
+++ b/package.json
@@ -293,6 +293,15 @@
         "title": "%cmake-tools.command.cmake.cleanRebuildAll.title%"
       },
       {
+        "command": "cmake.openConfiguration",
+        "title": "%cmake-tools.command.cmake.openConfiguration.title%",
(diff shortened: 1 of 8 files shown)


bobbrow:
Please double check that the same changes as #1019 are present.  Everything looks fine on my end, but I'd like to be sure.

andreeis (approved):
I tested and it looks good. Thanks very much for your help with this. It was a good idea to stop fixing the tests in the previous state and port the changes over.

bobbrow:
We now build an installable extension with each commit. You can test this out by downloading cmake-tools.vsix from https://github.com/microsoft/vscode-cmake-tools/actions/runs/310550791 and using the "install from vsix" command in VS Code.

Issue #1492, opened 2020-10-16
Visual Studio kits are not detected when temp folder has a space in it


sam20908:
### Brief Issue Summary

CMake Tools is unable to detect Visual Studio kits. It doesn't matter if it's Release or Preview version, because I tried them all. I can use "unspecified" to make it use the Visual Studio kit, but that doesn't really solve the main problem.

### Expected:
CMake Tools should detect Visual Studio kits

1. Try scanning for kits

### Apparent Behavior:

It finds every other compiler EXCEPT Visual Studio (it even finds Clang with MSVC)

### CMake Tools Log

```
[variant] Loaded new set of variants
[kit] Successfully loaded 2 kits from C:\Users\Sam Huang\AppData\Local\CMakeTools\cmake-tools-kits.json
[main] Configuring folder: Super- 
[cmakefileapi-parser] Code model version (2.1) of cmake-file-api is unexpected. Expecting (2.0). IntelliSense configuration may be incorrect.
[cmakefileapi-parser] Code model version (2.1) of cmake-file-api is unexpected. Expecting (2.0). IntelliSense configuration may be incorrect.
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.EXE" --no-warn-unused-cli -DARCANE_BUILD_TEST=TRUE -DCMAKE_TOOLCHAIN_FILE=C:\Development\Arcane\vcpkg\scripts\buildsystems\vcpkg.cmake -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_C_COMPILER:FILEPATH=C:\msys64\mingw64\bin\gcc.exe -DCMAKE_CXX_COMPILER:FILEPATH=C:\msys64\mingw64\bin\g++.exe -Hc:/Development/Super- -Bc:/Development/Super-/build -G Ninja
[cmake] Not searching for unused variables given on the command line.
[cmake] -- Configuring done
[cmake] -- Generating done
[cmake] -- Build files have been written to: C:/Development/Super-/build
[cmakefileapi-parser] Code model version (2.1) of cmake-file-api is unexpected. Expecting (2.0). IntelliSense configuration may be incorrect.
[cmakefileapi-parser] Code model version (2.1) of cmake-file-api is unexpected. Expecting (2.0). IntelliSense configuration may be incorrect.
[...]

PR #1495, opened 2020-10-20
Fix shell script and vcvars devbat when TEMP folder has a space in it

andreeis:
Fix for https://github.com/microsoft/vscode-cmake-tools/issues/1492
Quote and unquote various paths in various circumstances.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -549,14 +549,29 @@ async function collectDevBatVars(devbat: string, args: string[], major_version: 
   for (const envvar of MSVC_ENVIRONMENT_VARIABLES) {
     bat.push(`echo ${envvar} := %${envvar}% >> ${envfname}`);
   }
-  const batContent = bat.join('\r\n');
-  const batpath = path.join(paths.tmpDir, batfname);
-  const envpath = path.join(paths.tmpDir, envfname);
+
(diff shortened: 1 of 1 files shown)


src/kit.ts
```diff
@@ -549,14 +549,29 @@ async function collectDevBatVars(devbat: string, args: string[], major_version:
...
   }
-  const batContent = bat.join('\r\n');
-  const batpath = path.join(paths.tmpDir, batfname);
-  const envpath = path.join(paths.tmpDir, envfname);
+
+  // writeFile and unlink don't need quotes (they work just fine with an unquoted path with space)
+  // but they might fail sometimes if quotes are present, so remove for now any quotes
+  // that may have been defined by the user (the command prompt experience makes it very likely
+  // for the user to use quotes when defining an environment variable with a space containing path).
+  const tmpDir: string = paths.tmpDir?.replace(/"/mg, "");
```
bobbrow, line 557: I think we should only trim quotes at the beginning or end of the string.  Linux lets you put quotes in folder and file names.

bobbrow (changes requested):

src/kit.ts
andreeis: Oh, that's right, good catch, thank you.

bobbrow (approved):

Issue #1498, opened 2020-10-21
Fix the kits file watcher to detect when a .vscode/cmake-kits.json is created for the first time


andreeis:
https://github.com/microsoft/vscode-cmake-tools/pull/1417 is causing various issues: unnecessary reading of kits (as many as 5-6 sometimes), of variants, popups associated with these operations.
That PR was reverted with https://github.com/microsoft/vscode-cmake-tools/pull/1499 and we need to figure out a different way to fix it.

PR #1417, opened 2020-08-22
Tool kit not update after create .vscode/cmake-kits.json on new workspace/folder. 

alan-wr:
### This changes [[visible behavior/performance/documentation/etc.]]

The following changes are proposed:

- 
- 
-

PR #1499, opened 2020-10-21
Audit for immediate effect of settings

andreeis:
Ensure immediate effect for settings: cmakeCommunicationMode, generator and preferredGenerators.
More to be investigated and eventually fixed in a following commit.

Also reverted https://github.com/microsoft/vscode-cmake-tools/pull/1417 since we need to fix it in a different way.
Opened https://github.com/microsoft/vscode-cmake-tools/issues/1498 to track the alternative fix of the above PR.


bobbrow:
There's something really weird with this branch.  It has all of the CMake GUI stuff in it.  Can you start over with a fresh branch off of develop and copy in your last commit?

PR #1500, opened 2020-10-21
Audit for immediate effect of settings

andreeis:
Ensure immediate effect for settings: cmakeCommunicationMode, generator and preferredGenerators.
More to be investigated and eventually fixed in a following commit.

Also reverted #1417 since we need to fix it in a different way.
Opened #1498 to track the alternative fix of the above PR.

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -152,6 +152,21 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
   private readonly _codeModel = new Property<codemodel_api.CodeModelContent|null>(null);
   private _codeModelDriverSub: vscode.Disposable|null = null;
 
+  private readonly _communicationModeSub = this.workspaceContext.config.onChange('cmakeCommunicationMode', () => {
+    log.info(localize('communication.changed.restart.driver', "Restarting the CMake driver after a communication mode change."));
+    return this._reloadCMakeDriver();
+  });
+
+  private readonly _generatorSub = this.workspaceContext.config.onChange('generator', () => {
+    log.info(localize('generator.changed.restart.driver', "Restarting the CMake driver after a generator change."));
+    return this._reloadCMakeDriver();
+  });
+
+  private readonly _preferredGeneratorsSub = this.workspaceContext.config.onChange('preferredGenerators', () => {
+    log.info(localize('preferredGenerator.changed.restart.driver', "Restarting the CMake driver after a preferredGenerators change."));
+    return this._reloadCMakeDriver();
+  });
+
   /**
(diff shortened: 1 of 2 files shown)


src/cmake-tools.ts
```diff
@@ -152,6 +152,21 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
   private readonly _codeModel = new Property<codemodel_api.CodeModelContent|null>(null);
   private _codeModelDriverSub: vscode.Disposable|null = null;
 
+  private readonly _communicationModeSub = this.workspaceContext.config.onChange('cmakeCommunicationMode', () => {
```
bobbrow, line 155: I don't really know what 'sub' means here.  I know it's reused from other places, but it doesn't really make sense to me.  Can you open an issue to rename some of these things?

src/cmake-tools.ts
```diff
@@ -152,6 +152,21 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+  private readonly _generatorSub = this.workspaceContext.config.onChange('generator', () => {
+    log.info(localize('generator.changed.restart.driver', "Restarting the CMake driver after a generator change."));
+    return this._reloadCMakeDriver();
+  });
+
+  private readonly _preferredGeneratorsSub = this.workspaceContext.config.onChange('preferredGenerators', () => {
+    log.info(localize('preferredGenerator.changed.restart.driver', "Restarting the CMake driver after a preferredGenerators change."));
+    return this._reloadCMakeDriver();
+  });
+
```
bobbrow, line 169: I expected to see watchers for configureArgs/Settings.  I noticed there is something in driver.ts that should be doing it, but there are still reports of those config settings not initiating a reload of the driver.  Are you able to repro that at all?

bobbrow (changes requested):

src/cmake-tools.ts
andreeis: No, I saw the reports and I tried but I wasn't able to reproduce. They seemed to update fine for me. I tried windows and linux, fileApi and serverApi. Can you also try to reproduce configureArgs/Settings not getting updated between a settings change and a configure?

src/cmake-tools.ts
andreeis: https://github.com/microsoft/vscode-cmake-tools/issues/1509

src/cmake-tools.ts
bobbrow: It looks like there may be a bug with deleting the whole setting not taking effect, but I'm not going to block this PR on that.  We can open a bug for it.  The important scenario of being able to add settings to a fresh folder appears to be working, so I'll approve.

Issue #1428, opened 2020-09-08
Support CUDA language


Synxis:
### Brief Issue Summary

I have a project that contains some CUDA source files (`*.cu`). I use CMake 3.18, and when configuring the project, I get spurious messages about the c++ standard. Please note that building still works as expected, it is just an issue of cluttered output.

### Apparent Behavior:

Log from CMake (in the Output window), when configuring:

    ...
    [cmake] -- Build files have been written to: /home/...
    [cpptools] Unknown language: -std=c++14
    [cpptools] Unknown language: -std=c++14
    [cpptools] Unknown language: -std=c++14
    [cpptools] Unknown language: -std=c++14
    [cpptools] Unknown language: -std=c++14

Basically, one `[cpptools] Unknown language: -std=c++14` for each CUDA source file.

### Expected:

Configuring CMake should *not* output `[cpptools] Unknown language: -std=c++14`, and especially not once for each source file.

### Platform and Versions

- **Operating System**: Ubuntu 18.04
- **CMake Version**: 3.18.2
- **VSCode Version**: 1.48.2
- **CMake Tools Extension Version**: 1.4.1
- **Compiler/Toolchain**: Clang 10 (*not relevant*)

### Other Notes/Information

I tracked this log to `cpptools.ts`, line 193:

https://github.com/microsoft/vscode-cmake-tools/blob/1def55f2ad3d866a57cfdee257b721911736a987/src/cpptools.ts#L193

Indeed, the code checks for `lang` being `CXX`, `OBJCXX`, `C`, `OBJC`, `undefined` **but not `CUDA`**, which I imagine is the source of the problem.

bobbrow:
Thank you for your feature idea. We may not be able to implement it immediately, but we will track the community reactions to determine where this might fit into our backlog.

Synxis:
@bobbrow Thanks for the answer !

I'm OK with no real support for CUDA until you have more demand for it. **In the mean time, can you just change the error message to either be less annoying, or to disable it completely ?**

This is the current concerned code, around line 193 in `cpptools.ts`: https://github.com/microsoft/vscode-cmake-tools/blob/c398e0b5acfe0b006d0eef94e882686165c6c690/src/cpptools.ts#L166-L195

Please change line 168 from

    if (lang === 'CXX' || lang === 'OBJCXX' ) {

to

    if (lang === 'CXX' || lang === 'OBJCXX' || lang === 'CUDA' ) {

[...]

PR #1502, opened 2020-10-22
Hide "Unknown language: ..." occurring for CUDA source files

Synxis:
**Addresses the useless message "Unknown language..." shown for CUDA source files, cf #1428**

**Details**: when parsing compiler flags, in particular for the C++ standard flag, we need to account for CUDA. According to the nvcc documentation here: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-altering-compiler-linker-behavior-std , nvcc accepts the same C++ standard command line options than gcc/clang. Thus we can simply handle it as for C++, and this avoid getting the message "Unknown language: -std=c++14" (for example) once for each *.cu

**Test**: Here is a sample folder containing `CMakeLists.txt` and `main.cu`: [test-cmake-tools-cuda.zip](https://github.com/microsoft/vscode-cmake-tools/files/5428063/test-cmake-tools-cuda.zip)

### DIFF
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -165,7 +165,7 @@ export function parseCompileFlags(cptVersion: cpt.Version, args: string[], lang?
       extraDefinitions.push(def);
     } else if (value.startsWith('-std=') || lower.startsWith('-std:') || lower.startsWith('/std:')) {
       const std = value.substring(5);
-      if (lang === 'CXX' || lang === 'OBJCXX' ) {
+      if (lang === 'CXX' || lang === 'OBJCXX' || lang === 'CUDA' ) {
         const s = parseCppStandard(std, can_use_gnu_std);
         if (s === null) {
           log.warning(localize('unknown.control.gflag.cpp', 'Unknown C++ standard control flag: {0}', value));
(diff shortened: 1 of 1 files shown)


bobbrow:
Thanks for the PR and the test project.  Sorry for being a total n00b here, but I get this when I try to configure (even after installing the CUDA tools).
```
[cmake] CMake Error at C:/Program Files/CMake/share/cmake-3.17/Modules/CMakeDetermineCompilerId.cmake:367 (message):
[cmake]   No CUDA toolset found.
```
Is there something I need to do to make sure CMake knows about CUDA?  I have the tools installed at `C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.1`

bobbrow:
I figured it out.  I didn't install the Visual Studio integration package.

Synxis:
Glad you've worked out this toolset problem ! Seems like there is an error in the nvidia installer btw...

Did this PR work for you (removing the message) ?

bobbrow:
Yes, it works.  Could you do us a favor and fix line 193 of cpptools.ts while you're in the file?  The warning message prints the stdver flag instead of the "unknown" language.  Just replace `value` with `lang` and then I'll merge it in.

image

Synxis:
Done !

bobbrow (approved):

bobbrow:
Wonderful!  Thank you for your help here!

PR #1503, opened 2020-10-22
Additional configure telemetry

andreeis:
First commit: trigger, Kit(toolchain, vs arch, c/cxx compiler).


src/cmake-tools.ts
```diff
@@ -51,12 +51,30 @@ const log = logging.createLogger('main');
...
 
-enum ConfigureType {
+export enum ConfigureType {
   Normal,
   Clean,
   Cache,
 }
 
+export enum ConfigureTrigger {
+  api = "Triggered by calling configure commands via the CMake Tools API",
```
bobbrow, line 61: I would rather not have long complicated strings for this.  Using the enum name as the string is best, I think.

src/drivers/driver.ts
```diff
@@ -601,7 +601,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
    * Perform a clean configure. Deletes cached files before running the config
    * @param consumer The output consumer
    */
-  public async cleanConfigure(extra_args: string[], consumer?: proc.OutputConsumer): Promise<number> {
+  public async cleanConfigure(trigger: string, extra_args: string[], consumer?: proc.OutputConsumer): Promise<number> {
```
bobbrow, line 604: Can you use the enum type name instead of `string` here?

src/drivers/driver.ts
```diff
@@ -614,10 +614,10 @@ export abstract class CMakeDriver implements vscode.Disposable {
     await this.doPreCleanConfigure();
     this.configRunning = false;
 
-    return this.configure(extra_args, consumer);
+    return this.configure(trigger, extra_args, consumer);
   }
 
-  async configure(extra_args: string[], consumer?: proc.OutputConsumer, withoutCmakeSettings:boolean = false): Promise<number> {
+  async configure(trigger: string, extra_args: string[], consumer?: proc.OutputConsumer, withoutCmakeSettings:boolean = false): Promise<number> {
```
bobbrow, line 620: Can you use the enum type name instead of `string` here?

src/drivers/driver.ts
```diff
@@ -665,7 +665,24 @@ export abstract class CMakeDriver implements vscode.Disposable {
         CMakeExecutableVersion: cmakeVersion ? `${cmakeVersion.major}.${cmakeVersion.minor}.${cmakeVersion.patch}` : '',
         CMakeGenerator: this.generatorName || '',
         ConfigType: this.isMultiConf ? 'MultiConf' : this.currentBuildType || '',
+        Toolchain: this._kit?.toolchainFile ? "true" : "false", // UseToolchain?
+        Trigger: trigger
       };
+
+      if (this._kit?.compilers) {
+        if (this._kit.compilers["C"]) {
+          telemetryProperties.CCompiler = this._kit.compilers["C"];
```
bobbrow, line 674: This needs to be scrubbed for PII. We can only report on known compilers and must exclude the path information.  cpptools has a list if you want to grab from that.  I can show you where it is.  I think we might want to simplify and remove version numbers from compiler names in favor of querying the version and logging that too.

src/cmake-tools.ts
```diff
@@ -463,9 +481,21 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
 
     this._statusMessage.set(localize('ready.status', 'Ready'));
 
+    this.extensionContext.subscriptions.push(vscode.workspace.onDidOpenTextDocument(async td => {
+      const str = td.uri.fsPath;
+      const sourceDirectory = await this.sourceDir;
+      if ((str.endsWith("CMakeLists.txt") || str.endsWith(".cmake")) && lightNormalizePath(str).startsWith(sourceDirectory)) {
```
bobbrow, line 487: Does it matter if it's in the workspace?  We should probably just log it?

src/cmake-tools.ts
```diff
@@ -463,9 +481,21 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+        telemetry.logEvent("cmakeFileOpen");
+      }
+    }));
+
     this.extensionContext.subscriptions.push(vscode.workspace.onDidSaveTextDocument(async td => {
       const str = td.uri.fsPath;
       const sourceDirectory = await this.sourceDir;
+      if ((str.endsWith("CMakeLists.txt") || str.endsWith(".cmake")) && lightNormalizePath(str).startsWith(sourceDirectory)) {
```
bobbrow, line 495: Does it matter if it's in the workspace?

src/variant.ts
```diff
@@ -383,6 +384,20 @@ export class VariantManager implements vscode.Disposable {
       if (!chosen) {
         return false;
       }
+
+      const cfg = vscode.workspace.getConfiguration('cmake', this.folder.uri).inspect<object>('defaultVariants');
+      const telemetryProperties: telemetry.Properties = {
+        customFile: (util.checkFileExists("cmake-variants.json") ||
```
bobbrow, line 390: Instead of checking file paths each time, can this state be saved during `_reloadVariantsFile` and then reported here?

test/extension-tests/successful-build/test/configure-and-build.test.ts
```diff
@@ -83,7 +83,7 @@ suite('Build', async () => {
   });
 
   test('Configure', async () => {
-    expect(await cmt.configure()).to.be.eq(0);
+    expect(await cmt.configure([], ConfigureType.Normal, ConfigureTrigger.runTests)).to.be.eq(0);
```
bobbrow, line 86: I probably would have made the trigger the first parameter since all are optional. This would also make it match `driver.configure`

bobbrow (changes requested):

src/variant.ts
bobbrow: An in any case, these functions return promises and would need to be awaited.

bobbrow: And in any case, these functions return promises and would need to be awaited.

src/cmake-tools.ts
andreeis: Oh, I guess it doesn't. For some reason I assumed this from the start.

src/variant.ts
andreeis: You are right, there is no need to check for the variants files and the defaultVariants setting every time a new variant is selected. It can all be moved to _reloadVariantsFile (no state var to be saved), do you agree?

andreeis: My mistake, _reloadVariantsFile is called too many times. I understand, a state var can be saved then and only when changing a variant we can report the telemetry. Exactly what you suggested.

test/extension-tests/successful-build/test/configure-and-build.test.ts
andreeis: Making the trigger the first parameter causes compilation errors with the configure prototype in CMake Tools API. I changed the prototype to be able to achieve what you suggest. Luckily not a breaking change since the first parameter I added is still optional. We'll discuss if there is a better way.

bobbrow: The CMake Tools API is not currently enabled for external callers and has some problems.  There is a "leak" in the API.  `type` is not part of the declaration, but it is in the definition.

What I think needs to happen here is we need to forward the public API call to an internal version of `configure` that accepts additional parameters.  When calling the public version it will automatically add the correct `trigger` and `type`.

src/drivers/driver.ts
```diff
@@ -614,10 +615,64 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+      throw e;
+    }
+  }
+
+  async getCompilerVersion(compilerPath: string) : Promise<string | undefined> {
+    const compilerName = path.parse(compilerPath).name;
+    const compilerDir = path.parse(compilerPath).dir;
+    let versionSwitch;
+    let versionRegexp;
+    switch (compilerName) {
```
bobbrow, line 649: I don't think a `switch` will work very well for us because there are tons of variants of gcc compilers out there and we will miss a lot of them if we only take the ones with these exact names.  Take a look at kit.ts which does some version parsing.  Maybe there's a way to reuse some of the code.  In the cpptools closed-source code, there is some logic you should look at to get these extra compiler names.

src/drivers/driver.ts
```diff
@@ -614,10 +615,64 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+        versionRegexp = RegExp(`${compilerName} .* (.*)`, "mg");
+        break;
+      case "clang":
+      case "clang-cl":
+      case "clang++":
+        versionSwitch = "--version";
+        versionRegexp = /clang version (.*) /mg;
+        break;
+      default:
+        return; // don't disclose information for compilers not in the known list
```
bobbrow, line 669: We should mark them as "other" so we get a sense of how many we're missing.

src/drivers/driver.ts
```diff
@@ -614,10 +615,64 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+    }
+
+    return `${compilerName}(${await this.testCompilerVersion(compilerName, compilerDir, versionSwitch, versionRegexp)})`;
```
bobbrow, line 672: Can you please make the compiler name and compiler version be separate properties instead of putting them together in the same string?

src/drivers/driver.ts
```diff
@@ -614,10 +615,64 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
-  async configure(extra_args: string[], consumer?: proc.OutputConsumer, withoutCmakeSettings:boolean = false): Promise<number> {
+  async testCompilerVersion(program: string, cwd: string, arg: string | undefined, regexp: RegExp): Promise<string> {
+    const args = [];
+    if (arg) {
+      args.push(arg);
+    }
+    const child = this.executeCommand(program, args, undefined, {silent: true, cwd});
+    try {
+      const result = await child.result;
+      log.debug(localize('command.version.test.return.code', 'Command version test return code {0}', nullableValueToString(result.retc)));
```
bobbrow, line 629: I don't think we should put output for telemetry errors in the user log.  This could go in the console.log if you wanted to keep it.

src/drivers/driver.ts
```diff
@@ -614,10 +615,64 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+      const versionLine = (program === "cl") ? result.stderr : result.stdout; //.split("\\n")[0];
+      const match = regexp.exec(versionLine);
+      return match ? match[1] : "error";
+    } catch (e) {
+      const e2: NodeJS.ErrnoException = e;
+      log.debug(localize('compiler.version.return.code', 'Compiler version test return code {0}', nullableValueToString(e2.code)));
+      if (e2.code === 'error') {
+        return e2.code;
+      }
+      throw e;
```
bobbrow, line 640: This function should not be allowed to throw because this is just telemetry.  It needs to return a string.  If there's an error, then return an error string that is easy for telemetry to filter on (i.e. don't make it different per compiler).

src/drivers/driver.ts
```diff
@@ -614,10 +615,64 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+      case "cpp:":
+      case "c++:":
+        versionSwitch = "--version";
+        versionRegexp = RegExp(`${compilerName} .* (.*)`, "mg");
+        break;
+      case "clang":
+      case "clang-cl":
+      case "clang++":
+        versionSwitch = "--version";
+        versionRegexp = /clang version (.*) /mg;
```
bobbrow, line 666: Look at kit.ts for clang regex.  This will miss Apple clang.

src/drivers/driver.ts
```diff
@@ -614,10 +615,64 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+        // cl does not have a version switch but it outputs the compiler version on stderr
+        // when no source files arguments are given
+        versionRegexp = /.* Compiler Version (.*) for .*/mg;
+        break;
+      case "gcc":
+      case "g++:":
+      case "cpp:":
+      case "c++:":
+        versionSwitch = "--version";
+        versionRegexp = RegExp(`${compilerName} .* (.*)`, "mg");
```
bobbrow, line 660: The first `.*` should probably be a `[^\s]+` instead.  `.` matches whitespace too and you specifically don't want to match whitespace for this.

bobbrow (changes requested):

src/drivers/driver.ts
andreeis: Not sure. Here is the version output I started from: 
**cpp (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0**
With the expression above, I am looking at match[1] and obtain **9.3.0**
With the suggestion I get this whole thing: **9.3.0-17ubuntu1~20.04) 9.3.0**

src/drivers/driver.ts
andreeis: Good catch, I didn't know.

src/drivers/driver.ts
andreeis: Inspired by the areas you recommended me to look at, I realized -v is more general version switch and it produces different kind of output than --version, in which case your suggestion is perfect and you'll see in the next update.

src/drivers/driver.ts
andreeis: Ok, I got the list of compiler names from cpptools (src\vc\projbld\CMake\Package\Telemetry\Telemetry.cs) and I hardcoded each version switch and regexp scenario based on man commands and internet searches (I wasn't able to install all of them to really see what works).

src/api.ts
```diff
@@ -8,7 +8,7 @@
  */ /** */
 
 import {DebugSession, Disposable, Event, Terminal} from 'vscode';
-
+//import {ConfigureTrigger} from '@cmt/cmake-tools';
```
bobbrow, line 11: please remove commented code

src/drivers/driver.ts
```diff
@@ -614,10 +620,114 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+      args.push(arg);
+    }
+    const child = this.executeCommand(program, args, undefined, {silent: true, cwd});
+    try {
+      const result = await child.result;
+      console.log(localize('command.version.test.return.code', 'Command version test return code {0}', nullableValueToString(result.retc)));
+      // cl.exe version information is printed on stderr when no source files are given
+      const versionLine = (program === "cl") ? result.stderr : result.stdout; //.split("\\n")[0];
+      const match = regexp.exec(versionLine);
+      return match ? match[2] : "error";
```
bobbrow, line 638: how do we know the array has two elements?

andreeis: From the regular expressions constructed in getCompilerVersion. But indeed, there needs to be a way to be verified or enforced.

andreeis: And I'll remove that commented code in this snippet.

src/drivers/driver.ts
bobbrow: So if you want to capture the thing after the last space, then you would use `\`${compilerName}.*([^\s]+)$`

---
In reply to: [516683687](https://github.com/Microsoft/vscode-cmake-tools/pull/1503#discussion_r516683687) [](ancestors = 516683687)

bobbrow: Oh, but if you're using multiline regex, then I don't know.  I guess you can leave it if it's doing the right thing.

src/drivers/driver.ts
bobbrow: This doesn't cover cross compilers with triplet names (and I don't think a `switch` is the right tool for this since there are a lot of combinations)
For example: `arm-none-eabi-gcc`

---
In reply to: [517393239](https://github.com/Microsoft/vscode-cmake-tools/pull/1503#discussion_r517393239) [](ancestors = 517393239)

bobbrow: Maybe this logic can be added to the `default` case.  Look at the code for cpptools for how we report this compilers.  Search for the variable `compiler_allowed_list` in the code.

---
In reply to: [517487911](https://github.com/Microsoft/vscode-cmake-tools/pull/1503#discussion_r517487911) [](ancestors = 517487911,517393239)

andreeis: I didn't find any compiler_allowed_list, but SupportedDefaultCompilers. Did you mean that one?

bobbrow: Are you looking in the C++ code, or the TypeScript?  It's not in the TypeScript.

src/drivers/driver.ts
bobbrow: You could add a field to your compilerAllowList entries that specifies which capture group the version string is in.

bobbrow (approved):
One optional comment about the regex

bobbrow (approved):

PR #1514, opened 2020-10-28
Cache GUI localization and UI changes

andreeis:
Make sure all the strings in the cache GUI are localized.
Change the UI appearance so that the web view looks more similar with CppTools.
I was not able so far to change the following:
- the active border (the border when any UI element is in focus). For CppTools is blue, for CMake Tools is a yellow/orange but I tried to change it in various ways without success
- the checkboxes have a blue background when set and remain in sync with the current theme when not set. CppTools checkboxes are in sync with the current theme also when set.

### DIFF
--- a/src/cache-view.ts
+++ b/src/cache-view.ts
@@ -162,45 +162,17 @@ export class ConfigurationWebview {
               border: 1px solid var(--vscode-settings-dropdownBorder);
           }
 
-          .cmake-input-string {
-              position: relative;
-              background-color: var(--vscode-settings-textInputBackground);
-              color: var(--vscode-settings-textInputForeground);
-              width: 800px;
-              height: 25px;
+          input {
+            height: 17px;
+            padding: 6px;
+            border: solid 1px;
+            font-size: 13px;
+            font-family: Menlo, Monaco, Consolas, "Droid Sans Mono", "Courier New", monospace, "Droid Sans Fallback";
+            color: var(--vscode-settings-textInputForeground);
+            background: var(--vscode-settings-textInputBackground);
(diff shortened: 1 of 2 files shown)


src/cache-view.ts
```diff
@@ -161,46 +243,83 @@ export class ConfigurationWebview {
...
+              contrastBorder: var(--vscode-focusBorder);
+              contrastActiveBorder: var(--vscode-focusBorder);
           }
         </style>
         <script>
           const vscode = acquireVsCodeApi();
           function toggleKey(id) {
             const label = document.getElementById('LABEL_' + id);
-            label.textContent = label.textContent == 'ON' ? 'OFF' : 'ON';
+            label.textContent = label.textContent == '${onButtonText}' ? '${offButtonText}' : '${onButtonText}';
```
bobbrow, line 322: I think these are CMake enum values.  I don't believe they should be localized.

src/cache-view.ts
```diff
@@ -40,6 +41,16 @@ export class ConfigurationWebview {
     );
   }
 
+  async persistCacheEntries() {
```
bobbrow, line 44: This seems to save the temporary changes directly into the cache. I thought that you were going to persist them in memory and then reload the changes the next time the page renders. I don't think we should be saving stuff without the user pushing the "save" button.

andreeis: I agree. I noticed that the settings get persisted in settings.json when changing focus or closing the tab so I thought to be matching that but then the SAVE button has no reason to exist.

src/cache-view.ts
andreeis: Ups, yes

bobbrow (approved):

PR #1520, opened 2020-11-05
CMake Tools 1.5 Release Notes

andreeis:


CHANGELOG.md
```diff
@@ -2,12 +2,35 @@
...
+- Add option to set CMAKE_BUILD_TYPE also on multi-config generators. [PR #1393](https://github.com/microsoft/vscode-cmake-tools/pull/1393) [@tonka3000](https://github.com/tonka3000)
+- Detect Clang for MSVC (GNU CLI) kits. [#823](https://github.com/microsoft/vscode-cmake-tools/issues/823) [@omcnoe](https://github.com/omcnoe)
+- GUI support for CMake Tools cache. [#513](https://github.com/microsoft/vscode-cmake-tools/issues/513) [@nieroger](https://github.com/nieroger)
+- Tasks support. [PR #1268](https://github.com/microsoft/vscode-cmake-tools/pull/1268) [@vptrbv](https://github.com/vptrbv)
+- MacBook Pro touchbar support. [#499](https://github.com/microsoft/vscode-cmake-tools/issues/499) [@vptrbv](https://github.com/vptrbv)
+
+Bug Fixes:
+- Set right base_path for variant config files. [PR #1462](https://github.com/microsoft/vscode-cmake-tools/pull/1462) [@leolcao](https://github.com/leolcao)
+- Remove duplicate option start signs. [#1280](https://github.com/microsoft/vscode-cmake-tools/issues/1280) [@dnkbln](https://github.com/dnkbln)
```
bobbrow, line 15: Did you go back and test this on linux?  I think it needs to be backed out.

andreeis: Yes I did test it again, windows and linux, several different generators... The "--" was really an extra. In all cases I tested, what was following the "--" was also having a switch prefix, either "/" or "-", depending on generator.

andreeis: I'll get soon on MAC, I'll test it there too.

bobbrow: Did you test with the build tool args setting set?

bobbrow: Maybe the `--` just needs to go after the generator args.

bobbrow: It looks like maybe we should just use -j as a cmake option instead of the native build tool switch.  https://cmake.org/cmake/help/v3.18/manual/cmake.1.html#build-a-project

bobbrow (approved):
Please remove the line for the change we backed out, and then submit.

bobbrow (approved):

andreeis:
> Please remove the line for the change we backed out, and then submit.

I did. Did I miss something else?

bobbrow:
No, nothing missed.  I sent the message around the same time you pushed an update so it didn't take that into account.

PR #1521, opened 2020-11-05
Use newer CMake CLI for job parallelism

bobbrow:
This addresses some confusion around arguments that can be passed to cmake related to build parallelism.

A bug was introduced in PR #1421 as a result.


tonka3000 (approved):

src/drivers/driver.ts
```diff
@@ -1235,23 +1235,32 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+    if (this.cmake.version && this.cmake.version.major >= 3 && this.cmake.version.minor > 13) {
+      buildArgs.push('-j');
+      if (this.config.numJobs) {
+        buildArgs.push(this.config.numJobs.toString());
+      }
+    } else {
+      if (gen) {
+        if (/(Unix|MinGW) Makefiles|Ninja/.test(gen) && target !== 'clean') {
+          buildToolArgs.push('-j', this.config.numJobs.toString());
+        } else if (/Visual Studio/.test(gen) && target !== 'clean') {
```
andreeis, line 1252: Can this.config.numJobs be null here? I am asking because I noticed the check higher up when version >= 3.14.
Otherwise looks good.

andreeis (approved):

src/drivers/driver.ts
bobbrow: It's not checking for null.  It's checking for non-zero.

andreeis: Got it, I understand.

PR #1546, opened 2020-11-12
improve japanese translation

SSW-SCIENTIFIC:
##  The purpose of this change

Improvement for Japanese translation.

I know that the translation files are auto-generated, so this is suggestion for translation against obvious mistranslations.


bobbrow:
Thank you! I will share these suggestions with our translators.  When they are accepted a new PR will be generated with the updated translations and we will close this one.  Since these files are generated, we don't want to change them manually without updating the source.

bobbrow:
This was added to PR #1560

Issue #1548, opened 2020-11-12
Loc issue with Kits


bobbrow:
I just accepted PR #1538, but didn't notice that there was a new loc string added for the Clang kits which will break sorting in the Kits list.  We need to undo that part of the loc change (for "clang.for.msvc" entries) and get a new translation that keeps "Clang" at the beginning of the string.

bobbrow:
Something like this needs to be added to the code (but I'm not 100% sure).  Please check with cpptools on the syntax since they have a few "hints" as well.

```ts
localize({ key: 'clang.for.msvc', comment: ['Clang should stay at the beginning of the string as it is used in UI sorting']}, 'Clang {0} {1} with {2} ({3})', version.version, clang_cli, install_name, vs_arch),
```

Issue #1542, opened 2020-11-10
1.5.0 appears to have broken "Clean All Projects"


schnort:
### Brief Issue Summary

1.5.0, 1.5.1, and 1.5.2 seem to have broken the "Clean All projects" menu item from the CMake Project Outline tab

### Expected:

1. Open VSCode in cmake workspace
2. Choose 'configure' or 'clean configure', etc. from the '...' menu item on the top right of the CMake Project Outline tab
3. Choose 'Clean All Projects' from the '...' menu item

### Apparent Behavior:

On 1.5.0 and beyond, it calls cmake with --target all 
On 1.4.2 and before, it calls cmake with --target clean

### Platform and Versions

Version: 1.51.0 (user setup)
Commit: fcac248b077b55bae4ba5bab613fd6e9156c2f0c
Date: 2020-11-05T18:18:23.642Z
Electron: 9.3.3
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Windows_NT x64 10.0.18363

- **CMake Version**: 3.18.0
- **CMake Tools Extension Version**: 1.5.0, 1.5.1, 1.5.2
- **Compiler/Toolchain**: [cmake] -- The C compiler identification is MSVC 19.26.28806.0

andreeis:
Thank you for opening this issue. We reproduce on our side and identified the PR that introduced this bug. We'll keep you updated about the fix for this.

andreeis:
This is fixed in the 1.5.3 release that was just published earlier.
Upgrade your CMake Tools extension in VSCode and let us know if you encounter any other problems.

PR #1554, opened 2020-11-17
1.5.3 bug fixes

andreeis:
Fixes for:
https://github.com/microsoft/vscode-cmake-tools/issues/1542
https://github.com/microsoft/vscode-cmake-tools/issues/1548
https://github.com/microsoft/vscode-cmake-tools/issues/1548

Version bump to 1.5.3. Changelog updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # What's New?
 
+## 1.5.3
+Bug Fixes:
+- "Clean all projects" broken since 1.5.0. [#1542](https://github.com/microsoft/vscode-cmake-tools/issues/1542)
+- Localization fix related to "clang for MSVC" strings in kits.  [#1548](https://github.com/microsoft/vscode-cmake-tools/issues/1548)
+- CMake task provider should not attempt to register until the CMake driver is available.  [#1549](https://github.com/microsoft/vscode-cmake-tools/issues/1549)
+
 ## 1.5.2
 Bug Fixes:
-- Fix deadlock caused by commands invoked in string expansion during activation. [PR #1532](https://github.com/microsoft/vscode-cmake-tools/issues/1532)
+- Fix deadlock caused by commands invoked in string expansion during activation. [PR #1532](https://github.com/microsoft/vscode-cmake-tools/pull/1532)
 
 ## 1.5.1
 Bug Fixes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## 1.5.3
 Bug Fixes:
 - "Clean all projects" broken since 1.5.0. [#1542](https://github.com/microsoft/vscode-cmake-tools/issues/1542)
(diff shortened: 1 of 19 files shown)


src/extension.ts
```diff
@@ -1083,6 +1083,14 @@ class ExtensionManager implements vscode.Disposable {
 let _EXT_MANAGER: ExtensionManager|null = null;
 let cmakeTaskProvider: vscode.Disposable | undefined;
 
+export async function registerTaskProvider(command: string | null) {
+  if (command) {
+    rollbar.invokeAsync(localize('registerTaskProvider', 'Register the task provider.'), async () => {
+      cmakeTaskProvider = vscode.tasks.registerTaskProvider(CMakeTaskProvider.CMakeType, new CMakeTaskProvider({ build: command }));
```
bobbrow, line 1089: I think this should check if one is already provided and dispose the old one. Now that I think of it, I think this is already broken for multi-root today since there can be multiple cmake drivers alive. Can you open a bug? For now, let's just `dispose` the existing one and replace it with a new one.

src/folders.ts
```diff
@@ -85,7 +85,7 @@ export class CMakeToolsFolderController implements vscode.Disposable {
   get(ws: vscode.WorkspaceFolder | Array<string> | undefined): CMakeToolsFolder | undefined {
     if (ws) {
       if (util.isArrayOfString(ws)) {
-        return this._instances.get(ws[-1]);
+        return this._instances.get(ws.slice(-1)[0]);
```
bobbrow, line 88: Can it just be `ws[ws.length-1]`? Do we need to copy the reference to the last element?

src/extension.ts
```diff
@@ -833,7 +833,7 @@ class ExtensionManager implements vscode.Disposable {
 
   build(folder?: vscode.WorkspaceFolder, name?: string) { return this.mapCMakeToolsFolder(cmt => cmt.build(name), folder, true); }
 
-  buildAll(name: string[]) { return this.mapCMakeToolsAll(cmt => cmt.build(util.isArrayOfString(name) ? name[-1] : name), true); }
+  buildAll(name: string[]) { return this.mapCMakeToolsAll(cmt => cmt.build(util.isArrayOfString(name) ? name.slice(-1)[0] : name), true); }
```
bobbrow, line 836: Can it just be `name[name.length-1]`?

CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
 # What's New?
 
+## 1.5.3
+Bug Fixes:
+- "Clean all projects" broken since 1.5.0. [#1542](https://github.com/microsoft/vscode-cmake-tools/issues/1542)
+- Localization fix related to "clang for MSVC" strings in kits.  [#1548](https://github.com/microsoft/vscode-cmake-tools/issues/1548)
```
bobbrow, line 6: This never shipped incorrectly, so it doesn't need to be mentioned.

bobbrow (changes requested):

bobbrow (approved):

Issue #1493, opened 2020-10-19
"Compile Active File" does not work with an empty define


dslijepcevic:
### Brief Issue Summary

It looks like "Compile Active File" incorrectly parses empty defines from `compile_commands.json`. Instead of passing `-DMY_DEFINE=` to the compiler, it passes an empty argument.

`CMakeLists.txt`:
```cmake
cmake_minimum_required(VERSION 3.8 FATAL_ERROR)
project(test C)
add_executable(test test.c)
target_compile_definitions(test PUBLIC MY_DEFINE=)
```

`test.c`:
```c
int main()
{
    return 0;
}
```

`build/compile_commands.json`:
```json
[
{
  "directory": "/proj/test/build",
  "command": "/usr/bin/cc -DMY_DEFINE=\"\"  -g   -o CMakeFiles/test.dir/test.c.o   -c /proj/test/test.c",
  "file": "/proj/test/test.c"
}
]
```

"Compile Active File" output:
```sh
/usr/bin/cc "" -g -o CMakeFiles/test.dir/test.c.o -c /proj/test/test.c
/proj/test/build$ /usr/bin/cc "" -g -o CMakeFiles/test.dir/test.c.o -c /proj/test/test.c
cc: error: : No such file or directory
```

andreeis:
Thank you for reporting this issue. We'll let you know if we need more info to investigate it.

elahehrashedi:
"Compile Active File" does not work with any string define.

Issue #969, opened 2019-12-12
Can not compile active file if definition has quoted text


dmyger:
### Brief Issue Summary
If CMakeLists.txt has any C++ definition with ampersands char it brokes compiling active file.

It does not mater what kind of cmake command used no one of `add_definitions`, `add_compile_definitions` nor `target_compile_definitions`.

### Apparent Behavior:
Just add in the file
```
add_compile_definitions(_WRONG_CHAR_='&')
```
First try configure the project and check `compile_commands.json` to ensure that it was correctly processed in compilers option `-D_WRONG_CHAR_=\"'&'\"`
And you can successfully build entire project as expected.

But if you try run compile only one active file with command `cmake.compileFile` you will get an error, because this definition will lost first part and leave only `"'&'"`

### Expected:
Compile active file should use the same command from `compile_commands.json` as it was used to compile entire project.

### Other Notes
Originally the issue was noticed with following cmake code:
```
string(LENGTH "${CMAKE_SOURCE_DIR}/" __SOURCE_PATH_SIZE)
add_compile_definitions(__FILENAME__=\(&__FILE__[${__SOURCE_PATH_SIZE}]\))
```
To set `__FILENAME__` in relative path of current sorce file.

### Platform and Versions
- **Operating System**: openSUSE 15.0
- **CMake Version**: 3.13.4
- **VSCode Version**: 1.40.2
- **CMake Tools Extension Version**: 1.2.3
- **Compiler/Toolchain**: GCC 7.4

bobbrow:
The problem appears to be in the argument splitting logic where the current token is cleared when a quote char is encountered. https://github.com/microsoft/vscode-cmake-tools/blob/de7672c1c8e2e14ecb23abd4a2128b943147d64c/src/shlex.ts#L57

This might not have ever worked in the extension.  Thanks for reporting it!

ThorstenKleinSmith:
Looks like this is the same as issue #605

Trass3r:
> The problem appears to be in the argument splitting logic where the current token is cleared when a quote char is encountered.
> 
> https://github.com/microsoft/vscode-cmake-tools/blob/de7672c1c8e2e14ecb23abd4a2128b943147d64c/src/shlex.ts#L57

It's much easier to just let the shell do its job via [`shell: true`](https://nodejs.org/api/child_process.html) or similar.
Had the same problem in another extension.

stevenr3:
[...]

PR #1556, opened 2020-11-17
support parsing sub-string

elahehrashedi:
Add support parsing define variables that are in the form of a string, e.g.:
`target_compile_definitions(prname PUBLIC AWESOME=\"foo\")`
bug:
main:
https://github.com/microsoft/vscode-cmake-tools/issues/969

duplicates: 
https://github.com/microsoft/vscode-cmake-tools/issues/1493
https://github.com/microsoft/vscode-cmake-tools/issues/1565


src/shlex.ts
```diff
@@ -41,17 +42,29 @@ export function* split(str: string, opt?: ShlexOptions): Iterable<string> {
...
       }
-
       // Another quoted char
       token = (token || '') + char;
       continue;
     }
 
     if (quoteChars.includes(char)) {
+      // Beginning of a sub-quoted token
+      if (i > 1 && str.substring(i - 1, i + 1) === "\\\"") {
```
sean-mcmanus, line 61: It seems like `"\\\'"` should be handled too, maybe just when when opt.mode == "posix".

sean-mcmanus (changes requested):

sean-mcmanus (changes requested):
The posix case can use ‘’ or “” quotes.

bobbrow:
Can you add a test that covers this @elahehrashedi?

andreeis (approved):

sean-mcmanus (approved):

sean-mcmanus:
> Can you add a test that covers this @elahehrashedi?

A test case appears to exist already at https://github.com/microsoft/vscode-cmake-tools/blob/80a745aeb7a12c76f5976a7b4ebca9674d17b5dd/test/unit-tests/shlex.test.ts, so it looks like you'd just need to add some lines to cover this scenario.

src/shlex.ts
```diff
@@ -41,17 +42,29 @@ export function* split(str: string, opt?: ShlexOptions): Iterable<string> {
...
       }
-
       // Another quoted char
       token = (token || '') + char;
       continue;
     }
 
     if (quoteChars.includes(char)) {
+      // Beginning of a sub-quoted token
+      if (i > 0 && (opt.mode === 'posix' ? (str.substring(i - 1, i + 1) === "\\\'" || str.substring(i - 1, i + 1) === "\\\"") : str.substring(i - 1, i + 1) === "\\\"")) {
```
michelleangela, line 61: `str.substring(i - 1, i + 1)` repeats 3 times, Can it be saved into a variable? And then compare that variable with the different `\\\`

michelleangela (changes requested):

src/shlex.ts
michelleangela: `"\\\""` also repeats, Can the conditional be simplified to sometime like:

```
const quote: string = str.substring(i - 1, i + 1);
if (i > 0 && (quote === "\\\"" || ( opt.mode === 'posix' && === "\\\'"))

elahehrashedi:
I am closing this PR, the content of this PR are included here: https://github.com/microsoft/vscode-cmake-tools/pull/1648

Issue #1480, opened 2020-10-09
Compile errors in status.ts using ts 4.0+


xisui-MSFT:
Since classes overloading Button is defining members such as 'icon' as property, but Button defines them as accessor, there would be compile error when suing ts 4.0+.

This is not very urgent since cmake tools is still using 3.7. However, 4.0+ could be selected in VS Code and F5 uses this setting by default:
![image](https://user-images.githubusercontent.com/44103947/95630660-0fd49080-0a37-11eb-8072-5cbc93ab3ff0.png)

Would still be nice to change these members anyway just for coding style.

bobbrow:
Basically, move the initialization into a constructor in the child classes.

E.g: 
![image](https://user-images.githubusercontent.com/12818240/96522038-7459d180-1227-11eb-8f9d-421a96ca1842.png)

PR #1558, opened 2020-11-17
change property to accessor property

elahehrashedi:
fixing Compile errors in status.ts using ts 4.0+: https://github.com/microsoft/vscode-cmake-tools/issues/1480

### DIFF
--- a/src/status.ts
+++ b/src/status.ts
@@ -24,8 +24,8 @@ abstract class Button {
   private _tooltip: string|null = null;
   private _icon: string|null = null;
 
-  constructor(protected readonly config: ConfigurationReader, protected readonly _priority: number) {
-    this.button = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, this._priority);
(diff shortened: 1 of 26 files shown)


src/status.ts
```diff
@@ -24,7 +24,7 @@ abstract class Button {
   private _tooltip: string|null = null;
   private _icon: string|null = null;
 
-  constructor(protected readonly config: ConfigurationReader, private readonly _priority: number) {
+  constructor(protected readonly config: ConfigurationReader, protected readonly _priority: number) {
```
sean-mcmanus, line 27: It seems like _priority should be renamed to priority if it's no longer private.

sean-mcmanus (changes requested):

sean-mcmanus (changes requested):
The changes from https://github.com/microsoft/vscode-cmake-tools/pull/1558/commits/f72aec66abc3fdb835db8e331732ff37c6dd2077 should be reverted.

sean-mcmanus (approved):

Issue #1440, opened 2020-09-17
Hardcoded CMAKE_EXPORT_COMPILE_COMMANDS getting in the way


treilly-nuodb:
### Brief Issue Summary

I had to hack the this extension to turn off CMAKE_EXPORT_COMPILE_COMMANDS.    It needs to be turned off when doing unity builds.  

### Expected:

There should be a way to override the default behavior of hardcoding CMAKE_EXPORT_COMPILE_COMMANDS to true.  Our typical workflow is to run cmake once with it on and then turn it off and run again with unity on.

andreeis:
We almost have CMake Tools 1.6.0 release out the door. Until then, if you want, you can also verify if this problem has been fixed correctly by installing the vsix built here: https://github.com/microsoft/vscode-cmake-tools/actions/runs/557242002#artifacts.

PR #1559, opened 2020-11-18
add cmake.exportCompileCommandFile to package.json

elahehrashedi:
feature request: https://github.com/microsoft/vscode-cmake-tools/issues/1440

### DIFF
--- a/package.json
+++ b/package.json
@@ -1581,10 +1581,10 @@
             "never"
           ]
         },
-        "cmake.exportCompileCommandFile": {
+        "cmake.exportCompileCommandsFile": {
           "type": "boolean",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1580,6 +1580,11 @@
             "always",
             "never"
           ]
+        },
+        "cmake.exportCompileCommandFile": {
```
sean-mcmanus, line 1584: Seems like this should be exportComileCommandsFile (i.e. with an 's').

sean-mcmanus (changes requested):

sean-mcmanus (approved):

PR #1568, opened 2020-11-23
Improve french translation

sjoubert:
This improves a translation string which does not make much sense to me.

### DIFF
--- a/i18n/fra/package.i18n.json
+++ b/i18n/fra/package.i18n.json
@@ -12,7 +12,7 @@
 	"cmake-tools.command.cmake.setVariantAll.title": "Sélectionner une variante pour tous les projets",
 	"cmake-tools.command.cmake.configure.title": "Configurer",
 	"cmake-tools.command.cmake.configureAll.title": "Configurer tous les projets",
-	"cmake-tools.command.cmake.build.title": "Version",
+	"cmake-tools.command.cmake.build.title": "Générer",
 	"cmake-tools.command.cmake.buildAll.title": "Générer tous les projets",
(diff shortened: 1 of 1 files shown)


bobbrow:
Thank you! I will share this suggestion with our translators. When they are accepted a new PR will be generated with the updated translation and we will close this one. Since these files are generated, we don't want to change them manually without updating the source.

i18n/fra/package.i18n.json
```diff
@@ -12,7 +12,7 @@
 	"cmake-tools.command.cmake.setVariantAll.title": "Sélectionner une variante pour tous les projets",
 	"cmake-tools.command.cmake.configure.title": "Configurer",
 	"cmake-tools.command.cmake.configureAll.title": "Configurer tous les projets",
-	"cmake-tools.command.cmake.build.title": "Version",
```
sean-mcmanus, line 15: This is currently "Build" -- not sure if that changes the translation.

sjoubert: It shouldn't. "Build" can be translated in various ways here such as "Construire" or "Générer". I've just followed the terminology already used for the "buildAll" entry.

bobbrow:
PR #1699 incorporates this feedback.  The localizers said this:
```
This is now translated not per suggestion. A better translation was placed and will be sync'd with the original repo.
Linguistic answer: "This is a very old issue. IT pros and developers in France use "Build". If a a verb, please use "Générer un build""
```
When the bug came back to me as resolved I reactivated it because they picked "Build" instead of "Générer un build."  It looks like those changes have now landed and I merged them in this morning.  This PR can now be closed.  Thanks again for your translation suggestion!

PR #1581, opened 2020-12-03
Update 3rd party notices

bobbrow:


sean-mcmanus (changes requested):
yarn and other dev-only dependencies should not be in the NOTICE.txt.

andreeis (approved):

PR #1587, opened 2020-12-09
Use non greedy search for the = sign when replacing cmake cache entries

teajay-fr:
## The purpose of this change
When editing the cmake cache through the UI, cmake cache entries which contain a = in the value are incorrectly split because of the greedy match operator used.
```
CMAKE_CXX_COMPILE_FLAGS:STRING=-mcpu=cortex-m4
```
This change makes sure the first = sign is used for the lookup of the cmake cache string.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -188,7 +188,7 @@ export class CMakeCache {
    * @param value Boolean value
    */
   private replace(content: string, key: string, value: string): string {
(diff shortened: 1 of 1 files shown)


andreeis:
Thank you for fixing this. We'll take a look, test and merge this in shortly.

bobbrow:
This was also fixed by PR #1629.  Thank you for your help!

PR #1602, opened 2020-12-17
pass args when we Launch the executable without Debug

jbdamiano:
## The purpose of this change

  When we launch the executable in the debuger, it uses the args parameter. But when we launch it without the debuger 
it doesn't. 

  This change provides the same behaviour and allow to pass args when we launch without debug the executable/


jbdamiano:
I will open another to fix error of check

PR #1598, opened 2020-12-16
Fixes environment expanding and document testEnvironment,

lygstate:
They share a single computeExpandedEnvironment for expanding.

TODO: fixes debug launch and run program the same as CTest,
they should share the same environment variables.

Fixes 
https://github.com/microsoft/vscode-cmake-tools/issues/1536
https://github.com/microsoft/vscode-cmake-tools/issues/1246
https://github.com/microsoft/vscode-cmake-tools/issues/1250
https://github.com/microsoft/vscode-cmake-tools/issues/1246

PR #1603, opened 2020-12-17
Pass args

jbdamiano:
## The purpose of this change
When we launch the executable in the debuger, it uses the args parameter. But when we launch it without the debuger
it doesn't.

This change provides the same behaviour and allow to pass args when we launch without debug the executable/


andreeis:
Thank you for addressing the launch args problem in this PR. This was previously attempted with https://github.com/microsoft/vscode-cmake-tools/pull/1201 and we still need to figure out how this would work in more complicated terminal and quoting scenarios. We will revisit this after the holidays to make a decision.

lygstate:
https://github.com/microsoft/vscode-cmake-tools/pull/1598 are also ralted, launch program also need specify proper environment variables.

andreeis (approved):

Issue #1342, opened 2020-07-01
Scanning for Kits on WSL with clang 11 installed fails to detect clang.


danielmrussell:
Issue Type: <b>Bug</b>

Linux Ubuntu 20.04 under WSL2, fresh install.
Freshly installed Clang-11 and clang, clang++ symlinks via sudo apt install clang clang-11. Running clang, clang++ etc work from commandline. It is in the path as well. Scanning for kits refuses to notice clang whatsoever, only picking up gcc.

Extension version: 1.4.1
VS Code version: Code 1.46.1 (cd9ea6488829f560dc949a8b2fb789f3cdc05f5d, 2020-06-17T21:13:20.174Z)
OS version: Windows_NT x64 10.0.19041
Remote OS version: Linux x64 4.19.104-microsoft-standard

bobbrow:
What is the output of `clang -v` for you?

danielmrussell:
I'm not sure what it was when I wrote this yesterday and had been using apt
packages.

But now I deinstalled those and built clang because I needed newer features
for C++20. Nonetheless, kit scanning still fails to look for it.
Here's the output now when running clang -v:

`
clang version 11.0.0 (https://github.com/llvm/llvm-project.git
425fb21e03b0ec1cac40a415b6417ff5ba04d076)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/local/bin
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/8
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/9
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/9
Candidate multilib: .;@m64
Selected multilib: .;@m64
`

bobbrow:
Thanks. This output looks normal, so I'm not sure why it wouldn't pick it up.  We'll need to give WSL2 a try and see what's going on there.  Are you able to make it work by adding your own kit?

```
  {
    "name": "clang 11",
    "compilers": {
      "C": "<path/to/clang>",
      "CXX": "<path/to/clang++>"
    }
  }
```

danielmrussell:
I was indeed able to do that. To be clear, this also happened on WSL1. I
tried to upgrade everything to get less broken.

bobbrow:
Thanks for the update.  We have WSL1 already and can try there.

BMWestrich:
I am having the same issue with native linux:

[...]

PR #1620, opened 2021-01-15
Detect clang version at end of line

falbrechtskirchinger:
The clang version regex expects the version number to be followed by white space or a dash. This is not the case for Gentoo builds.
Example output:
```
$ clang -v
clang version 12.0.0
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm/12/bin
(snip)
```
This patch modifies the regex to detect version numbers ending at end-of-line.


tichun:
This should fix https://github.com/microsoft/vscode-cmake-tools/issues/1342

bobbrow (approved):

Issue #1562, opened 2020-11-19
Extension activated on Show Tasks/Run Build Task even if not in a CMake project


rapgenic:
### Brief Issue Summary

When working on non CMake projects and using the "Tasks: Run Build Task" or simply "Show tasks" command, the cmake extension activates automatically asking to choose a kit, creating a new view and filling the statusbar with useless information (since there is no CMake project)

### Expected:

For a non CMake project I would expect that even though I run the "Build" task, the extension wouldn't turn on.

### Other Notes/Information

Looking at the package.json I can see that one of the activation events is `    onCommand:workbench.action.tasks.runTask`. This suggests to me that the originally desired behaviour was:

1. User asks to show tasks
2. Register a cmake task provider in case there were any tasks to provide

However i think that this activation event should be removed, leading to the following new behaviour:

1. If there is some reason to activate cmake on folder open (e.g. CMakeLists.txt present) load the full cmake extension (which is the actual behaviour as of now, for what I can see)
2. Since CMake is used and configured in the project, _only now_ register a task provider for cmake tasks

This would allow to reduce the UX clutter, because the CMake extension doesn't turn on apparently for no reason adding a lot of unwanted UI elements, and in my opinion is more correct, because there is no reason to register a cmake task provider if there is no cmake project in the first place

andreeis:
Thank you for opening this issue, we will investigate and let you know if we need any other info.

Trass3r:
Same here, for example this extension is active when working on TypeScript projects (vscode extensions).
It overrides Ctrl+F5 and asks for a kit for example, so I always have to go through the Run menu.
There are also annoying popups or configure attempts every now and then.

bobbrow:
Sorry for the inconvenience.  We will be fixing this for the next release.

andreeis:
We almost have CMake Tools 1.6.0 release out the door. Until then, if you want, you can also verify if this problem has been fixed correctly by installing the vsix built here: https://github.com/microsoft/vscode-cmake-tools/actions/runs/557242002#artifacts.

rapgenic:
I tried it and I believe the issue is fixed now, thank you!

PR #1642, opened 2021-02-02
remove Cmake activation on tasks.runTask

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/1562

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,9 +77,8 @@
     "onCommand:cmake.outline.configureAll",
     "onCommand:cmake.outline.buildAll",
     "workspaceContains:**/CMakeLists.txt",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1736,5 +1735,7 @@
     "webpack/terser-webpack-plugin": "^1.4.5",
     "yargs-parser": "^15.0.1",
     "node-fetch": "^2.6.1"
-  }
+  },
+  "extensionPack": [
```
bobbrow, line 1739: This section should be part of a separate PR.

bobbrow (changes requested):

bobbrow (approved):

Issue #1576, opened 2020-11-27
"Build" button builds wrong target when using tasks.json


sclukey:
### Brief Issue Summary

When using a build task, the "Build" buttons in the CMake Project Outline (both the one in-line with the target, and the one in the target's context menu) do not build the correct target. It seems to build the target that was marked as the "Build Target" when the window was reloaded.

Note, some form of this issue also occurred before the new tasks integration, but the buttons used to at least track the current Build Target. Now, it seems completely broken.

### Expected:

Build buttons associated with each target build the target they are associated with

### Apparent Behavior:

![BuildWrongTarget](https://user-images.githubusercontent.com/3012369/100478226-1e8f0980-30b0-11eb-99ad-a0a214fb3711.png)

### CMake Tools Log

```
> Executing task: /usr/bin/cmake --build /path/cmake-tools-issue-repros/build --config Debug --target Program1 -- -j 14 <

ninja: no work to do.

Terminal will be reused by tasks, press any key to close it.
```

### Developer Tools Log

```
[main] Building folder: cmake-tools-issue-repros Program2
[main] Saving open files before configure/build
[build] Starting build
[driver] Start build Program2
[driver] Runnnig pre-configure checks and steps
[proc] Executing command: /usr/bin/cmake --build /path/cmake-tools-issue-repros/build --config Debug --target Program2 -- -j 14
[cmakefileapi-parser] Read reply folder: /path/cmake-tools-issue-repros/build/.cmake/api/v1/reply
[cmakefileapi-parser] Found index files: ["cache-v2-6a5795a39105ba64c2b9.json","codemodel-v2-735653c4ee88b71fb057.json","index-2020-11-27T19-01-15-0378.json","target-Program1-Debug-0ba81c5d52bd99a4c4fe.json","target-Program2-Debug-4b3c411d0f9a00614b05.json"]
[driver] Run _refreshExpansions
[driver] Run _refreshExpansions cb
[build] Build finished with exit code 0
[extension] [2402] cmake.build finished (returned 0)
[cache] Reading CMake cache file /path/cmake-tools-issue-repros/build/CMakeCache.txt
[cache] Parsing CMake cache string
```

### Platform and Versions

- **Operating System**: Linux, Pop!_OS
- **CMake Version**: 3.16.3
- **VSCode Version**: 1.51.1
- **CMake Tools Extension Version**: 1.5.3
- **Compiler/Toolchain**: GCC 9.3.0

[...]

PR #1645, opened 2021-02-03
re-register task provider for new target

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/1576

using tasks.json:
```
{
	"type": "cmake",
	"label": "CMake Build test",
	"command": "build",
	"problemMatcher": [],
	"group": {
		"kind": "build",
		"isDefault": true
	}
}
```
This code will fix these commands (when `"cmake.buildTask": true`)

> `Cmake: build` (builds the default task)
> `Cmake: build target` (chooses a task to build)
> `Tasks: Run Task`
> `Tasks: Run Build Task`

When we use `Cmake: set build target` to change the default build target. 
or when we change the build target by right-clicking:
image


src/cmake-tools.ts
```diff
@@ -1208,6 +1209,16 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
       return;
     }
     await this._setDefaultBuildTarget(target);
+    const drv = await this._cmakeDriver;
+    await this.reRegisterTaskProviderforNewTarget(drv, target);
```
andreeis, line 1213: I wonder if we need a call to reRegister here (when the default target is changed) since we have a call in run build. Is there a scenario where a build does not execute that other call to reRegister and the call here (setDefaultTarget) would matter?

elahehrashedi: Yes, we need to re-register here. The taskprovider is hard-coded for only one target in the code. Ideally, we have to have only one task provider and register it only once, and keep multiple tasks in it. I will open an issue to refactor the task provider.
https://github.com/microsoft/vscode-cmake-tools/issues/1656

andreeis (approved):

PR #1650, opened 2021-02-05
Add support for File API "toolchains" object

benmcmorran:
This changes the way compiler paths are discovered. CMake 3.20 introduces the [`toolchains` object kind in the File API](https://github.com/benmcmorran/vscode-cmake-tools), which directly report the compiler path for each language, among other features. When it's available, using this information is preferable to reading variables directly out `CMakeCache.txt`, because compiler paths are not always persisted to the cache.

For example, the extension would currently fail to discover the correct compiler path with a toolchain file like this:

```cmake
set(CMAKE_CXX_COMPILER arm-none-eabi-g++)
```

This a valid toolchain file, meaning configure and build will succeed, but the value of `CMAKE_CXX_COMPILER` is never persisted to `CMakeCache.txt` because the `set()` command lacks the `CACHE` argument. We see toolchain files like this in the wild, particularly provided by vendors for embedded projects.

With the changes in this PR, the runtime value of `CMAKE_CXX_COMPILER` will be written into the `toolchains` File API object, where we later read it.


andreeis:
Thank you Ben for this contribution. Makefile Tools extension first release kept us busier than usual and we got behind with reviewing PRs. We are catching up now and we will get to this early next week.

andreeis (approved):

PR #1657, opened 2021-02-09
Gcc/clang version analysis improvements

andreeis:
Fix for https://github.com/microsoft/vscode-cmake-tools/issues/1575.
Gcc/clang version analysis should account for localization of word "version", for case variations and accept a dash variant.
Examples of what was not working before:
gcc-Version 10.2.0 (GCC)
gcc 版本 10.2.0 (GCC)

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -130,7 +130,9 @@ async function getClangVersion(binPath: string): Promise<ClangVersion|null> {
     return null;
   }
   const lines = exec.stderr.split('\n');
-  const version_re = /^(?:Apple LLVM|.*clang) version ([^\s-]+)(?:[\s-]|$)/;
+  const versionWord: string = localize("version.word", "version");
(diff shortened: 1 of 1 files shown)


src/kit.ts
```diff
@@ -130,7 +130,9 @@ async function getClangVersion(binPath: string): Promise<ClangVersion|null> {
     return null;
   }
   const lines = exec.stderr.split('\n');
-  const version_re = /^(?:Apple LLVM|.*clang) version ([^\s-]+)(?:[\s-]|$)/;
+  const versionWord: string = localize("version.word", "version");
```
bobbrow, line 133: I think we need to check for both the English and the localized word.  Perhaps we do multiple passes on this?  Does the regex stuff support Unicode characters too?

andreeis: I just verified that this regexp functionality works fine with a string like 版本, so we can count on that.
And yes, good catch, check for both.
I also investigated the LANG environment variables (to enforce that we get english output when running grcc/clang with -v) but since it's not straightforward for me to reproduce and you also said you had problems in the past, including one customer, I think checking twice with 2 regexp is acceptable.

Issue #1578, opened 2020-11-30
Disable the extension when no CMakeLists is present


XsongyangX:
Hello! I like this extension but sometimes I am not working on a CMake project in my workspace and this extension starts to search for kits and configurations right away. It would be nice if the extension would not aggressively start on every workspace in VSCode.

bobbrow:
Thank you for your issue report. I had thought we disabled the extension in cases where there is no CMakeLists.txt found, but we will go back and look at the code to see what is missing.

To help us understand better, are you opening a folder that happens to have a CMakeLists.txt buried in one of the subfolders?

andreeis:
I vaguely remember that there is some kits searching logic triggered before the project has the chance to first configure. We need to fix that, of course.
@XsongyangX , as a workaround you can select a kit, doesn't matter which one (let's say 'unspecified'), then you should see a popup notifying you that there is no CMakeLists found, with an option button to ignore for the current project. Press that and you should not see any activation happening after that. Let us know how that works.

XsongyangX:
> 
> 
> To help us understand better, are you opening a folder that happens to have a CMakeLists.txt buried in one of the subfolders?

I was in a leaf folder. So no subfolders.

@andreeis I'll try that next time.

XsongyangX:
I believe this issue is linked with #1564.

andreeis:
Fix in PR now. Duplicate of https://github.com/microsoft/vscode-cmake-tools/issues/1007.

JanHildenbrandBosch:
Can this in any way be disabled over a configuration? 
We are having currently the issue that the cmakelist.txt will be created after the extension is loaded. But then the extension bar is never loaded.

bobbrow:
@JanHildenbrandBosch this issue is closed so I moved your comment to a new issue.

PR #1689, opened 2021-03-01
Ensure kit selection quickPick doesn't show on non CMake projects

andreeis:
Fix for https://github.com/microsoft/vscode-cmake-tools/issues/1578.
Plus the following:
- rename for commands related to edit cache UI, they were confusingly named openConfiguration
- edit cache was not guarded by full feature set, being available in the UI for non CMake projects
- resetState command is useful even if the current project is seen as non CMake (mostly to identify bugs when the extension analyzes wrong this aspect). It's inconvenient to have resetState not available (it's hard to get out of a bad state when investigating some bugs)
- the extension would show full features (until the next reload) even if the single project in the workspace would be disabled, because we were calculating the state of the workspace too early


src/cmake-tools.ts
```diff
@@ -74,6 +74,7 @@ export enum ConfigureTrigger {
   commandCleanConfigure = "commandCleanConfigure",
   commandConfigureAll = "commandConfigureAll",
   commandCleanConfigureAll = "commandCleanConfigureAll",
+  kitSelection = "kitSelection",
```
bobbrow, line 77: This doesn't appear to be needed anymore

src/extension.ts
```diff
@@ -304,7 +306,7 @@ class ExtensionManager implements vscode.Disposable {
       // We have an active kit. We're good.
       return true;
     }
-    // No kit? Ask the user what they want.
+    // No kit? Ask the user what they want, if showQuickPick allows.
```
bobbrow, line 309: I think this comment should be reverted now

src/extension.ts
```diff
@@ -406,30 +416,56 @@ class ExtensionManager implements vscode.Disposable {
...
+      buildKitTargetArch: "",
+      buildKitVersionMajor: "",
+      buildKitVersionMinor: "",
+      workspaceHash: "",
+    };
+
+    // Don't configure if the current project is not CMake based (it doesn't have a CMakeLists.txt in the sourceDirectory).
+    const sourceDirectory: string = cmt.workspaceContext.config.sourceDirectory;
+    const expandedSourceDirectory: string = util.lightNormalizePath(await expandString(sourceDirectory, { vars: optsVars }));
+    if (await fs.exists(expandedSourceDirectory + "/CMakeLists.txt")) {
```
bobbrow, line 445: We also support sourceDirectory with `CMakeLists.txt` at the end (even though it's technically incorrect), so you should take that into account here as well.

bobbrow (changes requested):

src/extension.ts
andreeis: Ok, can it ever be a different file name than "CMakeLists.txt"? I wonder if we should error for such a setting.

src/extension.ts
```diff
@@ -441,8 +441,11 @@ class ExtensionManager implements vscode.Disposable {
 
     // Don't configure if the current project is not CMake based (it doesn't have a CMakeLists.txt in the sourceDirectory).
     const sourceDirectory: string = cmt.workspaceContext.config.sourceDirectory;
-    const expandedSourceDirectory: string = util.lightNormalizePath(await expandString(sourceDirectory, { vars: optsVars }));
-    if (await fs.exists(expandedSourceDirectory + "/CMakeLists.txt")) {
+    let expandedSourceDirectory: string = util.lightNormalizePath(await expandString(sourceDirectory, { vars: optsVars }));
+    if (path.basename(expandedSourceDirectory).toLocaleLowerCase() !== "cmakelists.txt") {
+      expandedSourceDirectory = path.join(expandedSourceDirectory, "/CMakeLists.txt");
```
bobbrow, line 446: `path.join` shouldn't need you to add the directory separator in `"/CMakeLists.txt"`

bobbrow (approved):

bobbrow (approved):

PR #1690, opened 2021-03-01
Use variables (not hardcoded values) for system path references (extended from PR1063)

andreeis:
Started from PR https://github.com/microsoft/vscode-cmake-tools/pull/1063.


docs/cmake-settings.md
```diff
@@ -22,7 +22,7 @@ Options that support substitution, in the table below, allow variable references
 | `cmake.configureSettings` | An object containing `key:value` pairs, which will be passed to CMake when configuring. The same as passing `-DVAR_NAME=ON` via `cmake.configureArgs`. | `null` (no values) | yes |
 | `cmake.copyCompileCommands`| If not `null`, copies the `compile_commands.json` file generated by CMake to the path specified by this setting whenever CMake successfully configures. |  `null` (do not copy the file) | yes |
 | `cmake.defaultVariants` | Override the default set of variants that will be supplied when no variants file is present. See [CMake variants](variants.md). | | no |
-| `cmake.environment` | An object containing `key:value` pairs of environment variables, which will be passed onto CMake when configuring, build to the compiler, run/debug program and CTest. | `null` (no environment variables) | yes |
+| `cmake.environment` | An object containing `key:value` pairs of environment variables, which will be available when configuring, building, debug, running or testing with CTest. | `null` (no environment variables) | yes |
```
andreeis, line 25: Debugging

package.json
```diff
@@ -1298,7 +1298,7 @@
             "description": "%cmake-tools.configuration.cmake.searchDirs.items.description%"
           },
           "default": [
-            "C:\\MinGW"
+            "${rootPath}/MinGW"
```
andreeis, line 1301: systemDrive

andreeis: Actually, remove rootPath, the same can be achieved by env.SystemDrive.

package.json
```diff
@@ -1298,7 +1298,7 @@
             "description": "%cmake-tools.configuration.cmake.searchDirs.items.description%"
           },
           "default": [
-            "C:\\MinGW"
+            "${env.SystemDrive}/MinGW"
```
bobbrow, line 1301: VS Code variable syntax asks for `env:` instead of `env.`.  We support both because VS did it with `.` instead of `:` and people get confused.  But for VS Code we should use a colon `:` if we set this as a default.

bobbrow (approved):
One minor nit, then it's good to check in.

bobbrow (approved):

Issue #1518, opened 2020-11-04
Suggestion：Do not pop up a message to remind that cmake tools is initializing


chouzz:
Suggestion：

Please don't bother me when cmake tools extension activate.  Every time I open vscode, a pop-up window will pop up in the lower right corner to prompt **cmake tools initializing**, it may not be very user friendly for some people. There's no reason for me to know that cmake tools is initializing.

Maybe you can cancel this prompt or replace it with the status bar.

andreeis:
Thank you for your idea. We may not be able to implement it immediately, but we will track the community reactions to determine where this might fit into our backlog.

PR #1695, opened 2021-03-02
Do not pop up a message to remind that cmake tools is initializing, for all users (codespaces or non-codespaces)

elahehrashedi:
bug fix for https://github.com/microsoft/vscode-cmake-tools/issues/1518

Edit: the initialization message is removed altogether.


package.json
```diff
@@ -1703,7 +1703,7 @@
     "tslint": "^5.20.1",
     "typedoc": "^0.20.25",
     "typescript": "^4.1.5",
-    "vscode": "^1.1.36",
+    "vscode": "^1.53.0",
     "vscode-nls-dev": "^3.3.2",
     "webpack": "^5.22.0",
```
andreeis, line 1708: These version changes are causing CI to fail. No idea why but lets' experiment with versions and see if we can unblock the build.

bobbrow: There shouldn't be a need to update the version so high.  I'm not sure when this API went in, but you don't need latest.

I believe "vscode" doesn't go above a certain number.  They changed the way they distribute the API around 1.40 or so, I think.  cpptools uses this now: 
```
   "@types/vscode": "1.52.0",
```
...but I think there's more to it than just making this small change.  You might want to sync with @Colengms since he made that change a while back.

src/extension.ts
```diff
@@ -1329,14 +1329,17 @@ export async function activate(context: vscode.ExtensionContext) {
   vscode.workspace.registerTextDocumentContentProvider('cmake-tools-schema', new SchemaProvider());
   vscode.commands.executeCommand("setContext", "inCMakeProject", true);
 
-  await vscode.window.withProgress(
+  if (vscode.env.uiKind === vscode.UIKind.Web) {
+    // Do not prompt for users of Web-based Codespaces.
```
bobbrow, line 1333: I think we also want to check for Codespaces on Desktop too.  Talk to @Colengms about what needs to be checked for that. You may not need to do the UIKind check at all...

elahehrashedi: I have changed the way we find codespaces users based on cpp-tools

package.json
elahehrashedi: It is changed back to older version

andreeis (approved):

Issue #1477, opened 2020-10-08
c11 is being used instead of gnu17


sean-mcmanus:
See the repro at https://github.com/microsoft/vscode-cpptools/issues/6279 .

andreeis:
Possibly related https://github.com/microsoft/vscode-cmake-tools/issues/1489.

Colengms:
This issue is due to falling back to a hard-coded default, and not querying the compiler to determine what its default actually is.  To avoid requiring all config providers from having to do this, we've made a change to the cpptools-api to allow IntelliSenseMode, cStandard and cppStandard to be omitted, and automatically determined by cpptools based on the compilerPath and compilerArgs.  I think we have some other changes we'd like make to the API before publishing a new version.  Once a new version of cpptools-api is published, CMake Tools can be updated to omit these fields.

PR #1696, opened 2021-03-02
not send a default standard to IntelliSense

elahehrashedi:
bug fix https://github.com/microsoft/vscode-cmake-tools/issues/1477


src/cpptools.ts
```diff
@@ -117,7 +116,7 @@ export function parseCompileFlags(cptVersion: cpt.Version, args: string[], lang?
   const can_use_gnu_std = (cptVersion >= cpt.Version.v4);
   const iter = args[Symbol.iterator]();
   const extraDefinitions: string[] = [];
-  let standard: StandardVersion = (lang === 'C') ? 'c11' : 'c++17';
+  let standard: StandardVersion | undefined = undefined;
```
andreeis, line 119: remove " = undefined", declaring with " | undefined" and not initializing means initialized with undefined. The CI fails with a warning about this also.

src/cpptools.ts
```diff
@@ -22,13 +22,12 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 const log = createLogger('cpptools');
 
 type Architecture = 'x86' | 'x64' | 'arm' | 'arm64' | undefined;
-type StandardVersion = "c89" | "c99" | "c11" | "c17" | "c++98" | "c++03" | "c++11" | "c++14" | "c++17" | "c++20"
-  | "gnu89" | "gnu99" | "gnu11" | "gnu17" | "gnu++98" | "gnu++03" | "gnu++11" | "gnu++14" | "gnu++17" | "gnu++20";
+type StandardVersion = "c89" | "c99" | "c11" | "c17" | "c++98" | "c++03" | "c++11" | "c++14" | "c++17" | "c++20" | "gnu89" | "gnu99" | "gnu11" | "gnu17" | "gnu++98" | "gnu++03" | "gnu++11" | "gnu++14" | "gnu++17" | "gnu++20";
```
bobbrow, line 25: I think the C++ and C versions were split onto two lines to make it easier to find where they start and end.  Can you change this back?

src/cpptools.ts
```diff
@@ -22,13 +22,12 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
...
 
 
 export interface CompileFlagInformation {
   extraDefinitions: string[];
-  standard: StandardVersion;
+  standard: StandardVersion | undefined;
```
bobbrow, line 30: The more common way to do this when you're dealing with optional fields in an interface is:
```ts
  standard?: StandardVersion;
```

src/cpptools.ts
elahehrashedi: I copied this line from cpptools API (SourceFileConfiguration interface in api.d.ts) to make sure we are not missing any item. My intention was not to remove the new line. I can keep it like this to be similar to cpptools API, or split it between C and C++.

bobbrow: I would like to keep the c++ and c versions on different lines.  [In api.ts of the cpptools API](https://github.com/microsoft/vscode-cpptools-api/blob/8ae425a623649b3f26bebad8e0917f11848506e8/api.ts#L176), you will see that they are split this way.  When api.d.ts is generated, the generator removes the newline.

andreeis (approved):

PR #1697, opened 2021-03-02

csigs:


translations_auto_pr.js
```diff
@@ -130,10 +130,7 @@ console.log(`pushing to remove branch (git push -f origin ${branchName})`);
 cp.execSync(`git push -f origin ${branchName}`);
 
 console.log("Checking if there is already a pull request...");
-const octokit = new Octokit({auth: {
-    username: authUser,
-    password: authToken}
-});
+const octokit = new Octokit.Octokit({auth: authToken});
```
sean-mcmanus, line 133: This is the fix I made...looks like that change got bundled into this one when I ran the build to test the change.

sean-mcmanus: Fixes a regression from updating the octokit version in https://github.com/microsoft/vscode-cmake-tools/pull/1475

bobbrow (approved):

PR #1701, opened 2021-03-04
User defined additional kits

andreeis:
Inspired from the idea of https://github.com/microsoft/vscode-cmake-tools/pull/744.


package.json
```diff
@@ -1573,6 +1573,12 @@
             }
           }
         },
+        "cmake.additionalKitsFile": {
```
bobbrow, line 1576: >additionalKitsFile [](start = 15, length = 18)

Let's make it `additionalKitsFiles` and have it be array of string.

package.nls.json
```diff
@@ -119,5 +119,6 @@
     "cmake-tools.configuration.cmake.statusbar.advanced.length.description": "Configures the maximum length of visible text in 'compact' mode.",
     "cmake-tools.configuration.views.cmake.folders.description": "Folders",
     "cmake-tools.configuration.views.cmake.outline.description": "Project Outline",
+    "cmake-tools.configuration.cmake.additionalKitsFile.description": "Path to a custom kit file",
```
bobbrow, line 122: Update the description to match the array type

src/kitsController.ts
```diff
@@ -46,17 +47,59 @@ export class KitsController {
...
     }
-    const kitsWatcher = chokidar.watch(KitsController._workspaceKitsPath(cmakeTools.folder), {ignoreInitial: true});
+
+    const folderKitsFiles: string[] = [KitsController._workspaceKitsPath(cmakeTools.folder)];
+    const additionalKitsFile: string | null = await KitsController.expandAdditionalKitsFile(cmakeTools);
+    if (additionalKitsFile) {
+      folderKitsFiles.push(additionalKitsFile);
+    }
+
+    const kitsWatcher = chokidar.watch(folderKitsFiles, {ignoreInitial: true});
```
bobbrow, line 98: Please double check this works with files outside the workspace folder

src/kitsController.ts
```diff
@@ -119,17 +162,30 @@ export class KitsController {
...
-      // Read folder kits
+      // Read default folder kits
       this.folderKits = await readKitsFile(KitsController._workspaceKitsPath(this.folder));
-      const current = this.cmakeTools.activeKit;
-      if (current) {
-        const already_active_kit = this.availableKits.find(kit => kit.name === current.name);
-        // Set the current kit to the one we have named
-        await this.setFolderActiveKit(already_active_kit || null);
+
+      // Read additional folder kits
```
bobbrow, line 170: >// Read additional folder kits [](start = 6, length = 30)

Maybe the reading of kits can be factored out and reused by `init`

andreeis:
The tests fail on clean develop as well, not sure what could have introduced this. The recent commits don't look like touching this area. Investigating...

package.json
bobbrow: Please add `"Items"` property to enforce string for all array items.

---
In reply to: [587658754](https://github.com/Microsoft/vscode-cmake-tools/pull/1701#discussion_r587658754) [](ancestors = 587658754)

src/kitsController.ts
```diff
@@ -46,17 +48,56 @@ export class KitsController {
...
+        userHome: paths.userHome,
+        workspaceFolder: cmakeTools.workspaceContext.folder.uri.fsPath,
+        workspaceFolderBasename: path.basename(cmakeTools.workspaceContext.folder.uri.fsPath),
+        workspaceHash: "",
+        workspaceRoot: cmakeTools.workspaceContext.folder.uri.fsPath,
+        workspaceRootFolderName: path.basename(cmakeTools.workspaceContext.folder.uri.fsPath),
+      }
+    };
+
+    await Promise.resolve(additionalKitFiles.forEach(async kitFile => {
```
bobbrow, line 81: We talked about using Promise.all here instead.  See `scanForVSKits` for an example.

package.json
```diff
@@ -1631,6 +1631,12 @@
             }
           }
         },
+        "cmake.additionalKits": {
+          "type": "array",
+          "default": [],
```
bobbrow, line 1636: It looks like my other comment about adding the `"items"` property here was resolved by accident.  We still need to provide stronger validation here.

bobbrow (changes requested):

package.json
andreeis: Oh right, I missed that.

bobbrow (approved):

PR #1725, opened 2021-03-19
CMake Presets

xisui-MSFT:
Added [CMake Presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) support.


H-G-Hristov:
It looks like there are more features added to presets for the next CMake version? Do you plan to support them with this PR too? https://cmake.org/cmake/help/git-stage/release/index.html
Maybe the next version will be released about the same time.

bobbrow:
CMake 3.20 is available now.  We intend to have support for configure, build, and test presets in 1.7 of CMake Tools.

src/drivers/legacy-driver.ts
```diff
@@ -94,9 +111,24 @@ export class LegacyCMakeDriver extends CMakeDriver {
     });
   }
 
-  static async create(cmake: CMakeExecutable, config: ConfigurationReader, kit: Kit|null, workspaceFolder: string | null, preconditionHandler: CMakePreconditionProblemSolver, preferredGenerators: CMakeGenerator[]): Promise<LegacyCMakeDriver> {
+  static async create(cmake: CMakeExecutable,
+                      config: ConfigurationReader,
+                      useCMakePresets: boolean,
+                      kit: Kit|null,
+                      configurePreset: ConfigurePreset | null,
```
bobbrow, line 118: Why does the legacy driver need these arguments in the constructor?  I didn't think they would be used.  Or do we try to invoke cmake with the values in the presets?

schemas/kits-schema.json
```diff
@@ -45,7 +45,11 @@
             },
             "visualStudio": {
                 "type": "string",
-                "description": "Name of the Visual Studio product"
+                "description": "Instance ID of the Visual Studio product"
+            },
+            "visualStudioVersion": {
```
bobbrow, line 50: Changing kits is not recommended if we can manage it.  Can this information be supplied some other way, or be internal for when we hydrate the Kit objects?  I thought that Kits wouldn't be used for Presets.

src/expand.ts
```diff
@@ -22,14 +22,20 @@ const log = createLogger('expand');
  * references will be available when performing an expansion. Those guaranteed
  * variables are specified as properties on this interface.
  */
-export interface RequiredExpansionContextVars {
-  buildKit: string;
-  buildType: string;
+interface RequiredExpansionContextVars {
+  [key: string]: string;
```
bobbrow, line 26: It seems odd to put this in the "required" section since it's unbounded.  I think it makes more sense to put it in the derived classes even though it means duplicating it for each.

src/expand.ts
```diff
@@ -79,9 +91,33 @@ export interface ExpansionOptions {
  * @returns A string with the variable references replaced
  */
 export async function expandString(tmpl: string, opts: ExpansionOptions) {
+  if (!tmpl) {
+    return tmpl;
+  }
+
+  const MAX_RECURSION = 50;
```
bobbrow, line 98: This seems a bit high for variable expansion.  A smaller number like 5 or 10 seems more reasonable.

src/expand.ts
```diff
@@ -161,5 +228,5 @@ export async function expandString(tmpl: string, opts: ExpansionOptions) {
 
   let final_str = tmpl;
   subs.forEach((value, key) => { final_str = replaceAll(final_str, key, value); });
-  return final_str;
+  return { result: final_str, didReplacement: subs.size > 0};
```
bobbrow, line 231: I think we have a few places where we say to replace "${var}" with "${var}" because it's effectively not supported for that expansion.  Is this going to say those are replaced and "recurse" until the MAX is hit?

src/drivers/legacy-driver.ts
xisui-MSFT: They are used the same way in all drivers. `create()` calls `_baseInit()`, which sets the initial value for all presets. And then they are used in driver.ts (but may not be used directly in *-driver.ts). The reason we need them on initialization is that drivers may not exist at the same time as CMakeTools instances.

schemas/kits-schema.json
xisui-MSFT: Kits are reused as storage for compiler information. It's an optional field, so it's compatible with the previous scans, and therefore rescan is not required if the users don't need some of the new features. So I think it's acceptable to make this change.

The other option we have is to invoke and parse vswhere by calling `vsInstallations()` in visual-studio.ts, and then iterate through all the instances for each VS kit to look for the versions. This feels like a less ideal way to me.

src/expand.ts
xisui-MSFT: Yeah randomly picked 50... 10 sounds good to me, and we can easily adjust this based on the customers' feedbacks.

src/expand.ts
xisui-MSFT: Recursion is off by default, and so it's not a concern now. However, we can add a check to filter out the case where $*{*} === $*{*}

src/cmake-tools.ts
```diff
@@ -123,6 +141,127 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+      const drv = await this._cmakeDriver;  // Use only an existing driver, do not create one
+      if (drv) {
+        try {
+          this._statusMessage.set(localize('reloading.status', 'Reloading...'));
+          await drv.setConfigurePreset(this._configurePreset);
+          this.workspaceContext.state.configurePresetName = configurePreset;
+          this._statusMessage.set(localize('ready.status', 'Ready'));
+        } catch (error) {
+          vscode.window.showErrorMessage(localize('unable.to.set.config.preset', 'Unable to set configure preset "{0}".', error));
+          this._statusMessage.set(localize('error.on.switch.config.preset', 'Error on switch of configure preset ({0})', error.message));
```
andreeis, line 178: {0} in between " is intentional? I get the popup message with {0} as text and not resolved with error.

src/presetsController.ts
```diff
@@ -0,0 +1,844 @@
...
+    }
+  }
+
+  async setBuildPreset(presetName: string, needToCheckConfigurePreset: boolean = true): Promise<void> {
+    if (needToCheckConfigurePreset && presetName !== preset.defaultBuildPreset.name) {
+      preset.expandConfigurePresetForPresets('build');
+      const _preset = preset.getPresetByName(preset.allBuildPresets(), presetName);
+      if (_preset?.configurePreset !== this._cmakeTools.configurePreset?.name) {
+        log.error(localize('build.preset.configure.preset.not.match', 'Build preset {0}: The configure preset does not mach the selected configure preset', presetName));
+        return;
```
andreeis, line 668: typo: "does not mach" instead of match

src/cmake-tools.ts
andreeis: and it's in several other places, just check for pattern "{

xisui-MSFT: This code is actually copied from kit (cmekt-tools.ts, line 732). How did you get that popup message?

schemas/kits-schema.json
xisui-MSFT: Sounds good I'll change this.

src/cmake-tools.ts
andreeis: By connecting the preset to a fake artificial kit I had tested for something else (and forgot that it was not valid when I selected it). It had only "name" in it. This will not be a true scenario but here are all the issues I found through this, which I believe should be fixed because who knows how else the following may be reachable via a real scenario.

1. add a new preset --> scan for compilers --> select invalid kit --> name the new preset --> select again the new preset (by the way, I felt a bit confused that by "add a new preset" I didn't automatically select it and I have to press again the button in the status bar) --> the message
2. add a new valid preset and corresponding build/test created from it --> change the configure preset to the invalid one above --> no more message (inconsistent with the above, but message will appear after any future reload) --> build/test presets show unset in status bar but from output channel I think they're set (incompatibility message). Also if I select the original good configure preset, the build/test presets magically change from unset to what they were before.
3. Continuing (2.), a reset state now is working but if I reset state after a reload, it will never do a good reset, it will pop up the configure preset quickPick: if I cancel --> reset does reload but from output channel I deduce that it didn't reset, if I select a good configure preset now then it will properly reset all the state
4. Continuing (2.): after a few more steps which I can't reproduce anymore, any reload will show the bottom right {0} message (expected, configure preset can't be set, it's invalid) but the quickpick isn't automatically shown as I was used to see it before every time a configure preset is invalid (can't be set) or there isn't one at all yet.
5. I have several presets tied to similar invalid kits and after playing with them (set, unset, etc...) I go through various inconsistent states: sometimes when I set such preset I don't see a bottom right message at all (unless I reload), or I see the popup but UI button takes the name of the new invalid preset.... or I see the popup but the button is not changing from "No configure preset selected".

andreeis:
More issues I notice while testing:

1. not sure how important, changes of a preset directly into the json don't immediately reflect on the status bar UI. But output channel has messages that show the changes were processed. If I change the display name of a preset, it will be updated in the status bar button after reselecting from the quick pick (no big deal) but change of name (equivalent to deletion) will leave the status bar buttons for build/test presets (if associated to the previous) in incorrect states, also the configure preset button. All will be truthful again after a reload.

2. add new preset --> scan for compilers --> scan for compilers. It's confusing to see 2 scan for compilers when only the second one performs a scan. Also, at some point I saw "scan for compilers" progress popup on bottom right, followed by scan for kits... now I see only scan for kits. What is "scan for compilers" and why I don't see it anymore, it's too fast and only the first time it was notice-able or...?

3.

src/folders.ts
```diff
@@ -1,31 +1,101 @@
...
 
+  // Go through the decision tree here since there would be dependency issues if we do this in config.ts
+  get useCMakePresets(): boolean {
+    if (this.cmakeTools.workspaceContext.config.useCMakePresets === 'automatic') {
+      // TODO (P1): check if configured with kits + vars
+      // // Always check if configured before since the state could be reset
+      // const state = this.cmakeTools.workspaceContext.state;
+      // const configuredWithKitsVars = !!(state.activeKitName || state.activeVariantSettings?.size);
+      // return !configuredWithKitsVars || (configuredWithKitsVars && (this.presetsController.cmakePresetsExist || this.presetsController.cmakeUserPresetsExist));
+      return this.presetsController.presetsFileExist;
```
andreeis, line 62: Double commented comment, commented typescript.... maybe you forgot to delete a block of source code here?

src/preset.ts
```diff
@@ -0,0 +1,1222 @@
...
+  displayName?: string;
+  description?: string;
+  hidden?: boolean;
+  inherits?: string | string[];
+  environment?: { [key: string]: null | string };
+  vendor?: VendorType;
+
+  __expanded?: boolean; // Private field to indicate if we have already expaned thie preset.
+}
+
```
andreeis, line 39: "expanded this"

src/preset.ts
```diff
@@ -0,0 +1,1222 @@
...
+function getVendorForConfigurePresetHelper(preset: ConfigurePreset, allowUserPreset: boolean = false): VendorType | Vendor_VsSettings | null {
+  if (preset.__expanded) {
+    return preset.vendor || null;
+  }
+
+  if (referencedConfigurePresets.has(preset.name)) {
+    // Refernced this preset before, but it doesn't have a configure preset. This is a circular inheritance.
+    log.error('circular.inherits.in.config.preset', 'Circular inherits in configure preset {0}', preset.name);
+    return null;
+  }
```
andreeis, line 359: "referenced"... see it in 5 more places throughout this PR, no need to point them individually.

src/folders.ts
xisui-MSFT: Commented out part is P1, which is what this code will look like later.

test/extension-tests/single-root-UI/project-folder/CMakeUserPresets.json
```diff
@@ -0,0 +1,19 @@
...
+      "description": "Sets generator, build and install directory, vcpkg",
+      "generator": "Unix Makefiles",
+      "binaryDir": "${workspaceFolder}/build",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_C_COMPILER": "gcc",
+        "CMAKE_CXX_COMPILER": "g++"
+      },
+      "environment": {
+        "TEST_VARIANT_ENV": "0cbfb6ae-f2ec-4017-8ded-89df8759c502"
```
bobbrow, line 15: should it be different from the non-user var so you can make sure it is set?

src/preset.ts
```diff
@@ -0,0 +1,1222 @@
...
+            arch = preset.architecture.value;
+          } else {
+            log.warning(localize('no.cl.arch', 'Configure preset {0}: No architecture specified for cl.exe, using x86 by default', preset.name));
+          }
+          const toolsetArchRegex = /(host=\w+),?/i;
+          const toolsetVsVersionRegex = /version=(\w+),?/i;
+          const noToolsetArchWarning = localize('no.cl.toolset.arch', 'Configure preset {0}: No toolset architecture specified for cl.exe, using x86 by default', preset.name);
+          const noToolsetVsVersionWarning = localize('no.cl.toolset.version', 'Configure preset {0}: No toolset version specified for cl.exe, using latest by default', preset.name);
+          const matchToolsetArchAndVersion = (toolset: string) => {
+            const tollsetArchMatches = toolset.match(toolsetArchRegex);
```
bobbrow, line 505: [sp] `toolsetArchMatches`

package.json
```diff
@@ -1666,6 +1806,16 @@
           "type": "boolean",
           "default": true,
           "description": "%cmake-tools.configuration.cmake.exportCompileCommandsFile.description%"
+        },
+        "cmake.useCMakePresets":  {
+          "type": "string",
+          "enum": [
+            "true",
```
bobbrow, line 1813: I think it looks a little weird in the settings.json to have these booleans be strings.  I have the setting right next to another boolean setting and one has quotes while the other does not. Is there a better way?

src/presetsController.ts
```diff
@@ -0,0 +1,844 @@
...
+    const tabSize = config.get<number>('tabSize') || 4;
+    const insertSpaces = config.get<boolean>('insertSpaces') || true;
+    return { insertSpaces, tabSize };
+  }
+
+  async updatePresetsFile(presetsFile: preset.PresetsFile, isUserPresets = false): Promise<boolean> {
+    const presetsFilePath = isUserPresets? this.userPresetsPath : this.presetsPath;
+    const indent = this.getIndentationSettings();
+    try {
+        await fs.writeFile(presetsFilePath, JSON.stringify(presetsFile, null, indent.insertSpaces ? indent.tabSize : '\t'));
```
bobbrow, line 824: When I created my first preset, I was expecting the CMakePresets.json file to open in the editor so I could confirm the values in the preset.  Should we open (or refocus) the file at this point?

src/presetsController.ts
```diff
@@ -0,0 +1,844 @@
...
+
+        newPreset.name = name;
+        await this.addPresetAddUpdate(newPreset, 'buildPresets');
+      }
+
+      return true;
+    }
+  }
+
+  async addTestPreset(): Promise<boolean> {
```
bobbrow, line 447: Something is broken with this function.  I ran the command to add a test preset and got two entries added (neither of them looked correct):
```
        {
            "name": "__placeholder__",
            "description": "",
            "displayName": "",
            "configurePreset": "debug-x64"
        },
        {
            "name": "debug-x64-test",
            "description": "",
            "displayName": "",
            "configurePreset": "debug-x64"
        },
```

src/presetsController.ts
```diff
@@ -0,0 +1,844 @@
...
+      const option_later = localize('later', 'later');
+      const result = await vscode.window.showErrorMessage(message_noConfigurePreset, option_selectConfigurePreset, option_later);
+      if (result === option_selectConfigurePreset && await vscode.commands.executeCommand('cmake.selectConfigurePreset')) {
+        return this._cmakeTools.configurePreset;
+      }
+    }
+    return selectedConfigurePreset;
+  }
+
+  async selectBuildPreset(): Promise<boolean> {
```
bobbrow, line 629: I ran into an issue where both my presets did not appear in the quick pick:
image

xisui-MSFT: Only the build presets with configurePreset set to the currently active configure preset will show up in the selector

src/presetsController.ts
xisui-MSFT: Sure we can do this

src/presetsController.ts
xisui-MSFT: Probably forgot something when moving the name input box to the end of the function. I'll take a look.

package.json
xisui-MSFT: We can use any strings with similar meanings such as 'yes' and 'no'. true/false/undefined is a harder choice since config.ts doesn't distinguish the source of the settings, so we need to read config only for this setting, which makes maintenance harder.

src/presetsController.ts
bobbrow: Haha, ok.  I get it.

package.json
bobbrow: I agree.  We have a bunch of settings that are type: boolean with default: null and vs code kicks the user back to json because it can't be edited in the UI.  If we used "oneOf" with boolean or enum: auto, would that help?

src/presetsController.ts
```diff
@@ -0,0 +1,844 @@
...
+  async setConfigurePreset(presetName: string): Promise<void> {
+    // Load the configure preset into the backend
+    await vscode.window.withProgress(
+        {
+          location: vscode.ProgressLocation.Notification,
+          title: localize('loading.config.preset', 'Loading configure preset {0}', presetName),
+        },
+        () => this._cmakeTools.setConfigurePreset(presetName),
+    );
+    await vscode.window.withProgress(
```
bobbrow, line 597: When I change configure presets but have a build preset from another configure preset active, it looks like the status bar does not update and still shows the build preset from the old configure preset as active.  Should we clear it?

src/drivers/driver.ts
```diff
@@ -1231,47 +1359,75 @@ export abstract class CMakeDriver implements vscode.Disposable {
     }
   }
 
-  async getCMakeBuildCommand(target: string): Promise<proc.BuildCommand|null> {
-    const gen = this.generatorName;
-    target = this.correctAllTargetName(target);
+  async getCMakeBuildCommand(target?: string): Promise<proc.BuildCommand|null> {
+    if (this.useCMakePresets) {
```
bobbrow, line 1363: I'm having a problem here because this is always returning false for me even though cmake.useCMakePresets is set to true in my settings.  It doesn't look like `this._useCMakePresets` is ever updated once the driver is created.

src/cmake-tools.ts
```diff
@@ -406,16 +544,39 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+                                               this.useCMakePresets,
+                                               this.activeKit,
+                                               this.configurePreset,
+                                               this.buildPreset,
+                                               this.testPreset,
+                                               workspace,
+                                               preConditionHandler,
+                                               preferredGenerators);
         }
     } finally { this._statusMessage.set(localize('ready.status', 'Ready')); }
```
bobbrow, line 582: I can't add a comment on line 588, but there is a problem with `registerTaskProvider` when using presets because `this.tasksBuildCommandDrv` will try to generate the build command, but the build preset is not yet set, so it fails and returns null.

src/presetsController.ts
```diff
@@ -0,0 +1,844 @@
...
+            } else {
+              log.debug(localize('user.selected.compiler', 'User selected compiler {0}', JSON.stringify(chosen_kit)));
+              newPreset = {
+                name: '__placeholder__',
+                displayName: chosen_kit.kit.name,
+                description: chosen_kit.description,
+                generator: chosen_kit.kit.preferredGenerator?.name,
+                toolset: chosen_kit.kit.preferredGenerator?.toolset,
+                architecture: chosen_kit.kit.preferredGenerator?.platform,
+                binaryDir: '${sourceDir}/out/build/${presetName}',
```
bobbrow, line 293: This is at odds with the default value of `cmake.buildDirectory`.  We don't normally add `out` to the path.  I'll send Erika an email and ask what she thinks.

src/presetsController.ts
xisui-MSFT: Yeah I only added an error message in that case. I think we should clear it.

src/drivers/driver.ts
xisui-MSFT: The driver is disposed when switching... Looks like I need to reset it to null as well... This is great finding and might be the reason why my test was failing yesterday.

src/cmake-tools.ts
xisui-MSFT: I didn't notice this function. I'll do reregisterTaskProvider in setBuildPresets.

src/preset.ts
```diff
@@ -0,0 +1,1229 @@
...
+      if (preset.environment[key]) {
+        preset.environment[key] = await expandString(preset.environment[key]!, expansionOpts);
+      }
+    }
+  }
+
+  removeNullEnvVars(preset);
+
+  // Expand other fields
+  if (preset.binaryDir) {
```
bobbrow, line 615: The spec says that binaryDir must always be set unless it is inherited from another preset, but it is required to be set somewhere. I just got a crash when creating a new configure preset that just had a "name".
```
[rollbar] Unhandled exception: Failed writing to file .cmake\api\v1\query\client-vscode\query.json Error: EPERM: operation not permitted, mkdir 'C:\Program Files\Microsoft VS Code\.cmake' {}
```

bobbrow: binaryDir

An optional string representing the path to the output binary directory. This field supports macro expansion. If a relative path is specified, it is calculated relative to the source directory. If binaryDir is not specified, it must be inherited from the inherits preset (unless this preset is hidden).

bobbrow: I don't think this PR does the relative path support yet either.

package.json
xisui-MSFT: Yeah this is a good idea.

bobbrow: It seems the oneOf approach requires editing the setting in JSON, but I opened a feature request on the VS Code team to evaluate whether we can get tri-state bools somehow.  https://github.com/microsoft/vscode/issues/120991

xisui-MSFT:
> More issues I notice while testing:
> 
> 1. not sure how important, changes of a preset directly into the json don't immediately reflect on the status bar UI. But output channel has messages that show the changes were processed. If I change the display name of a preset, it will be updated in the status bar button after reselecting from the quick pick (no big deal) but change of name (equivalent to deletion) will leave the status bar buttons for build/test presets (if associated to the previous) in incorrect states, also the configure preset button. All will be truthful again after a reload.
> 2. add new preset --> scan for compilers --> scan for compilers. It's confusing to see 2 scan for compilers when only the second one performs a scan. Also, at some point I saw "scan for compilers" progress popup on bottom right, followed by scan for kits... now I see only scan for kits. What is "scan for compilers" and why I don't see it anymore, it's too fast and only the first time it was notice-able or...?

@andreeis Thanks for the feedback!
1. Fixed.
2.  Changed the first one to "create from compilers". `scanForCompilers` reuses `scanForKits`.  I think the message diff is acceptable.

src/presetsController.ts
bobbrow: We can leave it as 'out'

bobbrow (approved):

PR #1728, opened 2021-03-23
Fix configure hang and full/partial feature set view

andreeis:
Fix configure hang when locate missing CMakeLists.txt.
Fix bug about full feature set.


src/cmake-tools.ts
```diff
@@ -319,18 +319,12 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
-              drv.config.updatePartial({sourceDirectory: joinedPath});
-            } else {
-              const reloadWindowButton = localize('reload.window', 'Reload Window');
-              const reload = await vscode.window.showErrorMessage(localize('setting.sourceDirectory.failed.needs.reload.window',
-                    'Something went wrong while updating the cmake.sourceDirectory setting. Please run the "Reload window" command for the change to take effect.'),
-                    reloadWindowButton);
-              if (reload === reloadWindowButton) {
-                vscode.commands.executeCommand('workbench.action.reloadWindow');
+            // tslint:disable-next-line: no-floating-promises
+            this.getCMakeDriverInstance().then (d => {
```
bobbrow, line 323: `cmakePreConditionProblemHandler` is invoked by CMakeDriver (in `_beforeConfigureOrBuild`).  It would be nice if you could just pass in the `config` or the `driver` directly to avoid making this call.

src/extension.ts
```diff
@@ -202,8 +202,7 @@ class ExtensionManager implements vscode.Disposable {
   // because currently there is no way of reading a context variable
   // like cmake:enableFullFeatureSet, to apply the OR operation on it.
   private _enableFullFeatureSetOnWorkspace = false;
-  public enableFullFeatureSet(fullFeatureSet: boolean, folder: vscode.WorkspaceFolder) {
-    this.getFolderContext(folder).ignoreCMakeListsMissing = !fullFeatureSet;
+  public enableFullFeatureSet(fullFeatureSet: boolean) {
```
bobbrow, line 205: Just for my benefit, it looks like we can't do the enable/disable per workspace in multi-root. Is it ever the case that one workspace might not want it enabled and end up disabling it for the whole workspace?

src/extension.ts
```diff
@@ -1140,7 +1139,7 @@ class ExtensionManager implements vscode.Disposable {
   // depending on their 'ignoreCMakeListsMissing' state variable.
   enableWorkspaceFoldersFullFeatureSet() {
     for (const cmtFolder of this._folders) {
-      this.enableFullFeatureSet(!this.getFolderContext(cmtFolder.folder)?.ignoreCMakeListsMissing, cmtFolder.folder);
+      this.enableFullFeatureSet(!this.getFolderContext(cmtFolder.folder)?.ignoreCMakeListsMissing);
```
bobbrow, line 1142: Does it make sense to call this function N times if we don't store state per folder?

src/cmake-tools.ts
andreeis: I tried but config.updatePartial generates another configure during this configure, causing error popup. I wasn't able to quickly rewire everything for this not to happen. It looks like the ".then" ensures updatePartial happens at a better timing. Do you think I overlook something?

src/extension.ts
andreeis: Once I read your latest comments I though so too and I removed much of the functionality but in the end with testing I brought it all back. If the last folder that is attempted to be configured (via a configureAll for example) prompts and the user chooses to ignore, then without the workspace folder by folder check we end up with the UI disabled.

src/extension.ts
andreeis: We can call it once but only if we do a boolean processing folder by folder before that. I'll submit an update shortly.

src/cmake-tools.ts
bobbrow: Can you add an optional parameter to `updatePartial` that instruct it not to fire the events that trigger a new configure?  The configure will happen after this function returns so it should be safe to disable those events.  (you can correct me if I'm wrong though)

src/extension.ts
bobbrow: If we track all of the "enable" states for all folders in an array `[true, false, true]` (just an example -- it doesn't need to be an array), then whenever this array changes, we set enabled to true if any folder is true, and false if all folders are false.

src/extension.ts
```diff
@@ -1109,9 +1114,15 @@ class ExtensionManager implements vscode.Disposable {
...
     telemetry.logEvent("resetExtension");
-    return this.mapCMakeToolsFolder(cmt => cmt.resetState(), folder);
+    if (folder) {
+      await this.mapCMakeToolsFolder(cmt => cmt.resetState(), folder);
+    } else {
+      await this.mapCMakeToolsAll(cmt => cmt.resetState());
+    }
+
+    vscode.commands.executeCommand('workbench.action.reloadWindow');
   }
```
andreeis, line 1127: Only the active folder was resetting state. I imagine that with this command, the user wants all projects to reset state. Without this change it was super annoying to test this PR in multi root.

bobbrow (approved):

Issue #1707, opened 2021-03-06
suggestion: remove -T flag when running CTest


leungyauming:
## Current

CMake Extension will run CTest with flag -T when running tests through status bar.
And this will lead to `Cannot find file: D:/Projects/cpp/bound/build/DartConfiguration.tcl`.

## Suggestion

Remove -T flag when running tests.

sureshjoshi:
I don't know if this is a new bug, but it's definitely been a problem for me on two computers right now. The DartConfiguration.tcl can be found, assuming that `include (CTest)` exists in CMakeFiles.txt, but the action of running `Run Tests` now fails. 

When I go to the terminal, to my build directory, I can use `ctest` and all my tests run. Using the `Run Tests` command of `/usr/local/bin/ctest -j8 -C Debug -T test --output-on-failure` doesn't run the tests.

When I manually go into the terminal and run `/usr/local/bin/ctest -j8 -C Debug -T start --output-on-failure`, and then run `Run Tests` - everything works again.

andreeis:
Thank you for opening this report. With a quick try on windows I wasn't able to reproduce this but I will explore ctest on linux and on more recent versions as well.

In the meantime, maybe we can notice something by looking at the logs in the output channel. Set the logging verbosity to debug "cmake.loggingLevel": "Debug" and run the test command again.

@leungyauming, when you see that DartConfiguration.tcl is missing, do you have "include (CTest)" in your CMakeLists.txt?
Please share us the output channel logs with the ctest command with and without -T, since you say that simply removing -T unblocks you.

@sureshjoshi, regarding running the ctest with "-T start" first. Once you do that, how do you get again into the state of "-T test" not running the tests? Can you share the terminal output when you run all these commands?

sureshjoshi:
@andreeis Once I manually did the start, I think it continued to work using Run Tests (until I clear cache or something?)

I'll experiment more and report back. I also had an open ticket re: custom ctest commands that I want to test out too

sureshjoshi:
Okay, so, I think I found out what might be happening. I think when jumping between ctest commands on the terminal and using the VS Code integration causes some weird behaviour. I wasn't able to come up with a full repro, but essentially, I do some commands from the terminal, and when I want to run everything I just use the VSCode button.

At some point, maybe some terminal settings corrupt the DartConfiguration.tcl (or some other files), and then the VSCode button shows that there are no tests to run. Wiping everything and re-doing configuration and build fixes this.

[...]

PR #1729, opened 2021-03-23
Changed ctest --output-on-failure from hardcode to default argument

PedroLima92:
## This changes CTest hardcoded argument `--output-on-failure` to a default argument

This a really minor change. There's not much detail really. Only 2 lines changed.

## The purpose of this change

This is VSCode so we can only gain from keeping things configurable :smiley:


bobbrow:
The only trouble with this approach is that it will change behavior for folks who have already changed the ctestArgs setting in their workspaces.  A safer approach might be to add a new setting for default CTest args and move this setting and perhaps the `-T` one there as well.  There is this issue requesting removal of -T by default: #1707

PedroLima92:
> The only trouble with this approach is that it will change behavior for folks who have already changed the ctestArgs setting in their workspaces.

Yes... I see it now.  Your suggestion seems simple enough that I could add a commit that corrects it.

Although, adding a new setting for default arguments feels kinda 'bloaty' to me. Is there any way to add these defaults to the user configurations when they update the extension? (might have to be done for each workspace too)

Thanks for the quick response by the way!

bobbrow:
Yeah, I'm not super excited about adding a new setting to deal with this either, but at the same time it could also provide the benefit of users being able to merge "common" args that they always use for CTest with "workspace specific" settings since workspace settings *override* user settings.

> Is there any way to add these defaults to the user configurations when they update the extension? (might have to be done for each workspace too)

There is, but that requires reading and checking the setting and adding a Memento to every workspace that is ever opened going forward so that we don't insert the defaults into the user settings by mistake.  It's quite complicated and not worth the effort in my opinion, so I didn't suggest it.

There's also the potential for [CMake test presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#test-preset) to come in and make all of this obsolete, but I'm not currently seeing the options I expect in that documentation.  I'm following up with a few people about that.

bobbrow:
Following up:
It looks like support for `-T` was intentionally left out of the test presets ([reference](https://gitlab.kitware.com/cmake/cmake/-/issues/21391#note_870758)), but `outputOnFailure` is there.

PedroLima92:
Cool. I haven't tried the presets yet, but I see how they might turn these obsolete.

Anyway, I have implemented it to your suggestions in my machine with apparently no trouble. But now I've noticed the translations needed hahah. I assume those were done with some tool I'm unfamiliar with (?). 

I can confidently do portuguese, french and spanish on my own, but I'm hesitant to trust AI translation on this.

bobbrow:
> I can confidently do portuguese, french and spanish on my own, but I'm hesitant to trust AI translation on this.

Você não precisa fazer as traduções 😉.  We have an internal process that will take care of it.  You just provide the changes in the English `package.nls.json` file.

PedroLima92:
hahah great! I've added the new setting. Was that what you had in mind?

package.json
```diff
@@ -1253,6 +1253,19 @@
...
+          "items": {
+            "type": "string"
+          },
+          "description": "%cmake-tools.configuration.cmake.ctestDefaultArgs.description%",
+          "default": [
+            "-T",
+            "test",
+            "--output-on-failure"
+          ],
+          "scope": "resource"
```
bobbrow, line 1267: Can you set the scope to `machine`?  I think this will help the setting make more sense to people.  This setting is meant to apply to all instances of vscode on the machine and should not be overridable by the workspace or folder.

bobbrow (changes requested):
Just one minor thing and then this is good to go.  Thank you!

package.json
PedroLima92: Done!

bobbrow (approved):

bobbrow:
Thank you!  We'll do some testing on our side and then merge this in.

Issue #1691, opened 2021-03-01
Problems with using gdb debugger on MAC


andreeis:
Opened from comment in report https://github.com/microsoft/vscode-cmake-tools/issues/884.

Zrealshadow commented 3 hours ago
I still face this problem in MacOS Catalina 10.15.4
And the cmake-tool verison is 1.6.0

When i try to debug,
the log is:
[build] Starting build
[proc] Executing command: /snap/cmake/current/bin/cmake --build /home/hristo/Projects/GitHub/GunBox/GunBox/GunBox/build --config Debug --target GunBox_Game -- -j 10
[build] ninja: no work to do.
[build] Build finished with exit code 0
[debugger] Unable to automatically determine debugger corresponding to compiler: /usr/local/bin/g++-8
[main] Failed to generate debugger configuration

My cmake tool setting:

    "C_Cpp.default.configurationProvider": "ms-vscode.cmake-tools",
    "cmake.configureOnOpen":true,
    "cmake.debugConfig": {
        "stopAtEntry": false,
        "MIMode": "gdb",
        "miDebuggerPath": "/usr/local/bin/gdb",
    }
Any solution to this problem ?

andreeis:
This is not a CMake Tools bug. We recommend that you open a report in the VSCode repository on GitHub: https://github.com/microsoft/vscode/issues.

andreeis:
Appologies, that text was meant for closing a different report, not this one.

timoscs:
Is there any workaround for this at the moment?

andreeis:
I am not sure it would work, but you can try to create an environment where lldb or lldb-mi would not be found in the path (hack the path environment variable). And also don't define miDebuggerPath (not an intuitive move, but I'm hoping it will go through a different code path that will enable you in the end to use gdb on mac).
We'll have a fix for this in 1.7 release and as soon as the fix is in I'll point you to a vsix to test (and unblock you) even earlier.

andreeis:
CMake Tools 1.7.0 was released today and it contains a fix for this issue. Upgrade the extension in VSCode and let us know if you encounter any other problems.

PR #1739, opened 2021-04-01
Consider debugger MIMode override alongside debugger path override

andreeis:
This is to fix https://github.com/microsoft/vscode-cmake-tools/issues/1691.
Added 3 tests as well.

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -135,12 +135,12 @@ function searchForCompilerPathInCache(cache: CMakeCache): string|null {
   return null;
 }
 
-export enum miModes {
+export enum MIModes {
   lldb = 'lldb',
   gdb = 'gdb',
 }
 export async function getDebugConfigurationFromCache(cache: CMakeCache, target: ExecutableTarget, platform: string,
-                                                     modeOverride?: miModes, debuggerPathOverride?: string):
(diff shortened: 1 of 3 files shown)


src/debugger.ts
```diff
@@ -135,7 +135,12 @@ function searchForCompilerPathInCache(cache: CMakeCache): string|null {
   return null;
 }
 
-export async function getDebugConfigurationFromCache(cache: CMakeCache, target: ExecutableTarget, platform: string, debuggerPathOverride?: string):
+export enum miModes {
```
bobbrow, line 138: A type is added here, but I don't think we ever validate the user setting.  It's fine for now, but we should open a task to to validate everything that comes in as an enum to ensure the value is valid.  VS Code will squiggle, but I think we can still get any string value here.

Also, the type should start with capital letters per our coding conventions.  `MIModes`

bobbrow (approved):
Please rename `miModes` and then this is good to commit.

src/debugger.ts
andreeis: Good idea, opened work item https://github.com/microsoft/vscode-cmake-tools/issues/1741.

bobbrow (approved):

PR #1744, opened 2021-04-05
Fix retention telemetry for multi root

andreeis:
Since any document changed from within VSCode notifies all the projects in a multi-root workspace, count only the one sent by the active folder. An alternative was to aggregate the notification results after all notifications have been received and processed but it would have been more complicated and since there is always an active folder (and only one) it's not necessary to go that route.

Additionally, add a boolean for the modified file being inside (or associated with) the active folder.

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -499,23 +499,23 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
     this._statusMessage.set(localize('ready.status', 'Ready'));
 
     this.extensionContext.subscriptions.push(vscode.workspace.onDidOpenTextDocument(async td => {
-      const str = td.uri.fsPath;
-      if (str.endsWith("CMakeLists.txt") || str.endsWith(".cmake")) {
+      const str = td.uri.fsPath.toLowerCase();
+      if (str.endsWith("cmakelists.txt") || str.endsWith(".cmake")) {
         telemetry.logEvent("cmakeFileOpen");
       }
     }));
 
     this.extensionContext.subscriptions.push(vscode.workspace.onDidSaveTextDocument(async td => {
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -198,6 +198,10 @@ class ExtensionManager implements vscode.Disposable {
     return this._folders.get(folder);
   }
 
+  public isActiveFolder(cmt: CMakeToolsFolder): boolean {
+    return this._folders.activeFolder == cmt;
```
bobbrow, line 202: is double-equals intended here?

src/cmake-tools.ts
```diff
@@ -543,6 +530,49 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+      // For multi-root, the "onDidSaveTextDocument" will be received once for each project folder.
+      // To avoid misleading telemetry, consider the notification only for the active folder.
+      // There is always one active folder in a workspace and never more than one.
+      if (isActiveFolder(this.folder)) {
+        // "outside" evaluates whether the modified cmake file belongs to the active folder.
+        // Currently, we don't differentiate between outside active folder but inside any of the other
+        // workspace folders versus outside any folder referenced by the current workspace.
+        let outside: boolean = true;
+        let fileType: string | undefined;
+        if (str.endsWith("CMakeLists.txt")) {
```
bobbrow, line 543: For path string comparisons, should we be doing case insensitive?

src/extension.ts
andreeis: No, it's a mistake. I'll correct with the next commit.

src/cmake-tools.ts
andreeis: Good idea.

bobbrow (approved):

Issue #1715, opened 2021-03-16
Create commands to get/set the active folder of a workspace


andreeis:
Inspired from https://github.com/microsoft/vscode-cmake-tools/issues/1712. 
Such commands can be used in tasks.json for example.

Example PR for adding new commands: 
https://github.com/microsoft/vscode-cmake-tools/pull/1488

PR #1746, opened 2021-04-06
added command to get the active folder of a workspace (#1715)

guestieng:
## This change addresses item #1715

- adding command `cmake.activeFolder` for allowing to get the active folder of a workspace

## The purpose of this change

- This change can be helpful when working with tasks/macros/multi-commands.


andreeis:
Thank you for your contribution. At a quick glance it looks good. We intend to test and merge it in the repository after we release CMake Tools 1.7 next week.

src/extension.ts
```diff
@@ -1137,6 +1137,14 @@ class ExtensionManager implements vscode.Disposable {
     await logging.showLogFile();
   }
 
+  activeFolder() : string  {
+    let thisActiveFolder: string = '';
+    if(this._folders.activeFolder){
+      thisActiveFolder = this._folders.activeFolder.folder.name;
+    }
+    return thisActiveFolder;
+  }
```
xisui-MSFT, line 1146: You can do this which is equivalent:
```
activeFolder() : string { return this._folders.activeFolder?.folder.name || ''; }
```

Also, do you only need the name or the full path? Will there be problems when there are multiple folders with the same name?

package.json
```diff
@@ -79,10 +80,16 @@
...
     "workspaceContains:.vscode/cmake-kits.json"
-    ],
+  ],
   "main": "./dist/main",
   "contributes": {
     "commands": [
+      {
+        "command": "cmake.activeFolder",
+        "title": "Get active folder",
+        "when": "cmake:enableFullFeatureSet",
```
xisui-MSFT, line 90: You need to add 'when' to the menus/commandPalette list to make it effective (actually it does nothing in contributes/commands).

That being said, I'm not sure if enableFullFeatureSet is required here.

xisui-MSFT (dismissed):

src/extension.ts
guestieng: This is a point. For now both are implemented, the active folders name and path, respectively.

src/extension.ts
```diff
@@ -1506,7 +1513,9 @@ async function setup(context: vscode.ExtensionContext, progress?: ProgressHandle
 
   // List of functions that will be bound commands
   const funs: (keyof ExtensionManager)[] = [
-    "useCMakePresets",
+    'activeFolderName',
+    'activeFolderPath',
+	  "useCMakePresets",
```
xisui-MSFT, line 1518: Indentation?

guestieng: Interesting, in Vscode (1.54.1) this is not visible.

guestieng:
All requested changes are actually implemented. Still, the status display says "Changes requested" by now. 
This seems a bit misleading. 
Therefore, I kindly request a final review.

xisui-MSFT:
Sorry for the confusion. I have reset the review. Formal approval would be given once 1.8.0 release starts to accept PRs.

bobbrow:
I'll merge this in and fix the remaining issues with linting/documentation.

PR #1752, opened 2021-04-11
Add package scripts to package.json

lygstate:
So the developer can package vscode-cmake-tools easily.


package.json
```diff
@@ -1783,7 +1784,8 @@
     "tslint": "^5.20.1",
     "typedoc": "^0.20.25",
     "typescript": "^4.1.5",
-    "vscode": "^1.1.36",
+    "vsce": "^1.87.1",
+    "vscode": "^1.1.37",
```
bobbrow, line 1788: We typically don't like to upgrade the vscode version unless we have to.  We want to be able to run on the lowest possible version.

lygstate: Oh, sorry for that, I upgrade that because when I do yarn install it's failed locally

bobbrow: Are you saying that this change requires the update to 1.1.37?

lygstate: That's not necessary, fixed

bobbrow (approved):

PR #1755, opened 2021-04-12
cmake.useCMakePresets doc changes

esweet431:

- Reference "Use CMake Presets" docs from the README in the root of the repo 
- Reference always/never/auto instead of true/false/auto

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@
 
 ## Important doc links
 
-- [Use CMake Presets](docs/cmake-presets.md)
 - [CMake Tools quick start](https://code.visualstudio.com/docs/cpp/CMake-linux)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,6 +4,7 @@
 
 ## Important doc links
 
+- [Use CMake Presets](docs/cmake-presets.md)
```
bobbrow, line 7: I think this looks a little out of place coming before the "quick start" topic.  I think we should move it down one line and call it "Configure a project with CMake Presets".  We can rename the "Configure a project" item to "Configure a project with kits and variants".

esweet431: Fixed.

bobbrow (approved):

PR #1756, opened 2021-04-12
Change useCMakePreset setting to always, never, or automatic

xisui-MSFT:
Change useCMakePreset setting to always, never, or automatic

### DIFF
--- a/package.json
+++ b/package.json
@@ -1808,16 +1808,11 @@
           "description": "%cmake-tools.configuration.cmake.exportCompileCommandsFile.description%"
         },
         "cmake.useCMakePresets":  {
-          "oneOf": [
-            {
-              "type": "boolean"
-            },
-            {
-              "type": "string",
-              "enum": [
-                "automatic"
-              ]
-            }
+          "type": "string",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1808,16 +1808,11 @@
...
-              "type": "string",
-              "enum": [
-                "automatic"
-              ]
-            }
+          "type": "string",
+          "enum": [
+            "always",
+            "never",
+            "automatic"
```
bobbrow, line 1815: Erika's documentation shortens it to 'auto'.  Can you rename it?

xisui-MSFT: CMakeCommunicationMode also have an 'automatic' option, and I want to make them consistent.

bobbrow: We also have:
```
"cmake.outputLogEncoding": {
          "type": "string",
          "default": "auto",
          "description": "%cmake-tools.configuration.cmake.outputLogEncoding.description%",
          "scope": "window"
        },
```

So it seems we're not 100% consistent right now already.

bobbrow (approved):

PR #1758, opened 2021-04-13
Add a docker container definition

xisui-MSFT:
We can use this docker container definition to create docker containers, and debug the extension in it. Main purpose for now is to test the extension on Linux when using Windows, or develop the extension using Codespaces.


xisui-MSFT:
The CMake installed using this docker file is 3.14.5. We can change it to a newer/older version.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        apt \
+        curl \
+        file \
+        git \
+        ninja-build \
+        net-tools \
+        ninja-build \
```
bobbrow, line 18: ninja-build is here twice

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+        xz-utils && \
+    apt-get autoremove -y && \
+    apt-get clean -y && \
+    rm -rf /var/lib/apt/lists/*; \
+    curl -L -O https://github.com/Kitware/CMake/releases/download/v3.14.5/cmake-3.14.5-Linux-x86_64.tar.gz && \
```
bobbrow, line 23: I would prefer a newer version of CMake if it's compatible with this distro.  3.20 is the newest and we're releasing a lot of support for that version, so I would recommend that version.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,27 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+# To fully customize the contents of this image, use the following Dockerfile instead:
+# https://github.com/microsoft/vscode-dev-containers/tree/v0.128.0/containers/typescript-node-10/.devcontainer/Dockerfile
+FROM mcr.microsoft.com/vscode/devcontainers/typescript-node:12
```
bobbrow, line 8: I don't know what all is in here, but does it have C++ compilers?  This dockerfile installs cmake and ninja which I assume are for testing, but I didn't see any actual compilers.

bobbrow (changes requested):

.devcontainer/Dockerfile
xisui-MSFT: Probably accidental copy paste... I'll remove it.

.devcontainer/Dockerfile
xisui-MSFT: It has gcc. Didn't check where it comes from in detail though. But we can add it (+gdb) to this file anyway since they would just be no-op if they are already installed, and so we won't be confused if the base image is changed for some reason in the future.

bobbrow: I like that idea

bobbrow (approved):

bobbrow (approved):

Issue #1592, opened 2020-12-11
environmentSetupScript capitalizes environment variable names


cetygamer:
### Brief Issue Summary

`environmentSetupScript` in `cmake-kits.json` converts all exported environment variable names to uppercase.

### Expected:

Variable names are preserved as is.

### Apparent Behavior:

Variable names are capitalized.

### CMake Tools Log

```
[main] Configuring folder: cmake-test 
[proc] Executing command: /usr/bin/cmake --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -DCMAKE_BUILD_TYPE:STRING=Release -H/home/aqtest/cmake-test -B/home/aqtest/cmake-test/build -G "Unix Makefiles"
[cmake] Not searching for unused variables given on the command line.
[cmake] LOWERCASEVAR=value
[cmake] -- Configuring done
[cmake] -- Generating done
[cmake] -- Build files have been written to: /home/aqtest/cmake-test/build
```

### Platform and Versions

- **Operating System**: Ubuntu
- **CMake Version**: 3.16.3
- **VSCode Version**: 1.51.1
- **CMake Tools Extension Version**: 1.5.3
- **Compiler/Toolchain**: Any

### Minimal project example to reproduce:

```
.vscode
|   cmake-kits.json
CMakeLists.txt
envsetup.sh
```

**cmake-kits.json**

```json
[
    {
        "name": "TestKit",
        "environmentSetupScript": "${workspaceFolder}/envsetup.sh"
    }
]
```

**CMakeLists.txt**

```cmake
cmake_minimum_required(VERSION 3.10)

project(TEST)

execute_process(COMMAND "bash" "-c" "env | grep -i lowercasevar")
```

**envsetup.sh**

```bash
#!/bin/bash
export lowercasevar=value
```

Steps:

1. Select kit as `TestKit`
2. Configure.

andreeis:
Thank you for opening this report. We will fix this.

bobbrow:
The fix for this issue is available in 1.10 "pre-release".  You can help us validate it by:
1. Opening the extensions panel in VS Code
2. Selecting the CMake Tools extension
3. Clicking the "Switch to Pre-release Version" button

![image](https://user-images.githubusercontent.com/12818240/156218538-7bce4a1f-4fc9-43cc-afb0-7f423aa5980e.png)

PR #1763, opened 2021-04-13
Using consistence function for environment set and get to fix the gap between windows and unix.

lygstate:
Fixes
https://github.com/microsoft/vscode-cmake-tools/issues/1592

### DIFF
--- a/src/drivers/driver.ts
+++ b/src/drivers/driver.ts
@@ -16,7 +16,7 @@ import {CMakeOutputConsumer} from '@cmt/diagnostics/cmake';
 import {RawDiagnosticParser} from '@cmt/diagnostics/util';
 import {ProgressMessage} from '@cmt/drivers/cms-client';
 import * as expand from '@cmt/expand';
-import {CMakeGenerator, effectiveKitEnvironment, Kit, kitChangeNeedsClean, KitDetect, getKitDetect, getKitEnvironmentVariablesObject} from '@cmt/kit';
+import {CMakeGenerator, effectiveKitEnvironment, Kit, kitChangeNeedsClean, KitDetect, getKitDetect} from '@cmt/kit';
 import * as logging from '@cmt/logging';
 import paths from '@cmt/paths';
 import {fs} from '@cmt/pr';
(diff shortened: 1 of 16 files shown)


bobbrow:
Thanks @lygstate!  We'll take a look at this for 1.8.

src/drivers/driver.ts
```diff
@@ -169,69 +169,56 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
-    const env = {} as {[key: string]: string};
-    const opts = this.expansionOptions;
-
-    await Promise.resolve(
-      util.objectPairs(in_env)
-      .forEach(async ([key, value]) => env[key] = await expand.expandString(value, {...opts, envOverride: expanded_env}))
-    );
-    return env;
-  }
+  private _kitEnvironmentVariables: proc.EnvironmentVariables = {};
```
elahehrashedi, line 174: the plan is not to use _ anymore, I know you didn't name this property, but as you are touching it, that would be good to rename it at the end. https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines

src/drivers/driver.ts
```diff
@@ -169,69 +169,56 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+    envs_to_merge.push(this._variantEnv);
+    return expand.mergeEnvironmentWithExpand(false, envs_to_merge, this.expansionOptions);
   }
 
   /**
    * Get the environment variables that should be set at CMake-build time.
    */
   async getCMakeBuildCommandEnvironment(in_env: proc.EnvironmentVariables): Promise<proc.EnvironmentVariables> {
-    let envs: proc.EnvironmentVariables;
+    const envs_to_merge: (proc.EnvironmentVariables | undefined)[] = [in_env];
```
elahehrashedi, line 196: Thank you for your contribution. 
Would you please provide some explanation on why `mergeEnvironmentWithExpand` is needed here?
There is a `mergeEnvironment`  and `splitEnvironmentVars` and `computeExpandedEnvironment`, and it is a little confusing having another function. We can add more comments/samples like what you did for `mergeEnvironmentWithExpand` to add some clarity, or choose some more clear names.

src/proc.ts
```diff
@@ -76,6 +76,8 @@ export interface BuildCommand {
 
 export interface EnvironmentVariables { [key: string]: string }
 
+export type EnvironmentMapType = EnvironmentVariables | Map<string, string> | NodeJS.ProcessEnv;
```
elahehrashedi, line 79: the name is confusing. Could you change it to something like GeneralEnvironmentType?

src/drivers/driver.ts
lygstate: Sorry, I touched too much variable with leading _, I think it's better to apply the naming guidelines with a single short, maybe by the link rules, otherwise it's would get the code hard to review

src/drivers/driver.ts
lygstate:
```ts

/**
 * mergeEnvironmentWithExpand will merge new env variable and
 * expand it one by one, so that the environment like this
 * will expand properly:
 * config.environment
 * { PATH:`${env.PATH};C:\\MinGW\\bin` }
 * config.configureEnvironment
 * { PATH:`${env.PATH};C:\\OtherPath\\bin` }
 *
 * @param intermediateExpand If this is the intermediate expanding
 * procedure or the final expanding procedure, if it's the
 * intermediate expanding procedure, the variable that not found
 * won't not expaned, if it's the final expanding procedure, the variable
 * that not found would expand to ''.
 * @param envs The list of environment variables to expand
 * @param opts Environment expand options
 *
 * NOTE: By mergeEnvironment one by one to enable expanding self
 * containd variable such as PATH properly. If configureEnvironment
 * and environment both configured different PATH, doing this will
 * preserve them all
 */
export async function mergeEnvironmentWithExpand(
  intermediateExpand: boolean,
  envs: (EnvironmentVariables | undefined)[],
  opts?: ExpansionOptions)
```
I've document it in mergeEnvironmentWithExpand, don't know if this is proper

src/proc.ts
lygstate: done

src/drivers/driver.ts
elahehrashedi: You are right, it's easier if we start applying new guidelines in future PRs.

src/drivers/driver.ts
elahehrashedi: I appreciate if you could add some comments/samples to `mergeEnvironment` and `splitEnvironmentVars` and `computeExpandedEnvironment`, similar to what you did for `mergeEnvironmentWithExpand` to add some clarity.

lygstate: I didn't touch and use of splitEnvironmentVars, other function are documented

lygstate:
Hope this can be merged in 1.8, as it fixes a lot of issues

src/expand.ts
```diff
@@ -84,14 +84,95 @@ export interface ExpansionOptions {
...
+ * procedure or the final expanding procedure, if it's the
+ * intermediate expanding procedure, the variable that not found
+ * won't not expaned, if it's the final expanding procedure, the variable
+ * that not found would expand to ''.
+ * @param opts Environment expand options
+ */
+export async function computeExpandedEnvironment(
+  in_env: EnvironmentVariables,
+  expanded_env: EnvironmentVariables,
+  intermediateExpand: boolean,
```
bobbrow, line 106: Can you rename `intermediateExpand` to `ignoreUnknownVars`?  I think that better conveys the meaning.

src/extension.ts
```diff
@@ -490,7 +491,7 @@ class ExtensionManager implements vscode.Disposable {
                                              await scanForKitsIfNeeded(cmt);
 
     let should_configure = cmt.workspaceContext.config.configureOnOpen;
-    if (should_configure === null && process.env['CMT_TESTING'] !== '1') {
+    if (should_configure === null && util.envGetValue(process.env, 'CMT_TESTING') !== '1') {
```
bobbrow, line 494: The names of these new functions and the way we access the environment seem a little weird now.  I wouldn't think to go to `util` to get environment values.  I think it would be better if we had an `environment` or `env` module that managed all of this.

And accessing the `process.env` directly should be ok.  `process.env` looks up environment variables in a [case-insensitive way on Windows](https://nodejs.org/api/process.html#process_process_env) which should be faster than the new function.

bobbrow (changes requested):
The names of these new functions and the way we access the environment seem a little weird now.  I wouldn't think to go to `util` to get environment values.  I think it would be better if we had an `environment` or `env` module that managed all of this.

Accessing the `process.env` directly should be ok.  `process.env` looks up environment variables in a [case-insensitive way on Windows](https://nodejs.org/api/process.html#process_process_env) which should be faster than the new function.  The real problem is with the `proc.EnvironmentVariables` type which does not do the same thing.  There is also new expansion behavior added that seems it should be in a separate PR.  I would prefer for this PR to scale back the changes and focus on the problem with accessing variables in `proc.EnvironmentVariables`.  I worry that the changes proposed are risky for 1.8 given that we want to release it soon.

src/expand.ts
lygstate: good name, thanks

lygstate:
> The names of these new functions and the way we access the environment seem a little weird now. I wouldn't think to go to `util` to get environment values. I think it would be better if we had an `environment` or `env` module that managed all of this.
> 
> Accessing the `process.env` directly should be ok. `process.env` looks up environment variables in a [case-insensitive way on Windows](https://nodejs.org/api/process.html#process_process_env) which should be faster than the new function. The real problem is with the `proc.EnvironmentVariables` type which does not do the same thing. There is also new expansion behavior added that seems it should be in a separate PR. I would prefer for this PR to scale back the changes and focus on the problem with accessing variables in `proc.EnvironmentVariables`. I worry that the changes proposed are risky for 1.8 given that we want to release it soon.

It's OK to merge this in 1.9, I'll apply your suggestions

bobbrow:
I didn't finish this prototype -- it doesn't compile yet.  But it's kind of what I was thinking for more consistent access to the environment.

Draft PR: #2025

lygstate:
> I didn't finish this prototype -- it doesn't compile yet. But it's kind of what I was thinking for more consistent access to the environment.
> 
> Draft PR: #2025

OK, would you implement this or I can reference your draft and re-implement this PR

bobbrow:
Because I manage a lot of projects, I don't always have time to work on it.  If you would like to complete this, that would be the fastest path.

lygstate:
> Because I manage a lot of projects, I don't always have time to work on it. If you would like to complete this, that would be the fastest path.

Sorry, it's a accident close

bobbrow:
This appears to have been closed automatically because we deleted the 'develop' branch in favor of working directly in 'main'.  Draft PR: #2025 has been retargeted to 'main'.

PR #1767, opened 2021-04-14
Add dry run commands

xisui-MSFT:
Added configure and build dry run commands.

According to Erika, these commands are mainly in response to feedback (primarily from VS users) that they didn't understand how settings map to the actual CMake command line, and the goal is to provide transparency so users can see how specific settings map to the CLI.

Current implementation requires cmake installed on the machine, since cmake-tools.ts and driver.ts aren't in a good state to get rid of this requirement, and I think it's okay as the target scenario isn't checking configure/build commands without cmake installed.


bobbrow:
It's not really a "dry run" so much as it is just logging the command line.  I would prefer we come up with a better name for the command.  "CMake: Log Configure Command" or something like that would be more accurate, I think.  You might also want to put focus on the output window when the command is run.

elahehrashedi:
I am getting this error while testing the branch:
image

I think it is related to the latest regression we had, as your branch is up to date with the develop: https://github.com/microsoft/vscode-cmake-tools/issues/2006

I will test an older version of your branch i.e. <d51d531>. I have two questions:
1. do you also see the same error while testing your latest version or not? just wondering. 
2. Is it like a Demo or a Log? In my mind `log build command` would have given me some info about the past build command. While this is more like giving insight into what a build command would do. I appreciate some info.

xisui-MSFT:
1. I haven't tested this for a long time. I'll test it today.
2. This is a great feedback. I have the same feeling as well. I'll change it to "show config/build command".

xisui-MSFT:
@elahehrashedi  I tested the latest changes, and didn't see the issue in your comment.

elahehrashedi:
Copying the error message here: 
```
Error: Failed to get build command
    at CMakeTools.runBuild (d:\vscode-cmake-tools\src\cmake-tools.ts:1276:15)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at d:\vscode-cmake-tools\src\extension.ts:1521:21
```

xisui-MSFT:
@elahehrashedi The problem should be fixed now. Can you please test it?

elahehrashedi:
@andreeis I have tested the code and it is working. We can have this checked in for 1.9.

andreeis:
Not sure how important this is, but for showConfigureCommand we write more (unuseful?) lines than for showBuildCommand.

[main] Configuring folder: Project1 
[main] Saving open files before configure/build
[driver] Run _refreshExpansions
[driver] Run _refreshExpansions cb
[driver] Start configure 
[driver] Runnnig pre-configure checks and steps
[driver] Run _refreshExpansions
[driver] Run _refreshExpansions cb
[cmakefileapi-driver] "c:/Program Files/CMake/bin/cmake.exe" --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -Hc:/Work/Projects/CMakeTools/GitHub2145/Project1 -Bc:/Work/Projects/CMakeTools/GitHub2145/Project1/build -G "Visual Studio 16 2019" -T host=x64 -A x64
[extension] [8858] cmake.showConfigureCommand finished (returned 0)

I am approving and will follow up whether we should hide the lines that are not the cmake configure command itself. It would just be querying in a few extra places for the boolean showCommandOnly or the trigger value and it would also require passing these further to some functions that do write these lines and which had no reason yet to have this parameter added to their prototype.

andreeis (approved):

PR #1774, opened 2021-04-16
setup environment variable for debugTarget and launchTarget properly

lygstate:

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -1774,7 +1774,7 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
       return null;
     }
 
-    let debug_config;
+    let debug_config: debugger_mod.Configuration | null;
     try {
       const cache = await CMakeCache.fromPath(drv.cachePath);
       debug_config = await debugger_mod.getDebugConfigurationFromCache(cache, targetExecutable, process.platform,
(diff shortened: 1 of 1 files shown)


src/cmake-tools.ts
```diff
@@ -1777,6 +1777,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
     }
+    const debug_environment = [];
+    const debug_env_entries = Object.entries(await drv.getCTestCommandEnvironment());
+    for (const [key, value] of debug_env_entries) {
+      debug_environment.push({
+        name: key,
+        value
+      });
+    }
+    debug_config.environment = debug_environment;
```
xisui-MSFT, line 1788: Nit: feels like Object.entries(...).map() is a more concise way?

lygstate: If you insist, otherwise this is more readable:0

lygstate: done

andreeis (approved):

bobbrow:
There is some internal debate as to which preset should supply the environment by default and how we might provide an override for people who don't like that decision.  Right now, Visual Studio uses the configurePreset environment instead of the build or test preset environment.

andreeis:
@lygstate, we may postpone or close this PR since for now we need to implement the same but using the configure preset instead of the test preset. Thank you for creating this PR, it helped us with developing the equivalent that we're going to include in 1.8.
https://github.com/microsoft/vscode-cmake-tools/pull/1977

elahehrashedi:
@lygstate, thank you for your contribution :) The final decision is to follow VS and use configurePresets.

PR #1777, opened 2021-04-16
Don't use for..in for array iteration

bobbrow:
This fixes an issue while trying to set a build preset.  for..in should not be used when iterating through an array.  It will return back other properties (e.g. attached functions) that you don't intend to index out of the array object.

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -878,13 +878,13 @@ async function expandBuildPresetHelper(preset: BuildPreset,
     if (util.isString(preset.targets)) {
       preset.targets = await expandString(preset.targets, expansionOpts);
     } else {
-      for (const index in preset.targets) {
+      for (let index = 0; index < preset.targets?.length; index++) {
         preset.targets[index] = await expandString(preset.targets[index], expansionOpts);
       }
     }
   }
   if (preset.nativeToolOptions) {
-    for (const index in preset.nativeToolOptions) {
+    for (let index = 0; index < preset.nativeToolOptions?.length; index++) {
       preset.nativeToolOptions[index] = await expandString(preset.nativeToolOptions[index], expansionOpts);
     }
   }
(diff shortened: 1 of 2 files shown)


src/taskprovider.ts
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 import * as vscode from 'vscode';
-import { isString } from 'util';
+import {isString} from './util';
```
bobbrow, line 6: Saw this while searching for `for..in` usages in the codebase so I just bundled it in.  We were using the wrong `isString`.

andreeis (approved):

src/preset.ts
```diff
@@ -878,13 +878,13 @@ async function expandBuildPresetHelper(preset: BuildPreset,
     if (util.isString(preset.targets)) {
       preset.targets = await expandString(preset.targets, expansionOpts);
     } else {
-      for (const index in preset.targets) {
+      for (let index = 0; index < preset.targets?.length; index++) {
```
xisui-MSFT, line 881: else if (preset.targets), and remove all the `?`s in this change.

xisui-MSFT (changes requested):

src/preset.ts
xisui-MSFT: actually it's already checked (hidden in compare view...). Can you remove all the '?'s in your changes?

bobbrow: Yes. I noticed the same. We've already validated the types of the presets file at this point, right?  I don't need to make sure it's an array or anything do I?

xisui-MSFT: That's correct.

xisui-MSFT (approved):

PR #1779, opened 2021-04-16
Clarify issue in CHANGELOG. Fix launch executable command.

andreeis:
It looks that this wasn't actually a regression (although I don't remember seeing the symptom) so I won't object if we triage it for 1.8.
In many scenarios, the first time would launch an executable we wouldn't have its real path as a command but "Utility target" which is not recognized by the shell. Since now build before run is defaulted to true, the binary will be present when invoked, even if when evaluated it wasn't yet there. And even if build before run is false, I don't understand why we would ever want "Utility target" as a command for launch executable, let me know if I miss anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ Bug fixes:
 - Dependencies package versions upgrade. [PR #1475](https://github.com/microsoft/vscode-cmake-tools/pull/1475) [@lygstate](https://github.com/lygstate)
 - Add vendor hostOs targetOs targetArch versionMajor versionMinor attributes for kit. [PR #1337](https://github.com/microsoft/vscode-cmake-tools/pull/1337) [@lygstate](https://github.com/lygstate)
 - Always correctly build target executable path. [PR #1674](https://github.com/microsoft/vscode-cmake-tools/pull/1674) [@falbrechtskirchinger](https://github.com/falbrechtskirchinger)
-- Use variables instead of hardcoded values for system path references. [PR #1690](https://github.com/microsoft/vscode-cmake-tools/pull/1690)
+- Use variables instead of hardcoded values for system path references. [#883](https://github.com/microsoft/vscode-cmake-tools/issues/883)(https://github.com/Zingam)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -22,7 +22,7 @@ Bug fixes:
 - Dependencies package versions upgrade. [PR #1475](https://github.com/microsoft/vscode-cmake-tools/pull/1475) [@lygstate](https://github.com/lygstate)
 - Add vendor hostOs targetOs targetArch versionMajor versionMinor attributes for kit. [PR #1337](https://github.com/microsoft/vscode-cmake-tools/pull/1337) [@lygstate](https://github.com/lygstate)
 - Always correctly build target executable path. [PR #1674](https://github.com/microsoft/vscode-cmake-tools/pull/1674) [@falbrechtskirchinger](https://github.com/falbrechtskirchinger)
-- Use variables instead of hardcoded values for system path references. [PR #1690](https://github.com/microsoft/vscode-cmake-tools/pull/1690)
+- Use variables instead of hardcoded values for system path references. [#883](https://github.com/microsoft/vscode-cmake-tools/issues/883)(https://github.com/Zingam)
```
bobbrow, line 25: missing `[@Zingam]` for the second link

bobbrow:
> Since now build before run is defaulted to true, the binary will be present when invoked, even if when evaluated it wasn't yet there. And even if build before run is false, I don't understand why we would ever want "Utility target" as a command for launch executable, let me know if I miss anything.

I don't know much about the scenario, but it seems ok.

bobbrow (approved):

Zingam:
Document which Presets/Presets version are supported? I don't know if that's is actually important.

bobbrow:
Documentation is here: https://github.com/microsoft/vscode-cmake-tools/blob/develop/docs/cmake-presets.md

You can actually test this out today using one of the latest "Build VSIX" builds on our [Actions tab](https://github.com/microsoft/vscode-cmake-tools/actions/workflows/build-vsix.yml).

Zingam:
@bobbrow Thank you! I still tend to look for help at: https://vector-of-bool.github.io/docs/vscode-cmake-tools/index.html
I never even think about looking at the repo for docs.

The following command appears to be dead:
>Command 'CMake: Online Help' resulted in an error (command 'cmake.onlineHelp' not found)

PR #1780, opened 2021-04-16
disallow for..in for arrays

bobbrow:
It turns out that there is a tslint rule to catch `for..in` with arrays. I'm enabling this rule for the repo.


tslint.json
```diff
@@ -34,6 +34,7 @@
 		"no-trailing-whitespace": true,
 		"object-literal-shorthand": true,
 		"object-literal-key-quotes": [true, "as-needed"],
-		"prefer-template": [true, "allow-single-concat"]
+		"prefer-template": [true, "allow-single-concat"],
+    "no-for-in-array": true
```
xisui-MSFT, line 38: This file uses tab.

bobbrow: fixed

bobbrow:
Extension is now using eslint. This needs to be ported. Closing...

PR #1789, opened 2021-04-20
CppTools-API v5 integration

michelleangela:
### This changes IntelliSenseMode and Standard values as optional (undefined) for CppTools-API v5+

- Return undefined IntelliSenseMode for CppTools-API v5+
- If Standard C or C++ cannot be parsed, returns undefined for CppTools-API v5+

## The purpose of this change

CppTools-API v5 integration

### DIFF
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -23,7 +23,7 @@ const log = createLogger('cpptools');
 
 type Architecture = 'x86' | 'x64' | 'arm' | 'arm64' | undefined;
 type StandardVersion = "c89" | "c99" | "c11" | "c17" | "c++98" | "c++03" | "c++11" | "c++14" | "c++17" | "c++20"
-  | "gnu89" | "gnu99" | "gnu11" | "gnu17" | "gnu++98" | "gnu++03" | "gnu++11" | "gnu++14" | "gnu++17" | "gnu++20";
+  | "gnu89" | "gnu99" | "gnu11" | "gnu17" | "gnu++98" | "gnu++03" | "gnu++11" | "gnu++14" | "gnu++17" | "gnu++20" | undefined;
 
 export interface CompileFlagInformation {
   extraDefinitions: string[];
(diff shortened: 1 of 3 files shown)


test/unit-tests/cpptools.test.ts
```diff
@@ -19,9 +19,14 @@ suite('CppTools tests', () => {
     // Parse definition
     const cpptoolsVersion3 = Version.v3;
     const cpptoolsVersion4 = Version.v4;
+    const cpptoolsVersion5 = Version.v5;
+
+    // Verify CppTools API version 5
+    let info = parseCompileFlags(cpptoolsVersion5, ['-target', 'arm-arm-none-eabi']);
+    expect(info.targetArch).to.eql(undefined);
```
bobbrow, line 26: Can you add a check for `info.standard` == undefined too?

src/cpptools.ts
```diff
@@ -119,37 +119,38 @@ function parseTargetArch(target: string): Architecture {
 }
 
 export function parseCompileFlags(cptVersion: cpt.Version, args: string[], lang?: string): CompileFlagInformation {
+  const require_standard_target = (cptVersion < cpt.Version.v5);
```
bobbrow, line 122: I'm looking through the change history for cpptools-api and it looks like this should have been made optional in v4.  Cpptools requires v5 for standard to be optional though.

I think the bug is actually in cpptools `Client.isSourceFileConfigurationItem` requires v5 for intelliSenseMode and standard to be optional, but at this point we probably have to leave it alone.

bobbrow:
>   private _cpptoolsVersion: cpt.Version = cpt.Version.latest;

I'm not sure how this worked previously because it looks like this should have already been v5 when someone upgraded the package.json a month ago.

---
Refers to: src/cpptools.ts:391 in 79de84a. [](commit_id = 79de84ab573da8b4de3aa7b2dd5ff788cafd3441, deletion_comment = False)

bobbrow:
I figured it out.  You also need to update `extension.ts` line 661.  We're requesting v4 and we need to request v5 specifically.

src/cpptools.ts
michelleangela: It's okay if a value sent for IntelliSenseMode for v4 even though it's optional. If the incorrect value is sent to cpptools, cpptools v1.3.0 can correct it from the #defines from compiler query.

From this PR https://github.com/microsoft/vscode-cpptools-api/pull/28, IntelliSenseMode and Standard were made optional when cpptools-api version was changed to 4.0.2, so I just required these properties for v4 or under.

bobbrow (approved):

Issue #1787, opened 2021-04-20
Wrong evaluation of CMakePresets.json


RedDwarf69:
https://github.com/microsoft/vscode-cmake-tools/blob/develop/docs/cmake-presets.md#supported-cmake-and-cmakepresetsjson-versions says "CMake Tools reads and evaluates CMakePresets.json and CMakeUserPresets.json, and does not invoke CMake directly with the --preset option.", unfortunately with CMake Tools 1.7.0 the evaluation is not equivalent to the one done by CMake.

https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#macro-expansion says "All macros are evaluated in the context of the preset being used, even if the macro is in a field that was inherited from another preset.". And when using

```
{
    "version": 2,
    "cmakeMinimumRequired": {
        "major": 3,
        "minor": 20,
        "patch": 0
    },
    "configurePresets": [
        {
            "name": "Base",
            "hidden": true,
            "generator": "Ninja",
            "binaryDir": "build/${presetName}"
        },
        {
            "name": "Complete",
            "inherits": "Base"
        }
    ]
}
```
and using the "Complete" preset CMake Tools uses `build/Base` as binaryDir while CMake uses `build/Complete`.

bobbrow:
Thanks for reporting this.  @xisui-MSFT can you take a look?

andreeis:
The PR with a fix for this problem was merged in the GitHub repository and, if you'd like, you can validate it by installing this custom vsix before we release the official 1.7.1.
https://github.com/microsoft/vscode-cmake-tools/actions/runs/774916180

andreeis:
CMake Tools 1.7.1 was released today and it contains a fix for this issue. Upgrade the extension in VSCode and let us know if you encounter any other problems.

PR #1791, opened 2021-04-20
Expand macros inherited under the context of current preset

xisui-MSFT:
## This change addresses item #1787 

- Expand macros at the end of `expand*Preset()`, using a shallow copy of the preset caches. The caches now only contain expanded fields (inherits, etc.), but not expanded macros.


gugavaro (approved):

xisui-MSFT:
#1793

Issue #1250, opened 2020-05-19
Environment variable substituion: Kit environment and buildDirectory


alex-che:
### Brief Issue Summary

`"cmake.buildDirectory" : "${env.VARNAME}"` does not work if `VARNAME` is set in active kit's `"environmentVariables"`. Empty string is used instead. Ditto for `"cmake.installPrefix"`.

Although, the following works fine:
1. `"cmake.buildDirectory" : "${buildKit}"`
2. `"cmake.configureSettings: { "SOME_SETTING" : "${env.VARNAME}" }`
3. `"cmake.buildDirectory" : "${env.VARNAME}"` - if `VARNAME` is set in environment before running vscode, instead of the kit. 

### Expected:

Values of variables set in active kit's `"environmentVariables"` to be used in `${env.XXX}` substitution for `"cmake.buildDirectory"` and `"cmake.installPrefix"`.

### Apparent Behavior:

1: `.vscode/cmake-kits.json`:
```
[
   {
      "name": "TEST_KIT",
      "environmentVariables": {"TEST_VAR": "TEST"}
   }
]
```
2: `.vscode/settings.json`:
```
   "cmake.configureSettings": {
      "TEST_SETTING": "${env.TEST_VAR}"
   },
   "cmake.buildDirectory": "${workspaceFolder}/__build/${env.TEST_VAR}-${buildKit}-${buildType}",
```
3. CMake: Select Kit => TEST_KIT

4. CMake: Delete Cache and Reconfigure

5. Output: CMake/Build 
see below

### CMake Tools Log
```
[proc] Executing command: /usr/local/bin/cmake --no-warn-unused-cli -DTEST_SETTING:STRING=TEST ... -B/home/user/project/__build/-TEST_KIT-Debug ...
```
### Platform and Versions

- **Operating System**: VSCode UI: Windows 10 1903, VSCode Server: Ubuntu 16 64bit
- **CMake Version**: 3.17.2 
- **VSCode Version**: 1.45.1
- **CMake Tools Extension Version**: 1.3.1
- **Compiler/Toolchain**: GCC 5.4.0

### Other Notes/Information

Is it expected bahaviour, am I doing something wrong, or is it a bug?

BTW, thanks for the great extension!

andreeis:
CMake Tools 1.7.0 was released today and it contains a fix for this issue. Upgrade the extension in VSCode and let us know if you encounter any other problems.

alex-che:
[...]

PR #1797, opened 2021-04-21
Inject environment merged from all sources into the string expansion …

andreeis:
https://github.com/microsoft/vscode-cmake-tools/issues/1250 is not fixed with 1.7.
I either had a loophole in testing the original PR that targeted the issue or it regressed with more recent PRs.
This fix is ensuring an up to date environment is sent every time we expand a string.

### DIFF
--- a/src/drivers/driver.ts
+++ b/src/drivers/driver.ts
@@ -549,6 +549,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
     return this.doRefreshExpansions(async () => {
       log.debug('Run _refreshExpansions cb');
       const opts = this.expansionOptions;
+      opts.envOverride = await this.getConfigureEnvironment();
       this._sourceDirectory = await util.normalizeAndVerifySourceDir(await expand.expandString(this.config.sourceDirectory, opts));
 
       if (!this.useCMakePresets) {


src/drivers/driver.ts
```diff
@@ -549,6 +549,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
     return this.doRefreshExpansions(async () => {
       log.debug('Run _refreshExpansions cb');
       const opts = this.expansionOptions;
+      opts.envOverride = await this.getConfigureEnvironment();
```
bobbrow, line 552: Should `.envOverride` be set as part of `this.expansionOptions`?  I think that's a bigger change but seems like if we do it here we should be doing it everywhere else too.

andreeis: I can't quickly answer this. Do you want to look at this more thoroughly for 1.8 and move the bug out of 1.7.1?  I am thinking we can also include this fix as is for 1.7.1, to unblock the main scenario we are aware of and revisit for 1.8 whether there is a better place to do this check?

bobbrow: That proposal sounds good to me.  I'll approve this and then let's make sure we're tracking the full fix for 1.8.

bobbrow (approved):

src/drivers/driver.ts
andreeis: https://github.com/microsoft/vscode-cmake-tools/issues/1798

Issue #1796, opened 2021-04-21
Error: Unknown diagnostic severity level with -fopt-info-vec/-O3 and GCC 10.2


fvanmaele:
### Brief Issue Summary

When using `-fopt-info-vec` to output diagnostics on vectorization with GCC 10.2, `cmake` configure or build steps result in an error after their completion:
```
​ [Extension Host] [CMakeTools] 2021-04-21T21:08:16.916Z [fatal] [rollbar] Unhandled exception: Unhandled Promise rejection: launchTarget Error: Unknown diagnostic severity level: optimized {}
```
The issue is not present with the clang11 kit.

### Expected:

1. Write a loop which can be (auto-)vectorized by GCC, for example
```
#include <vector>

double sum = 0;
int main() {
  std::vector<float> v(50000, 1.0);
  for (size_t i = 0; i < v.size(); ++i)
    sum += v[i];
}
```
2. Add `-fopt-info-vec` to compilation options:
```
add_compile_options(-fopt-info-vec)
```
3. Select GCC 10.2 as kit.
4. Select `Release` as build variant or add `-O3` to compilation options.
4. Perform `CMake: Build Target`.

### Apparent Behavior:

1. The build step completes successfully, but an error from the extension is shown.
2. `Launch the selected target in the terminal window` does not launch the target as a result.

### CMake Tools Log

```
[main] Building folder: hasc-code daxpy_bm
[build] Starting build
[proc] Executing command: /usr/bin/cmake --build /mnt/c/Users/Ferdinand/source/repos/hasc-code/build --config Release --target daxpy_bm -j 10 --
[build] [1/2  50% :: 1.498] Building CXX object CMakeFiles/daxpy_bm.dir/daxpy_bm.cc.o
[build] ../daxpy_bm.cc:30:39: optimized: loop vectorized using 32 byte vectors
[build] ../daxpy_bm.cc:30:39: optimized:  loop versioned for vectorization because of possible aliasing
[build] ../daxpy_bm.cc:30:39: optimized: loop vectorized using 16 byte vectors
[build] ../daxpy_bm.cc:37:28: optimized: loop vectorized using 32 byte vectors
[build] ../daxpy_bm.cc:37:28: optimized: loop vectorized using 16 byte vectors
[build] ../daxpy_bm.cc:37:28: optimized: loop vectorized using 32 byte vectors
[build] ../daxpy_bm.cc:37:28: optimized: loop vectorized using 16 byte vectors
[...]

PR #1799, opened 2021-04-21
Don't throw when unknown diagnostics appear

bobbrow:
Addresses #1796.
I'm not sure why this isn't throwing in the tests.  Will monitor test results to see if the french translation test is being skipped for some reason.


andreeis (approved):

test/unit-tests/diagnostics.test.ts
```diff
@@ -243,6 +243,11 @@ suite('Diagnostics', async () => {
     expect(path.posix.normalize(diag.file)).to.eq(diag.file);
     expect(path.posix.isAbsolute(diag.file)).to.be.true;
   });
+  test('Parsing non-diagnostic', () => {
+    const lines = ['/usr/include/c++/10/bits/stl_vector.h:98:47: optimized: basic block part vectorized using 32 byte vectors'];
+    feedLines(build_consumer, [], lines);
+    expect(build_consumer.compilers.gcc.diagnostics).to.have.length(0);
+  });
   test('Parsing linker error', () => {
```
andreeis, line 251: I think you need .to.have.length(1);

bobbrow: I figured out what I want. I needed to `resolveDiagnostics`.

src/diagnostics/build.ts
```diff
@@ -56,7 +56,9 @@ export class CompileOutputConsumer implements OutputConsumer {
       case 'remark':
         return vscode.DiagnosticSeverity.Information;
       }
-      throw new Error('Unknown diagnostic severity level: ' + p);
+      // tslint:disable-next-line
+      console.warn('Unknown diagnostic severity level: ' + p);
```
xisui-MSFT, line 60: Why not use log.warning instead? I think we can probably use a logger for this file, or accept a Logger in the constructor like CMakeBuildConsumer.

bobbrow: I think it would be too noisy on localized systems.  We have issues with the problem matchers not supporting translated versions of "error", "warning", etc coming from the compiler.

danishbelal:
I second this PR.  Only english diagnostics are recognized, therefore building anything that emits warnings on non-english systems fails launching.

On my german setup, warnings are not detected as such:
`[rollbar] Unhandled exception: Unhandled Promise rejection: launchTarget Error: Unknown diagnostic severity level: Warnung {}`

Issue #533, opened 2018-09-06
Show a quickPick of all the CMakeLists.txt inside the project (if not in the root)


Zingam:
### Brief Issue Summary

CMake Tools doesn't detect CMakeLists.txt automatically if it is not in the root directory on Linux (I haven't tried other OSs). The CMakeLists.txt is in a subdirectory of the root directory and cmake.sourceDirectory is set to point to that subdirectory.
If I execute the command to configure the project manually it runs as expected.

### Expected:

The same behavior as if CMakeLists.txt is in the root directory: when the project is opened CMake Tools should ask me if I want to configure the project.

jimbo1qaz:
Building a CMakeLists.txt in a subdirectory (Windows) is a pain.

First I have to go to VS Code settings, change the workspace `cmake.sourceDirectory`to the subfolder.

Then I get exception "[rollbar] Unhandled exception: Unhandled Promise rejection: configure [cmake-server] Failed to activate protocol version: "CMAKE_HOME_DIRECTORY" is set but incompatible with configured source directory value. {}".

To fix this, I have to

vector-of-bool:
VSCode extensions have something called "activation events," which they register in their `package.json`. Most activation events are when one of the command palette commands is invoked. When an activation event is triggered, only then will VSCode load the extension and it can register itself. CMake Tools has an activation event if it detects a `CMakeLists.txt` in the root directory of the workspace, but any subdirectories do not work. It is during extension activation that it generates this prompt.

There is a special `*` activation event that will _immediately_ and _always_ load the extension in VSCode. Its usage is dis-recommended, however. I would need to be careful about using it, but it may be an option.

mwheeler:
Doesn't this just require `"workspaceContains:**/CMakeLists.txt"` ? (though `cmake.sourceDirectory` is still needed to decide 'which' CMakeLists.txt to use if there are multiple)

Zingam:
cmake.sourceDirectory

Doesn't seem to work now:

```
[cmake] CMake version: 3.10.2
[cmake] The C compiler identification is unknown
[cmake] The C compiler identification is unknown
[cmake] CMake Error: Could not find cmake module file: ./CMakeFiles/3.10.2/CMakeCCompiler.cmake
[cmake] CMake Error: Could not find cmake module file: ./CMakeFiles/3.10.2/CMakeCCompiler.cmake
[cmake] The CXX compiler identification is unknown
[...]

PR #1803, opened 2021-04-22
locate CmakeLists.txt in root's subfolders

elahehrashedi:
fix: https://github.com/microsoft/vscode-cmake-tools/issues/533

if there is no cmakelists in the root this will be shown:
image

Then you hit on "Locate",

in case there are few cmakelists files in subdirectories this will be shown:
image

in case there is no cmakelists file in the root this will be shown:
image


src/util.ts
```diff
@@ -627,3 +629,30 @@ export async function normalizeAndVerifySourceDir(sourceDir: string): Promise<st
 export function isCodespaces(): boolean {
   return !!process.env["CODESPACES"];
 }
+
+export function getAllFilePaths(dir: vscode.Uri, filename: string): string[] | undefined {
+  const regex: RegExp = new RegExp(`(\\/|\\\\)${escapeStringRegExp(filename)}$`);
```
bobbrow, line 634: It's unclear to me why we need the `(\\/|\\\\)` as part of the regular expression.  Can you comment this, or remove it if it's not necessary?

src/util.ts
```diff
@@ -627,3 +629,30 @@ export async function normalizeAndVerifySourceDir(sourceDir: string): Promise<st
...
+  return recGetAllFilePaths(dir.fsPath, filename, regex, fs.readdirSync(dir.fsPath), []);
+}
+
+function recGetAllFilePaths(dir: string, filename: string, regex: RegExp, files: string[], result: string[]) {
+  for (const item of files) {
+    const file = path.join(dir, item);
+    if (fs.statSync(file).isDirectory()) {
```
bobbrow, line 641: >fs.statSync(file) [](http://example.com/codeflow?start=8&length=17)

You should save the result of this `fs.statSync` in a variable so that you don't have to call it twice (for the isFile case)

src/cmake-tools.ts
```diff
@@ -494,24 +495,42 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+            fullPath: file
+          })) : [];
+          items.push({ label: localize("browse.for.cmakelists", "[Browse for CMakeLists.txt]") , fullPath: "", description: "Search for CMakeLists.txt on this computer" });
+          const selection: FileItem | undefined = await vscode.window.showQuickPick(items, {
+            placeHolder: (items.length === 1 ? localize("cmakelists.not.found", "No CMakeLists.txt was found.") : localize("select.cmakelists", "Select CMakeLists.txt"))
+          });
+          let selectedFile: string | undefined;
+          if (!selection) {
+            break; // User canceled it.
+          } else if (selection.label == "[Scan for CmakeLists.txt]") {
```
bobbrow, line 513: >label == "[Scan for CmakeLists.txt] [](http://example.com/codeflow?start=31&length=35)

label is localized and doesn't match this string.  Can you save the localized string in a variable and test against that instead?

src/cmake-tools.ts
```diff
@@ -494,24 +495,42 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+            const openOpts: vscode.OpenDialogOptions = {
+              canSelectMany: false,
+              defaultUri: vscode.Uri.file(this.folder.uri.fsPath),
+              filters: { "CMake files": ["txt"], "All files": ["*"] },
+              openLabel: "Load",
+            };
+            const cmakeListsFile = await vscode.window.showOpenDialog(openOpts);
+            if (cmakeListsFile) { selectedFile = cmakeListsFile[0].fsPath; }
```
bobbrow, line 521: >{ selectedFile = cmakeListsFile[0].fsPath; } [](http://example.com/codeflow?start=32&length=44)

nit: fix the formatting here

bobbrow (changes requested):
Submitted a few comments.

src/util.ts
elahehrashedi: `(\\/|\\\\)` and `(\/|\\\\)`  work similar !

having that checks for having `\\` or `/` before string `CMakeLists.txt` so that these paths don't get matched:
`D:\\project\CMakeLists.txt` or `D:\\project\\_CMakeLists.txt`

gulpfile.js
```diff
@@ -268,4 +268,15 @@ gulp.task('lint', function () {
         .pipe(eslint.format())
         //.pipe(gulp.dest(file => file.base))
         .pipe(eslint.failAfterError());
-});
\ No newline at end of file
+});
+
+gulp.task('pr-check', (done) => {
```
bobbrow, line 273: We don't need this task. It's a cpptools thing. Can you remove it?

src/util.ts
```diff
@@ -648,9 +647,9 @@ export function isCodespaces(): boolean {
   return !!process.env["CODESPACES"];
 }
 
-export async function getAllFilePaths(dir: vscode.Uri, filename: string): Promise<string[] | undefined> {
-  const regex: RegExp = new RegExp(`(\/|\\\\)${escapeStringRegexp(filename)}$`);
-  return recGetAllFilePaths(dir.fsPath, filename, regex, fs.readdirSync(dir.fsPath), []);
+export async function getAllCMakeListsPaths(dir: vscode.Uri): Promise<string[] | undefined> {
+  const regex: RegExp = new RegExp(/(\/|\\)CMakeLists\.txt$/);
+  return recGetAllFilePaths(dir.fsPath, "CMakeLists.txt", regex, fs.readdirSync(dir.fsPath), []);
```
bobbrow, line 652: I don't think you should need to pass in "CMakeLists.txt" as the second parameter.  It doesn't look like it's used.  Can you just remove the 2nd parameter from the recursive function?

bobbrow: And should the 4th parameter be `await readDir(dir.fsPath)`?

src/util.ts
```diff
@@ -563,6 +562,28 @@ export function checkDirectoryExists(filePath: string): Promise<boolean> {
   });
 }
 
+// Read the files in a directory.
+export function readDir(dirPath: string): Promise<string[]> {
+  return new Promise((resolve) => {
+    fs.readdir(dirPath, (_err, list) => {
```
bobbrow, line 568: In the case of `_err`, is the list an empty array?  Maybe it's best to check for `_err` and force return `[]`.

PR #1808, opened 2021-04-23
Fix Operator Precedence when Getting Code Page

taoyouh:

- Fix operator precedence when getting code page. The current version cannot get correct code page (the `replace` doesn't work) thus encoding is incorrect.
  - Before: when `cmake.outputLogEncoding` was set to "auto", the output log is always treeated as UTF-8. Thus, the output log contains garbage characters (on Windows in Chinese or other non-English language).
  - After: the default code page is correctly detected, and the output log is correct.

## Other Notes/Information

This problem affects version 1.7.0 and 1.7.1.


lygstate (approved):
LGTM

bobbrow (approved):

PR #1635, opened 2021-01-27
The 'path' member of 'target' or 'source' could be relative

elsonwei:
Fix #1504

Refer to https://cmake.org/cmake/help/git-stage/manual/cmake-file-api.7.html#codemodel-version-2-target-object.
The target path and source file path could be relative, so the convertToExtCodeModelFileGroup() must convert the paths to absolute first.

PR #1816, opened 2021-04-28
Fix paths of target sources outside the workspace

sleiner:

Target sources from outside the VS Code workspace folder are now represented with their correct paths.
Furthermore, if the user clicks on them, instead of an error message, the correct file opens 😃

## Other Notes/Information

This is not thoroughly tested yet. But everything looks fine for with VS Code 1.55.2 on macOS and a very simple example project.


bobbrow:
Interesting.  I actually stumbled on this old PR that was closed without review/merge and it appears to address the same issue.  https://github.com/microsoft/vscode-cmake-tools/pull/1635/files

sleiner:
Oh yeah, I see. As far as I see, it does exactly the same thing (although in a slightly different way) 👍 
How would you like to proceed?

sleiner:
I just tested this with a more complex project. Turns out: My approach fails when the target's source root is not the same as the workspace root. (In context of the `convertToExtCodeModelFileGroup` function: `targetObject.paths.source != "."` still produces incorrect paths with my patch. On the other hand, #1635 handles it correctly.
So I propose closing this PR and instead applying the patches from #1635. Is that alright with you?

bobbrow:
That user closed the PR and we didn't hear back, so if you could update your PR with the changes needed, we'll approve it and get it in.

sleiner:
Alright, I'm ready 👍 Could you allow CI execution?

bobbrow:
@andreeis we forgot to mark the milestone on this one.  The code change is pretty small - this is mostly adding tests.  Should we take it in 1.7.2?

andreeis:
I think we can, of course, I am looking at it now.

andreeis (approved):

Issue #1815, opened 2021-04-28
CMakePresets.json parsing: Simplified bool cache variable syntax ignored


RedDwarf69:
https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#configure-preset says the following about `cacheVariables`: The key is the variable name (which may not be an empty string), and **the value is** either null, **a boolean (which is equivalent to a value of "TRUE" or "FALSE" and a type of BOOL)**

But when presented with
```
{
    "version": 2,
    "cmakeMinimumRequired": {
        "major": 3,
        "minor": 20,
        "patch": 0
    },
    "configurePresets": [
        {
            "name": "test_preset",
            "generator": "Ninja",
            "binaryDir": "build",
            "cacheVariables": {
                "VAR_A": true
            }
        }
    ]
}
```
cmake-tools 1.7.1, will not report any error, but it will ignore VAR_A, simply executing `cmake -H<source_dir> -B<binary_dir> -G Ninja`.

If instead 
```
{
    "version": 2,
    "cmakeMinimumRequired": {
        "major": 3,
        "minor": 20,
        "patch": 0
    },
    "configurePresets": [
        {
            "name": "test_preset",
            "generator": "Ninja",
            "binaryDir": "build",
            "cacheVariables": {
                "VAR_A": {
                    "type": "BOOL",
                    "value": "TRUE"
                }
            }
        }
    ]
}
```
is used then it uses `cmake -DVAR_A:BOOL=TRUE -H<source_dir> -B<binary_dir> -G Ninja`

PS. I'm not opening an issue for it because it seems to work, but I would be expecting `-S<source_dir>` instead of `-H<source_dir>`, I can't find any documentation for such an usage of `-H`.

andreeis:
Thank you for opening this bug report. @xisui-MSFT, can you have a look?

xisui-MSFT:
I guess the '-H' was copied from the legacy driver and we never changed it. https://stackoverflow.com/questions/31090821/what-does-the-h-option-means-for-cmake

I'll take a look at that cacheVar issue.

andreeis:
@RedDwarf69, if you would like, you can try this vsix (https://github.com/microsoft/vscode-cmake-tools/actions/runs/811030625) to get unblocked with a fix for this issue before we release 1.7.2.

mizvekov:
Another related issue here is that cache entries which are just an empty string also don't get created, they are ignored and not added to the cmake configure command line.

[...]

PR #1817, opened 2021-04-28
Dev/xisui/stricter type check presets

xisui-MSFT:
### This corrects presets expansion behavior

- Fix the missing boolean case reported in #1815, which is introduced when moving string expansion out of presets expansion recursion.
- Fixed some places that may cause incorrect presets expansion because of falsy checks.


xisui-MSFT:
Wrong branch...? discarding this and opening another one...

PR #1818, opened 2021-04-28
Use stricter type checks in presets expansion

xisui-MSFT:
### This corrects presets expansion behavior

- Fix the missing boolean case reported in #1815, which is introduced when moving string expansion out of presets expansion recursion.
- Fixed some places that may cause incorrect presets expansion because of falsy checks.


src/preset.ts
```diff
@@ -550,7 +555,7 @@ async function expandConfigurePresetHelper(folder: string,
   // [Windows Only] If CMAKE_CXX_COMPILER or CMAKE_C_COMPILER is set as 'cl' or 'cl.exe', but they are not on PATH,
   // then set the env automatically
   if (process.platform === 'win32') {
-    const getStringValueFromCacheVar = (variable: CacheVarType) => {
+    const getStringValueFromCacheVar = (variable: CacheVarType | undefined) => {
```
bobbrow, line 558: I prefer `variable?: CacheVarType` over `variable: CacheVarType | undefined`

src/preset.ts
```diff
@@ -448,13 +449,17 @@ export async function expandConfigurePreset(folder: string,
     expandedPreset.cacheVariables = { };
     for (const cacheVarName in preset.cacheVariables) {
       const cacheVar = preset.cacheVariables[cacheVarName];
-      if (cacheVar && typeof cacheVar !== 'boolean') {
-        if (util.isString(cacheVar)) {
-          expandedPreset.cacheVariables[cacheVarName] = await expandString(cacheVar, expansionOpts);
-        } else if (util.isString(cacheVar.value)) {
-          expandedPreset.cacheVariables[cacheVarName] = { type: cacheVar.type, value: await expandString(cacheVar.value, expansionOpts) };
+      if (cacheVar) {
```
bobbrow, line 452: If `cacheVar === false` we won't enter this block.  Is that ok?

bobbrow: It might be good to also add a `util.isBoolean` helper function.

xisui-MSFT: Ah that's a good point

bobbrow (approved):

Issue #1821, opened 2021-04-29
GCC not recognized anymore


riri:
* OS: GNU/Linux (5.11.16-arch1-1 #1 SMP PREEMPT Wed, 21 Apr 2021 17:22:13 +0000 x86_64 GNU/Linux)
* VS Code: 1.55.2
* CMakeTools: 1.7.0
* Installed compilers: gcc 10.2.0, clang 11.1.0

Since yesterday, when I scanned available Kits, GCC dropped out of the list. Setting cmake.loggingLevel to debug permitted me to see GCC is not recognized (sorry output is in french):
```
[kit] Bad GCC binary "/usr/bin/gcc -v" version:10.2.0 output: Utilisation des specs internes.
COLLECT_GCC=/usr/bin/gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/lto-wrapper
Cible : x86_64-pc-linux-gnu
Configuré avec: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++,d --with-isl --with-linker-hash-style=gnu --with-system-zlib --enable-__cxa_atexit --enable-cet=auto --enable-checking=release --enable-clocale=gnu --enable-default-pie --enable-default-ssp --enable-gnu-indirect-function --enable-gnu-unique-object --enable-install-libiberty --enable-linker-build-id --enable-lto --enable-multilib --enable-plugin --enable-shared --enable-threads=posix --disable-libssp --disable-libstdcxx-pch --disable-libunwind-exceptions --disable-werror gdc_include_dir=/usr/include/dlang/gdc
Modèle de thread: posix
Algorithmes de compression LTO supportés: zlib zstd
gcc version 10.2.0 (GCC) 
```
After reading issue #1575, I checked the extension code and actual gcc output and found the failure here:
https://github.com/microsoft/vscode-cmake-tools/blob/d5fa7327a0ecde7dfb86c96263d089c04cfb24b0/src/kit.ts#L156

gcc help (man gcc) explains `-v` is not used to get the version (although it is in the output) but information:
```
-v        Print (on standard error output) the commands executed to run the stages of
           compilation.  Also print the version number of the compiler driver program
[...]

PR #1824, opened 2021-04-30
override the locale when querying compiler versions

bobbrow:
## This change addresses item #1821

This changes how compilers are queried when scanning for kits.

#### The following changes are proposed:
Set LANG and LC_ALL to "C" in the environment before invoking gcc or clang.

### DIFF
--- a/.github/workflows/build-vsix.yml
+++ b/.github/workflows/build-vsix.yml
@@ -2,7 +2,9 @@ name: Build VSIX
 
 on:
   push:
-    branches: [ develop ]
+    branches:
+      - develop
+      - bobbrow/gccLocale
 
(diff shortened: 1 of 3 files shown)


andreeis (approved):
Cool. I should pick this up for Makefile Tools.

bobbrow:
Yes, and I think there are a few other places in CMake Tools where we might want to override the locale as well.  In particular, I'm thinking of the problem matchers since we don't match all of the translated versions of the word "error", etc right now.  We might want to add a setting to switch the errors/warnings to English so that the problem matchers work, or we need to add a setting to map the translated errors to the English ones we know about.  Not sure what would be best...

PR #1828, opened 2021-05-03
clang-cl warning/error format in problem view

ki-bo:
### Changes parsing in the msvc diagnostics module to allow for empty warning/error codes

- adapt regex in msvc.ts to make Cxxxx warning and error codes optional

## The purpose of this change

Clang-cl allows clang to run in an msvc emulation mode, but the output is slightly different compared to cl.exe, since clang-cl is not reporting error and warning codes in its output


andreeis:
Thank you for this contribution. We will review and validate this change once we are done with shipping the CMake Tools 1.7.2 patch release.

andreeis (approved):

andreeis:
I noticed that the diagnostic functionality is not triggered by compile active file the same as it does by build. Not sure if it's by design or not, I opened issue https://github.com/microsoft/vscode-cmake-tools/issues/1984 to discuss in triage.

Issue #1832, opened 2021-05-04
CMake Tools Not Available as C++ Configuration Provider


KyleFromKitware:
### Brief Issue Summary

I'm trying to use CMake Tools as the C++ configuration provider, but it is not available when I go to "C/C++: Change Configuration Provider..."

### Expected:

1. Open command palette
2. Select "C/C++: Change Configuration Provider..."
3. "CMake Tools" should be an option

### Apparent Behavior:

1. Open command palette
2. Select "C/C++: Change Configuration Provider..."
3. "CMake Tools" is not an option

### Platform and Versions

- **Operating System**: Ubuntu
- **CMake Version**: 3.20.2
- **VSCode Version**: 1.55.2
- **CMake Tools Extension Version**: 1.7.1
- **Compiler/Toolchain**: GCC 9.3.0

### Other Notes/Information

I am using `CMakeUserPresets.json`, not sure if this is relevant or not.

bobbrow:
CMake Tools does not currently register itself until a successful "configure" has happened.  We recommend setting `cmake.configureOnOpen` to `true` until we make this the default value.  Otherwise, you can run the `CMake: Configure` command and that will start the provider.  We're also looking at #1149 and #477 which will help for the "unconfigured" or "configured outside of VS Code" situations.

KyleFromKitware:
Thanks. I tried running "CMake: Configure" and then changing the provider, but CMake Tools was still not available.

bobbrow:
Did the configure fail?  I don't know of any other reason why this shouldn't work.  Is this on the CMake repo, or some other one?

KyleFromKitware:
The configure succeeded. I just tried it a second time and was still not able to change the configuration provider.

I am working on the CMake repository, but I'm using `/usr/bin/cmake` for the configure.

bobbrow:
Hmm... A fresh clone of Kitware/CMake configured fine for me.  

![image](https://user-images.githubusercontent.com/12818240/117078308-7b8d8c00-acee-11eb-9070-777b54039623.png)

Would you be able to set the `cmake.loggingLevel` to `debug`, reload the window, run the "configure" command and then share the CMake Tools logs with me?  I think you have my email address if you don't want to share them publicly.

KyleFromKitware:
Thanks. I emailed you the debug log in private.

andreeis:
[...]

PR #1852, opened 2021-05-07
Make sure CMake Tools configuration provider gets registered with presets on

andreeis:
Fix for https://github.com/microsoft/vscode-cmake-tools/issues/1832.
The configuration provider was registered only when a kit was set. Presets don't have kits, so this code was skipped.
Ensure registration happens for presets on as well. The only thing that was used from the active kit was its environment and we can use the environment of the current preset.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -660,7 +660,7 @@ class ExtensionManager implements vscode.Disposable {
       if (!this._cppToolsAPI) {
         this._cppToolsAPI = await cpt.getCppToolsApi(cpt.Version.v5);
       }
-      if (this._cppToolsAPI && cmt.codeModel && cmt.activeKit) {
+      if (this._cppToolsAPI && cmt.codeModel && (cmt.activeKit || cmt.configurePreset)) {
         const codeModel = cmt.codeModel;
         const kit = cmt.activeKit;
         const cpptools = this._cppToolsAPI;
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -673,7 +673,18 @@ class ExtensionManager implements vscode.Disposable {
         }
         const drv = await cmt.getCMakeDriverInstance();
         const opts = drv ? drv.expansionOptions : undefined;
-        const env = await effectiveKitEnvironment(kit, opts);
+        let env: Map<string, string> = new Map<string, string>();
+        if (kit) {
+          env = await effectiveKitEnvironment(kit, opts);
+        } else if (cmt.configurePreset) {
+          for (const key in cmt.configurePreset.environment) {
+            const value = cmt.configurePreset.environment[key];
```
bobbrow, line 681: Just double checking... The vcvars are set in here, right?  I think I saw code that populates the environment with the stuff we need to find cl.exe, but just wanted to make sure you saw that too.

src/extension.ts
```diff
@@ -673,7 +673,18 @@ class ExtensionManager implements vscode.Disposable {
         }
         const drv = await cmt.getCMakeDriverInstance();
         const opts = drv ? drv.expansionOptions : undefined;
-        const env = await effectiveKitEnvironment(kit, opts);
+        let env: Map<string, string> = new Map<string, string>();
+        if (kit) {
+          env = await effectiveKitEnvironment(kit, opts);
+        } else if (cmt.configurePreset) {
```
bobbrow, line 679: I don't think you should need `else if`, just `else`, unless you plan to validate a second time that kit || preset was used (i.e. in case the `if` on line 663 changes and you want to add a final `else` that asserts or something)

bobbrow (approved):

src/extension.ts
xisui-MSFT: Yes, they are inserted to configurePreset.environment during expansion.

src/extension.ts
```diff
@@ -673,7 +673,18 @@ class ExtensionManager implements vscode.Disposable {
...
+          for (const key in cmt.configurePreset.environment) {
+            const value = cmt.configurePreset.environment[key];
+            if (value) {
```
xisui-MSFT, line 682: Can we do an util.isString(value) here instead? I think some falsy checks in presets.ts also need to be fixed at some point.

xisui-MSFT (approved):

src/extension.ts
andreeis: I had that if because of red squiggles (typescript didn't figure out that cmt.configurePreset was not null) and I didn't realize that I can use !.

src/extension.ts
andreeis: So, nothing for me to change here, right?

xisui-MSFT (approved):

bobbrow (approved):

PR #1855, opened 2021-05-07

csigs:


i18n/cht/src/cmake-tools.i18n.json
```diff
@@ -88,7 +88,7 @@
 	"waiting.on.variant": "正在等候變化選取項目",
 	"no.variant.abort": "未選取任何變化。中止設定",
 	"cannot.configure.no.config.preset": "無法設定: 此 CMake 工具沒有使用中的設定預設",
-	"warn.skip.configure.when.cache.present": "延伸模組判斷目前需要設定\\\n                          但因為設定 cmake.skipConfigureWhenCachePresent 已開啟，所以正在跳過設定。\\\n                          請確定 CMake 快取與最新的設定變更同步。",
+	"warn.skip.configure.when.cache.present": "The extension determined that a configuration is needed at this moment \\\r\n                          but we are skipping because the setting cmake.skipConfigureWhenCachePresent is ON. \\\r\n                          Make sure the CMake cache is in sync with the latest configuration changes.",
```
Colengms, line 91: I've noticed that localized strings have been inserting "\r" before "\n".  In this case, there appears to be some additional escaping, so I'm concerned that the \r being inserted (without that escaping) may be causing a problem...  One possible solution would be to avoid newlines in localized strings, and instead compose multiple lines from separate localized strings.  (Or, if it's a single sentence, maybe find some way to avoid needing to manually insert newlines).

bobbrow: There's definitely a bug here because even if the `\r` is wanted, it shouldn't leave the `\n` without and escaping slash.  It should have been `\\r\\n` (just two slashes for each), right?

bobbrow: Oh wait, is this just a really long string that doesn't actually want newlines added?  @andreeis, it looks like you touched this string last.  Do you remember?

Colengms: I think there are 2 problems here.  The backslash at the end of each line exists in code, where it's used to allow the string to be continued on the next line (however, the indent level gets included in the string as well, which does not seem desirable in this case).  The slashes are being captured by the loc export process, and included in the exported string (perhaps due to how vscode-nls-dev 'rewrites' the calls to `localize`).  (The other issue being the unwanted addition of the `\r`.)

Colengms: I opened: https://github.com/microsoft/vscode-nls-dev/issues/34

bobbrow: Thanks.  Did you also open an issue about the unwanted `\r`?  I updated this particular string instance so that the newlines go away, but there are a few other strings that have explicit `\n` in them.  It's probably ok for UI strings because I think extra whitespace is removed automatically, but it's not for logging.

Colengms: I'm working with our loc contact to investigate the `\r` issue.  He said he would open an issue for it, but we're still digging into it.  I'm pretty sure it's on their side, as I've confirmed we're exporting the string with only `\n`, and not doing anything (such as checking xlf files into repos) that might cause the change to come from us.

Colengms:
Closing this for now, as the Loc team works on some changes.  There may be a PR after this that I may need to cancel, if their changes did not work as expected.

PR #1866, opened 2021-05-11
Develop --> Main merge for CMake Tools 1.7.3

andreeis:


andreeis:
Yes I remember doing that by mistake. I will revert that first, thanks for the suggestion. I was wondering why this PR feels different. I solved the conflicts one by one but then it told me that this will check into develop which I didn't understand why and I didn't finalize on the conflicts. Hopefully the revert of the squash in main will make develop-->main FI conflict-less.

Issue #1656, opened 2021-02-09
refactor Task Provider


elahehrashedi:
we need to re-factor the task provider to utilize the latest updates on the task provider from Vs Code.

Will cover these features:
Keeping the list of multiple tasks in only one registered TaskProvider
Running tasks from different terminals
....

PR #1880, opened 2021-05-17
Task Provider for Cmake build tasks

elahehrashedi:
feature-request fix: https://github.com/microsoft/vscode-cmake-tools/issues/1656
the defined task:
```
{
	"label": "CMake: cmake example",
	"type": "cmake",
	"command": "build",
	"group": "build",
	"problemMatcher": "$msCompile",
	"detail": "user defined"
}
```

Invoke : `"Task: Run build Task"` and choose the `CMake: cmake example` task:

a build finished with error (target is chosen from the user defined target)
image

a build finished successfully (here the project was not configured yet and the target is set to all by default)
image

and this is a successful build after changing target:
image


package.json
```diff
@@ -489,12 +489,35 @@
       {
         "type": "cmake",
         "required": [
-          "command"
+          "command",
+          "label"
```
bobbrow, line 493: >  "label" [](http://example.com/codeflow?start=8&length=9)

Since we are modifying the existing task provider, we can't make this required without breaking people.  We should have the label be optional (it is optional for tasks.json in general) and create our own label for it when it is not set (e.g. "cmake: build")

package.json
```diff
@@ -492,9 +492,35 @@
...
+          "label": {
+            "type": "string",
+            "description": "The name of the task"
+          },
           "command": {
             "type": "string",
-            "description": "CMake command"
+            "description": "CMake build command"
+          },
+          "target": {
```
bobbrow, line 503: Let's save this for later.  I think we want to do "targets" as an array of string instead anyway.  You also mentioned support for multiple targets in the status bar that we don't have.  We should have issues for both.

The target we select for this task will be the active target (or "all" if no target selected)

package.json
```diff
@@ -492,9 +492,35 @@
...
+            "type": "string",
+            "description": "The build target"
+          },
+          "args": {
+            "type": "array",
+            "description": "Additional arguments"
+          },
```
bobbrow, line 510: Let's remove "args" for now as well.

src/cmake-tools.ts
```diff
@@ -244,7 +244,7 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         try {
           this._statusMessage.set(localize('reloading.status', 'Reloading...'));
           await drv.setBuildPreset(expandedBuildPreset);
-          await this.reRegisterTaskProviderforNewTarget(drv, undefined);
+          //this.updateDefaultTargetInTaskProvider(drv, undefined);
```
bobbrow, line 247: Uncomment this.  You also shouldn't need to pass "undefined" for the second parameter if it has correct type set up in the signature.

src/cmake-tools.ts
```diff
@@ -1487,17 +1487,16 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
     }
     await this._setDefaultBuildTarget(target);
-    const drv = await this._cmakeDriver;
-    await this.reRegisterTaskProviderforNewTarget(drv, target);
+    /*const drv = await this._cmakeDriver;
+    this.updateDefaultTargetInTaskProvider(drv, target);*/
   }
 
-  async reRegisterTaskProviderforNewTarget(drv: CMakeDriver | null, target: string | undefined) {
+  /*updateDefaultTargetInTaskProvider(drv: CMakeDriver | null, target: string | undefined) {
```
bobbrow, line 1494: per convention, prefer `target?: string` instead of `target: string | undefined`

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import {
+  TaskDefinition, Task, TaskGroup, TaskProvider, TaskScope, CustomExecution, WorkspaceFolder, Pseudoterminal, EventEmitter, Event, TerminalDimensions, window, TextEditor
```
bobbrow, line 6: import * as vscode from 'vscode';

package.json
```diff
@@ -492,9 +492,35 @@
...
         ],
         "properties": {
+          "label": {
+            "type": "string",
+            "description": "The name of the task"
+          },
           "command": {
             "type": "string",
-            "description": "CMake command"
+            "description": "CMake build command"
```
bobbrow, line 501: let's add:
```
"enum": [ "build" ],
```
to this command object so that people know what's valid to set here.  We can add others as we support them (e.g. "install", "test")

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+import path = require('path');
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+interface CMakeTaskDefinition extends TaskDefinition {
+  type: string;
+  label: string;
+  command: string; // Command is either "build" or "install".
+  args?: string[];
```
bobbrow, line 21: args can be removed

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+  options?: cp.ExecOptions | undefined;
```
bobbrow, line 22: >cp.ExecOptions [](http://example.com/codeflow?start=12&length=14)

{ cwd: string }

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+  public updateCMakeDriver(cmakeDriver: CMakeDriver | undefined) {
+    this.cmakeDriver = cmakeDriver;
+  }
+
+  public async provideTasks(): Promise<CMakeTask[]> {
+    const emptyTasks: CMakeTask[] = [];
+    const editor: TextEditor | undefined = window.activeTextEditor;
+    if (!editor) {
+        return emptyTasks;
+    }
```
bobbrow, line 46: We can provide tasks even if no files are open, so you can remove this.

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+      if (buildCommand) {
+        cmakePath = buildCommand.command;
+        args = buildCommand.args;
+      }
+    }*/
+    const definition: CMakeTaskDefinition = {
+      type: CMakeTaskProvider.CMakeScriptType,
+      label: taskName,
+      command: "build",
+      target: "all"
```
bobbrow, line 65: >target: "all" [](http://example.com/codeflow?start=6&length=13)

remove

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+    };
+    const task = new Task(definition, TaskScope.Workspace, taskName, CMakeTaskProvider.CMakeSourceStr);
+    task.group = TaskGroup.Build;
+    task.detail = "CMake build task, with build target set to \"all\".";
+    result.push(task);
+    return result;
+  }
+
+  public resolveTask(_task: CMakeTask): CMakeTask | undefined {
```
bobbrow, line 74: We try not to have `_` in variable names. Can this be removed?

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+    /*let buildCommand: BuildCommand | null;
+    let cmakePath: string = "CMake.EXE";
+    let args: string[] | undefined = [];
+    if (this.cmakeDriver) {
+      buildCommand = await this.cmakeDriver.getCMakeBuildCommand(this.defaultBuildTarget || CMakeTaskProvider.allTargetName);
+      if (buildCommand) {
+        cmakePath = buildCommand.command;
+        args = buildCommand.args;
+      }
+    }*/
```
bobbrow, line 60: delete commented code

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+  public get onDidWrite(): Event<string> { return this.writeEmitter.event; }
+  public get onDidClose(): Event<number> { return this.closeEmitter.event; }
+  private endOfLine: string = "\r\n";
+
+  constructor(private command: string, private target: string | null, private cmakeDriver: CMakeDriver | undefined) {
+  }
+
+  async open(_initialDimensions: TerminalDimensions | undefined): Promise<void> {
+      // At this point we can start using the terminal.
+      this.writeEmitter.fire(localize("starting_build", "Starting build...") + this.endOfLine);
```
bobbrow, line 102: nit: convention for localization keys is to use `.` not `_` between words.

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+      await this.doBuild();
+  }
+
+  close(): void {
+      // The terminal has been closed. Shutdown the build.
+  }
+
+  private async doBuild(): Promise<any> {
+    if (this.command !== "build") {
+      this.writeEmitter.fire(localize("not_a_build_command", "Not a recognized build command.") + this.endOfLine);
```
bobbrow, line 112: >not_a_build_command [](http://example.com/codeflow?start=39&length=19)

same here

bobbrow, line 112: we should return here too

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+    let buildCommand: BuildCommand | null;
+    let cmakePath: string = "CMake.EXE";
+    let args: string[] | undefined = [];
```
bobbrow, line 116: > | undefined  [](http://example.com/codeflow?start=22&length=13)

remove `| undefined` here since you are initializing it.

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+    if (this.cmakeDriver) {
+      buildCommand = await this.cmakeDriver.getCMakeBuildCommand(defaultBuildTarget);
+      if (buildCommand) {
+        cmakePath = buildCommand.command;
+        args = buildCommand.args;
+      }
+    }
+    // TODO: find the correct cwd
+    const cwd: string = path.dirname(cmakePath) ? path.dirname(cmakePath) : path.dirname(defaultBuildTarget);
+    const options: cp.ExecOptions | undefined = { cwd: cwd };
```
bobbrow, line 129: >| undefined [](http://example.com/codeflow?start=34&length=11)

remove `| undefined` here since you are initializing it

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+    const defaultBuildTarget: string = this.target ? this.target : "all";
+    if (this.cmakeDriver) {
+      buildCommand = await this.cmakeDriver.getCMakeBuildCommand(defaultBuildTarget);
+      if (buildCommand) {
+        cmakePath = buildCommand.command;
+        args = buildCommand.args;
+      }
+    }
+    // TODO: find the correct cwd
+    const cwd: string = path.dirname(cmakePath) ? path.dirname(cmakePath) : path.dirname(defaultBuildTarget);
```
bobbrow, line 128: The cwd should come from the task definition, if available.

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    const splitWriteEmitter = (lines: string | Buffer) => {
+      for (const line of lines.toString().split(/\r?\n/g)) {
+        this.writeEmitter.fire(line + this.endOfLine);
+      }
+    };
+    this.writeEmitter.fire(activeCommand + this.endOfLine);
+    try {
+      const result: number = await new Promise<number>((resolve) => {
+        cp.exec(activeCommand, options, (_error, stdout, _stderr) => {
```
bobbrow, line 148: >_error, stdout, _stderr) [](http://example.com/codeflow?start=41&length=24)

prefer not to have `_` in variable names

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    const splitWriteEmitter = (lines: string | Buffer) => {
+      for (const line of lines.toString().split(/\r?\n/g)) {
+        this.writeEmitter.fire(line + this.endOfLine);
+      }
+    };
+    this.writeEmitter.fire(activeCommand + this.endOfLine);
+    try {
+      const result: number = await new Promise<number>((resolve) => {
```
bobbrow, line 147: >await  [](http://example.com/codeflow?start=29&length=6)

We shouldn't await a new Promise here.  We can just return the promise and fire the closeEmitter with the result before calling "resolve".

src/cMakeTaskProvider.ts
bobbrow, line 148: >exec [](http://example.com/codeflow?start=11&length=4)

I think we should use `child_process.spawn` instead. [https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options)

Otherwise it is too long to wait for all of the stdout to be processed.  The user will be confused and think the build is hung.  With `spawn` you can register callbacks for stdout and stderr that will fire each time a line is written.

src/cMakeTaskProvider.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    if (this.cmakeDriver) {
+      buildCommand = await this.cmakeDriver.getCMakeBuildCommand(this.defaultTarget);
+      if (buildCommand) {
+        cmakePath = buildCommand.command;
+        args = buildCommand.args ? buildCommand.args : [];
+      }
+    }
+    if (this.options?.cwd) {
+      this.options.cwd = util.resolveVariables(this.options.cwd);
```
bobbrow, line 126: >util.resolveVariables [](http://example.com/codeflow?start=25&length=21)

I see you copied this over from cpptools.  cmake-tools has it's own resolver.  Look at `expand.expandStringHelper`.  To supply the `opts`, do a FindAllReferences and see what the other functions are doing.  There may be a way to make this better in the future (e.g. pass in a driver or extension object and we'll fill it out for you), but that would be for a separate PR.

elahehrashedi: in future we can use `expandStringHelper(str, this.cmakeDriver.expansionOptions)` to resolve variables other than `cwd`, 
currently vscode resolves `cwd `in `resolveTask`.

bobbrow (approved):

hwmaier:
I just saw this new feature being available in 1.8.0, great addition. It would be nice also to be able to set the "target" in the task definition.

bobbrow:
We talked about this and kept it "out of scope" for this PR, but I don't think an issue was opened to track it.  Thank you for the suggestion.  We'll open an issue now.

bobbrow:
#2122

PR #1881, opened 2021-05-17
(#1 in order) upgrade the tests [ready]

elahehrashedi:
upgrading the test using the new vscode API

A question: do we want to add "yarn run docs" to our yml files ?
The last two changes named "minor changes" and "testing" were efforts to resolve the build error, but they were not successful.


test/helpers/vscodefake/memento.ts
```diff
@@ -1,9 +1,29 @@
...
+  }
+  public update(key: string, value: any): Thenable<void> {
+    this.storage.set(key, value);
+    return Promise.resolve();
+  }
+  public containsKey(key: string): boolean { return this.storage.hasOwnProperty(key); }
+  public clear() { this.storage.clear(); }
+}
+
+export class StatetMemento implements vscode.Memento {
```
bobbrow, line 23: misspelled?  Should it be `StateMemento`?

bobbrow: Do we need both this class and `TestMemento`?  Would the tests still work if we just kept one of them?

package.json
```diff
@@ -1969,6 +1978,7 @@
     "vscode-cpptools": "^5.0.0",
     "vscode-extension-telemetry": "^0.1.6",
     "vscode-nls": "^5.0.0",
+    "webpack": "^5.38.1",
```
bobbrow, line 1981: I think webpack is supposed to stay in `devDependencies`. cpptools has it there too.

test.tsconfig.json
```diff
@@ -11,20 +11,11 @@
             "@cmt/*": ["src/*"],
             "@test/*": ["test/*"]
         },
-        "typeRoots": [
-            "node_modules/@types",
-            "types/"
-        ],
         "sourceMap": true,
-        "rootDir": ".",
-        "strict": true,
```
bobbrow, line 20: why are these deleted here, but kept in `tsconfig.json`?

package.json
```diff
@@ -1897,14 +1897,21 @@
...
+    "compile-watch": "yarn install && webpack --mode development --watch --progress",
     "compile-production": "yarn install && yarn run translations-generate && webpack --env BUILD_VSCODE_NLS=true --mode production",
     "translations-export": "gulp translations-export",
     "translations-generate": "gulp translations-generate",
     "translations-import": "gulp translations-import",
     "pretest": "tsc -p test.tsconfig.json",
-    "postinstall": "node ./node_modules/vscode/bin/install",
+    "pr-check": "gulp pr-check",
     "lint": "gulp lint",
+    "unitTests": "tsc -p test.tsconfig.json && node ./out/test/unit-tests/runTest.js",
```
bobbrow, line 1909: instead of invoking `tsc` directly in all these test scripts, can you do `yarn run pretest` instead?

package.json
```diff
@@ -1897,14 +1897,21 @@
...
+    "extensionTestsNoCMkeListsFile": "tsc -p test.tsconfig.json && node ./out/test/unit-tests/runTest.js",
```
bobbrow, line 1910: >extensionTestsNoCMkeListsFile [](http://example.com/codeflow?start=5&length=29)

CMake is mispelled and I think it's just running the unit test again.

.github/workflows/ci-develop-linux.yml
```diff
@@ -14,13 +14,51 @@ jobs:
...
+        uses: GabrielBB/xvfb-action@v1.4
+        with:
+          run: yarn run extensionTestsSingleRoot
+
+      - name: Run multioot tests
+        uses: GabrielBB/xvfb-action@v1.4
+        with:
+          run: yarn run extensionTestsMultioot
+
+      - name: Run smoke tests
```
bobbrow, line 61: I think it makes more sense to run the smoke tests first.

test.tsconfig.json
elahehrashedi: typeRoots were the one that was unnecessary and I removed that one, I brought back the other ones so that the two config file stay similar.

bobbrow:
> function Invoke-DocsBuild {

You can delete this function

---
Refers to: scripts/ci.ps1:58 in f461595. [](commit_id = f46159556a35dcc9a0d54b29caa1d5a2aa2b6802, deletion_comment = False)

scripts/ci.ps1
```diff
@@ -138,33 +113,6 @@ $ninja_binary = Install-TestNinjaMakeSystem -Version "1.8.2"
...
-    foreach ($name in @("multi-root-UI"; )) {
-        Invoke-VSCodeTest "CMake Tools: $name" `
-            -TestsPath "$REPO_DIR/out/test/extension-tests/$name" `
-            -Workspace "$REPO_DIR/test/extension-tests/$name/project-workspace.code-workspace"
-    }
-}
-
-Invoke-DocsBuild
 
 $vsce = Find-Program vsce
```
bobbrow, line 117: don't need this either

bobbrow:
> function Build-SphinxDocumentation {

don't need this anymore

---
Refers to: scripts/cmt.psm1:122 in f461595. [](commit_id = f46159556a35dcc9a0d54b29caa1d5a2aa2b6802, deletion_comment = False)

scripts/ci.ps1
```diff
@@ -138,33 +113,6 @@ $ninja_binary = Install-TestNinjaMakeSystem -Version "1.8.2"
 # Add ninja to search path environment variable
 $Env:PATH = (get-item $ninja_binary).Directory.FullName + [System.IO.Path]::PathSeparator + $Env:PATH
 
-if (! $NoTest) {
-    # Prepare to run our tests
-    Invoke-TestPreparation -CMakePath $cmake_binary
```
bobbrow, line 143: This command needs to run.

bobbrow:
>     $NoTest,

remove this

---
Refers to: scripts/ci.ps1:14 in f461595. [](commit_id = f46159556a35dcc9a0d54b29caa1d5a2aa2b6802, deletion_comment = False)

test/extension-tests/multi-root-UI/runTest.ts
```diff
@@ -14,9 +15,11 @@ async function main() {
 
     const testWorkspace = path.resolve(extensionDevelopmentPath, 'test/extension-tests/multi-root-UI/project-workspace.code-workspace');
 
-    const launchArgs = [ "--disable-extensions", testWorkspace ];
-
-    const extensionTestsEnv: { [key: string]: string | undefined } = { "CMT_TESTING": "1" };
+    const launchArgs = ["--disable-extensions", testWorkspace];
+    const extensionTestsEnv: { [key: string]: string | undefined } = {
+      "CMT_TESTING": "1",
+      "HasVs": os.platform().toLowerCase().includes("win") ? "true" : "false"
```
bobbrow, line 21: I don't see `HasVs` used anywhere in the product.  You can probably remove this variable from the test configs.

bobbrow:
It seems many of the tests are failing because the VM has CMake 3.20.3 and cmake-server support is removed from that version.  I think the ci script is able to install a specific version of CMake for the tests.  We might need to rely on that until we can update cmake-server tests to be skipped when the version doesn't support it.

elahehrashedi:
> It seems many of the tests are failing because the VM has CMake 3.20.3 and cmake-server support is removed from that version. I think the ci script is able to install a specific version of CMake for the tests. We might need to rely on that until we can update cmake-server tests to be skipped when the version doesn't support it.
I reverted invoking the tests from the yml file back to the shell script so that the tests can use the CMake 3.18.2

test/backend-unit-tests/cmake/strand.test.ts
```diff
@@ -6,13 +6,12 @@ import * as chaiAsPromised from 'chai-as-promised';
 
 chai.use(chaiAsPromised);
 
-import {Strand} from '@cmt/strand';
+import {Strand} from '../../../src/strand';
```
bobbrow, line 9: >../../../src [](http://example.com/codeflow?start=22&length=12)

can you revert this?

test/unit-tests/config.test.ts
```diff
@@ -1,5 +1,5 @@
-import {ConfigurationReader, ExtensionConfigurationSettings} from '@cmt/config';
-import {expect} from '@test/util';
+import {ConfigurationReader, ExtensionConfigurationSettings} from '../../src/config';
```
bobbrow, line 1: revert these

bobbrow (approved):

Issue #1800, opened 2021-04-22
Intellisense is not working when integrating conan packages using cmake_multi


andredsm:
### Brief Issue Summary

I am using vscode + cmake + conan on Windows 10 with MSBuild. Conan provides different ways to integrate their packages with cmake. One of them is named "cmake_multi" which aims multi configurations build tools, such as Visual Studio. Cmake_multi generates cmake scripts which in turn use cmake generator expressions to select the proper libraries, depending on if I am building a release or debug target. Using cmake_multi, the intellisense does not work as expected, it shows red squiggles on the include directives. Nonetheless, the cmake is able to generate the artifact successfully. 

![image](https://user-images.githubusercontent.com/877080/115707270-efba4e00-a344-11eb-80fd-f6c6de06dcae.png)

### CMake Tools Log

```
[main] Configuring folder: teste-conan 
[proc] Executing command: D:\bin\CMake\bin\cmake.EXE --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -Hd:/projects/teste-conan -Bd:/projects/teste-conan/build -G "Visual Studio 16 2019" -T host=x64 -A x64
[cmake] Not searching for unused variables given on the command line.
[cmake] -- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.19041.
[cmake] -- Conan: Using cmake global configuration
[cmake] -- Library Qt5AccessibilitySupportd found D:/.conan/68a569/1/lib/Qt5AccessibilitySupportd.lib
[cmake] -- Library Qt5Concurrentd found D:/.conan/68a569/1/lib/Qt5Concurrentd.lib
[cmake] -- Library Qt5Cored found D:/.conan/68a569/1/lib/Qt5Cored.lib
[cmake] -- Library Qt5DBusd found D:/.conan/68a569/1/lib/Qt5DBusd.lib
[cmake] -- Library Qt5DeviceDiscoverySupportd found D:/.conan/68a569/1/lib/Qt5DeviceDiscoverySupportd.lib
[cmake] -- Library Qt5EdidSupportd found D:/.conan/68a569/1/lib/Qt5EdidSupportd.lib
[cmake] -- Library Qt5EglSupportd found D:/.conan/68a569/1/lib/Qt5EglSupportd.lib
[...]

PR #1900, opened 2021-05-24
Preventing file index overwritting in multi-config generators

andredsm:
## This change addresses item #1800

bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/1800
### This changes the way the configuration data is updated. Now, in multi-config environments, only the current active variant is considered while updating internals. This way, I am preventing overwriting _fileIndex variable with other non-active variants.

- Add a new variable named name on CodeModelConfiguration type, which represents the name of the variant. 
- Add a new variable named activeVariant on CodeModelParams type, indicating the current active variant
- A condition on updateConfigurationData method that only allows the configuration proceeds whether the configuration variant is the same as the current active variant.

## The purpose of this change

As described on issue #1800, I was having a problem with the Intellisense when using the [conan package manager](https://conan.io/) + MSBuild. Basically, Conan has several ways to integrate the dependencies with my CMake file. One of them is the [cmake_multi](https://docs.conan.io/en/latest/integrations/build_system/cmake/cmake_multi_generator.html), which copes with multi-config build systems, relying on [cmake generator expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html). When updating the configuration data on cpptools module, we overwrite the file index several times, one per variant. This way, only the last processed configuration remains.  This is likely to point to wrong paths. In  #1800, I had only the libraries in release mode, but my last processed configuration was relWithDebInfo, which had nothing configurated, but we still overwrite the file index with empty values. Thus the missing paths and red squiggles on my include directives.


andreeis:
Thank you for your PR submission. We will get to review and test it soon.
PS: You have some easy to fix linter errors.

andredsm:
I fixed a typo on test cases that was causing the test to fail. I believe now they are all ok.

elahehrashedi (approved):
I tested the code.

Issue #1896, opened 2021-05-21
CMake Tools doesn't send --target to C/C++ tools


thunderstorm010:
See microsoft/vscode-cpptools#7519 for more info.

[TL;DR](https://github.com/microsoft/vscode-cmake-tools/issues/1896#issuecomment-954878454)

andreeis:
@thunderstorm010, this could be the same as https://github.com/microsoft/vscode-cmake-tools/issues/1879. We are almost done with a fix for that and once it's merged in, we will send you a vsix to install and try. If that will not solve your scenario then we will investigate further.

Colengms:
Hi @thunderstorm010 .  Would you be able to provide a simplified repro that we might use to validate a potential fix?  Perhaps a minimal project configured with the same target, using the same toolset, as a github repo, with setup instructions?

andreeis:
@thunderstorm010, can you try the CMake Tools vsix from [here](https://github.com/microsoft/vscode-cmake-tools/actions/runs/920800165) and confirm if it solves anything for you?

Colengms:
@thunderstorm010 When you build with CMake, it should generate 'target' files under a `.cmake\api\v1\reply` path, under the build directory.  CMake Tools collects compiler arguments from these files.  If you could provide one of these files from your repro, we may be able to determine why the `--target` arg was not picked up.  Otherwise, we seem to be in need of a repro we can use to investigate further.

yuxinyuan:
@andreeis I'm having pretty much the same issue. I'm not sure if I completely followed all the discussions, but the vsix you mentioned didn't solve the issue.

@Colengms I didn't see a .cmake/api/v1 path under the build directory

andreeis:
@yuxinyuan, there is no .cmake/api/v1 path under the build directory not even after you have a successful configure?
Also, maybe you can share with us a standalone project where we can reproduce this problem on our side, for easier debugging and investigating?

bobbrow:
@andreeis @yuxinyuan that path is created when you have cmake version 3.15 and newer as well as the `cmake.cmakeCommunicationMode` set to `automatic` or `fileApi`

yuxinyuan:
@andreeis @bobbrow I was using cmake version 3.10. After upgrading to version 3.18, I can now see the .cmake/api/v1 path. However, there is still no "target" field in the generated files. It seems to me that the extension has parsed --sysroot option from the compile commands, but fail to parse --target option.

[...]

Issue #1879, opened 2021-05-17
Check for target architecture when reading toolchain FileAPI


andreeis:
Opening from https://github.com/microsoft/vscode-cmake-tools/issues/1188, fix for toolchain FileAPI target.
A fix that almost works is here https://github.com/benmcmorran/vscode-cmake-tools/commit/1619ffcddebe35e5024e12b6b03e693b2e7f9a6a developed by @benmcmorran.
We will pick up and finalize how this interacts with getIntelliSenseMode.

PR #1902, opened 2021-05-25
Prefer toolchain fileApi target information over parsing of flags when sending intelliSenseMode to CppTools

andreeis:
Most of this comes from @benmcmorran. I added the undefined check in getIntelliSenseMode from cpptools.ts. We are supposed to let CppTools decide on a default when we don't have target information but when we do we should still send it.

This fixes the following issues:
https://github.com/microsoft/vscode-cmake-tools/issues/1896
https://github.com/microsoft/vscode-cmake-tools/issues/1879

EDIT:
Actually, this PR will be merged only with a fix for 1879. We will create a vsix and send to users that created 1879 and 1896 and if the "--target" issue is not fixed, we may end up implementing the reverse of parseCompilerArgs and parseTargetArch (which would read target information from toolchain fileApi and deduce what is the target switch for the given compiler: -m or -arch or --target or ... etc ...).


Colengms:
Will this _always_ pass the `--target` arg on to cpptools when present?  If not, I think that would be necessary.  As a compiler arg, it influences the include paths and defines returned by the compiler when we query it.

andreeis:
> Will this _always_ pass the `--target` arg on to cpptools when present? If not, I think that would be necessary. As a compiler arg, it influences the include paths and defines returned by the compiler when we query it.

Yes. We will leave CppTools to decide on a default only when CMake Tools doesn't know anything about the intended arch.

src/cpptools.ts
```diff
@@ -113,6 +113,10 @@ function parseTargetArch(target: string): Architecture {
     return 'arm64';
   } else if (target.indexOf('arm') >= 0 || is_arm_32(target)) {
     return 'arm';
+  } else if (target.indexOf('i686') >= 0) {
+    return 'x86';
+  } else if (target.indexOf('amd64') >= 0 || target.indexOf('x86_64')) {
```
michelleangela, line 118: Should `target.indexOf('x86_64')` have a check `>= 0`?

andreeis: Yes definitely. Thanks for catching that.

Issue #1149, opened 2020-03-31
Use cached cmake-file-api response to configure IntelliSense on startup


ghuser404:
I have `configureOnOpen` set to false, because I don't want it to happen automatically. However, I still expect to not see red squiggles when i open my project. To make them go away I need to configure project every time I open it.

The errors are with includes: this show for first unrecognized include `#include errors detected. Please update your includePath. Squiggles are disabled for this translation unit (C:\foo.cpp).` and this show for every include `cannot open source file "include/bar.h"`

bobbrow:
That is currently the designed behavior.  CMake needs to configure in order for the code model to be generated so that IntelliSense configurations can be provided.  This is also the root cause for your other issue #1148.

We need to do the work to persist the configurations.  Fortunately, with cmake-file-api (CMake 3.14.6+), they are already persisted.  We could easily just re-read them if the configure step has not happened yet.

ghuser404:
Yeah, I figured they are somewhat related. I probably wouldn't be as concerned if configuring a project would be a little easier, i.e. currently it's select CMake Tools extension -> press configure -> go back to Explorer view. I particularly don't like switching between views just to configure project. Another way, of course, is to use variants, i.e. press on variants -> select variant. But that requires to go through the list of variants and press on the one that you were previously working. Maybe there is an easier method that I just don't know about.

Overall, I'm quite happy with the extension, but I just think UI could be a bit better :)

KyleFromKitware:
+1 for reading from the existing File API response on startup.

Issue #1148, opened 2020-03-31
Last selected target isn't read on start up


ghuser404:
When I select any target but [all], and exit vscode, then on start up it will still show [all] as selected target. When I press Build, it switches to my previously selected target and starts building it.

I expect the extension to show correct selected target on start up.

bobbrow:
As mentioned in the other issue, the status bar is not currently updated until after the project has been configured.

Issue #477, opened 2018-07-12
Detect and support pre-configured build directories


Flamefire:
### Brief Issue Summary

Allow Usage of a pre-configured build directory.

### Expected:

1. Configure project via `cmake`/`ccmake`/`cmake-gui`
2. Open project source folder in VSC
3. (optional) Set the build directory in the configuration if non-default
4. CMake Tools picks up the CMakeCache.txt and determines the used variant and Kit or shows something useful if this is not possible automatically
5. Intellisense/Include directories etc. work (requires setting the compilation database and possibly a reconfigure with the `CMAKE_EXPORT_COMPILE_COMMANDS` define)

### Apparent Behavior:

- No sign that the CMakeCache is deteced
- Choosing a Kit matching the one used manually deletes the preconfigured build dir (see #392)

vector-of-bool:
This is one that people have asked for in the past.

spinicist:
Hello,

I'm being hit by this problem again quite badly. I have created a build script for new devs to run when they setup this project. This sets a couple of CMake variables, that in particular, determine how CMake calls vcpkg.

If said dev (or me) then opens the freshly configure repo in VS Code, the cache is blown away on the first configure, and then CMake proceeds to call vcpkg to rebuild all the dependencies, and does it with the incorrect settings.

Is there any solution?

PR #1909, opened 2021-05-27
Read cmake file API on open

elahehrashedi:
feature-request fix: https://github.com/microsoft/vscode-cmake-tools/issues/477
feature-request fix:  https://github.com/microsoft/vscode-cmake-tools/issues/1148
feature-request fix: https://github.com/microsoft/vscode-cmake-tools/issues/1149

This PR, make sure Intellisense is working before configure. 
If there is a pre-existing build folder, the code model will be picked up as well, for example, the latest chosen target will be shown in the bar at the buttom. If a build folder is copied from another path this will work, as long as all the paths inside files of the build folder are updated.


src/extension.ts
```diff
@@ -414,12 +415,15 @@ class ExtensionManager implements vscode.Disposable {
...
   }
 
-  async configureInternal(trigger: ConfigureTrigger, cmt: CMakeTools): Promise<void> {
+  async configureExtensionInternal(trigger: ConfigureTrigger, cmt: CMakeTools): Promise<void> {
     if (!await this._ensureActiveConfigurePresetOrKit(cmt)) {
       return;
     }
-
-    await cmt.configureInternal(trigger, [], ConfigureType.Normal);
+    if (trigger === ConfigureTrigger.configureOnOpen && !cmt.workspaceContext.config.configureOnOpen) {
```
bobbrow, line 422: I think this check could be moved into `configureInternal` and you could remove `ConfigureType.Auto` from the enum.

src/drivers/driver.ts
```diff
@@ -1095,6 +1104,8 @@ export abstract class CMakeDriver implements vscode.Disposable {
   }
 
   async configure(trigger: ConfigureTrigger, extra_args: string[], consumer?: proc.OutputConsumer, withoutCmakeSettings: boolean = false): Promise<number> {
+    // Check if this is an automatic configuration and adjust the logging messages based on that.
```
bobbrow, line 1107: I struggle with the term "automatic configuration".  To me this doesn't suggest that "CMake wasn't configured." I would prefer that we rename this to something else.  "configureFromCache" or "useCachedConfiguration" are some ideas.

elahehrashedi: changed to `readCachedConfiguration`

src/extension.ts
elahehrashedi: Removing the enum from CMakeDriver was more straightforward, as we have only one driver of type `CMakeFileApiDriver` and we could add a flag `_isConfiguredatLeastOnce` to it, and this flag shows if we have called configure at least once or not.
but in the `configureInternal` in CMakeTools, we have to check if the driver instance is type of `CMakeFileApiDriver` and then check if it's the first time or not. I found that to be a more confusing check, compare to having `ConfigureType.Auto`. 
I changed the code to remove `ConfigureType.Auto`. Let me know which one you think is more clear. I added some comments for clarity, in case we go for the second approach.

elahehrashedi: I made the change. the check is a little long but it works.

src/drivers/driver.ts
bobbrow: Can you rename the variable too?

src/extension.ts
bobbrow: Do you actually need to check the `type`?  I would think the `trigger` check makes that unnecessary.

You could also move some of the extra checks into an abstract method that returns true|false if it would be acceptable to load a cached configuration.  (e.g. drv is CMakeFileApiDriver && !configureOnOpen && !isConfiguredAtLeastOnce)

src/drivers/driver.ts
elahehrashedi: I renamed the variables as well

src/extension.ts
elahehrashedi: I added a simpler check, i.e. `drv.shouldUseCachedConfiguration(trigger)` that calls to the function `isCacheConfigSupported` and this approach won't need to check the instance of the driver directly.

src/drivers/driver.ts
```diff
@@ -69,6 +70,12 @@ export abstract class CMakeDriver implements vscode.Disposable {
    * @returns The exit code from CMake
    */
   protected abstract doConfigure(extra_args: string[], consumer?: proc.OutputConsumer): Promise<number>;
+  protected abstract doCacheConfigure(): Promise<number>;
+
+  protected _isConfiguredAtLeastOnce = false;
```
bobbrow, line 75: You only set this to true in the file-api driver, but it is owned by the base driver class.  I think you should make this private (the 'get' below can be protected), and set it to true when configure is called here in the base class.

src/drivers/driver.ts
```diff
@@ -69,6 +70,12 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+  protected isConfiguredAtLeastOnce(): boolean {
```
bobbrow, line 76: nit: you can make this a 'get' property so that you don't have to access it like a function (i.e. no parenthesis)

src/drivers/driver.ts
```diff
@@ -82,16 +89,29 @@ export abstract class CMakeDriver implements vscode.Disposable {
     return Promise.resolve(true);
   }
 
+  /**
+   * Check if using cached configuration is supported.
+   */
+  protected abstract isCacheConfigSupported(): boolean;
```
bobbrow, line 95: nit: you can make this a 'get' property so that you don't have to access it like a function (i.e. no parenthesis)

bobbrow (approved):

src/drivers/driver.ts
```diff
@@ -1139,13 +1173,23 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
 
       // A more complete round of expansions
       await this._refreshExpansions();
 
       const timeStart: number = new Date().getTime();
-      const retc = await this.doConfigure(expanded_flags, consumer);
+      let retc: number;
+      if (usingCachedConfiguration) {
+        retc = await this.doCacheConfigure();
+        this._isConfiguredAtLeastOnce = true;
```
bobbrow, line 1187:
```ts
this._isConfiguredAtLeastOnce = true;
```
Since you set this the same in both branches of the code, you can move this outside of the if/else.

bobbrow (approved):

Issue #1934, opened 2021-06-16
Existing presets are not shown


KUGA2:
### Brief Issue Summary

The plugin does not read existing (and working - tested via cli) presets in CMakePresets.json.

### Expected:

From what I read [here](CMakePresets.json), klicking on the button "No Configure Preset Selected" should list my presets. It does not.

### Apparent Behavior:

It wants me to create a new preset. See video.
https://streamable.com/19sk8n

### CMake Tools Log

There is an error here:

```
[presetController] Invalid kit contents CMakePresets.json (d:\git\platform_rts\CMakePresets.json):
[presetController]  >> /configurePresets/0/cacheVariables/CMAKE_BUILD_TYPE: should be equal to one of the allowed values
[presetController]  >> /configurePresets/0/cacheVariables/CMAKE_BUILD_TYPE: should be object
[presetController]  >> /configurePresets/0/cacheVariables/CMAKE_BUILD_TYPE: should match some schema in anyOf
```

So the reason is, that I specified a variable for CMAKE_BUILD_TYPE.

```
            "cacheVariables": {
                "CMAKE_INSTALL_PREFIX": "Build/EXPORT_PLATFORM",
                "CMAKE_BUILD_TYPE": "$env{BUILD_TYPE}",
                "CMAKE_TOOLCHAIN_FILE": "${sourceDir}/Scripts/build_resources/Components/CMake/Toolchain/$env{TOOLCHAIN}.cmake"
            },
            "environment": {
                "TOOLCHAIN": "placeholder",
                "BUILD_TYPE": "Release"
            }
```

There is also a tooltip warning:
![image](https://user-images.githubusercontent.com/10581931/122242422-99403a80-cec3-11eb-8af0-44161862ffab.png)

So I guess, I must disable this check somehow. How?

### Platform and Versions

- **Operating System**:  Windows
- **CMake Version**: 3.20.4
- **VSCode Version**: 1.57
- **CMake Tools Extension Version**: 1.7.3
- **Compiler/Toolchain**: Visual C++ 2019

bobbrow:
@xisui-MSFT @gugavaro Did we add `CMAKE_BUILD_TYPE` to the CMake schema when we pulled from their repo so we could get IntelliSense for it?  What I see now seems to reject environment references.

```
                          "CMAKE_BUILD_TYPE": {
                              "anyOf": [
                                  {
                                      "type": "string",
                                      "description": "Specifies the build type on single-configuration generators.",
[...]

PR #1935, opened 2021-06-16
Update schema for cacheVariables

bobbrow:
## This change addresses item #1934

The schema was too strict for CMAKE_BUILD_TYPE and rejected environment references. It was also missing support for `null` which is a valid way to clear the value of a variable for a given preset.

The object support was removed as I don't believe it is valid for `CMAKE_BUILD_TYPE`.

NOTE FOR REVIEWERS: You should "hide whitespaces changes" to see the actual changes in the file.  The indentation of the entire file was off by 2 which caused enough pain editing the schema that I reformatted the file as part of this.


bobbrow:
@gugavaro This PR affects the schema and we will want to bring the changes back to VS too.

Issue #1905, opened 2021-05-27
detecting vs2015 comes with vs2019 properly


lygstate:
According to C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat
We can specify options for vs2015 or vs 2017
```bat

:usage_error
echo [ERROR:%~nx0] Error in script usage. The correct usage is:
goto :usage

:usage
echo Syntax:
echo     %~nx0 [arch] [platform_type] [winsdk_version] [-vcvars_ver=vc_version] [-vcvars_spectre_libs=spectre_mode]
echo where :
echo     [arch]: x86 ^| amd64 ^| x86_amd64 ^| x86_arm ^| x86_arm64 ^| amd64_x86 ^| amd64_arm ^| amd64_arm64
echo     [platform_type]: {empty} ^| store ^| uwp
echo     [winsdk_version] : full Windows 10 SDK number (e.g. 10.0.10240.0) or "8.1" to use the Windows 8.1 SDK.
echo     [vc_version] : {none} for latest installed VC++ compiler toolset ^|
echo                    "14.0" for VC++ 2015 Compiler Toolset ^|
echo                    "14.xx" for the latest 14.xx.yyyyy toolset installed (e.g. "14.11") ^|
echo                    "14.xx.yyyyy" for a specific full version number (e.g. "14.11.25503")
echo     [spectre_mode] : {none} for libraries without spectre mitigations ^|
echo                      "spectre" for libraries with spectre mitigations
echo:
echo The store parameter sets environment variables to support Universal Windows Platform application
echo development and is an alias for 'uwp'.
echo:
echo For example:
echo     %~nx0 x86_amd64
echo     %~nx0 x86_amd64 10.0.10240.0
echo     %~nx0 x86_arm uwp 10.0.10240.0
echo     %~nx0 x86_arm onecore 10.0.10240.0 -vcvars_ver=14.0
echo     %~nx0 x64 8.1
echo     %~nx0 x64 store 8.1
echo:
echo Please make sure either Visual Studio or C++ Build SKU is installed.
goto :end

:missing
echo The specified configuration type is missing.  The tools for the
echo configuration might not be installed.
goto :end

:end
set __VCVARSALL_TARGET_ARCH=
set __VCVARSALL_HOST_ARCH=
set __VCVARSALL_STORE=
set __VCVARSALL_WINSDK=
set __VCVARSALL_PARSE_ERROR=
[...]

PR #1955, opened 2021-06-26
Fixes msvc2015 detection when only vs2019 are installed.

lygstate:

Fixed:
https://github.com/microsoft/vscode-cmake-tools/issues/1905

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -554,41 +554,86 @@ export function kitHostTargetArch(hostArch: string, targetArch?: string, amd64Al
 }
 
 /**
+ * Possible msvc host architectures
+ */
+export const MSVC_HOST_ARCHES: string[] = ['x86', 'x64'];
+
+/*
  * List of environment variables required for Visual C++ to run as expected for
  * a VS installation.
+ * The diff of vcvarsall.bat output env and system env:
+    DevEnvDir=C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\
+    Framework40Version=v4.0
+    FrameworkDir=C:\Windows\Microsoft.NET\Framework\
+    FrameworkDIR32=C:\Windows\Microsoft.NET\Framework\
+    FrameworkVersion=v4.0.30319
+    FrameworkVersion32=v4.0.30319
+    INCLUDE=C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE;C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\ATLMFC\INCLUDE;C:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\ucrt;C:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um;C:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\shared;C:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\um;C:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\winrt;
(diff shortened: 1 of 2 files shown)


andreeis:
@lygstate, we got to review this for 1.8. I merged with the latest changes in develop and I am testing.
Even your first commit did not include util.envSet and util.envGetValue in util.ts. First I thought that I merged something wrong.
I am did some changes on my side to unblock testing but if you had anything special in mind for envSet and engGetValue please push an update.

andreeis:
> Also fixes the output encoding.
@lygstate, the output encoding fix is in util.parseVersion? Do we have a GitHub issue created for that?

src/kit.ts
```diff
@@ -779,25 +847,28 @@ const VsGenerators: {[key: string]: string} = {
...
   console.log(`varsForVSInstallation path:'${inst.installationPath}' version:${inst.installationVersion} host arch:${hostArch} - target arch:${targetArch}`);
   const common_dir = path.join(inst.installationPath, 'Common7', 'Tools');
+  const majorVersion = parseInt(inst.installationVersion);
+
   let vcvarsScript: string = 'vcvarsall.bat';
-  if (targetArch === "arm" || targetArch === "arm64") {
-    // The arm(64) vcvars filename for x64 hosted toolset is using the 'amd64' alias.
-    vcvarsScript = `vcvars${kitHostTargetArch(hostArch, targetArch, true)}.bat`;
+  if (majorVersion < 14) {
+    if (targetArch === "arm" || targetArch === "arm64") {
```
andreeis, line 854: @lygstate, did you have anything in mind with moving the whole targetArch === "arm" or "arm64" inside the if majorVersion < 14? It causes arm/arm64 kits to be created in cmake-user-kits.json even when arm/arm64 is not installed.

lygstate: Indeed, I don't know why this is needed, as for  majyorVersion>=14, we can always using 'vcvarsall.bat'.
And for vs2010, 'vcvarsall.bat' also exist. I even think vcvarsScript = `vcvars${kitHostTargetArch(hostArch, targetArch, true)}.bat`; should be removed totally, I don't know why this present

lygstate:
> > Also fixes the output encoding.
> 
> @lygstate, the output encoding fix is in util.parseVersion? Do we have a GitHub issue created for that?

I split the fix of encoding problem in separate PR https://github.com/microsoft/vscode-cmake-tools/pull/1985

src/kit.ts
andreeis: The reason for invoking vcvars target custom name is that, when arm/arm64 is not installed, that script will not exist. If installed, the script will exist. If we use vcvarsall, it exists regardless of arm/arm64 installed and also it doesn't fail when run. On top of that, it produces INCLUDE and other env vars so it will cause arm/arm64 kits to be created, listed even if there is no arm/arm64 support in the VS installation. Users can be very confused by this.

andreeis:
@lygstate, if you want you can push an update. I tested this and I sign off with 2 comments:
- revert the vcvars file name change as it was before this PR, for arm32/arm64
- figure out what you wanted with util.envGetValue and util.envSet. For now, when testing, I simply used var.get and var.set.
If you won't have the chance, I'll push an update or start a new PR from my own branch (derived from yours) later tomorrow.

andreeis:
@lygstate, I noticed that you mentioned that this PR depends on https://github.com/microsoft/vscode-cmake-tools/pull/1763. We may not be able to test thoroughly the env var consistency PR before we release 1.8. It was a large PR with interesting scenarios to cover. When I tested this msvc PR it worked fine. Let me know if I overlooked something. Do you think this msvc PR can work without the env var consistency one?

Another question, can you describe the encoding problem? Just to make sure I properly test that one too and I don't miss something. What was happening.

andreeis:
@lygstate, for an example of what I needed to change in order to have successful testing see branch dev/andris/cmake-tools/fromPR1955.

andreeis (approved):

bobbrow:
@lygstate Is it necessary to add 'Platform' to the list of MSVC variables to make your PR succeed?  We are experiencing a regression in building some of our projects and this blocks the release of 1.8.0.  @elahehrashedi has PR #2056 out where she removes this from the list and it fixes our problem.  Do you see any problem with us removing it?

Issue #1956, opened 2021-06-28
Some hardcoded (previsously made optional) directives were brought back


PedroLima92:
### Brief Issue Summary

Commit 4293388 brought back the **ctest** hardcoded directives `'-T', 'test', '--output-on-failure'` which were made optional on PR #1729. Some of the code we wrote was kept, though. Should I interpret this as an accident?

### Quick fix

It seems from my side that fixing it will just be a matter of deleting them from here:

https://github.com/microsoft/vscode-cmake-tools/blob/b71fb38d89faa256e416c11092c93d8c74a4edbe/src/ctest.ts#L366-L367

I would've submit a pull request, but if this is all it takes to fix it you guys should get to it much faster than I

bobbrow:
It looks like an incorrect merge happened.  Thanks for spotting this!

PR #1962, opened 2021-06-28
Fix a regression in ctestDefaultArgs

bobbrow:
#1956

### DIFF
--- a/src/ctest.ts
+++ b/src/ctest.ts
@@ -363,7 +363,7 @@ export class CTestDriver implements vscode.Disposable {
       ctestArgs = ['-T', 'test'].concat(testArgs(driver.testPreset));
     } else {
       const configuration = driver.currentBuildType;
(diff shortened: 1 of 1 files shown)


xisui-MSFT:
I believe `-T Test` is required to make the test status in the status bar correct (something like `9/10 passed`). Notice that since it writes to a file, you may need to make a failed test success / make a successful test fail to see the difference.

xisui-MSFT (changes requested):
Can you please test if the test result in the status bar works correctly?

bobbrow:
`-T Test` is in `this.ws.config.ctestDefaultArgs` by default.  This change was made so that users can override the defaults and set different options if this is not what they wanted.

gugavaro (approved):

xisui-MSFT (approved):

PR #1967, opened 2021-07-01
Add pre-commit hook to run linter before git commit

xisui-MSFT:
[Husky](https://www.npmjs.com/package/husky/v/4.3.8) is a package that makes git hooks easier. I decide to use V4 since V5+ requires a specific config file for it, and so doesn't feel like a good practice (VS Code is using 0.13.1 so I think we are safe here).

In this PR, use husky to run linter as a pre-task for git commit since apparently we all tend to forget to run it...


elahehrashedi:
I have a question, I tested the branch, it makes committing takes longer. Is it necessary? I normally run "gulp lint" or "yarn run lint" before pushing my branch, not on every commit I have in the middle stages.

xisui-MSFT:
The idea is that, people (like me) forget to run it from time to time. Plus, we can add more checks to this hook when needed. If you often commit incomplete work, we can use a pre-push hook instead of a pre-commit one. Does this sound good to you?

elahehrashedi:
@xisui-MSFT yes pre-push would be a better idea

package.json
```diff
@@ -1967,6 +1972,7 @@
     "chai-as-promised": "^7.1.1",
     "chai-string": "^1.5.0",
     "clang-format": "^1.5.0",
+    "husky": "^4.3.8",
```
elahehrashedi, line 1975: Could you please move it few lines lower to keep the alphabetic order?

elahehrashedi (approved):

bobbrow:
So I just tried to push and got a generic failure from VS Code.  The experience was confusing.  Is there a better way to make it clear why the push fails?  I ended up looking at the git log and saw this:

image

Is there a way to make it log where the problems are?

xisui-MSFT:
@bobbrow  The output should be exactly the same as running `yarn run lint` in the command line, for example:
```
PS D:\CMakeTools> git push
warning cmake-tools@1.7.3: The engine "vscode" appears to be invalid.
husky > pre-push (node v14.16.0)
warning cmake-tools@1.7.3: The engine "vscode" appears to be invalid.
$ gulp lint
[16:57:39] Using gulpfile D:\CMakeTools\gulpfile.js
[16:57:39] Starting 'lint'...
[16:57:48]
D:\CMakeTools\src\drivers\driver.ts
  1181:38  error  Missing semicolon  @typescript-eslint/semi

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

[16:57:48] 'lint' errored after 8.6 s
[16:57:48] ESLintError in plugin "gulp-eslint"
Message:
    Failed with 1 error
Details:
    domainEmitter: [object Object]
    domainThrown: false

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
husky > pre-push hook failed (add --no-verify to bypass)
error Command failed with exit code 1.
error: failed to push some refs to 'https://github.com/xisui-MSFT/vscode-cmake-tools.git'
```

Can you provide the repro steps of that error you encountered?

bobbrow:
I didn't push on the command line.  I pushed by using the "sync" button in the status bar.

xisui-MSFT:
@bobbrow Only stderr is stored in that file: https://github.com/microsoft/vscode/blob/c8d5c905bfdf6c78e84c6df501ff3ad2b79db006/extensions/git/src/commands.ts#L2789. We probably should ask them to write stdout to the same file or another one as well.

Alternatively, we can forward that yarn output to stderr.

PR #1969, opened 2021-07-01
Fix a race in the variants test

bobbrow:


test/extension-tests/successful-build/test/variant-envs.test.ts
```diff
@@ -80,8 +81,17 @@ suite('[Environment Variables in Variants]', async () => {
         }
       }
     });
+    console.log('done updating defaultVariants');
+    const sleep = new Promise<void>(resolve =>{
```
andreeis, line 85: Very ingenious :). I think it fixed the env var problem, although a few reruns may confirm this one wasn't just a lucky pass. Maybe we can figure out a similar fix for the setKitByName issue? Both that and the env var used to fail in the past but recently it's much harder to get a pass, lots of reruns needed.

bobbrow: I have a suspicion that setKitByName is affected by the target triplet bug.  When that PR goes in, we can try again. For now, I think this PR is ready to go.

andreeis (approved):

PR #1973, opened 2021-07-02
Updating some parsing/logging for toolset

bobbrow:
I saw a few warnings in my presets file and realized I had incorrectly told people to set the toolset object to 'x64' instead of 'host=x64'.  I'm not sure why it worked before, but it doesn't now. I'm adding some logging that will help guide folks to the right syntax for toolset.


elahehrashedi:
without preset, it was working before, in `kit.ts` we had added this code: 

`const host: string = hostArch.toLowerCase().replace(/ /g, "").startsWith("host=") ? hostArch : "host=" + hostArch;`

but it looks like we are not accepting the same error to happen in presets.

bobbrow:
> without preset, it was working before, because in `kit.ts` we had added this code:
> 
> `const host: string = hostArch.toLowerCase().replace(/ /g, "").startsWith("host=") ? hostArch : "host=" + hostArch;`
> 
> but it looks like we are not accepting the same error to happen in presets.

Correct.  The reason not to accept this is because the presets file is intended to be consumed the same way by all IDE's and CMake itself.  So if we patch it up here, then the customer tries to use it on the command line or in Visual Studio, it will be broken.  Because of this, I thought it would be best to add a warning to help them fix it instead.

src/preset.ts
```diff
@@ -572,43 +651,9 @@ async function expandConfigurePresetHelper(folder: string,
...
-            log.warning(noToolsetArchWarning);
-          } else if (util.isString(preset.toolset)) {
-            matchToolsetArchAndVersion(preset.toolset);
-          } else if (!preset.toolset.value) {
-            log.warning(noToolsetArchWarning);
-          } else {
-            matchToolsetArchAndVersion(preset.toolset.value);
-          }
+          const arch = getArchitecture(preset);
+          const toolset = getToolset(preset);
```
elahehrashedi, line 655: I can not hit this code. the cl location is found and then I won't see any error related to toolset.
Could you please share your `configurePresets`?

bobbrow: I believe these were the important parts to get into this situation:
```
{
  "generator": "Ninja",
  "architecture": { ... },
  "toolset": { ... },
  "cacheVariables": {
    "CMAKE_C_COMPILER": "cl.exe",
    "CMAKE_CXX_COMPILER": "cl.exe"
  }
}
```

elahehrashedi: I still need to run the code like this:
```
        const arch = getArchitecture(preset);
        const toolset = getToolset(preset);
        if (!clLoc.stdout) {
```

to be able to see the warning messages. 

image

elahehrashedi (approved):
I tested the code.

Issue #1884, opened 2021-05-18
Environment variables from cmake presets do not seem to carry over to debugger 


moyo1997:
I specified LD_LIBRARY_PATH as an environment variable in my cmake presets. However, when I tried to debug with the “CMake: Debug” command and also right clicking the target and selecting debug, both didn’t work as I got this error : “error while loading shared libraries: libPocoCrypto.so.62: cannot open shared object file: No such file or directory”.
Once I created the launch.json and specified the LD_LIBRARY_PATH and then selected that launch.json configuration at the bottom to debug, then debugging worked.

xisui-MSFT:
Best place to add this:

For debugging: https://github.com/microsoft/vscode-cmake-tools/blob/951ec68872e6f8a08fba9be378e19034b274071b/src/cmake-tools.ts#L1769

For launching: https://github.com/microsoft/vscode-cmake-tools/blob/951ec68872e6f8a08fba9be378e19034b274071b/src/cmake-tools.ts#L1812

elahehrashedi:
we can carry over the `build` preset environment variables to the `debug`.

xisui-MSFT:
> we can carry over the `build` preset environment variables to the `debug`.

Not sure if build preset is the best candidate for this... Maybe test preset is better?

bobbrow:
I think build preset is the preferred choice.  Test presets (or even tests configured by CMake) may not exist.

PR #1977, opened 2021-07-06
adding configurePreset environment to "debug" and "launch" target

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/1884
still debating whether to use config presets or build presets.
Update: we are using the config presets.


src/drivers/driver.ts
```diff
@@ -207,23 +207,24 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+    return envs;
   }
 
   /**
    * Get the environment variables that should be set at CTest and running program time.
    */
   async getCTestCommandEnvironment(): Promise<proc.EnvironmentVariables> {
     if (this.useCMakePresets) {
-      return this._testPreset?.environment as proc.EnvironmentVariables;
+      return (this._testPreset?.environment ? this._testPreset?.environment : {}) as proc.EnvironmentVariables;
```
elahehrashedi, line 227: this was throwing an error when there was no `this._testPreset?.environment`. fixed it in this PR although not related to the change.

src/drivers/driver.ts
```diff
@@ -207,23 +207,24 @@ export abstract class CMakeDriver implements vscode.Disposable {
    * Get the environment variables that should be set at CMake-build time.
    */
   async getCMakeBuildCommandEnvironment(in_env: proc.EnvironmentVariables): Promise<proc.EnvironmentVariables> {
```
elahehrashedi, line 209: Should we revert this in case we decide to use Config Preset Environment variables using `getConfigureEnvironment` ? Or do we still want these in_env to be used?

src/drivers/driver.ts
```diff
@@ -207,23 +207,24 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+    let envs: proc.EnvironmentVariables;
```
elahehrashedi, line 210: Should we revert this in case we decide to use Config Preset Environment variables using `getConfigureEnvironment `? Or do we still want the `in_env` to be used?

src/drivers/driver.ts
```diff
@@ -207,23 +207,24 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
     if (this.useCMakePresets) {
-      return this._buildPreset?.environment as proc.EnvironmentVariables;
+      envs =  util.mergeEnvironment(in_env, this._buildPreset?.environment as proc.EnvironmentVariables);
```
elahehrashedi, line 212: Should we revert this in case we decide to use Config Preset Environment variables using `getConfigureEnvironment `? Or do we still want the `in_env `to be used?

bobbrow:
@elahehrashedi Can you update this PR to get the "configure" environment instead of the "build" environment?

src/drivers/driver.ts
andreeis: I believe we want still in_env.

andreeis: @xisui-MSFT, maybe you want to confirm whether we should or shouldn't use in_env in getCMakeBuildCommandEnvironment here? Originally, before presets, in_env was used. With the support of presets, I don't know if you intentionally skipped merging with the incoming in_env or simply missed this doing a merge.

xisui-MSFT: This is intentional. All other sources of env are ignored when using presets.

andreeis: Ok, but the functionality fixed before presets in getCMakeBuildCommandEnvironment will not match when presets are on. I understand the difference between getCMake[Configure][BuildCommand][Test]Environment family but talking only about getCMakeBuildCommandEnvironment, with presets on and off, I am not sure about the difference.

xisui-MSFT: I think matching between presets on/off had never been the goal here, but we are trying more to match with the behavior of `cmake --preset` command when presets are on. And so this is why we ignore a large number of settings and defaults, including but not limited to environments, in pre-presets era.

andreeis: Makes sense, thank you @xisui-MSFT for your input.

elahehrashedi:
> @elahehrashedi Can you update this PR to get the "configure" environment instead of the "build" environment?

One of the tests is broken. I am working on that.

src/util.ts
```diff
@@ -359,7 +359,7 @@ export function* flatMap<In, Out>(rng: Iterable<In>, fn: (item: In) => Iterable<
   }
 }
 
-export function splitEnvironmentVars(env: EnvironmentVariables): EnvironmentVariables[] {
+export function expandEnvironmentVars(env: EnvironmentVariables): EnvironmentVariables[] {
```
xisui-MSFT, line 362: I would suggest to change this back to split. Term 'expand' may cause confusion with terms used in expand.ts.

andreeis (approved):

src/drivers/driver.ts
```diff
@@ -207,23 +207,24 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
-      envs = util.mergeEnvironment(envs, await this.computeExpandedEnvironment(this.config.environment, envs));
-      envs = util.mergeEnvironment(envs, await this.computeExpandedEnvironment(this.config.buildEnvironment, envs));
-      envs = util.mergeEnvironment(envs, await this.computeExpandedEnvironment(this._variantEnv, envs));
-      return envs;
+      envs = util.mergeEnvironment(in_env, getKitEnvironmentVariablesObject(this._kitEnvironmentVariables));
     }
+    envs = util.mergeEnvironment(envs, await this.computeExpandedEnvironment(this.config.environment, envs));
+    envs = util.mergeEnvironment(envs, await this.computeExpandedEnvironment(this.config.buildEnvironment, envs));
+    envs = util.mergeEnvironment(envs, await this.computeExpandedEnvironment(this._variantEnv, envs));
+    return envs;
```
lygstate, line 219: Why getCMakeBuildCommandEnvironment are using
this.computeExpandedEnvironment(this.config.environment
this.config.buildEnvironment
this._variantEnv

Why 
getCTestCommandEnvironment 
getConfigureEnvironment didn't update

Issue #1923, opened 2021-06-08
"Inherit from configure presets" - maybe show also the "hidden" presets?


Zingam:
The meaning I find in the "hidden" parameter of a configure preset is to hide it from the IDE for presets that are used for setting up common properties of presets that will inherit from them. I that case I find it meaningful when invoking the command "New Preset" and then selecting the option "Inherit from configure presets" to also be able to see these base hidden presets.

Does that make sense?

```json
    "configurePresets": [
        {
            "name": "default",
            "hidden": true,
            "binaryDir": "${sourceDir}/build/${presetName}",
        },
        {
            "name": "default-x64-windows-ninja",
            "hidden": true,
            "inherits": [
                "default"
            ],
            "generator": "Ninja",
            "architecture": {
                "value": "x64",
                "strategy": "external"
            },
            "vendor": {
                "microsoft.com/VisualStudioSettings/CMake/1.0": {
                  "hostOS": [ "Windows" ]
                }
              }
        },
        {
            "name": "windows-x64-ninja-debug",
            "displayName": "Windows x64 Debug",
            "description": "Default build using Ninja generator",
            "inherits": [
                "default-x64-windows-ninja"
            ],
            "cacheVariables": {
                "CMAKE_BUILD_TYPE": "Debug"
            }
        },
    ]
}
```

bobbrow:
Yes, this makes perfect sense.  Thanks for reporting this issue.

PR #1979, opened 2021-07-06
"Inherit from configure presets" - show also the "hidden" presets

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/1923


andreeis (approved):

src/presetsController.ts
```diff
@@ -552,14 +552,24 @@ export class PresetsController {
     }
   }
 
-  // Returns the name of the preset selected
-  private async showPresetSelector(presets: preset.Preset[], options: vscode.QuickPickOptions & { canPickMany: true }): Promise<string[] | undefined>;
-  private async showPresetSelector(presets: preset.Preset[], options: vscode.QuickPickOptions): Promise<string | undefined>;
-  private async showPresetSelector(presets: preset.Preset[], options: vscode.QuickPickOptions): Promise<string | string[] | undefined> {
+  // Returns the name of preset selected from the list of non-hidden presets.
+  private async selectNonHiddenPreset(presets: preset.Preset[], options: vscode.QuickPickOptions): Promise<string | undefined> {
+    const chosenPreset: string | undefined = await this.selectPreset(presets, options, false);
```
bobbrow, line 557: It looks like you could just return here instead of adding a temp variable.

bobbrow (approved):

PR #1995, opened 2021-07-13
No floating promise lint rule, and fixed a timing issue

xisui-MSFT:
No floating promise lint rule, and fixed a timing issue caused by omitted `Thenable`'s in state.ts.

no-floating-promises: forbids usage of Promise-like values in statements without handling their errors appropriately.
await-thenable: disallows awaiting a value that is not a "Thenable".
no-return-await: disallows unnecessary return await.


xisui-MSFT:
Trying to fix test

src/cache-view.ts
```diff
@@ -69,7 +69,7 @@ export class ConfigurationWebview {
     if (this.dirty) {
       telemetry.logEvent("editCMakeCache", {command: "saveCMakeCacheUI"});
       await this.saveCmakeCache(this._options);
-      vscode.window.showInformationMessage(localize('cmake.cache.saved', 'CMake options have been saved.'));
+      void vscode.window.showInformationMessage(localize('cmake.cache.saved', 'CMake options have been saved.'));
```
elahehrashedi, line 72: I don't think void is needed here and similar places in the code, i.e. before `void vscode.window.showInformationMessage`.

.eslintrc.js
```diff
@@ -56,6 +56,9 @@ module.exports = {
         "@typescript-eslint/triple-slash-reference": "error",
         "@typescript-eslint/type-annotation-spacing": "error",
         //"@typescript-eslint/unified-signatures": "error",
+        "@typescript-eslint/no-floating-promises": "error",
```
elahehrashedi, line 59: `no-floating-promises` doesn't seem to be necessary. the code is not handling the values that are returned from `vscode.window.showInformationMessage` or `vscode.window.showErrorMessage`. You can remove this.

.eslintrc.js
```diff
@@ -56,6 +56,9 @@ module.exports = {
...
+        "@typescript-eslint/await-thenable": "error",
+        "no-return-await": "error",
```
elahehrashedi, line 61: this rule already exists in line 90.

.eslintrc.js
xisui-MSFT: The important ones are in state.ts, and those had caused a number of troubles :)

src/cache-view.ts
xisui-MSFT: required by no-floating-promises

.eslintrc.js
xisui-MSFT: Nice catch! I'll remove it.

.eslintrc.js
```diff
@@ -56,6 +56,9 @@ module.exports = {
         "@typescript-eslint/triple-slash-reference": "error",
         "@typescript-eslint/type-annotation-spacing": "error",
         //"@typescript-eslint/unified-signatures": "error",
+        "@typescript-eslint/no-floating-promises": "error",
+        "@typescript-eslint/await-thenable": "error",
```
elahehrashedi, line 60: `"@typescript-eslint/await-thenable": "error"` this rule didn't change anything in the code. Was there a specific reason to add it?

xisui-MSFT: It helps cleaning up the code when doing refactoring: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/await-thenable.md

src/cache-view.ts
elahehrashedi: I see that this rule mostly applies to `vscode.window.showInformationMessage` and `vscode.window.showErrorMessage`. Trying to understand why this rule is needed? as we normally don't handle the value from these two.

.eslintrc.js
elahehrashedi: For future?

src/state.ts
```diff
@@ -28,8 +28,9 @@ export class StateManager {
   get ignoreCMakeListsMissing(): boolean {
     return this._get<boolean>('ignoreCMakeListsMissing') || false;
   }
-  set ignoreCMakeListsMissing(v: boolean) {
-    this._update('ignoreCMakeListsMissing', v);
+
```
elahehrashedi, line 31: Which test this change will fix? I am asking as I am working on some other failing tests.

src/cache-view.ts
xisui-MSFT: The important ones are in state.ts. Those has caused a number of problems, including many flaky test :). Adding 'void' keyword is a very cheap way to avoid these errors in the future.

src/state.ts
xisui-MSFT: Not anymore since Bob's changes worked around (and thus hided) the real issue.
Also this had been causing test failures in our extension (azure sphere) for a rather long time.

xisui-MSFT: A typical issue caused by this is in multi root mode, there's a chance the extension asks for variant before doing configure, even the default is clearly set to 'debug'.

src/cache-view.ts
elahehrashedi: There is an option ignoreVoid you can add to this rule

.eslintrc.js
xisui-MSFT: Right

src/cache-view.ts
xisui-MSFT: I'd like to avoid using that option. That option ignores Promise<void>. And all those update functions in state.ts return Promise<void>.

.eslintrc.js
elahehrashedi: we can keep it, I was wondering if it was fixing anything for tests right now.

src/cache-view.ts
elahehrashedi: It looks like I deleted my comment before knowing you already replied to it. Yes, I realized adding 'ignoreVoid' won't help.

.eslintrc.js
xisui-MSFT: This one doesn't really affect execution. Sorry for the confusion.

src/cache-view.ts
elahehrashedi: The rule originally helped in finding where we need to add await instead of just ignoring the promise, right? IMHO, we can remove the rule now, and remove all the void terms that are added to the code and only keep the awaits that are necessary. The reason is that the code needs some cleaning already, and adding so many voids looks unnecessary in my opinion.

src/state.ts
elahehrashedi: So the failure this will fix is the race problem in multi-root? Would you mind explaining that more? Or if that is easier, you can create an issue for the PR and add some details. whichever is more convenient. meanwhile, I am going through all the waits that are being added and trying to figure out if it affects other scenarios.

src/cache-view.ts
xisui-MSFT: The thing is, it's really hard to find if anyone encounters this issue. For example, it took Bob and me months to figure out these problems in state.ts. Comparing to the debugging effort, it's much easier to write some 'void's.

src/state.ts
xisui-MSFT: It's hard to explain... The behavior is when the variant is "updated" in global state, it's not actually "completely updated". So in the immediate next "get", the value returned is correct, but the "get" function returns undefined later at certain point, although it's never set to undefined. If you would like to know more, ping me on teams and I'll send you some logs.

src/cache-view.ts
elahehrashedi: you are right. I will close this comment as resolved.

src/state.ts
elahehrashedi: thank you for the explanation. it helps me understand what the code is doing.

elahehrashedi: I approved the code. I added @andreeis to do the final approval.

xisui-MSFT: Thanks for the review!

andreeis (approved):

elahehrashedi:
In the latest update, the test folder is being checked for linter errors, I think the build is failing because of those new errors.

xisui-MSFT:
Not sure why those errors are not shown when not running the tests. I'll take a look today.

xisui-MSFT:
I see. Those lint errors are because of recent test change. I'll fix those.

PR #1997, opened 2021-07-15
modify unit test

elahehrashedi:
the main change is in runTests.ts


test/unit-tests/kitmanager.test.ts
```diff
@@ -1,7 +1,8 @@
+/* eslint-disable no-unused-expressions */
 import {readKitsFile, getShellScriptEnvironment} from '@cmt/kit';
 import {expect} from '@test/util';
 import * as path from 'path';
-import * as vscode from 'vscode';
+//import * as vscode from 'vscode';
```
andreeis, line 5: Minor. If we don't need vscode anymore, let's remove instead of leave it commented out.

andreeis:
Besides the minor comment (about a commented include), can you explain what are the changes in the exe, ilk and pdb of TestBuildProcessOtherOutputDir? Are those meant to be source controlled?

andreeis (approved):

Issue #1414, opened 2020-08-19
Compiler flags containing spaces not passed correctly to IntelliSense


patrislav1:
### Brief Issue Summary

Using custom CMake toolchain which is referenced in project-local kit `.vscode\cmake-kits.json`:
```
[
    {
      "name": "arm-gcc from CMake Toolchain",
      "toolchainFile": "${workspaceRoot}/cmake/arm-gcc-toolchain.cmake"
    }
]
```
Intellisense is configured via `.vscode\settings.json` to receive configuration from cmake-tools:
````
{
    "C_Cpp.default.configurationProvider": "ms-vscode.cmake-tools"
}
````

The cmake toolchain succeeds in locating the compiler and cmake-tools receives the compiler information, after I implemented the [cmake cache workaround](https://github.com/microsoft/vscode-cmake-tools/issues/1188#issue-602562754).
The project build works as expected.

However, Intellisense C++ fails to locate the standard C library include files and marks the `#include` statements with squiggles.

The arm-gcc toolchain is on a Windows machine on a `C:\Program Files (x86)\...` path, thus containing a whitespace.
When I move it to `C:\arm-gnu-toolchain`, eliminating the whitespace in the path, and reconfigure the project, then IntelliSense works and the affected `#include` files are found.

I assume the compiler path is not passed correctly from cmake-tools to IntelliSense, maybe there are quotes missing somewhere along the way.

### Platform and Versions

- **Operating System**:  Windows
- **CMake Version**: 3.17.4
- **VSCode Version**: 1.48.0
- **CMake Tools Extension Version**: 1.4.1
- **Compiler/Toolchain**: arm-none-eabi-gcc 6.3.1.508

**Comment by bobbrow** [comment]
Can you [enable Debug logging](https://code.visualstudio.com/docs/cpp/enable-logging-cpp#_enable-logging-for-the-language-server) for cpptools which will print out the configurations sent CMake Tools?

After enabling the logging, please run the "Reload Window" command to reset the extension which will cause the configuration to be resent.  If you can send that log and also the log printed when you run the "Log Diagnostics" command, that would be super helpful.  Thanks!

**Comment by patrislav1** [comment]
Surprise: the compiler path seems to be passed correctly to cpptools, AFAICS.
cpptools even lists the compiler include paths to be indexed ...
Maybe it is not a cmake-tools bug, but a cpptools one?
```
Custom browse configuration received: {
  "browsePath": [
    <bunch of source paths>
[...]

PR #2003, opened 2021-07-20
double quote sysroot flag if necessary [don't merge yet]

elahehrashedi:
bug fix :https://github.com/microsoft/vscode-cmake-tools/issues/1414
The `compiler flags` , `include paths` , `defines` are passed correctly to the intellisense. 
> for example an include path that has space in the path, 
> or a define string that includes space or double quote. 

However, the sysroot flag was treated differently in the code. This PR should fix that.


andreeis (approved):

src/compdb.ts
```diff
@@ -29,8 +26,7 @@ export class CompilationDatabase {
           directory: cur.directory,
           file: cur.file,
           output: cur.output,
-          command: cur.command,
-          arguments: cur.arguments ? cur.arguments : [...shlex.split(cur.command)]
```
bobbrow, line 33: I don't think `arguments` should be removed here

PR #2010, opened 2021-07-23
Trigger a reconfiguration, on all cmake files change

AbdullahAmrSobh:
## The purpose of this change
When any CMakeLists.txt or **.cmake file is changed in the project, it triggers a reconfiguration


elahehrashedi:
this will trigger configure even when the toolchain files are modified.

src/cmake-tools.ts
```diff
@@ -761,7 +761,7 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
       }
 
       const sourceDirectory = (await this.sourceDir).toLowerCase();
-      if (str === path.join(sourceDirectory, "cmakelists.txt")) {
+      if ((str.includes("cmakelists.txt") || str.includes(".cmake"))) {
```
bobbrow, line 764: instead of `includes`, I think `endsWith` would be better.

elahehrashedi: I made a small change, we want only the cmakelists that are in the workspace folder or in the sourcedirectory.

andreeis (approved):

PR #2012, opened 2021-07-23
Write lint message to stderr

xisui-MSFT:
This resolves @bobbrow 's concern in #1967. Turns out gulp-eslint has a way to write to stderr.

First parameter is formatter. Passing undefined makes it to use the default.


bobbrow:
Do we need this PR if we take #2013 instead?

xisui-MSFT:
> Do we need this PR if we take #2013 instead?

No, just I'm not sure if you want that one or not.

bobbrow:
Can you show me screenshots of both approaches?

xisui-MSFT:
See the other PR for screenshots 😊

bobbrow:
Closing in favor of #2013

PR #2013, opened 2021-07-23
Make lint output easier to use in VS Code

xisui-MSFT:
By default the lint output looks like:
```
file.ts
line:column message
```

It's annoying to use this output in VS Code because ctrl + left click only works for `file.ts`. This is not a big issue since this could be easily worked around by open the corresponding file and then switching to `problems` tab (target file must be opened). But I feel it easier to just ctrl + left click file name + line number.

Maybe this is just my habit, so feel free to decline this PR. Also, there are alternatives out there, such as [eslint-path-formatter](https://github.com/Bartvds/eslint-path-formatter) and [jshint-path-reporter](https://github.com/Bartvds/jshint-path-reporter).


bobbrow:
Do you have a screenshot of what it looks like?

xisui-MSFT:
Before:
image


After:
image

bobbrow:
Is there an easy way to keep the red error message?  Or maybe make the errors themselves red?

image

xisui-MSFT:
> Is there an easy way to keep the red error message? Or maybe make the errors themselves red?
> 
> image

The message itself isn't hard to add. But need to investigate how to make characters red. I guess that shouldn't be too hard though.

xisui-MSFT:
Need to do color wrap like this: https://github.com/Bartvds/eslint-path-formatter/blob/5122538587e215af327d5caa91c5676105266ba0/index.js#L7-L19. I'll add them.

xisui-MSFT:
Now it looks like this:
image

bobbrow:
Awesome!

bobbrow (approved):

PR #2017, opened 2021-07-26
Correcting translations

bobbrow:
Also removing the source of confusion for future translations


Colengms:
Just a thought: If you have a substring that doesn't need to be localized, you can break it out of the loc string and use a placeholder (`{0}`, `{1}`) in it's place.  Translators should be familiar with that style of placeholder.

bobbrow:
I think that's what I did in this PR (in addition to updating the existing translations). You have to scroll down to the last file. 😄

Colengms (approved):

i18n/plk/src/preset.i18n.json
```diff
@@ -9,8 +9,8 @@
 	"default.test.preset": "[Domyślna]",
 	"default.test.preset.description": "Pusty predefiniowany test, które nie dodaje żadnych argumentów",
 	"no.cl.arch": "Predefiniowana konfiguracja {0}: nie określono architektury dla elementu cl.exe, używając domyślnej opcji x86",
-	"no.cl.toolset.arch": "Skonfiguruj ustawienie wstępne {0}: nie określono zestawu narzędzi dla elementu cl.exe, używając domyślnej opcji x86",
-	"invalid.cl.toolset.arch": "Skonfiguruj ustawienie wstępne {0}: Określono nieoczekiwaną architekturę zestawu narzędzi \"{1}\" czy chodziło Ci o \"host ={1}\"?",
+	"no.cl.toolset.arch": "Skonfiguruj ustawienie wstępne {0}: nie określono zestawu narzędzi dla elementu cl.exe, używając domyślnej opcji host=x86",
```
lygstate, line 12: weird, this is not updated

bobbrow: I added "host=" to that line, but it's not highlighted for some reason.

Issue #2009, opened 2021-07-23
CMakePresets: "warnings" field settings are not passed on to CMake


hwmaier:
When creating the command line to invoke CMake the `warnings` field is ignored and it is impossible to pass warning command line arguments to CMake. For example in this case `--no-warn-unused-cli` shall be passed on to CMake.

Example:
```
    "configurePresets": [
        {
           ...
            "warnings": {"unusedCli": false},
           ...
        },
```

As a side note, this issue would not occur if this plugin would use the `--preset` argument to invoke CMake rather than building its own command line.

bobbrow:
Thank you for reporting this issue.  Looks like a simple logic mistake in the code.  @xisui-MSFT could you fix this real quick?
https://github.com/microsoft/vscode-cmake-tools/blob/192fec8f871563e21952924b187f0fb22f88c4ba/src/preset.ts#L1247

<br/>

> As a side note, this issue would not occur if this plugin would use the `--preset` argument to invoke CMake rather than building its own command line.

Yes, this is likely possible, but Kitware has requested that IDE's/editors do their own parsing instead of using `--preset` ([reference](https://cmake.org/cmake/help/latest/guide/ide-integration/index.html#id4))

hwmaier:
> > As a side note, this issue would not occur if this plugin would use the `--preset` argument to invoke CMake rather than building its own command line.
> 
> Yes, this is likely possible, but Kitware has requested that IDE's/editors do their own parsing instead of using `--preset` ([reference](https://cmake.org/cmake/help/latest/guide/ide-integration/index.html#id4))

Also Thanks for this explanation. I have been wondering what the rationale was to reconstruct rather using the `--preset` argument. I suggest to put that explanation into the doc as I am sure it comes up again.

bobbrow:
I'm not sure which doc would be an appropriate place for it.  People would ask anyway. 😅

We committed a fix for this.  If you want to try it out, you can download the VSIX here: https://github.com/microsoft/vscode-cmake-tools/suites/3342925255/artifacts/78474075
and then run the "Install from VSIX..." command in VS Code.

PR #2019, opened 2021-07-26
Wrong logic for --no-warn-unused-cli

xisui-MSFT:
## This change addresses item #2009

Also fixed issues in other similar fields.

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -1244,7 +1244,7 @@ export function configureArgs(preset: ConfigurePreset): string[] {
     }
 
     preset.warnings.uninitialized && result.push('--warn-uninitialized');
-    preset.warnings.unusedCli && result.push('--no-warn-unused-cli');
+    !preset.warnings.unusedCli && result.push('--no-warn-unused-cli');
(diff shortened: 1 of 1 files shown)


src/preset.ts
```diff
@@ -1244,7 +1244,7 @@ export function configureArgs(preset: ConfigurePreset): string[] {
     }
 
     preset.warnings.uninitialized && result.push('--warn-uninitialized');
-    preset.warnings.unusedCli && result.push('--no-warn-unused-cli');
+    !preset.warnings.unusedCli && result.push('--no-warn-unused-cli');
```
bobbrow, line 1247: What is the default if it's undefined?  Should we just check if it's `=== false`?

bobbrow: The [manual](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) seems to imply we only add this switch if it's equal to false

xisui-MSFT: That's a good point. Will also fix other cases.

bobbrow (approved):

PR #2021, opened 2021-07-27
Patch Windows SDK bin path in a more reliable way.

lygstate:
Improve of https://github.com/microsoft/vscode-cmake-tools/pull/1955


src/kit.ts
```diff
@@ -734,8 +734,10 @@ async function collectDevBatVars(hostArch: string, devbat: string, args: string[
     const existPath = vars.get('PATH') ?? '';
     const oldWinSdkBinPath = path.join(WindowsSdkDir, 'bin', hostArch);
     const newWinSdkBinPath = path.join(WindowsSdkDir, 'bin', WindowsSDKVersion, hostArch);
-    if (existPath.toLowerCase().indexOf(oldWinSdkBinPath.toLowerCase()) >= 0
-      && existPath.toLowerCase().indexOf(newWinSdkBinPath.toLowerCase()) < 0) {
+    const newWinSdkBinPathExist = await fs.exists(newWinSdkBinPath);
+    if (newWinSdkBinPathExist &&
+        existPath !== '' &&
```
bobbrow, line 739: I wouldn't think this check is needed.  It should fail the next check without causing a crash.

lygstate: It's needed, as https://github.com/microsoft/vscode-cmake-tools/pull/1763 are didn't merged yet
On Windows 10, the ‘PATH' turn out to be 'Path', so existPath can be empty, when existPath is '', we can not override with ’PATH'

bobbrow: Ah. You're right. I misread the intent.

lygstate:
@bobbrow Hi, this is a relative small fix, can this be merged in 1.8.x?

bobbrow:
Sorry we didn't take a look at this yet.  We generally reserve the '.x' patches to fix regressions from the major release and I've been shorthanded for the past few weeks so we think to pull in any of the 1.9.0 stuff.  One consolation is that we plan to have a shorter 1.9 cycle (~1 month) compared to our normal release cycle (~3 months) because of the CMakePresets v3 addition.

andreeis (approved):

PR #2025, opened 2021-07-30
Enforce consistent access to the environment

bobbrow:
Draft

### DIFF
--- a/src/drivers/driver.ts
+++ b/src/drivers/driver.ts
@@ -175,14 +175,13 @@ export abstract class CMakeDriver implements vscode.Disposable {
   /**
    * Compute the environment variables that apply with substitutions by expansionOptions
    */
-  async computeExpandedEnvironment(in_env: proc.EnvironmentVariables, expanded_env: proc.EnvironmentVariables): Promise<proc.EnvironmentVariables> {
-    const env = {} as {[key: string]: string};
+  async computeExpandedEnvironment(toExpand: util.Dictionary<string>, expanded: proc.EnvironmentVariables): Promise<util.Dictionary<string>> {
+    const env: util.Dictionary<string> = {};
     const opts = this.expansionOptions;
 
-    await Promise.resolve(
-      util.objectPairs(in_env)
-      .forEach(async ([key, value]) => env[key] = await expand.expandString(value, {...opts, envOverride: expanded_env}))
-    );
+    for (const entry of Object.entries(toExpand)) {
+      toExpand[entry[0]] = await expand.expandString(entry[1], {...opts, envOverride: expanded});
+    }
     return env;
   }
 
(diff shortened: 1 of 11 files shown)


src/drivers/driver.ts
```diff
@@ -175,14 +175,13 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
    */
-  async computeExpandedEnvironment(in_env: proc.EnvironmentVariables, expanded_env: proc.EnvironmentVariables): Promise<proc.EnvironmentVariables> {
-    const env = {} as {[key: string]: string};
+  async computeExpandedEnvironment(toExpand: util.Dictionary<string>, expanded: proc.EnvironmentVariables): Promise<util.Dictionary<string>> {
+    const env: util.Dictionary<string> = {};
     const opts = this.expansionOptions;
 
-    await Promise.resolve(
-      util.objectPairs(in_env)
-      .forEach(async ([key, value]) => env[key] = await expand.expandString(value, {...opts, envOverride: expanded_env}))
```
bobbrow, line 184: Stumbled on this... I think the way this was written was incorrect and won't be properly awaited.

src/proc.ts
```diff
@@ -71,10 +72,70 @@ export interface Subprocess {
 export interface BuildCommand {
   command: string;
   args?: string[];
-  build_env?: {[key: string]: string};
+  build_env?: EnvironmentVariables;
 }
 
-export interface EnvironmentVariables { [key: string]: string }
+export class EnvironmentVariables {
```
bobbrow, line 78: Issue:  Need a way to keep "null"s because of the way presets use them to erase variables.  The way things are merged now, "null"s will be deleted immediately, but they need to stay around until we access the `.fullEnvironment` at which point they should be removed.  Could add a member variable to track whether the environment wants this behavior since most don't and it's a perf hit (though maybe insignificant) to have to scrub the environment before converting back to Dictionary<string>.

Issue: This class should be in its own module.

src/util.ts
```diff
@@ -16,6 +16,10 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
  */
 export function escapeStringForRegex(str: string): string { return str.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, '\\$1'); }
 
+export interface Dictionary<T> {
```
bobbrow, line 19: This should go in the environment module

src/proc.ts
```diff
@@ -71,10 +72,70 @@ export interface Subprocess {
...
+    return Object.entries(this.env).map(
+      ([key, value]) => ({
+        name: key,
+        value
+      })
+    );
+  }
+
+  public get(key: string): string | undefined {
+    return this.env[util.normalizeEnvironmentVarname(key)];
```
lygstate, line 103: That's what I am trying avoid, do not normalizeEnvironmentVarname,

src/proc.ts
```diff
@@ -71,10 +72,70 @@ export interface Subprocess {
...
+  }
+
+  public set(key: string, value?: string|null): void {
+    if (value !== undefined && value !== null) {
+      this.env[util.normalizeEnvironmentVarname(key)] = value;
```
lygstate, line 108: As on Windows, we need preserve the original case when genearting 
the final environmentMap to use

bobbrow: On Windows it shouldn't matter what the casing is, should it?

But if the goal is to preserve the casing, an additional object should be added that maps the normalized keys to the actual keys in the `this.env` object.  This would avoid us having to loop through all keys every time you look up a value.  That was my biggest concern with the original PR.

bobbrow: e.g. add a `this.keys`

bobbrow: I pushed a change that should handle the casing (though none of this code is tested yet as it is incomplete)

src/proc.ts
lygstate: How about implement EnvironmentVariables like NodeJS.ProcessEnv,
Getting EnvironmentVariables  as a alias of NodeJS.ProcessEnv,
as NodeJS.ProcessEnv handling windows case-insensitve properly

PR #2029, opened 2021-08-04
compdb: opt to use any compile_commands.json in build dir

Felix-El:

Add an option to collect all `compile_commands.json` from the build directory and merge them into one for consumption by `vscode-cpptools`.

## The purpose of this change

The option "Collect Compile Commands" will merge all `compile_commands.json` from the build directory into a new JSON file. This functionality is an enabler for IntelliSense in ExternalProject based "superbuilds".

## Other Notes/Information

In order to get IntelliSense support with superbuilds, few settings are required:

```json
{
    "cmake.collectCompileCommands": true,
    "cmake.copyCompileCommands": "${workspaceFolder}/build/combined_compile_commands.json",
    "C_Cpp.default.compileCommands": "${workspaceFolder}/build/combined_compile_commands.json",
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1440,6 +1440,12 @@
           "description": "%cmake-tools.configuration.cmake.emscriptenSearchDirs.description%",
           "scope": "window"
         },
+        "cmake.collectCompileCommands": {
+          "type": "boolean",
+          "default": false,
+          "description": "%cmake-tools.configuration.cmake.collectCompileCommands.description%",
+          "scope": "resource"
+        },
         "cmake.copyCompileCommands": {
           "type": "string",
           "default": null,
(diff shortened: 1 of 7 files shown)


andreeis:
@Felix-El, can you describe how you can have more than one compile_commands.json in the same project? I am not familiar with this scenario and I need a repro to test your PR, before we accept it.

Felix-El:
> @Felix-El, can you describe how you can have more than one compile_commands.json in the same project? I am not familiar with this scenario and I need a repro to test your PR, before we accept it.

Sure, this feature can help a lot if you are, like I am, working with CMake subprojects, i.e. using ExternalProject.
Each subproject typically has its own private build, say build/subproj1, build/subproject2, etc.
Now if I let CMake generate the compile_commands.json I receive one of those files in the build folder of each subproject
(with nested subprojects there can be again multiple).
As such projects evolve it is impractical to carry along a list of project names or compile_commands.json paths, hence I decided to find them by crawling the build directory.

Here is a multi-subproject test I've added with a CMake new featrue: https://gitlab.kitware.com/cmake/cmake/-/tree/master/Tests/InstallMode
It should serve well to verify this PR.

andreeis:
@Felix-El, we have a suggestion if you can update your PR. Let's have instead one path setting for the merged compile commands ("cmake.mergedCompileCommands") and no boolean because we can deduce everything based on the path being set or not.
What do you think of this approach?

bobbrow:
Yes. We wanted to avoid requiring setting both the `merge` and `copy` settings to get predictable behavior.

PR #2032, opened 2021-08-04
Partial activation A/B experiment

andreeis:


src/cmake-tools.ts
```diff
@@ -481,18 +484,31 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+
+        telemetryProperties["missingCMakeListsPopupType"] = showToast ? "toastCreateLocateIgnore" : "selectFromAllCMakeLists";
+
+        const result = showToast ? await vscode.window.showErrorMessage(
             localize('missing.cmakelists', 'CMakeLists.txt was not found in the root of the folder \'{0}\'. How would you like to proceed?', this.folderName),
-            quickStart, changeSetting, ignoreCMakeListsMissing);
+            quickStart, changeSetting, ignoreCMakeListsMissing) : changeSetting;
+
+        if (showToast) {
+          telemetryProperties["missingCMakeListsUserAction"] = result || "dismissed";
```
bobbrow, line 503: I like this, but we'll end up with localized properties which will be hard to aggregate. Let's get non-localized strings here.

src/extension.ts
```diff
@@ -187,7 +194,8 @@ class ExtensionManager implements vscode.Disposable {
       }
     }
     const telemetryProperties: telemetry.Properties = {
-      isMultiRoot: `${isMultiRoot}`
+      isMultiRoot: `${isMultiRoot}`,
+      isFullyActivated: `await ${this.workspaceHasCMakeProject()}`
```
bobbrow, line 198: I think the `await` has to be in the ${}, but it might be safer just to add a temp variable for this.

src/extension.ts
```diff
@@ -221,6 +229,11 @@ class ExtensionManager implements vscode.Disposable {
     return inst;
   }
 
+  private showToast: Promise<boolean | undefined> | undefined;
+  public async expShowToast(): Promise<boolean | undefined> {
```
bobbrow, line 233: Because we don't wait for the exp service here, this might return undefined which is not awaitable.  I think we should add a variable to hold the exp service promise and wait for that when this.showToast is undefined.

bobbrow: Actually, it looks like the variable exists in the telemetry module and we just have to call the method to get the exp service to access it.

src/telemetry.ts
```diff
@@ -16,41 +18,100 @@ interface IPackageInfo {
...
+      }
+
+      if (experimentationTelemetry) {
+          experimentationTelemetry.dispose();
+      }
+  }
+}
+
+export async function logEvent(eventName: string, properties?: Properties, measures?: Measures): Promise<void> {
+  try {
```
bobbrow, line 98: Might want to add `if (initializationPromise)` here to match `deactivate()`

bobbrow: And let Colen know.

src/telemetry.ts
```diff
@@ -16,41 +18,100 @@ interface IPackageInfo {
...
+  } catch (e) {
+    // Handle error with a try/catch, but do nothing for errors.
+  }
 }
 
-export function logEvent(eventName: string, properties?: Properties, measures?: Measures): void {
-    if (telemetryReporter) {
-        telemetryReporter.sendTelemetryEvent(eventName, properties, measures);
-    }
+export async function getExperimentationService(): Promise<IExperimentationService | undefined> {
```
bobbrow, line 79: This is potentially undefined

andreeis: I'm not sure what else to do here, were you recommending I change something?

bobbrow: Update the annotation to `Promise<IExperimentationService | undefined> | undefined` or call `getExperimentationServiceAsync` and assign `initializationPromise` here before returning it back.

bobbrow: If you assign it here, you'll need to update `activate()`

src/cache-view.ts
```diff
@@ -69,7 +69,7 @@ export class ConfigurationWebview {
   // Save from the UI table into the CMake cache file if there are any unsaved edits.
   async persistCacheEntries() {
     if (this.dirty) {
-      telemetry.logEvent("editCMakeCache", {command: "saveCMakeCacheUI"});
+      await telemetry.logEvent("editCMakeCache", {command: "saveCMakeCacheUI"});
```
bobbrow, line 72: Telemetry should be fire and forget.  I'm fixing `logEvent` so that it is not returning a Promise.

bobbrow:
>           const existingCmakeListsFiles: string[] | undefined = await util.getAllCMakeListsPaths(this.folder.uri);

It occurs to me that if there is no CMakeLists.txt in the folder, we would essentially just be prompting folks to search for one with the file dialog and skip the opportunity for them to do "Quick Start".  We might need to search for the CMakeLists.txt files before we decide to skip the first toast message.

---
Refers to: src/cmake-tools.ts:529 in df42259. [](commit_id = df42259dae62cd8bfcfb40057239703c7124f18e, deletion_comment = False)

src/cmake-tools.ts
```diff
@@ -572,6 +600,10 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
     // This CMT folder can go through various changes while executing this function
```
bobbrow, line 600: This comment block applies to line 607 and should be moved down

src/extension.ts
```diff
@@ -72,7 +73,22 @@ type CMakeToolsQueryMapFn = (cmt: CMakeTools) => Thenable<string | string[] | nu
...
+        void experimentationService
+              .then(expSrv => expSrv!.getTreatmentVariableAsync<boolean>("vscode", "partialActivation_showCMakeLists"))
+              .then(showCMakeLists => {
+                if (showCMakeLists !== undefined) {
+                  resolve(showCMakeLists);
+                } else {
+                  resolve(false);
+                }
+              });
+      }
```
bobbrow, line 89: something's wrong with our linter settings. It didn't catch the missing resolve for the "else" case. :(

src/telemetry.ts
bobbrow: I ported the changes over and started a PR

src/cmake-tools.ts
```diff
@@ -504,9 +504,13 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
         const changeSourceDirectory = localize('edit.setting', "Locate");
         const ignoreCMakeListsMissing = localize('ignore.activation', "Don't show again");
-        const showCMakeLists: boolean = await expShowCMakeLists();
 
-        telemetryProperties["missingCMakeListsPopupType"] = showCMakeLists ? "selectFromAllCMakeLists" : "toastCreateLocateIgnore";
+        const showCMakeLists: boolean = await expShowCMakeLists();
+        const existingCmakeListsFiles: string[] | undefined = await util.getAllCMakeListsPaths(this.folder.uri);
+        telemetryProperties["missingCMakeListsPopupType"] = (showCMakeLists && existingCmakeListsFiles && existingCmakeListsFiles.length > 0) ? "selectFromAllCMakeLists" : "toastCreateLocateIgnore";
+        if (showCMakeLists) {
+          telemetryProperties["ignoreExperiment"] = (!existingCmakeListsFiles || existingCmakeListsFiles.length === 0).toString();
```
bobbrow, line 512: `showCMakeLists` should also be updated based on the value of `existingCmakeListsFiles.length`

andreeis: good catch, thx. I just pushed an update.

src/cmake-tools.ts
```diff
@@ -505,13 +505,19 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         const changeSourceDirectory = localize('edit.setting', "Locate");
         const ignoreCMakeListsMissing = localize('ignore.activation', "Don't show again");
 
-        const showCMakeLists: boolean = await expShowCMakeLists();
+        let showCMakeLists: boolean = await expShowCMakeLists();
         const existingCmakeListsFiles: string[] | undefined = await util.getAllCMakeListsPaths(this.folder.uri);
-        telemetryProperties["missingCMakeListsPopupType"] = (showCMakeLists && existingCmakeListsFiles && existingCmakeListsFiles.length > 0) ? "selectFromAllCMakeLists" : "toastCreateLocateIgnore";
         if (showCMakeLists) {
-          telemetryProperties["ignoreExperiment"] = (!existingCmakeListsFiles || existingCmakeListsFiles.length === 0).toString();
+          if (existingCmakeListsFiles !== undefined && existingCmakeListsFiles.length > 0) {
```
bobbrow, line 511: Consider:
```ts
showCMakeLists = existingCmakeListsFiles !== undefined && existingCmakeListsFiles.length > 0;
telemetryProperties["ignoreExperiment"] = (!showCMakeLists).toString();
```

bobbrow (approved):

PR #2047, opened 2021-08-11
change resource to machine-overridable

elahehrashedi:
resolve issues on WSL

### DIFF
--- a/package.json
+++ b/package.json
@@ -1014,13 +1014,13 @@
           "type": "string",
           "default": "cmake",
           "description": "%cmake-tools.configuration.cmake.cmakePath.description%",
-          "scope": "resource"
+          "scope": "machine-overridable"
         },
         "cmake.buildDirectory": {
           "type": "string",
           "default": "${workspaceFolder}/build",
           "description": "%cmake-tools.configuration.cmake.buildDirectory.description%",
-          "scope": "resource"
+          "scope": "machine-overridable"
         },
         "cmake.installPrefix": {
           "type": "string",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1014,13 +1014,13 @@
...
+          "scope": "machine-overridable"
         },
         "cmake.buildDirectory": {
           "type": "string",
           "default": "${workspaceFolder}/build",
           "description": "%cmake-tools.configuration.cmake.buildDirectory.description%",
-          "scope": "resource"
+          "scope": "machine-overridable"
         },
         "cmake.installPrefix": {
```
bobbrow, line 1025: `install.Prefix` should also be machine-overridable

bobbrow (commented):
I think there's a number of settings that could be argued that they would need machine-overridable (e.g. environments, debugConfig, etc), but I'm not really seeing demand from our users for this, so maybe we should just leave it alone until someone asks for it.  The downside of doing this is that settings for remote SSH won't be able to grab from the USER settings anymore, so I'd rather not break that scenario if there's no reported problems with the way things are now.

sean-mcmanus:
I think all the settings should be reviewed to match the scopes descriptions: https://code.visualstudio.com/api/references/contribution-points , i.e. to match what vscode-cpptools does and what other extensions do in the remote scenarios. The scopes for cmake-tools were probably decided on before they added the new ones, i.e. https://github.com/microsoft/vscode-cpptools/issues/3774

andreeis (approved):

andreeis:
Closing this PR since such change requires users to reset the settings for all remote workspace. We may revisit when we see more upvotes from the other side and address the scenario impacting the most.

Issue #2026, opened 2021-07-31
Do not crash with cmake 3.7-3.13 with CMakePresets.json present


linquize:
### Brief Issue Summary

Do not crash with cmake 3.7-3.13 with `CMakePresets.json` present.
If I start VScode directly (cmake auto-detected version is 3.12), or start VSCode in VS2017 command prompt (cmake version is 3.12), this problem exists.
If I start VSCode in Ubuntu 18.04 (cmake from ubuntu bionic is 3.10), this problem exists.

If I start VSCode in VS2019 command prompt (cmake version is 3.20), there is no problem.
If I start VSCode in Ubuntu 20.04 (cmake from focal is 3.16), there is no problem.

Open a workspace with `CMakePresets.json` present.

CMakePresets.json

```json
{
  "version": 2,
  "cmakeMinimumRequired": { "major": 3, "minor": 20, "patch": 0 },
  "configurePresets": [
    {
      "name": "default",
      "generator": "Ninja",
      "binaryDir": "${sourceDir}/build/default"
    }
  ],
  "buildPresets": [ { "name": "default", "configurePreset": "default" }
  ]
}
```

### Expected:

**No Configure Preset Selected** is available for choosing configure preset.

### Apparent Behavior:

Configure Presets list is not available.
If **No Configure Preset Selected** clicked, a toast shows **command 'cmake.selectConfigurePreset' not found**

### CMake Tools Log

```
[variant] Loaded new set of variants
[kit] Successfully loaded 16 kits from C:\Users\User\AppData\Local\CMakeTools\cmake-tools-kits.json
[presetController] Successfully validated presets in c:\project\cmaketest\CMakePresets.json
[rollbar] CMake version 3.12.18081601 may not be supported. Minimum version required is 3.20.0. {}
[main] Unable to determine what CMake generator to use. Please install or configure a preferred generator, or update settings.json, your Kit configuration or PATH variable. Error: No usable generator found.
[expand] Invalid variable reference ${sourceDir} in string: ${sourceDir}/build/default
```

### Platform and Versions

- **Operating System**: Windows
- **CMake Version**: 3.12 (auto detected from VS2017)
- **VSCode Version**: 1.58
- **CMake Tools Extension Version**: 1.7.3
- **Compiler/Toolchain**: Both Visual C++ 2017 and 2019 installed

### Platform and Versions

- **Operating System**: Linux
- **CMake Version**: 3.10 (from Ubuntu bionic)
[...]

PR #2050, opened 2021-08-12
Fix presets using server API

xisui-MSFT:
## This change addresses item #2026

Source dir is required to evaluate presets. Presets contain generator info. Generator info is required for server api. Source dir currently requires driver. So this becomes a circular dependency. And I changed cmake-tools.sourceDir to use config.sourceDir instead.

Since generator could be changed, including changing to null, when switching between using kits and presets, and between different presets, added a function to shut down the driver, instead of restarting.

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -170,6 +170,8 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
    * Presets are loaded by PresetsController, so this function should only be called by PresetsController.
    */
   async setConfigurePreset(configurePreset: string | null) {
+    const previousGenerator = this.configurePreset?.generator;
+
     if (configurePreset) {
       log.debug(localize('resolving.config.preset', 'Resolving the selected configure preset'));
       const expandedConfigurePreset = await preset.expandConfigurePreset(this.folder.uri.fsPath,
@@ -178,6 +180,10 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
                                                                          await this.sourceDir,
                                                                          true);
       this._configurePreset.set(expandedConfigurePreset);
+      if (previousGenerator && previousGenerator !== expandedConfigurePreset?.generator) {
+        await this._shutDownCMakeDriver();
+      }
+
       if (!expandedConfigurePreset) {
         log.error(localize('failed.resolve.config.preset', 'Failed to resolve configure preset: {0}', configurePreset));
         await this.resetPresets();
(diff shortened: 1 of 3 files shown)


xisui-MSFT:
For some reason the server API mode doesn't work on my machine at all. @linquize @andreeis @elahehrashedi @bobbrow Can you please help testing this?

linquize:
cmake 3.20 dropped server API. 
You have to use cmake 3.7 - 3.19 to test this pull request.

xisui-MSFT:
@linquize I tried 3.13 as well, and it didn't work for me. It constantly tried to use VS 2019 as generator (which it doesn't support), even if the generator passed to it was VS 2017. Any idea why?

linquize:
What is the generator in your CMakePresets.json?
Is it Ninja or a specific VS version?

linquize:
Due to cmake bug, use cmake 3.13.3
https://cmake.org/cmake/help/latest/release/3.13.html#id19

linquize:
Use cmake 3.12 from vs 2017 and set generator to vs 2017 works

src/drivers/driver.ts
```diff
@@ -318,6 +318,23 @@ export abstract class CMakeDriver implements vscode.Disposable {
     return {vars, variantVars};
   }
 
+  static sourceDirDxpansionOptions(workspaceFolderFspath: string | null): expand.ExpansionOptions {
```
bobbrow, line 321: Looks like a typo here.

src/drivers/driver.ts
```diff
@@ -1619,13 +1637,13 @@ export abstract class CMakeDriver implements vscode.Disposable {
     // Load up kit or presets before starting any drivers.
     if (useCMakePresets) {
       if (configurePreset) {
-        await this.setConfigurePreset(configurePreset);
+        await this._setConfigurePreset(configurePreset);
```
bobbrow, line 1640: >_setConfigurePreset [](http://example.com/codeflow?start=19&length=19)

It seems we have two functions with basically the same name. I don't think an underscore helps differentiate. It seems there are more functions following this pattern and I don't want us to fall into this trap again.  Can we document the underscore ones, or at least open an issue to fix/rename these?

src/cmake-tools.ts
```diff
@@ -392,6 +398,14 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
     return this._reloadCMakeDriver();
   });
 
+  private readonly _sourceDirSub = this.workspaceContext.config.onChange('sourceDirectory', async () =>
+    this._sourceDir = Promise.resolve(
```
bobbrow, line 402: >_sourceDir [](http://example.com/codeflow?start=9&length=10)

We talked about not requiring this to be a Promise.

src/drivers/driver.ts
xisui-MSFT: #2069

PR #2051, opened 2021-08-12
Environment fixes for debug but mostly for launch.

andreeis:
The debug scenario was having a non critical bug (wrong preset environment checked before merging the configure environment into the VSCode debugger).

The launch scenario was broken when presets are on, because the assignment into user_config.environment would throw an exception about assigning into read-only property. Besides that, the new calculated environment was not sent to the VSCode terminal. Also, the VSCode terminal sees the environment as a map (same as presets and processes) and not as an array of "name"/"value" that we need to send to the VSCode debugger. So this is why we have some differences in how the environment is processed and passed further in the debug versus the launch scenario.

I plan to also add a test to verify the final environment before debug/launch but I didn't get to this yet. Sending out the fix PR first.

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -30,7 +30,7 @@ import {CMakeGenerator, Kit} from './kit';
 import {LegacyCMakeDriver} from '@cmt/drivers/legacy-driver';
 import * as logging from './logging';
 import {fs} from './pr';
-import {buildCmdStr} from './proc';
+import {buildCmdStr, EnvironmentVariables} from './proc';
 import {Property} from './prop';
 import rollbar from './rollbar';
 import * as telemetry from './telemetry';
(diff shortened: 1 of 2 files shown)


andreeis:
I also don't know why the CI is failing. I'll look into that later today. I haven't seen those errors before and I also didn't modify anything lint or gulp related.

src/cmake-tools.ts
```diff
@@ -1856,9 +1856,9 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
     const user_config = this.workspaceContext.config.debugConfig;
     Object.assign(debug_config, user_config);
     // Add environment variables from buildPreset.
-    if (this.buildPreset?.environment) {
-      const build_preset_environment = await drv.getConfigureEnvironment();
-      debug_config.environment = debug_config.environment ? debug_config.environment.concat(util.splitEnvironmentVars(build_preset_environment)) : {};
+    if (this.configurePreset?.environment) {
```
elahehrashedi, line 1859: So we decided to add the presets from the Config?

elahehrashedi (approved):

bobbrow (approved):

PR #2056, opened 2021-08-17
wrong project platform is being used

elahehrashedi:
An extra env variable `Platform` is being sent by `collectDevBatVars` function and it is causing the "test" project build to fail.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -603,7 +603,6 @@ const MSVC_ENVIRONMENT_VARIABLES = [
   'LIBPATH',
   'NETFXSDKDir',
   'Path',
-  //'Platform',
   'UCRTVersion',
   'UniversalCRTSdkDir',
   'user_inputversion',
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -603,7 +603,7 @@ const MSVC_ENVIRONMENT_VARIABLES = [
   'LIBPATH',
   'NETFXSDKDir',
   'Path',
-  'Platform',
+  //'Platform',
   'UCRTVersion',
   'UniversalCRTSdkDir',
   'user_inputversion',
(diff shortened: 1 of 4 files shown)


src/kit.ts
```diff
@@ -603,7 +603,7 @@ const MSVC_ENVIRONMENT_VARIABLES = [
   'LIBPATH',
   'NETFXSDKDir',
   'Path',
-  'Platform',
+  //'Platform',
```
bobbrow, line 606: I don't actually see Platform ever set in any of the Developer Command prompts I've opened (VS 2017-VS 2022).  We should just delete this line.

cc: @lygstate

lygstate: Platform  is set by VS2019, maybe we need a better way to detecting if Platform are set?

lygstate:
```
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.10.0
** Copyright (c) 2021 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x86_x64'

C:\Program Files (x86)\Microsoft Visual Studio\2019\Community>echo %Platform%
x64

C:\Program Files (x86)\Microsoft Visual Studio\2019\Community>
```

elahehrashedi:
test if we need to set platform when host and target kits are different? 
test building cpptools using just 64 without setting platform (setting host and target the same)

lygstate:
I have no object to remove `Platform`, I am just telling where the `Platform` comes from

bobbrow:
Thanks @lygstate, we just wanted to make sure removing it does the right thing for us, so these are a few notes @elahehrashedi wrote while we were discussing it this morning.  We were also hoping it doesn't break your fix for msvc2015 detection.

Issue #2053, opened 2021-08-13
`cmake -H` to generate build files?


wermos:
### Brief Issue Summary
The extension uses `cmake -H/path/to/source` to generate the build files. However, this flag is not documented in the [CMake Docs](https://cmake.org/cmake/help/v3.21/) as far as I know.

### Expected:
The extension should use `cmake -S/path/to/source/` to generate the build files. The `-H` flag is documented [here](https://cmake.org/cmake/help/latest/manual/cmake.1.html#view-help) and is only for printing the help text. See [here](https://stackoverflow.com/a/68699345/12591388) for the question I asked about it on Stack Overflow.

### CMake Tools Log
With `Clang 10.0.0 x86_64-pc-linux-gnu` kit:
```
[main] Building folder: Light 
[main] Configuring folder: Light 
[proc] Executing command: /usr/bin/cmake --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_C_COMPILER:FILEPATH=/bin/clang-10 -DCMAKE_CXX_COMPILER:FILEPATH=/bin/clang++-10 -H/home/tirthankar/projects/physicc/Light -B/home/tirthankar/projects/physicc/Light/build -G Ninja
[cmake] Not searching for unused variables given on the command line.
[cmake] -- The C compiler identification is Clang 10.0.0
[cmake] -- The CXX compiler identification is Clang 10.0.0
[cmake] -- Check for working C compiler: /bin/clang-10
[cmake] -- Check for working C compiler: /bin/clang-10 -- works
[cmake] -- Detecting C compiler ABI info
[cmake] -- Detecting C compiler ABI info - done
[cmake] -- Detecting C compile features
[cmake] -- Detecting C compile features - done
[cmake] -- Check for working CXX compiler: /bin/clang++-10
[cmake] -- Check for working CXX compiler: /bin/clang++-10 -- works
[cmake] -- Detecting CXX compiler ABI info
[cmake] -- Detecting CXX compiler ABI info - done
[cmake] -- Detecting CXX compile features
[cmake] -- Detecting CXX compile features - done
[cmake] -- Found OpenGL: /usr/lib/x86_64-linux-gnu/libOpenGL.so   
[cmake] -- Looking for pthread.h
[cmake] -- Looking for pthread.h - found
[...]

PR #2059, opened 2021-08-17
use -S to configure source directory

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/2053

### DIFF
--- a/src/drivers/cmfileapi-driver.ts
+++ b/src/drivers/cmfileapi-driver.ts
@@ -215,7 +215,7 @@ export class CMakeFileApiDriver extends CMakeDriver {
 
     // Dup args so we can modify them
     const args = Array.from(args_);
-    args.push(`-H${util.lightNormalizePath(this.sourceDir)}`);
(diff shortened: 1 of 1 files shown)


bobbrow (approved):

bobbrow:
Sorry, I approved this too soon before checking the docs.  We need to revert this or do a CMake version check before using this.  `-S` doesn't exist in CMake 3.10 which is still in use by many of our customers since it's the default in some versions of Ubuntu (18.04, I think).

alexreinking:
I don't understand this reversion. The code comments state that it's using the CMake File API for versions 3.15+, while `-S` has been in CMake since 3.13. The CMake file API itself was introduced in version 3.14, so there has never been a point in time where the File API worked, but `-S `did not.

See here:

https://github.com/microsoft/vscode-cmake-tools/blob/a6cbd3598acd908b34e9aa8b3b535a35d567f853/src/drivers/cmfileapi-driver.ts#L37-L39

PR #2061, opened 2021-08-17
revert using -S to configure source directory

elahehrashedi:
related to: https://github.com/microsoft/vscode-cmake-tools/issues/2053

### DIFF
--- a/src/drivers/cmfileapi-driver.ts
+++ b/src/drivers/cmfileapi-driver.ts
@@ -215,7 +215,7 @@ export class CMakeFileApiDriver extends CMakeDriver {
 
     // Dup args so we can modify them
     const args = Array.from(args_);
-    args.push(`-S${util.lightNormalizePath(this.sourceDir)}`);
+    args.push(`-H${util.lightNormalizePath(this.sourceDir)}`);
     const bindir = util.lightNormalizePath(this.binaryDir);
     args.push(`-B${bindir}`);
     const gen = this.generator;


bobbrow (approved):

alexreinking:
I don't understand this reversion. The code comments state that it's using the CMake File API for versions 3.15+, while `-S` has been in CMake since 3.13. The CMake File API itself was introduced in version 3.14, so there has never been a point in time where the File API worked, but `-S` did not.

See here:

https://github.com/microsoft/vscode-cmake-tools/blob/a6cbd3598acd908b34e9aa8b3b535a35d567f853/src/drivers/cmfileapi-driver.ts#L37-L39

In fact, it seems this class is protected by a check that `cmake.isFileApiModeSupported` is true, which _is already_ guarded by a version check.

https://github.com/microsoft/vscode-cmake-tools/blob/a6cbd3598acd908b34e9aa8b3b535a35d567f853/src/cmake/cmake-executable.ts#L37

https://github.com/microsoft/vscode-cmake-tools/blob/a6cbd3598acd908b34e9aa8b3b535a35d567f853/src/cmake-tools.ts#L689-L699

bobbrow:
You are correct. Because this was done in only one place I had assumed the file being edited was the base driver, but clearly this is not the case.  The other instance of this which should not be changed is here:
https://github.com/microsoft/vscode-cmake-tools/blob/a6cbd3598acd908b34e9aa8b3b535a35d567f853/src/drivers/legacy-driver.ts#L79

I assume that when this driver was added, it copied from there.  It is safe to make the change in the cmake-file-api driver without a version check.

Issue #2065, opened 2021-08-19
"Open" telemetry is broken


esweet431:
Version: v1.8.0-163e532

1. Open a folder that contains a CMakeLists.txt file in VS Code 
2. The cmake-tools extension fully activates 

Expected behavior: 
3. The EventName "open" fires with property.isFullyActivated == true

Actual behavior: 
3. The EventName "open" fires with property.isFullyActivated == false

Through my manual testing, I haven't found any instances where property.isFullyActivated returns true

bobbrow:
Fix for this is in PR now.

PR #2068, opened 2021-08-19
Fix "open" telemetry

bobbrow:
Addresses #2065


elahehrashedi (approved):
can we use this or the map is empty at this point anyway? 
`this._foldersAreCMake.get(cmtFolder.cmakeTools.folderName) || await this.folderIsCMakeProject(cmtFolder.cmakeTools)`

bobbrow:
The map is empty on open, so I just call the function.

PR #2072, opened 2021-08-24
Fix presets for msvc compilers with x86 outputs

bobbrow:
Fixes an issue I found while testing 1.8.0 with presets on Windows.

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -7,7 +7,7 @@ import * as logging from '@cmt/logging';
 import { EnvironmentVariables, execute } from '@cmt/proc';
 import { expandString, ExpansionOptions } from '@cmt/expand';
 import paths from '@cmt/paths';
-import { effectiveKitEnvironment, getKitEnvironmentVariablesObject, Kit } from '@cmt/kit';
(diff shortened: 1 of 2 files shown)


src/preset.ts
```diff
@@ -666,9 +666,8 @@ async function expandConfigurePresetHelper(folder: string,
             const kit = kits[i];
             if (kit.visualStudio && !kit.compilers) {
               const version = vsVersions.get(kit.visualStudio);
-              if (kit.preferredGenerator &&
-                  (kit.visualStudioArchitecture === arch || kit.preferredGenerator.platform === arch) &&
-                  kit.preferredGenerator.toolset === ('host=' + toolset.host)) {
+              if (kit.preferredGenerator && targetArchFromGeneratorPlatform(kit.preferredGenerator.platform) === arch &&
+                  (kit.visualStudioArchitecture === toolset.host || kit.preferredGenerator.toolset === ('host=' + toolset.host))) {
```
bobbrow, line 670: `kit.visualStudioArchitecture` is the "host" architecture, not the target architecture so this had to be switched around.  This also transforms 'win32' -> 'x86' for the target architecture check.

elahehrashedi (approved):

Issue #1053, opened 2020-02-08
MSVC_VERSION is incorrect when cmake configure project with tool kit clang-cl 


L-Sun:
### Brief Issue Summary
When I configure my project, the cmake variable `${MSVC_VERSION}` output incorrectly.
Here is my CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.15)

project(MyProject)

add_executable(main main.cpp)
if(MSVC)
    message("MSVC_VERSION: ${MSVC_VERSION}")
endif(MSVC)
```

### Expected:

1. run command CMake: Delete Cache and Reconfigure
2. and Output panel will show:
```
...
[cmake] MSVC_VERSION: 1924
[cmake]  
[cmake] -- Configuring done
[cmake] -- Generating done
```

### Apparent Behavior:

1. run command CMake: Delete Cache and Reconfigure
2. and Output panel show the following message
```
...
[cmake] MSVC_VERSION: 1911
[cmake]  
[cmake] -- Configuring done
[cmake] -- Generating done
```

### Other Notes/Information

When I add `'VCToolsInstallDir'` to `MSVC_ENVIRONMENT_VARIABLES` array in `\.vscode\extensions\ms-vscode.cmake-tools-1.3.0\dist\main.js`(or src/kit.ts) of extension.
and reconfigure my project. It works as expected. The cmake variable `${MSVC_VERSION}` is correct.

### Platform and Versions

- **Operating System**: Windows
- **CMake Version**: 3.16.4
- **VSCode Version**: 1.42.0
- **CMake Tools Extension Version**: 1.3.0
- **Compiler/Toolchain**:  clang-cl (11.0.0) + VS2019 (v16.4.4)
- **visualStudioArchitecture**: amd64

tklajnscek:
+1 
Just hit the same issue where Clang can't find the MSVC toolkit correctly. 
Traced it down to this after many hours.

Is there any specific reason the `VCToolsInstallDir` is omitted from `MSVC_ENVIRONMENT_VARIABLES `? Seems like someone generated this a while ago and it never got updated.

Clang uses `VCToolsInstallDir` to find the location of the toolkit.

See `https://clang.llvm.org/doxygen/MSVC_8cpp_source.html` -> `findVCToolChainViaEnvironment()`

tklajnscek:
Oh, just to be clear - in my case this doesn't simply mess up the MSVC_VERSION. 
[...]

PR #2089, opened 2021-08-30
Add missing environment variables to MSVC_ENVIRONMENT_VARIABLE

tklajnscek:

- add any missing environment variables to MSVC_ENVIRONMENT_VARIABLES so they get forwarded correctly

## The purpose of this change

This also fixes Clang's ability to find the MSVC toolkit correctly as noted in [my comment on the issue](https://github.com/microsoft/vscode-cmake-tools/issues/1053#issuecomment-908260301)


src/kit.ts
```diff
@@ -603,13 +603,19 @@ const MSVC_ENVIRONMENT_VARIABLES = [
   'LIBPATH',
   'NETFXSDKDir',
   'Path',
+  'Platform',
```
bobbrow, line 606: Please remove 'Platform' from the list. This causes some of our internal projects to fail to build. Is it required to make your clang-cl Kit work?

bobbrow (changes requested):

src/kit.ts
tklajnscek: No, it doesn't. I just need the `VCTools*` parts. 
I ran a full diff purely as a this-seems-like-a-good-idea thing :)
I'll remove it and if anyone ever needs it they can add it.

tklajnscek: Done!

lygstate: I suggest not remove it, but Comment it out with reason about 'Platform', so that future user won't  add it back

src/kit.ts
```diff
@@ -603,13 +603,19 @@ const MSVC_ENVIRONMENT_VARIABLES = [
   'LIBPATH',
   'NETFXSDKDir',
   'Path',
+  //'Platform', - disabled at the request of @bobbrow as this causes some of MS internal projects to fail to build
```
bobbrow, line 606: We don't normally annotate our code like this.  I think it is sufficient to say that it is disabled because it is currently unnecessary and causes some projects to fail to build.

tklajnscek: Changed as requested.

bobbrow (approved):

tklajnscek:
Not sure why the tests seem to have failed (`C++ extension not installed` etc.), let's see if merging latest develop fixes it :)

bobbrow:
'C++ extension not installed' is not an error. I saw some flakiness on one of the test platforms recently. We need to investigate that.

tklajnscek:
Yeah, seems to have passed now!

Issue #2088, opened 2021-08-30
1.8.0 release ignores cmake.configureOnOpen setting 


xikxp1:
### Brief Issue Summary

Newest 1.8.0 release ignores "cmake.configureOnOpen" setting. Every time I reload my workspace cmake reconfiguration happens.

### Expected:

1. Open workspace with `"cmake.configureOnOpen": false` in workspace settings
2. Cmake configure doesn't happen

### Apparent Behavior:

1. Open workspace with `"cmake.configureOnOpen": false` in workspace settings
2. Extension runs cmake configure

### CMake Tools Log

```
[variant] Loaded new set of variants
[kit] Successfully loaded 6 kits from /home/xikxp1/.local/share/CMakeTools/cmake-tools-kits.json
[proc] Executing command: /usr/bin/clang-7 -v
[main] Configuring folder: uservices 
[cmake] C compiler: /usr/bin/clang-7
[cmake] C++ compiler: /usr/bin/clang++-7
[cmake] Using LLVM lld linker
[cmake] LTO: ON
[cmake] LTO_CACHE: ON in folder /home/xikxp1/uservices-build/build/.ltocache
[cmake] ccache: enabled
[cmake] variant: 1
[cmake] Boost version: 1.58.0
[cmake] Checking -Wdisabled-optimization
[cmake] Checking -Wdisabled-optimization - found
[cmake] Checking -Winvalid-pch
[cmake] Checking -Winvalid-pch - found
[cmake] Checking -Wlogical-op
[cmake] Checking -Wlogical-op - not found
[cmake] Checking -Wuseless-cast
[cmake] Checking -Wuseless-cast - not found
[cmake] Checking -Wformat=2
[cmake] Checking -Wformat=2 - found
[cmake] Checking -Wno-error=deprecated-declarations
[cmake] Checking -Wno-error=deprecated-declarations - found
[cmake] Checking -ftemplate-depth=200
[cmake] Checking -ftemplate-depth=200 - found
[cmake] Checking -Wimplicit-fallthrough
[cmake] Checking -Wimplicit-fallthrough - found
[cmake] boost: 105800
[cmake] Python: /usr/lib/yandex/taxi-py3-2/bin/python3
[cmake] Checking python requirements: /home/xikxp1/arc/arcadia/taxi/uservices
[cmake] Checking python requirements: /home/xikxp1/arc/arcadia/taxi/uservices/testsuite
[...]

PR #2092, opened 2021-08-30
Fix cmake.configureOnOpen

bobbrow:
cmake.configureOnOpen is not respected after the "configure from existing cache" feature was added. The feature took us down the configuration path, but did not exit early when it was discovered that there was no cache to configure from.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # What's New?
 
+## 1.8.1
+Bug fixes:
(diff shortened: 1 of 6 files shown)


bobbrow:
Addresses #2088

elahehrashedi:
tested few scenarios on single-root and multi-root project and it worked

elahehrashedi (approved):

Issue #2091, opened 2021-08-30
1.8.0 breaks command substitutions in launch.json


fbridault:
### Brief Issue Summary

I used to employ many cmake command substitutions to launch and debug targets in my projects. Since 1.8.0, this no longer works. Reverting the extension to 1.7.3 solves the problem.

### Expected:

1. Use for instance ${command:cmake.buildDirectory} in the  "program" property of a target in `launch.json` (or any command:cmake* variable in any property)
2. Select this target and click on play button, or press F5
3. Programs launches

### Apparent Behavior:

1. Use for instance ${command:cmake.buildDirectory} in the  "program" attribute of a target in `launch.json`
2. Select this target and click on play button, or press F5
3. Programs does not launch, the debug bar very briefly pops in and immediately out. No error message is displayed anywhere. 

### CMake Tools Log

Not relevant I guess, the log is not filled in when launching a target.

### Developer Tools Log

Nothing moves in the log when trying to launch a target, but here is the log of extensions:
```
[2021-08-30 11:07:56.659] [exthost] [info] extension host started
[2021-08-30 11:07:56.659] [exthost] [info] Skipping acquiring lock for /home/fbridault/.config/Code/User/workspaceStorage/02745f54f69c82dedef436d43c94ae52.
[2021-08-30 11:07:56.984] [exthost] [info] ExtensionService#_doActivateExtension vscode.microsoft-authentication {"startup":false,"extensionId":{"value":"vscode.microsoft-authentication","_lower":"vscode.microsoft-authentication"},"activationEvent":"onAuthenticationRequest:microsoft"}
[2021-08-30 11:07:56.985] [exthost] [info] ExtensionService#loadCommonJSModule file:///snap/code/72/usr/share/code/resources/app/extensions/microsoft-authentication/dist/extension.js
[2021-08-30 11:07:57.038] [exthost] [info] ExtensionService#_doActivateExtension redhat.vscode-xml {"startup":false,"extensionId":{"value":"redhat.vscode-xml","_lower":"redhat.vscode-xml"},"activationEvent":"onLanguage:xml"}
[2021-08-30 11:07:57.038] [exthost] [info] ExtensionService#loadCommonJSModule file:///home/fbridault/.vscode/extensions/redhat.vscode-xml-0.18.0/dist/extension
[...]

PR #2093, opened 2021-08-30
launch a program in multiroot 

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/2091

the input of this function is not a folder when we are launching. but there was a check inside this function that checks if its a single root project, that is why it was working before. 
https://github.com/microsoft/vscode-cmake-tools/blob/114b910bf0c136ac50b7e78d44e3c4133bd4aff0/src/extension.ts#L1096

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1099,6 +1099,9 @@ class ExtensionManager implements vscode.Disposable {
       if (cmtFolder) {
         return fn(cmtFolder.cmakeTools);
       }
+      else {
+        rollbar.error(localize('no.active.folder', 'No active folder.'));
+      }
     } else {
       rollbar.error(localize('invalid.folder', 'Invalid folder.'));
     }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1098,8 +1098,7 @@ class ExtensionManager implements vscode.Disposable {
       const cmtFolder = this._folders.get(workspaceFolder);
       if (cmtFolder) {
         return fn(cmtFolder.cmakeTools);
-      }
-      else {
+      } else {
         rollbar.error(localize('no.active.folder', 'No active folder.'));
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -1098,6 +1098,8 @@ class ExtensionManager implements vscode.Disposable {
       const cmtFolder = this._folders.get(workspaceFolder);
       if (cmtFolder) {
         return fn(cmtFolder.cmakeTools);
+      } else {
+        rollbar.error(localize('no.active.folder', 'No active folder.'));
```
bobbrow, line 1102: Is it an error to not have an active folder?  I'm not sure the user can do anything about it, so it's probably best not to bug them about it.

src/util.ts
```diff
@@ -648,9 +648,7 @@ export function isArrayOfString(x: any): x is string[] {
 }
 
 export function isNullOrUndefined(x?: any): boolean {
-  // Double equals provides the correct answer for 'null' and 'undefined'
-  // http://www.ecma-international.org/ecma-262/6.0/index.html#sec-abstract-equality-comparison
-  return x === null;
+  return (x === null || x === undefined);
```
bobbrow, line 651: Ah.  It looks like this broke recently.  It used to be `x == null`.  I guess to keep the linter happy, this new approach is better.

bobbrow (changes requested):

src/extension.ts
elahehrashedi: before fixing the problem, this was where the error was happening (i.e. in `this._folders.get(workspaceFolder);`). I guess the error message is not necessary anymore as the error is prevented now.

bobbrow (approved):

PR #2102, opened 2021-09-01
Add support for CMakePresets version 3

sfreed141:
### This changes presets to support version 3

Update presets to support version 3 of CMakePresets.json, which has the following changes
- The `binaryDir` and `generator` properties are now optional, and will default to `${sourceDir}/out/build/${presetName}` and `Ninja` (or `cmake.generator` or `cmake.preferredGenerators` if set), respectively
- Add `installDir` and `toolchainFile` properties for configure presets, which map to the `CMAKE_INSTALL_PREFIX` and `CMAKE_TOOLCHAIN_FILE` cache variables, respectively
- Add a `condition` property to all presets, which allows more advanced control over enabling/disabling presets (compared to the `hidden` property)
- A new macro, `${hostSystemName}`, which evaluates to `uname -s` on systems that support that command (on Windows it will evaluate to `Windows`).

## Other Notes/Information

Since VSCode doesn't allow contributing multiple schemas for the same file, the v3 schema will be used (since it's more permissive). This means CMakePresets.json using v2 will not see error squiggles for invalid properties (like `installDir` or `toolchainFile`), however those errors will be caught be schema validation (and shown in an error message).

More details on CMakePresets.json can be found at https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html


src/preset.ts
```diff
@@ -392,28 +504,42 @@ function getVendorForConfigurePresetHelper(folder: string, preset: ConfigurePres
...
 
-export async function expandConfigurePreset(folder: string,
-                                            name: string,
-                                            workspaceFolder: string,
-                                            sourceDir: string,
-                                            allowUserPreset: boolean = false): Promise<ConfigurePreset | null> {
-  const refs = referencedConfigurePresets.get(folder);
-  if (!refs) {
-    referencedConfigurePresets.set(folder, new Set());
+async function getHostSystemName(): Promise<string> {
```
xisui-MSFT, line 507: Nit: you may want to put this to util.ts since this doesn't require anything in this file. Also, you can use memento in VS Code (`extensionContext.globalState`) to store the value so this value can be persisted in all VS Code instances.

xisui-MSFT (approved):

src/preset.ts
sfreed141: Thanks, I was wondering if VSCode had a built in way to handle that. Two questions though:
1. Does `extensionContext.globalState` store state independently for each platform? From [the docs](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.globalState) the storage is independent of any workspace but it doesn't mention the behavior if using the remote extensions. My guess is it is independent (my understanding is when using the remote extension the headless server process is an independent install, but I'm not sure if that's correct).
2. Is the only way to get a reference to `extensionContext` through `CMakeTools`? If so it means we'd have to pass that through the expand*Preset functions, which seems awkward to me.

src/preset.ts
```diff
@@ -392,28 +504,42 @@ function getVendorForConfigurePresetHelper(folder: string, preset: ConfigurePres
...
+  if (platform() === "win32") {
+    return "Windows";
   } else {
-    refs.clear();
+    return promisify(exec)('uname -s').then(
```
bobbrow, line 511: Can you use `proc.execute` instead?  There should be examples in the code of how to use it.

src/preset.ts
```diff
@@ -445,17 +600,53 @@ export async function expandConfigurePreset(folder: string,
 
   expansionOpts.envOverride = expandedPreset.environment as EnvironmentVariables;
 
+  const presetsFile = getPresetsFile(folder);
+  if (presetsFile && presetsFile.version >= 3) {
+    // For presets v3+ binaryDir and generator are optional, but cmake-tools needs a value. Default to something reasonable.
+    if (!preset.binaryDir) {
+      const defaultValue =  '${sourceDir}/out/build/${presetName}';
+
+      log.warning(localize('binaryDir.undefined', 'Configure preset {0}: No binaryDir specified, using default value "{1}"', preset.name, defaultValue));
```
bobbrow, line 609: We get complaints from customers when we warn too frequently.  If this is going to be printed every time the presets file is parsed, it might be too much. For cases like this our options are to track whether we've warned this or not already and only do it once, or to demote the warning to `log.debug`.

src/preset.ts
```diff
@@ -445,17 +600,53 @@ export async function expandConfigurePreset(folder: string,
...
+      preset.binaryDir = defaultValue;
+    }
+    if (!preset.generator) {
+      const defaultValue = 'Ninja';
```
bobbrow, line 613: The CMake docs say:
*In version 3 or above, this field may be omitted to fall back to regular generator discovery procedure.*

Are we supposed to emulate that or is there some strategy that will allow our generate command to match what might happen on the CLI?

src/preset.ts
```diff
@@ -392,28 +504,42 @@ function getVendorForConfigurePresetHelper(folder: string, preset: ConfigurePres
...
+
+const getHostSystemNameMemo = memoize(getHostSystemName);
+
+async function getExpansionOptions(folder: string,
+                                   workspaceFolder: string,
+                                   sourceDir: string,
+                                   preset: ConfigurePreset | BuildPreset | TestPreset) {
+  const generator = 'generator' in preset
+    ? preset.generator
+    : ('__generator' in preset ? preset.__generator : undefined);
```
bobbrow, line 538: Using the field name in a string here seems dangerous.  I don't expect that we'll rename it, but I would think this could break if we ever did a rename.  For this particular case since you only care about the property and not the `preset` object's type, I think you more safely achieve the same result by doing:
```ts
const generator = preset.generator || preset.__generator || undefined;
```

test/unit-tests/presets.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      { type: 'const', value: true },
+      { type: 'equals', lhs: 'foo', rhs: 'foo' }
+    ]};
+    expect(evaluateCondition(condition)).to.eql(true);
+
+    condition.type = 'anyOf';
+    expect(evaluateCondition(condition)).to.eql(true);
+
+    condition = { type: 'allOf', conditions: [
+      { type: 'const', value: false },
```
bobbrow, line 28: Since tests are pretty cheap, it might be good to throw in a test or two with a bogus type (e.g. emulate a misspelled preset condition in the JSON file)

Having one true and one false test for each operator would also be a good idea I think.

src/preset.ts
sfreed141: I agree it's not ideal but I get type errors if I just try using `preset.generator` because only `ConfigurePreset` contains the `generator` property. Using the field name seemed like the simplest way to deal with that.

Fwiw if you do indeed rename `generator` then there will be an error on `preset.generator` (with the current code).

src/preset.ts
sfreed141: That's fair, I can make it a debug warning.

src/preset.ts
xisui-MSFT: 1. It's independent, since extension is installed separately on the remote machine in remote case (for this extension).
2. That's a good questions. Actually you probably don't want to use CMakeTools in this file since it depends on this file. Then I guess the current way is probably good enough. And still, I don't have a strong preference on this but you may want move this to util.ts.

src/preset.ts
sfreed141: I'm not aware of any documented strategy for determining the generator. The closest I found was from [CMAKE_GENERATOR](https://cmake.org/cmake/help/latest/envvar/CMAKE_GENERATOR.html#envvar:CMAKE_GENERATOR)
> If the provided value doesn't name a generator known by CMake, the internal default is used.

which is pretty unhelpful. I went with `Ninja` just because that's what we do in VS. I can see what happens if I leave generator undefined, but I think there are a few places that assume generator is present. We could try matching CMake behavior experimentally too but, at least to me, it seems like more trouble than it's worth.

src/preset.ts
xisui-MSFT: I think your method is better, but I'm providing an alternative here anyway. You can try to use something like `(preset as ConfigurePreset).generator`. I don't think there would be any runtime errors this way but you probably want to double check.

bobbrow: It doesn't need to be changed since my rename concern is addressed. Ideally we'd have separate objects for the raw JSON and the expanded presets so that the properties could be the same on all 3 objects, but that's out of scope for this PR.

src/preset.ts
bobbrow: VS has the benefit of knowing that Ninja is always available.  Maybe we could check to see if the `cmake.generator` or `cmake.preferredGenerators` setting is set and use that as the first choice.

src/preset.ts
sfreed141: Thanks for confirming. I'll move it to util.ts but I'll keep the current memo code to avoid passing in CMakeTools

src/preset.ts
sfreed141: Good point, I changed it to try falling back to those settings first before falling back to Ninja.

bobbrow (approved):

Issue #2110, opened 2021-09-03
No project outline when targets have the same output name


fjacomme:
### Brief Issue Summary

Since last update 1.8.1.
The CMAKE PROJECT OUTLINE fails to be populated with the following error:

"Element with id test:C:\dev\test::mylib.dll:C:\dev\test is already registered"

![Capture d’écran 2021-09-03 112437](https://user-images.githubusercontent.com/56129513/131982866-43786a32-63e0-4c16-b16a-4eace0895ed1.png)

This occurs when several libraries have the same output name (CMake supports this so dynamic libraries can be swapped at runtime).

Everything else (build, configure, etc. ) works as expected.

### Expected:

The CMAKE PROJECT OUTLINE should show the targets.

The following CMakeLists.txt reproduces the issue.

```cmake
cmake_minimum_required(VERSION 3.10)

project(test)

add_library(lib1 SHARED lib1.cpp)
set_target_properties(lib1 PROPERTIES
    OUTPUT_NAME mylib
    RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/mylib1
    LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/mylib1
)

add_library(lib2 SHARED lib2.cpp)
set_target_properties(lib2 PROPERTIES
    OUTPUT_NAME mylib
    RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/mylib2
    LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/mylib2
)

add_executable(app main.cpp)
target_link_libraries(app lib1)

```

bobbrow:
Thanks for reporting this.  I didn't think we touched the outline this time, but we'll take a look.

bobbrow:
This also happens in 1.7.3, so it is not a regression from the 1.8 release.  Your example CMakeLists.txt made it easy to reproduce and debug the problem though.  I have a PR out for the fix.

You can help us with validating the fix by doing the following:
1. Download a version of the extension with the fix: https://github.com/microsoft/vscode-cmake-tools/suites/3678108347/artifacts/89553219
2. Unzip the VSIX file
3. Run the "Install from VSIX" command in VS Code and select the VSIX downloaded.

PR #2111, opened 2021-09-03
Fix an issue with non-unique ids

bobbrow:
The project outline fails to load when the item id's are not unique. When two targets have the same artifact name, we end up with non-unique ids. The fix is to include both the target name and the artifact name in the id.


bobbrow:
Addresses #2110

Issue #2121, opened 2021-09-12
Debug shows "No compiler found in cache file." dialog


anb0s:
### Brief Issue Summary

We are using CMakePresets and configure, build and run (Shift + F5) are working as expected. But using the debug-button shows dialog with only this error: "No compiler found in cache file."

### Expected:

The Debug session with builded application is starting or more detailed error is shown to investigate the configuration issue...

### Apparent Behavior:

Push the debug-button:

![image](https://user-images.githubusercontent.com/95811/132983667-6ad08e86-5a8f-48ba-910c-ab7397d66b12.png)

shows dialog with only this error:

"No compiler found in cache file."

![image](https://user-images.githubusercontent.com/95811/132983707-5d315513-e250-497e-a9cf-aba0123ece20.png)

### CMake Tools Log

```
main] Building folder: oae.jupiter.s7plcapp 
[build] Starting build
[proc] Executing command: /usr/bin/cmake --build /home/andre/git/oae.jupiter.s7plcapp/build/default
[build] ninja: no work to do.
[build] Build finished with exit code 0
[main] Building folder: oae.jupiter.s7plcapp s7plcapp
[build] Starting build
[proc] Executing command: /usr/bin/cmake --build /home/andre/git/oae.jupiter.s7plcapp/build/default --target s7plcapp
[build] ninja: no work to do.
[build] Build finished with exit code 0
```

### Developer Tools Log

```
???
```

### Platform and Versions

- **Operating System**: WSL2 with Debian 11 Bullseye (VSCode running @ Windows 10 1909)
- **CMake Version**: 3.21.2
- **VSCode Version**: 1.60.0
- **CMake Tools Extension Version**: 1.8.1
- **Compiler/Toolchain**: gcc version 10.2.1 20210110 (Debian 10.2.1-6)

### Other Notes/Information

...

linquize:
What platform and compiler are you using?

anb0s:
> What platform and compiler are you using?

Ah sorry, somehow the description was not saved properly. I've updated the issues description now...

bobbrow:
The debug button at the bottom tries to guess which debugger to use based on information about your compiler found in the CMakeCache.txt and it generates the `launch.json` information on the fly.  If it can't find a `CMAKE_CXX_COMPILER` or `CMAKE_C_COMPILER` entry in your cache, it doesn't know where to look to find the debugger and you get this error message.

[...]

PR #2123, opened 2021-09-13
reorder some logic in the debugger override case

bobbrow:
When searching for a debugger in the "quick debug" case, it's possible the user tries to override the guessing by providing their own path to a debugger. However, if their compiler is not found in the cache, this override attempt will be thwarted.

This PR reorders some of the logic to ensure that we attempt to use the override before the cache is checked.

> Note that this only matters for non-MSVC debugging, so some code above the override check is not moved.


bobbrow:
Addresses: #2121

PR #2131, opened 2021-09-18
Add support for parsing IAR output.

willson556:
### This changes compiler output parsing

- No longer ignore `cmake.enabledOutputParsers` configuration value.
- Add IAR output parser.

## The purpose of this change

To allow users of IAR compiler to see compilation errors in the problems dialog.

## Other Notes/Information

- Tested with IAR ARM 8.
- Added unit test.


andreeis (approved):

andreeis:
@willson556, this looks good and thank you for implementing this support. Do you think you can add a few more tests? Maybe there are a few more outstanding scenarios specific to the IAR compiler output... how about relative paths for source files?

willson556:
Yeah, I'll look into adding more tests tomorrow. Relative paths are a good idea -- I haven't tested with those.

willson556:
Before https://github.com/microsoft/vscode-cpptools/pull/8174/commits/621ca46fdd04ff10684d7650526b4c54d9113d54, I was still getting an unused expression error from the linter on line 1825 on the old side of that commit.

Seems the `?.` expressions might confuse the linter?

willson556:
> 
> 
> Yeah, I'll look into adding more tests tomorrow. Relative paths are a good idea -- I haven't tested with those.

Even if relative paths are passed to IAR, it outputs an absolute path. So I believe that's handled.

When doing further testing, I did find the first line sometimes contained the start of the diagnostic message so I added a test to cover that and updated the code to correctly handle that case. I couldn't find any other interesting cases -- template errors, etc. all follow the normal format.

PR #2142, opened 2021-09-29
Set up localization for v3 presets schema

bobbrow:
Our localization pipeline for schemas expects the schema file name to end with `schema.json`. We missed that when adding support for v3 presets. 

This PR does not add the files to the i18n folder.  I want to copy those translations over to the v3 file, but it's going to take some time.


bobbrow:
Checking this in because I need to get started on the next part.

Issue #1927, opened 2021-06-10
Select MRU Build Preset/Test Preset when Configure Preset changes


esweet431:
I'm opening this feature suggestion on behalf of a customer. 

### Apparent Behavior:

1. Select a Configure Preset 
2. The active Build Preset (displayed in the status bar) defaults to "No Build Preset Selected". The active Test Preset defaults to "No Test Preset Selected".

### Expected:

1. Select a Configure Preset
2. The active Build Preset defaults to the last Build Preset used with the specified Configure Preset. The active Test Preset defaults to the last Test Preset used with the specified Configure Preset. The MRU persists between sessions of VS Code.

andreeis:
This improvement is included in the 1.9.0 CMake Tools release.

PR #2146, opened 2021-09-30
Auto select build and test preset when switching configure preset

xisui-MSFT:
When selecting a configure preset:
1. If this is the first time this configure preset is selected
  a. When there are valid build presets with `configurePreset` field setting to this configure preset, auto select the first one.
  b. When there are no such valid build presets, auto select the default (empty) build preset.
  c. Don't do anything for test presets, since some projects cannot use ctest.
1. If this is not the first time this configure preset is selected
  a. Auto select the build preset used last time this configure preset was active.
  b. If this build preset is not valid anymore, follow (1) steps.


bobbrow:
Addresses: #1927

Issue #1883, opened 2021-05-18
fix code formatting (help wanted from community)


elahehrashedi:
A reminder to ask the community to help us fix the code formatting, 
including indentation, wrapping the lines, ...

bobbrow:
I don't think this is something we'll ask the community for.  We should set up a formatting profile and just format the files using the VS Code Format File command.  I've been meaning to do it, but the last time we got down to 0 PR's we got a whole bunch of new ones before I had a chance to do it. 😄

elahehrashedi:
TODO:
- search for all  "tslint:disable ..." in the code.
- some rules are commented out in .eslintrc.js, bring them back one by one.

PR #2216, opened 2021-10-29
Style changes

gugavaro:
Adding .editorconfig

Removing some format settings form the setting files

Batch updating files to reflect spaces styles


src/api.ts
```diff
@@ -7,311 +7,311 @@
  * Copy the `api.ts` source file into your project to use it.
  */ /** */
 
-import {DebugSession, Disposable, Event, Terminal} from 'vscode';
+import { DebugSession, Disposable, Event, Terminal } from 'vscode';
```
bobbrow, line 10: It looks like there is some other formatter changing the spacing on `{}` blocks. I don't think we want to do that as part of this change.

src/api.ts
```diff
@@ -7,311 +7,311 @@
...
-  stdout: string;
-  /**
-   * Standard error output of the program. May be `` if standard error was
-   * not captured
-   */
-  stderr: string;
+    /**
+     * The return code of the program.
+     */
+    retc: number | null;
```
bobbrow, line 19: It looks like it is also spacing around the union operator `|`. I think this one is ok though.

src/cache.ts
```diff
@@ -20,306 +20,306 @@ const log = logging.createLogger('cache');
...
+
+    /**
+     * Create a new instance. This is **private**. You may only create an instance
+     * via the `fromPath` static method.
+     * @param _path Path to the cache
+     * @param _exists Whether the file exists
+     * @param _entries Entries in the cache
+     */
+    private constructor(private readonly _path: string,
+        private readonly _exists: boolean,
```
bobbrow, line 117: This is the kind of multi-line function signature spacing I was worried about introducing with a formatting change.

From the best I can tell, the TypeScript repo does not wrap lines in the function signatures even if they are long. I think we should follow their rules unless I can find a counter-example in their code.

bobbrow: But maybe we just let this go for now until we get all the formatting configured.

src/api.ts
bobbrow: Maybe add `"javascript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false` to settings.json

bobbrow:
ref: #1883

.vscode/settings.json
```diff
@@ -6,8 +6,5 @@
     "search.exclude": {
         "out": true // set this to false to include "out" folder in search results
     },
-    "typescript.tsdk": "./node_modules/typescript/lib", // we want to use the TS server from our node_modules folder to control its version
-    "files.trimTrailingWhitespace": true,
-    "clang-format.executable": "${workspaceFolder}/node_modules/.bin/clang-format",
```
bobbrow, line 11: There is a reference to clang-format in the CONTRIBUTING.md in the root folder. We'll need to keep the "formatting" section consistent with the changes we're making.

.editorconfig
```diff
@@ -0,0 +1,10 @@
+
+root = true
+
+[{src,scripts}/**.{ts,json,js}]
```
bobbrow, line 4: I'm confused by this. Wouldn't it be `{src,test}`?  The `test` folder appears to have participated in the formatting.

src/api.ts
bobbrow: It seems that breaks the linter for single-line code blocks since it wants whitespace after a semicolon. :(

bobbrow: But we could do a find/replace on `;}` and add a space.  Or just leave this for now and fix it when we set up a formatter. The built-in Typescript formatter just doesn't have enough options unfortunately.

bobbrow:
The merge conflicts look strange and not straightforward as I would have expected, so I started over with PR #2226.

PR #2234, opened 2021-11-08
Turn preset to use mergeEnvironmentWithExpand

lygstate:
Using consistence function for environment set and get to fix the
gap between windows and unix.

Directly use configureEnv, as now getConfigureEnvironment would not return environment with undefined/null value

Also expand with environment variable when get effectiveKitEnvironment

Fixes https://github.com/microsoft/vscode-cmake-tools/issues/1592

preset using the same environment variable merge strategy like kits does

Using GeneralEnvironmentType

Convert collectDevBatVars to use util.envGetValue

- 
- 
-

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -765,7 +765,7 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
   async executeCMakeCommand(args: string[], options?: ExecutionOptions): Promise<ExecutionResult> {
     const drv = await this.getCMakeDriverInstance();
     if (drv) {
-      return drv.executeCommand(drv.cmake.path, args, undefined, options).result;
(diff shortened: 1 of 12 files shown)


lygstate:
I want format these code to improve environment handling in vscode-cmake

bobbrow:
It looks like this just reimplements #1763 which wasn't the direction we wanted to take.  I started draft PR #2025 which is the way we want to go for this.  Sorry I haven't made any progress on it since then.

lygstate:
> It looks like this just reimplements #1763 which wasn't the direction we wanted to take. I started draft PR #2025 which is the way we want to go for this. Sorry I haven't made any progress on it since then.

Hi, I wanna rebase my work on PR #2025 and getting it finished, but I need resolve the conflict first.

Issue #1987, opened 2021-07-09
Override for inheriting configure environment when debugging/launching


andreeis:
There is some internal debate as to which preset should supply the environment by default and how we might provide an override for people who don't like that decision.  Right now, Visual Studio uses the configurePreset environment instead of the build or test preset environment.

_Originally posted by @bobbrow in https://github.com/microsoft/vscode-cmake-tools/issues/1774#issuecomment-873302811_

andreeis:
Using this report for implementing additional cmake commands for getting the environment variables from more precise sources (configure, build, test presets) and controlled by the user.

lygstate:
I think should use test environment  for debug/launch, and build environment for compile source file

PR #2236, opened 2021-11-09
Enforcing consistent access to environment variables

### LINKED ISSUE #1592: environmentSetupScript capitalizes environment variable names

### LINKED ISSUE #2237: debugTarget failed when debug_config.environment not present.
debug_config.environment = debug_config.environment ? debug_config.environment.concat(util.makeDebuggerEnvironmentVars(configure_preset_environment)) : [];

means when debug_config have no environment  then  we can not debugging at all
even thout preset or toolkist are already configured it properly

lygstate:
## This change addresses item
Fixes #1592
Replacement of  #2025

Fixes #2237

~~Fixes https://github.com/microsoft/vscode-cmake-tools/issues/1987~~

### This changes

- Refactoring environmentVariables.ts out for consistence environment variable handling with NodeJS.ProcessEnv

## The purpose of this change
- Get environment variables on win32 properly, that is we can set/get environment variables in case-insensitive way
- Add backendTests to the CI, and revise the launch.json for faster debugging utils funcitons
- NOTE: Future direction, moving tests that's not depends on VSCode into backendTests

### DIFF
--- a/src/cmake-tools.ts
+++ b/src/cmake-tools.ts
@@ -29,7 +29,7 @@ import { CMakeGenerator, Kit } from './kit';
 import { LegacyCMakeDriver } from '@cmt/drivers/legacy-driver';
 import * as logging from './logging';
 import { fs } from './pr';
-import { buildCmdStr, EnvironmentVariables } from './proc';
+import { buildCmdStr, DebuggerEnvironmentVariable } from './proc';
 import { Property } from './prop';
 import rollbar from './rollbar';
 import * as telemetry from './telemetry';
@@ -44,6 +44,7 @@ import { updateFullFeatureSetForFolder, updateCMakeDriverInTaskProvider, enableF
 import { ConfigurationReader } from './config';
 import * as preset from '@cmt/preset';
 import * as util from '@cmt/util';
+import { Environment, EnvironmentUtils } from './environmentVariables';
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
@@ -2037,6 +2038,20 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
     }
 
     /**
+     * Both debugTarget and launchTarget called this funciton, so it's refactored out
+     * Array.concat's performance would not beat the Dict.merge a lot.
+     * This is also the point to fixing the issue #1987
+     */
+    async _getTargetLaunchEnvironment(drv: CMakeDriver | null, debug_env?: DebuggerEnvironmentVariable[]): Promise<Environment> {
+        const env = util.fromDebuggerEnvironmentVars(debug_env);
+
+        // Add environment variables from ConfigureEnvironment.
+        const configureEnv = await drv?.getConfigureEnvironment();
+
+        return EnvironmentUtils.merge([env, configureEnv]);
+    }
+
+    /**
      * Implementation of `cmake.debugTarget`
      */
     async debugTarget(name?: string): Promise<vscode.DebugSession | null> {
@@ -2096,15 +2111,12 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         // Add debug configuration from settings.
         const user_config = this.workspaceContext.config.debugConfig;
         Object.assign(debug_config, user_config);
-        // Add environment variables from configurePreset.
-        if (this.configurePreset?.environment) {
-            const configure_preset_environment = await drv.getConfigureEnvironment();
-            debug_config.environment = debug_config.environment ? debug_config.environment.concat(util.makeDebuggerEnvironmentVars(configure_preset_environment)) : [];
-        }
-
+        const launchEnv = await this._getTargetLaunchEnvironment(drv, debug_config.environment);
+        debug_config.environment = util.makeDebuggerEnvironmentVars(launchEnv);
         log.debug(localize('starting.debugger.with', 'Starting debugger with following configuration.'), JSON.stringify({
             workspace: this.folder.uri.toString(),
-            config: debug_config
+            config: debug_config,
+            environment: debug_config.environment
         }));
 
         const cfg = vscode.workspace.getConfiguration('cmake', this.folder.uri).inspect<object>('debugConfig');
@@ -2146,22 +2158,9 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         }
 
         const user_config = this.workspaceContext.config.debugConfig;
-        let launchEnv: EnvironmentVariables = {};
 
         const drv = await this.getCMakeDriverInstance();
-        if (user_config.environment) {
-            const debugConfigEnvironment: { name: string; value: string }[] = user_config.environment;
-            debugConfigEnvironment.forEach(envVar => {
-                launchEnv[envVar.name] = envVar.value;
-            });
-        }
-
-        // Add environment variables from configurePreset.
-        if (drv && this.configurePreset?.environment) {
-            const configure_preset_environment = await drv.getConfigureEnvironment();
-            launchEnv = util.mergeEnvironment(launchEnv, configure_preset_environment);
-        }
-
+        const launchEnv = await this._getTargetLaunchEnvironment(drv, user_config.environment);
         const termOptions: vscode.TerminalOptions = {
             name: 'CMake/Launch',
             env: launchEnv,
--- a/src/config.ts
+++ b/src/config.ts
@@ -11,6 +11,7 @@ import * as telemetry from '@cmt/telemetry';
 import * as vscode from 'vscode';
 import * as nls from 'vscode-nls';
 import { CppDebugConfiguration } from './debugger';
+import { Environment } from './environmentVariables';
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
@@ -22,10 +23,6 @@ export type CMakeCommunicationMode = 'legacy' | 'serverApi' | 'fileApi' | 'autom
 export type StatusBarButtonVisibility = "default" | "compact" | "icon" | "hidden";
 export type TouchBarButtonVisibility = "default" | "hidden";
 export type UseCMakePresets = 'always' | 'never' | 'auto';
-
-interface HardEnv {
-    [key: string]: string;
-}
 
 export interface AdvancedTouchBarConfig {
     configure?: TouchBarButtonVisibility;
@@ -115,10 +112,10 @@ export interface ExtensionConfigurationSettings {
     defaultVariants: object;
     ctestArgs: string[];
     ctestDefaultArgs: string[];
-    environment: HardEnv;
-    configureEnvironment: HardEnv;
-    buildEnvironment: HardEnv;
-    testEnvironment: HardEnv;
+    environment: Environment;
+    configureEnvironment: Environment;
+    buildEnvironment: Environment;
+    testEnvironment: Environment;
     mingwSearchDirs: string[];
     emscriptenSearchDirs: string[];
     mergedCompileCommands: string | null;
@@ -445,10 +442,10 @@ export class ConfigurationReader implements vscode.Disposable {
         defaultVariants: new vscode.EventEmitter<object>(),
         ctestArgs: new vscode.EventEmitter<string[]>(),
         ctestDefaultArgs: new vscode.EventEmitter<string[]>(),
-        environment: new vscode.EventEmitter<HardEnv>(),
-        configureEnvironment: new vscode.EventEmitter<HardEnv>(),
-        buildEnvironment: new vscode.EventEmitter<HardEnv>(),
-        testEnvironment: new vscode.EventEmitter<HardEnv>(),
+        environment: new vscode.EventEmitter<Environment>(),
+        configureEnvironment: new vscode.EventEmitter<Environment>(),
+        buildEnvironment: new vscode.EventEmitter<Environment>(),
+        testEnvironment: new vscode.EventEmitter<Environment>(),
         mingwSearchDirs: new vscode.EventEmitter<string[]>(),
         emscriptenSearchDirs: new vscode.EventEmitter<string[]>(),
         mergedCompileCommands: new vscode.EventEmitter<string | null>(),
--- a/src/drivers/driver.ts
+++ b/src/drivers/driver.ts
@@ -16,7 +16,7 @@ import { CMakeOutputConsumer } from '@cmt/diagnostics/cmake';
 import { RawDiagnosticParser } from '@cmt/diagnostics/util';
 import { ProgressMessage } from '@cmt/drivers/cms-client';
 import * as expand from '@cmt/expand';
-import { CMakeGenerator, effectiveKitEnvironment, Kit, kitChangeNeedsClean, KitDetect, getKitDetect, getKitEnvironmentVariablesObject, getVSKitEnvironment } from '@cmt/kit';
+import { CMakeGenerator, effectiveKitEnvironment, Kit, kitChangeNeedsClean, KitDetect, getKitDetect, getVSKitEnvironment } from '@cmt/kit';
 import * as logging from '@cmt/logging';
 import paths from '@cmt/paths';
 import { fs } from '@cmt/pr';
@@ -30,6 +30,7 @@ import { majorVersionSemver, minorVersionSemver, parseTargetTriple, TargetTriple
 import * as preset from '@cmt/preset';
 import * as codemodel from '@cmt/drivers/codemodel-driver-interface';
 import { DiagnosticsConfiguration } from '@cmt/folders';
+import { Environment, EnvironmentUtils } from '@cmt/environmentVariables';
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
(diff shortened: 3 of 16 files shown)


bobbrow:
Thank you for continuing to work on this!  I'll try to take a look soon, but I currently have some vacation time planned and other time critical things to complete this month which may make that difficult.

lygstate:
@bobbrow Hi, ping for review to avoid conflict with the repository

bobbrow:
I will be looking at this today.

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
...
+    toString() {
+        return this[envProperty].toString();
+    }
+}
+
+export class EnvironmentVariablesUtils {
+
+    public static create(from?: Map<string, string> | EnvironmentVariables | null, preserveNull?: boolean, isWin32?: boolean): EnvironmentVariablesUndefined {
+        const env = new EnvironmentVariablesPrivate(preserveNull, isWin32);
+        const p = new Proxy(env, {
```
bobbrow, line 92: A lot of complication in this class is because of this Proxy. Can you help me understand why we need to use a Proxy?  I was assuming it was to allow for indexing syntax, but that seems to be [supported in TypeScript](https://www.typescriptlang.org/docs/handbook/2/classes.html#index-signatures) already.

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+        return p;
+    }
+
+    public static createPreserveNull(from?: Map<string, string> | EnvironmentVariables | null): EnvironmentVariables {
+        return EnvironmentVariablesUtils.create(from, true);
+    }
+
+    public static mergeImpl(preserveNull: boolean, isWin32: boolean | undefined, ...envs: EnvironmentVariablesItem[]): EnvironmentVariablesUndefined {
```
bobbrow, line 129: An `impl` method shouldn't be public.  Can we just have:
```ts
public static merge(options: EnvironmentOptions, ...envs: Environment[]): Environment
```
You can remove the other convenience methods.

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
...
+                    p[key] = value;
+                }
+            } else {
+                Object.assign(p, from);
+            }
+        }
+        return p;
+    }
+
+    public static createPreserveNull(from?: Map<string, string> | EnvironmentVariables | null): EnvironmentVariables {
```
bobbrow, line 125: I would rather not have another function if you can already pass the information to `create` which is also public.

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
...
+                return Reflect.get(target, p);
+            },
+            getOwnPropertyDescriptor: (target, p) => {
+                if (typeof p === 'string') {
+                    return Reflect.getOwnPropertyDescriptor(target[envProperty], target.getKey(p, false));
+                } else {
+                    return Reflect.getOwnPropertyDescriptor(target, p);
+                }
+            }
+            ,
```
bobbrow, line 108: comma should go up one line

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
...
+    [inspect]() {
+        return util.inspect(this[envProperty]);
+    }
+
+    toString() {
+        return this[envProperty].toString();
+    }
+}
+
+export class EnvironmentVariablesUtils {
```
bobbrow, line 88: It would be really nice if `create` and `merge` could be on `Environment` as static methods.  Because of the proxy, we can't do this right now. If we can't remove the proxy, then please rename this `EnvironmentUtils` instead.

src/expand.ts
```diff
@@ -95,9 +95,9 @@ export interface ExpansionOptions {
  * @param opts Options for the expansion process
  * @returns A string with the variable references replaced
  */
-export async function expandString(tmpl: string, opts: ExpansionOptions) {
+export async function expandString(tmpl: string | null | undefined, opts: ExpansionOptions): Promise<string> {
     if (!tmpl) {
-        return tmpl;
+        return tmpl as string;
```
bobbrow, line 100: `as` is not type coercion.  If we must return a string here, return `""` explicitly.

src/expand.ts
```diff
@@ -121,8 +121,8 @@ export async function expandString(tmpl: string, opts: ExpansionOptions) {
 }
 
 async function expandStringHelper(tmpl: string, opts: ExpansionOptions) {
-    const envPreNormalize = opts.envOverride ? opts.envOverride : process.env as EnvironmentVariables;
-    const env = mergeEnvironment(envPreNormalize);
+    const envPreNormalize = opts.envOverride ? opts.envOverride : process.env;
+    const env = EnvironmentVariablesUtils.merge(envPreNormalize);
```
bobbrow, line 125: `create`?

src/kit.ts
```diff
@@ -1108,32 +1108,31 @@ export async function getVSKitEnvironment(kit: Kit): Promise<Map<string, string>
     return varsForVSInstallation(requested, kit.visualStudioArchitecture!, kit.preferredGenerator?.platform);
 }
 
-export async function effectiveKitEnvironment(kit: Kit, opts?: expand.ExpansionOptions): Promise<Map<string, string>> {
-    let host_env;
-    const kit_env = objectPairs(kit.environmentVariables || {});
+export async function effectiveKitEnvironment(kit: Kit, opts?: expand.ExpansionOptions): Promise<EnvironmentVariablesUndefined> {
+    let host_env: EnvironmentVariablesUndefined | undefined;
+    const kit_env = EnvironmentVariablesUtils.merge(kit.environmentVariables);
```
bobbrow, line 1113: `create`?

src/preset.ts
```diff
@@ -925,7 +926,7 @@ async function expandConfigurePresetHelper(folder: string,
             const compilerName: string | undefined = util.isSupportedCompiler(cxxCompiler) || util.isSupportedCompiler(cCompiler);
             if (compilerName) {
                 const compilerLocation = await execute('where.exe', [compilerName], null, {
-                    environment: preset.environment as EnvironmentVariables,
+                    environment: EnvironmentVariablesUtils.merge(preset.environment),
```
bobbrow, line 929: `create`

src/util.ts
```diff
@@ -367,34 +371,30 @@ export function* flatMap<In, Out>(rng: Iterable<In>, fn: (item: In) => Iterable<
...
-            return { ...acc, ...norm_vars };
-        } else {
-            return { ...acc, ...vars };
+export function makeDebuggerEnvironmentVars(env?: EnvironmentVariablesUndefined): DebuggerEnvironmentVariable[] {
+    if (!env) {
+        return [];
+    }
+    const converted_env: DebuggerEnvironmentVariable[] = [];
+    for (const [key, value] of Object.entries(env)) {
+        if (value) {
```
bobbrow, line 380: I think it should be an explicit check against null `value === null`.  We want to keep `0` and `""`

src/util.ts
```diff
@@ -367,34 +371,30 @@ export function* flatMap<In, Out>(rng: Iterable<In>, fn: (item: In) => Iterable<
...
+            });
         }
-    }, {});
+    }
+    return converted_env;
 }
 
-export function normalizeEnvironmentVarname(varname: string) {
-    return process.platform === 'win32' ? varname.toUpperCase() : varname;
+export function fromDebuggerEnvironmentVars(debug_env?: DebuggerEnvironmentVariable[]): EnvironmentVariablesUndefined {
```
bobbrow, line 390: If you take my feedback from `cmake-tools.ts`, this function should be removed.

test/backend-unit-tests/cmake/environment.test.ts
```diff
@@ -0,0 +1,60 @@
...
+            u: 'BBQ'
+        };
+        const resultA = EnvironmentVariablesUtils.mergeImpl(false, false, envA, undefined, envB);
+        expect(resultA).to.deep.equal({A: 'x', a: 'T', u: 'BBQ'});
+        const resultB = EnvironmentVariablesUtils.mergeImpl(true, false, envA, undefined, envB);
+        expect(resultB).to.deep.equal({A: 'x', B: null, a: 'T', u: 'BBQ'});
+        const resultC = EnvironmentVariablesUtils.mergeImpl(false, true, envA, undefined, envB);
+        expect(resultC).to.deep.equal({A: 'T', u: 'BBQ'});
+        const resultD = EnvironmentVariablesUtils.mergeImpl(true, true, envA, undefined, envB);
+        expect(resultD).to.deep.equal({A: 'T', B: null, u: 'BBQ'});
```
bobbrow, line 21: Could you also add a test that has a `null` in envB override a value from envA? That's the presets case that I want to ensure doesn't break.

test/helpers/vscodefake/extensioncontext.ts
```diff
@@ -38,8 +38,11 @@ export class DefaultExtensionContext implements vscode.ExtensionContext {
     get extensionMode(): vscode.ExtensionMode {
         throw new Error(notImplementedErr);
     }
+    extension: vscode.Extension<any>;
```
bobbrow, line 41: I don't understand what the changes in this file are for or how they're used.

bobbrow (changes requested):
I still need to test it, but here is the initial review.

package.json
```diff
@@ -21,7 +21,7 @@
     "native"
   ],
   "engines": {
-    "vscode": "^1.53.0"
+    "vscode": "^1.56.0"
```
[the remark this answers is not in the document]
lygstate: How about upgrade this first?

As the new vscode version have valid environment parameter

please refer https://github.com/microsoft/vscode-cmake-tools/pull/2236/commits/18accb7e121fe4fc177db76c6f732739cb0a9d16

src/config.ts
```diff
@@ -23,9 +24,7 @@ export type StatusBarButtonVisibility = "default" | "compact" | "icon" | "hidden
 export type TouchBarButtonVisibility = "default" | "hidden";
 export type UseCMakePresets = 'always' | 'never' | 'auto';
 
-interface HardEnv {
-    [key: string]: string;
-}
+export type HardEnv = EnvironmentVariablesUndefined;
```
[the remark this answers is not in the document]
lygstate: We can remove it in latter PR,

src/cmake-tools.ts
```diff
@@ -2041,6 +2042,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         return this.getCurrentLaunchTarget();
     }
 
+    async _getTargetLaunchEnvironment(drv: CMakeDriver | null, debug_env?: DebuggerEnvironmentVariable[]): Promise<EnvironmentVariablesUndefined> {
+        const env = util.fromDebuggerEnvironmentVars(debug_env);
```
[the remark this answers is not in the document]
lygstate: Hi, this function is shared, the performance won't hurt by this, as we already using map

src/cmake-tools.ts
```diff
@@ -2041,6 +2042,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+
+        // Add environment variables from ConfigureEnvironment.
+        const ctestEnv = await drv?.getConfigureEnvironment();
```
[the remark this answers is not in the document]
lygstate: Hi, this doesn't change when it's presets, for non-presets, variants environment should include

src/cmake-tools.ts
```diff
@@ -2151,22 +2158,9 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
-            });
-        }
-
-        // Add environment variables from configurePreset.
-        if (drv && this.configurePreset?.environment) {
-            const configure_preset_environment = await drv.getConfigureEnvironment();
-            launchEnv = util.mergeEnvironment(launchEnv, configure_preset_environment);
-        }
-
+        const launchEnv = await this._getTargetLaunchEnvironment(drv, user_config.environment);
```
lygstate, line 2163: @bobbrow This is the refactored out place.

src/cmake-tools.ts
```diff
@@ -2101,15 +2111,12 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         // Add debug configuration from settings.
         const user_config = this.workspaceContext.config.debugConfig;
         Object.assign(debug_config, user_config);
-        // Add environment variables from configurePreset.
-        if (this.configurePreset?.environment) {
-            const configure_preset_environment = await drv.getConfigureEnvironment();
-            debug_config.environment = debug_config.environment ? debug_config.environment.concat(util.makeDebuggerEnvironmentVars(configure_preset_environment)) : [];
-        }
-
+        const launchEnv = await this._getTargetLaunchEnvironment(drv, debug_config.environment);
```
lygstate, line 2114: @bobbrow This is the refactored out place.

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
...
+export type EnvironmentVariablesItem = EnvironmentVariables | null | undefined;
+
+class EnvironmentVariablesPrivate {
+    private keyMapping: Map<string, string>;
+    public [envProperty]: EnvironmentVariables;
+    protected isWin32: boolean;
+    protected preserveNull: boolean;
+    constructor(
+        preserveNull: boolean | undefined,
+        isWin32: boolean | undefined) {
```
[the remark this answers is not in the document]
lygstate: Hi, this is placed here intensionlly, so that we can testing win32/non-win32 on windows and linux without
setup two operating system or waiting the github actions result.
That's simplify the test case a lot, we have no need remove it in the future.

src/environmentVariables.ts
lygstate: > 
EnvironmentVariablesPrivate is proxied because we need
mantain compatiable with NodeJS.ProcessEnv.
For example, supporse we have a env named with `get`, if we using
typescript `Index Signatures`, then what's the result of env.get will have
two meaning:
  * call the function `get`
  * get the environment variable `get`
But for environment variable, access any member with `name` should return the expeted
environment variable for that  `name`

src/util.ts
lygstate: fromDebuggerEnvironmentVars is paired with makeDebuggerEnvironmentVars

src/drivers/driver.ts
```diff
@@ -171,13 +172,13 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
     /**
      * Compute the environment variables that apply with substitutions by expansionOptions
      */
-    async computeExpandedEnvironment(in_env: proc.EnvironmentVariables, expanded_env: proc.EnvironmentVariables): Promise<proc.EnvironmentVariables> {
-        const env = {} as { [key: string]: string };
+    async computeExpandedEnvironment(in_env: EnvironmentVariablesUndefined, expanded_env: EnvironmentVariablesUndefined): Promise<EnvironmentVariablesUndefined> {
+        const env = EnvironmentVariablesUtils.create();
         const opts = this.expansionOptions;
 
         await Promise.resolve(
```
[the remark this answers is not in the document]
lygstate: done

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
+import * as os from 'os';
+import * as util from 'util';
+
+const inspect = util.inspect.custom;
+const envProperty = Symbol('envProperty');
+
+export type EnvironmentVariablesBasic = Record<string, string>;
+export type EnvironmentVariablesNull = Record<string, string | null>;
+export type EnvironmentVariablesUndefined = Record<string, string | undefined>; // alias of NodeJS.ProcessEnv
```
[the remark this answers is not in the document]
lygstate: > 
 Record<string, string | undefined> === Dict<string>

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
...
+export type EnvironmentVariables = Record<string, string | undefined | null>;
```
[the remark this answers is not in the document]
lygstate: How to name EnvironmentVariablesUndefined?

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
+import * as os from 'os';
+import * as util from 'util';
+
+const inspect = util.inspect.custom;
+const envProperty = Symbol('envProperty');
+
+export type EnvironmentVariablesBasic = Record<string, string>;
```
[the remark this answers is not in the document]
lygstate: Yeap, we can remove it

src/environmentVariables.ts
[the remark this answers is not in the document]
lygstate: I wanna better name for
export type EnvironmentVariablesUndefined = Record<string, string | undefined>; // alias of NodeJS.ProcessEnv
export type EnvironmentVariables = Record<string, string | undefined | null>;
Name to Environment  and `EnvironmentWithNull`

src/environmentVariables.ts
lygstate: It's called 11 times in preset, and maybe more in the future, that's worth another function

src/expand.ts
lygstate: OK, the intension is do template like things, input string, return string, input null return null, that's we want

src/environmentVariables.ts
```diff
@@ -0,0 +1,147 @@
...
+class EnvironmentVariablesPrivate {
+    private keyMapping: Map<string, string>;
+    public [envProperty]: EnvironmentVariables;
+    protected isWin32: boolean;
+    protected preserveNull: boolean;
+    constructor(
+        preserveNull: boolean | undefined,
+        isWin32: boolean | undefined) {
+        this.keyMapping = new Map<string, string>();
+        this[envProperty] = {};
```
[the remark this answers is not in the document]
lygstate: /* Using envProperty symbol is to provide valid implemention for [inspect]() */

test/helpers/vscodefake/extensioncontext.ts
lygstate: Split to https://github.com/microsoft/vscode-cmake-tools/pull/2262

test/backend-unit-tests/cmake/environment.test.ts
lygstate: Added

src/environmentVariables.ts
bobbrow: Ok, we can keep the one for presets.

bobbrow (approved):
Approved. After enabling the Windows CI #2295, I will merge with this, run one last test pass, and then commit.

bobbrow:
One last question.  I still see a lot of direct interaction with `process.env` after this change.  Is the goal to make all of those accesses consistent too?  In other words, should we keep a global `Environment` that corresponds to `process.env` and enforce using that everywhere in the code?

lygstate:
> process.env

Hi, there is no need modify `process.env`, I was using `Environment` to simulate typeof `process.env`, they indeed are the same thing after all.

lygstate:
Can this be merged now?

bobbrow:
@andreeis was almost done with the validation and then had to take some sick leave.  This will go in soon.  Thank you for your patience.

bobbrow:
I merged it. We will continue to test as we have time, but I believe it will be good if future development already has this change in as it will add additional coverage.

Issue #2120, opened 2021-09-12
IntelliSense uses the short name from the variants file for buildType


ghuser404:
I've been having many issues with this extension, and was always hoping it would improve with future versions. The changes came very slowly, but I was still hoping for better. The issues were mostly related to go-to-definition, it either jumped into wrong places or didn't work at all.

With 1.8.1 everything stopped working. All my includes are red-squiggled (except maybe one or two headers, system headers and standard headers), F12 doesn't open those includes, for the most part it doesn't work on function/objects - just says "No definition found for...", but those functions/objects are not even red-squiggled.

I went back to 1.7.3, for now it works for me.
Please let me know how I can assist you in fixing this.

PS. Unfortunately, I can't share the code, it's commercial.

bobbrow:
> Please let me know how I can assist you in fixing this.

We can start by looking at logs. With version 1.8.1 installed, can you:
1. Open the source file with red squiggles in it
2. Run the `C/C++: Log Diagnostics` command and paste it here?
    a. If any of the paths or project information cannot be shared in this forum, please redact/rename that information before sharing.

Please also share which version of CMake you are using and your operating system.

ghuser404:
CMake 3.21.1
Windows 10 19043

1.8.1
```
-------- Diagnostics - 13/09/2021, 20:59:11
Version: 1.6.0
Current Configuration:
{
    "name": "CMake",
    "configurationProvider": "ms-vscode.cmake-tools",
    "compilerPathIsExplicit": false,
    "cStandardIsExplicit": false,
    "cppStandardIsExplicit": false,
    "intelliSenseModeIsExplicit": false,
    "includePath": [
        "C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30037/include/*",
        "C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30037/atlmfc/include/*",
        "C:/Program Files (x86)/Windows Kits/10/Include/10.0.18362.0/um",
        "C:/Program Files (x86)/Windows Kits/10/Include/10.0.18362.0/ucrt",
        "C:/Program Files (x86)/Windows Kits/10/Include/10.0.18362.0/shared",
        "C:/Program Files (x86)/Windows Kits/10/Include/10.0.18362.0/winrt",
        "C:/Program Files (x86)/Windows Kits/10/Include/10.0.18362.0/cppwinrt"
[...]

PR #2247, opened 2021-11-16
Fix IntelliSense usage of the short name (#2120)

gost-serb:
IntelliSense now uses buidType from variants file instead of short name

- IntelliSense will use current build type from variants files instead of short name


bobbrow:
Thanks for the PR!  We'll evaluate it soon.

bobbrow (approved):
Thank you for your help fixing this bug. I will do a follow up PR that documents the `activeVariant` property better so that we don't hit this again.

gost-serb:
No problem

PR #2266, opened 2021-11-27
Allow to launch multiple targets with the launch button of the status bar

ColinDuquesnoy:
### This changes the behavior of the launch button of the status bar

- A new terminal is created each time the user triggers the launch button of the status bar with a different target. If the user launches the same target multiple times, the behavior is defined by  `cmake.launch.allowParallel` and  `cmake.launch.terminatePreviousInstance`.
- The launchTarget command gained two configuration options to fine-tune its behavior:
    -  `cmake.launch.allowParallel`: Allow launching multiple instances of the **same target** in parallel. By default, it is disabled. If enabled, a new terminal will be created every time the command is triggered.
    -  `cmake.launch.terminatePreviousInstance`: Controls what happens when you launch another instance of a target and `#cmake.launch.allowParallel#` is disabled. By default, it is enabled. If enabled, the previous instance will first be terminated before launching the new one. Otherwise, the new instance will wait until the previous one has finished *(this is the current behavior, prior to this PR)*.
- Terminals are now stored in a map (using the PID of the terminal as the key) 

## Other Notes/Information

When  `cmake.launch.allowParallel` is disabled and `cmake.launch.terminatePreviousInstance` is enabled, a keyboard interrupt sequence (`'\u0003'`) is sent to the terminal. I am not sure if that will work on Windows (cmd.exe).


bobbrow:
Thank you for your PR!  We will get to it as soon as we can. Please note that we are entering the holiday season in the US and this will impact how soon we are able to complete the review.

src/cmake-tools.ts
```diff
@@ -2138,14 +2136,37 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+                        term.sendText('\u0003');
+                    }
+                    return term;
+                }
+            }
+        }
+
+        if (options && options.env) {
+            options.env[this._launchTerminalTargetName] = executable.name;
+        }
```
andreeis, line 2165: This was missing from the original implementation. The above check for `creationOptions.env![this._launchTerminalTargetName]` wasn't working because the env var wasn't set to begin with, causing a previous instance to not be terminated when settings require.

package.nls.json
```diff
@@ -135,5 +135,7 @@
     "cmake-tools.configuration.cmake.additionalKits.description": "Array of paths to custom kit files",
     "cmake-tools.configuration.cmake.exportCompileCommandsFile.description": "Enables exporting compile_commands.json.",
     "cmake-tools.configuration.cmake.useCMakePresets.description": "Use CMakePresets.json to configure drive CMake configure, build, and test. When using CMakePresets.json, kits, variants, and some settings in settings.json will be ignored.",
-    "cmake-tools.configuration.cmake.allowCommentsInPresetsFile.description": "Allow the use of JSON extensions such as comments in CMakePresets.json. Please note that your CMakePresets.json file may be considered invalid by other IDEs or on the command line if you use non-standard JSON."
+    "cmake-tools.configuration.cmake.allowCommentsInPresetsFile.description": "Allow the use of JSON extensions such as comments in CMakePresets.json. Please note that your CMakePresets.json file may be considered invalid by other IDEs or on the command line if you use non-standard JSON.",
+    "cmake-tools.configuration.cmake.launch.allowParallel": "Allow launching multiple instances of the **same target** in parallel.",
```
bobbrow, line 139: I think we should merge the two settings into a single enum because we only need 3 states, not 4.  Not sure what the new name should be at the moment, but the enum values would be something like "waitForCompletion", "allowParallel", "terminatePrevious".  "waitForCompletion" would be the default value since that is the current behavior today.

src/cmake-tools.ts
```diff
@@ -2177,24 +2208,21 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
 
         let launchArgs = '';
         if (user_config && user_config.args) {
             launchArgs = user_config.args.join(" ");
         }
 
-        this._launchTerminal.sendText(`${executablePath} ${launchArgs}`);
-        this._launchTerminal.show(true);
-        return this._launchTerminal;
+        const quoted = shlex.quote(executable.path);
```
bobbrow, line 2218: It looks like this is a duplicate of line 2201.  We should use `executablePath` instead.

test/extension-tests/successful-build/test/debugger.test.ts
```diff
@@ -207,5 +207,66 @@ suite('Debug/Launch interface', async () => {
...
+        }
+
+        const term1 = await cmt.launchTarget();
+        expect(term1).to.be.not.null;
+        const term1Pid = await term1?.processId;
+
+        const term2 = await cmt.launchTarget();
+        expect(term2).to.be.not.null;
+        expect(term2!.name).of.be.eq('CMake/Launch');
+
```
bobbrow, line 237: Let's add this check to term1 as well.

bobbrow (approved):
:shipit:

bobbrow:
I can't seem to push to your fork right now, so I'm going to commit this with the broken tests and follow up with the fixes immediately after.

PR #2292, opened 2021-12-17
Stop using `-H` to configure projects

friendlyanon:
The -H flag is an undocumented feature that was never supposed to be
used by the public. This commit makes the drivers use the appropriate
methods of configuring depending on the CMake version they handle:

Legacy: The binary directory has to be created first and CMake must be
invoked with that folder as its working directory. The source directory
is passed without any flags.

ServerAPI: The binary directory was already ensured to exist. No changes
necessary.

FileAPI: The driver for this assumes CMake >= 3.15, so the -S and -B
flags can be used. These flags were introduced in CMake 3.13.

Fixes: #2053


bobbrow (approved):
Thank you!

PR #2295, opened 2021-12-20
Fix tests and enable Windows CI

bobbrow:
Running the tests now works reliably on Windows.  (Currently requires Visual Studio 2019)

There are still gaps with MinGW and it would be nice to support an arbitrary VS version.  For now we are locked to the `windows-2019` virtual environment.  GitHub is testing a 2022 environment with a newer VS.  We can move to that next year.

Some additional improvements were added that allow you to run just a single test file instead of the whole corpus in the event that you want to debug just a single test without deleting everything.


.vscode/settings.json
```diff
@@ -8,5 +8,9 @@
     },
     "typescript.tsdk": "./node_modules/typescript/lib", // we want to use the TS server from our node_modules folder to control its version
     "typescript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": true,
-    "typescript.format.insertSpaceAfterOpeningAndBeforeClosingEmptyBraces": false
+    "typescript.format.insertSpaceAfterOpeningAndBeforeClosingEmptyBraces": false,
+    "cmake.sourceDirectory": "${workspaceFolder}/test/fakeOutputGenerator",
```
sfreed141, line 12: What is this for? Is the fakeOutputGenerator test like a decoy so the extension doesn't prompt or do any significant work in CI?

src/drivers/driver.ts
```diff
@@ -780,11 +759,11 @@ export abstract class CMakeDriver implements vscode.Disposable {
         const child = this.executeCommand(program, args, undefined, { silent: true });
         try {
             const result = await child.result;
-            log.debug(localize('command.version.test.return.code', 'Command version test return code {0}', nullableValueToString(result.retc)));
+            log.trace(localize('command.version.test.return.code', '"{0}" returned code {1}', `${program} --version`, nullableValueToString(result.retc)));
```
sfreed141, line 762: should `--version` be replaced with `${args}`?

src/paths.ts
```diff
@@ -18,10 +19,16 @@ interface VSCMakePaths {
...
     get AppData(): string | undefined {
+        if (process.env['CMT_TESTING'] === '1') {
+            return path.join(vscode.workspace.workspaceFolders![0].uri.fsPath, '.vscode');
+        }
         return process.env['APPDATA'];
     }
 
     get LocalAppData(): string | undefined {
+        if (process.env['CMT_TESTING'] === '1') {
+            return path.join(vscode.workspace.workspaceFolders![0].uri.fsPath, '.vscode');
```
sfreed141, line 30: Might be nice to still have a separate subfolder like `.vscode/LocalAppData` to avoid polluting the .vscode folder (if tests ever write to %TEMP%)

src/triple.ts
```diff
@@ -129,17 +129,17 @@ export function computeTargetTriple(target: TargetTriple): string {
 
 export function parseTargetTriple(triple: string): TargetTriple | undefined {
     const triples = triple.split("-");
-    let foundArch = "unknow";
-    let foundOs = 'unknow';
-    let foundAbi = 'unknow';
-    let foundLibc = 'unknow';
+    let foundArch = "unknown";
```
sfreed141, line 132: nit: use single quotes like the others

sfreed141 (approved):

.vscode/settings.json
bobbrow: It pretends to be a compiler and spits out fake version information when we launch it.

src/paths.ts
bobbrow: The extension already creates a folder in `%LocalAppData%` called `CMakeTools`.  This is just for tests so it just gets deleted at the end of CI.  Though I should add that path to the `.gitignore`

src/drivers/driver.ts
bobbrow: It's currently always just `--version` with one exception where it's `/?`, but I'll fix it.

src/triple.ts
bobbrow: fixed

PR #2307, opened 2021-12-30
Port over some of the cpptools issue bots

bobbrow:


bobbrow:
@Colengms could you help me test these when you get back?  TypeScript complains when I compile this locally with an error in some dependency.  I wasn't sure how you got this working in cpptools.

PR #2309, opened 2022-01-01
Fix the kit detection issue

### LINKED ISSUE #1233: Selecting Clang as the kit compiles with /usr/bin/c++ (g++)
### Brief Issue Summary

I noticed that the binary `c++` was being used for compilation with the clang kit as opposed to `clang++` on Linux. Investigating this, I found that "scan for kits" populates an entry for clang with an empty CXX variable.

`cmake-tools-kits.json` is populated as follows:

```json
[
  {
    "name": "Clang 10.0.0",
    "compilers": {
      "C": "/usr/bin/clang-10"
    }
  },
  {
    "name": "GCC 9.1.0",
    "compilers": {
      "C": "/usr/bin/gcc",
      "CXX": "/usr/bin/g++"
    }
  },
  {
    "name": "GCC 9.3.0",
    "compilers": {
      "C": "/usr/bin/gcc",
      "CXX": "/usr/bin/g++"
    }
  },
  {
    "name": "GCC for i686-w64-mingw32 9.3.0",
    "compilers": {
      "C": "/usr/bin/i686-w64-mingw32-gcc-9.3.0"
    }
  },
  {
    "name": "GCC for x86_64-pc-linux-gnu 9.3.0",
    "compilers": {
      "C": "/usr/bin/x86_64-pc-linux-gnu-gcc-9.3.0"
    }
  },
  {
    "name": "GCC for x86_64-w64-mingw32 9.3.0",
    "compilers": {
      "C": "/usr/bin/x86_64-w64-mingw32-gcc-9.3.0"
    }
  }
]
```

`which clang++` reports `/usr/bin/clang++` on my system.

### Expected:

1. Clang kit variable CXX is populated with `/usr/bin/clang++` (or whatever is correct)
2. Selecting Clang kit compiles with `clang++`

### Apparent Behavior:

1. Selecting Clang as the kit compiles with `/usr/bin/c++`

### Platform and Versions

- **Operating System**: Arch Linux
- **CMake Version**: 3.17.2
- **VSCode Version**: 1.44.2
- **CMake Tools Extension Version**: 1.3.1
- **Compiler/Toolchain**: clang 10.0.0

### LINKED ISSUE #1410: Still clang kit detection does not work
Hi, about one year ago or so when I upraded to clang 11 the clang kit was not detected anymore. After some days I got it working then. Now I upgraded to clang 12 and the same problems still exists. The kit doesn't get detected. I tried everything I could think of. I also added manually a clang-12 section to the "cmake-tools-kits.json". Then it passes "-DCMAKE_C_COMPILER:FILEPATH=/usr/bin/clang-12 -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/clang++-12" to cmake. But if I print out the compiler cmake is really using it comes out that the clang 12 compiler is only used for C not for CXX. So what the hell is so difficult to detect these kits? It cannot be so difficult to scan some folders for clang program variants. Please could someone fix these totally annoying problems finally.

### LINKED ISSUE #1653: Issues with gcc/g++ detection and use on OSX
When I install gcc-10 on OSX (using `brew install gcc`), and `Scan for Kits`, CMake Tools will populate the `cmake-tools-kits.json` with 2 separate instances for gcc.   The one titled "GCC 10.2.0", which seems valid.  Another titled "GCC for x86_64-apple-darwin20-gcc-10.2.0" contains a "C" compiler but no CPP compiler.

```
{
    "name": "GCC for x86_64-apple-darwin20 10.2.0",
    "compilers": {
        "C": "/usr/local/bin/x86_64-apple-darwin20-gcc-10.2.0".
    }
}
```

There is a `g++` symlink present in the same folder, titled, `x86_64-apple-darwin20-g++10`, which reports a 10.2.0 version, so appears to be the matching g++.  Oddly, these appear to be separate distinct binaries (and not symlinks to the same binary used by the "GCC 10.2.0" configuration).  If it's desirable for both kits to be detected, it should probably pick up the corresponding g++ path for this one, otherwise the kit does not work.  With both C and C++ files, I see CMake Tools falls back to Apple Clang.

Also, when configuring CMake Tools with the valid "GCC 10.2.0" kit, I receive a browse config containing a `compilerPath` of `/usr/bin/clang++`.  Shouldn't CMake Tools be using the compiler specified in the selected kit for the `compilerPath` of the browse config?

### LINKED ISSUE #1759: Clang Kit scanning will accidentally use GCC for C++
### Brief Issue Summary

The Clang kit scan is broken with current LLVM 11. It will show as

image

and then call CMake with only the `CMAKE_C_COMPILER` set. This leads to the mixing of GCC and Clang toolchains

image

This is very subtle and causes a lot of confusion.

### Expected:

1. Kits without a C++ compiler should not show if the project is C++. Or such kits should have a different style to warn users.
2. LLVM 11.1 Kits should be properly detected

### CMake Tools Log

```
[kit] Found Kit: GCC for aarch64-linux-gnu 10.2.0
[kit] Found Kit: GCC for aarch64-linux-gnu 10.2.0
[kit] Found Kit: GCC for arm-none-eabi 10.2.0
[kit] Found Kit: GCC for arm-none-eabi 10.2.0
[kit] Found Kit: Clang 11.1.0
[kit] Found Kit: Clang 11.1.0
[kit] Found Kit: GCC 10.2.0
[kit] Found Kit: GCC 7.4.1
[kit] Found Kit: GCC 8.4.0
[kit] Found Kit: GCC 9.3.0
[kit] Found Kit: GCC for riscv64-linux-gnu 10.2.0
[kit] Found Kit: GCC for riscv64-linux-gnu 10.2.0
[kit] Found Kit: GCC for riscv64-unknown-elf 8.2.0
[kit] Found Kit: GCC for riscv64-unknown-elf 8.2.0
[kit] Found Kit: GCC for x86_64-pc-linux-gnu 10.2.0
[kit] Found Kit: GCC for x86_64-pc-linux-gnu 10.2.0
[kit] Found Kit: GCC for x86_64-pc-linux-gnu 7.4.1
[kit] Found Kit: GCC for x86_64-pc-linux-gnu 7.4.1
[kit] Found Kit: GCC for x86_64-pc-linux-gnu 8.4.0
[kit] Found Kit: GCC for x86_64-pc-linux-gnu 8.4.0
[kit] Found Kit: GCC for x86_64-pc-linux-gnu 9.3.0
[kit] Found Kit: GCC for x86_64-pc-linux-gnu 9.3.0
[kit] Found Kit: GCC 10.2.0
[kit] Successfully loaded 15 kits from /home/fan/.local/share/CMakeTools/cmake-tools-kits.json
```

### Platform and Versions

- **Operating System**: Arch Linux
- **CMake Version**: 3.20.1
- **VSCode Version**: 1.55.1
- **CMake Tools Extension Version**: 1.6.0
- **Compiler/Toolchain**: LLVM 11.1

### LINKED ISSUE #2246: CXX Variable not set
### Brief Issue Summary

Visual Studio Code Version: 1.62.2
OS Version: Linux x64 5.13.19-2-MANJARO
Extension Version: 1.9.1 (Latest?)

CXX variable is not set when I try to scan active kits.

Screenshot from 2021-11-15 14-41-33

As you can see, I set my kit to a compiler that is not a defaulted system one (my default gcc/g++ compiler is gcc-11)

Then, if I generate my project, cmake give me this output:
```
[proc] Executing command: /usr/bin/cmake --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_C_COMPILER:FILEPATH=/bin/x86_64-pc-linux-gnu-gcc-10.3.0 -H/home/pablo/personal/cppStuff/fileBin -B/home/pablo/personal/cppStuff/fileBin/build -G "Unix Makefiles"
[cmake] Not searching for unused variables given on the command line.
[cmake] -- The C compiler identification is GNU 10.3.0
[cmake] -- The CXX compiler identification is GNU 11.1.0
```

As you can see, the variable `CMAKE_C_COMPILER` is set, but not the `CMAKE_CXX_COMPILER` one.

If I try on a Debian OS, both variables are set when I scan for Active kits
thumb-Clipboard - November 15, 2021 1_50 PM

So to solve this "problem" I modified the file `${HOME}/.local/share/CMakeTools/cmake-tools-kits.json` and added manually all `CXX` variables.

For example:
Generated by automated searching kits:
```
{
    "name": "Clang 12.0.1 x86_64-pc-linux-gnu",
    "compilers": {
      "C": "/bin/clang-12"
    }
  }
```

CXX Variable is not set, so I added a new entry:
```
{
    "name": "Clang 12",
    "compilers": {
      "C": "/bin/clang",
      "CXX": "/bin/clang++"
    }
  }
```

This happens with al compilers, and versions.
CXX variable should not be added too automatically?

fourdim:
## This change addresses item

This closes #2246, closes #1759, closes #1653, closes #1410,
closes #1233.

### This changes

Detect the kits on linux.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -308,6 +308,10 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
             // Ensure the version is match
             const version2 = await fs.exists(bin2) ? await getCompilerVersion('GCC', bin2, pr) : null;
             const version_is_match = version2 === null ? false : version2.fullVersion === version.fullVersion;
+            // For the kits with only `x86_64-pc-linux-gnu-gcc` provided:
+            // We will have bin2 === bin1 because the regex did not make a replacement,
+            // then version_match will be true, but no C++ compiler will be found,
+            // so we will correctly skip setting the CXX compiler.
             if (version_is_match) {
                 // Names like x86_64-pc-linux-gnu-gcc
                 gccCompilers.C = bin2;
@@ -396,6 +400,10 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
             // Ensure the version is match
             const version2 = await fs.exists(bin2) ? await getCompilerVersion('Clang', bin2, pr) : null;
             const version_is_match = version2 === null ? false : version2.fullVersion === version.fullVersion;
+            // For the kits with only `clang` provided:
+            // We will have bin2 === bin1 because the regex did not make a replacement,
+            // then version_match will be true, but no C++ compiler will be found,
+            // so we will correctly skip setting the CXX compiler.
             if (version_is_match) {
                 // Names like clang
                 clangCompilers.C = bin2;
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -303,7 +303,7 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
         } else {
             const fname2 = fname.replace(/gcc(-\d+(\.\d+(\.\d+)?)?)/, 'gcc');
             const bin2 = path.join(path.dirname(bin), fname2);
-            const gxx_fname2 = fname.replace(/gcc(-\d+(\.\d+(\.\d+)?)?)/, 'g++');
+            const gxx_fname2 = fname2.replace(/gcc/, 'g++');
             const gxx_bin2 = path.join(path.dirname(bin), gxx_fname2);
             // Ensure the version is match
             const version2 = await getCompilerVersion('GCC', bin2, pr);
@@ -384,27 +384,27 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
         const clangxx_bin1 = path.join(path.dirname(bin), clangxx_fname);
         log.debug(localize('detected.clang.compiler', 'Detected Clang compiler: {0}', bin));
         if (await fs.exists(clangxx_bin1)) {
-            // names like clang-13
+            // Names like clang-13
             clangCompilers.C = bin;
-            // names like clang++-13
+            // Names like clang++-13
             clangCompilers.CXX = clangxx_bin1;
         } else {
             const fname2 = fname.replace(/clang(-\d+(\.\d+(\.\d+)?)?)/, 'clang');
             const bin2 = path.join(path.dirname(bin), fname2);
-            const clangxx_fname2 = fname.replace(/clang(-\d+(\.\d+(\.\d+)?)?)/, 'clang++');
+            const clangxx_fname2 = fname2.replace(/clang/, 'clang++');
             const clangxx_bin2 = path.join(path.dirname(bin), clangxx_fname2);
             // Ensure the version is match
             const version2 = await getCompilerVersion('Clang', bin2, pr);
             const version_is_match = version2 === null ? false : version2.fullVersion === version.fullVersion;
             if (version_is_match && await fs.exists(bin2)) {
-                // names like clang
+                // Names like clang
                 clangCompilers.C = bin2;
             } else {
-                // names like clang-13
+                // Names like clang-13
                 clangCompilers.C = bin;
             }
             if (version_is_match && await fs.exists(clangxx_bin2)) {
-                // names like clang++
+                // Names like clang++
                 clangCompilers.CXX = clangxx_bin2;
             }
         }
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -292,11 +292,33 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
         if (version === null) {
             return null;
         }
-        const gxx_fname = fname.replace(/gcc/, 'g++');
-        const gxx_bin = path.join(path.dirname(bin), gxx_fname);
-        const gccCompilers: { [lang: string]: string } = { C: bin };
-        if (await fs.exists(gxx_bin)) {
-            gccCompilers.CXX = gxx_bin;
+        const gccCompilers: { [lang: string]: string } = { };
+        const gxx_fname1 = fname.replace(/gcc/, 'g++');
+        const gxx_bin1 = path.join(path.dirname(bin), gxx_fname1);
+        if (await fs.exists(gxx_bin1)) {
+            // Names like x86_64-pc-linux-gnu-gcc-11.1.0
+            gccCompilers.C = bin;
+            // Names like x86_64-pc-linux-gnu-g++-11.1.0
+            gccCompilers.CXX = gxx_bin1;
+        } else {
+            const fname2 = fname.replace(/gcc(-\d+(\.\d+(\.\d+)?)?)/, 'gcc');
+            const bin2 = path.join(path.dirname(bin), fname2);
+            const gxx_fname2 = fname.replace(/gcc(-\d+(\.\d+(\.\d+)?)?)/, 'g++');
+            const gxx_bin2 = path.join(path.dirname(bin), gxx_fname2);
+            // Ensure the version is match
+            const version2 = await getCompilerVersion('GCC', bin2, pr);
+            const version_is_match = version2 === null ? false : version2.fullVersion === version.fullVersion;
+            if (version_is_match && await fs.exists(bin2)) {
+                // Names like x86_64-pc-linux-gnu-gcc
+                gccCompilers.C = bin2;
+            } else {
+                // Names like x86_64-pc-linux-gnu-gcc-11.1.0
+                gccCompilers.C = bin;
+            }
+            if (version_is_match && await fs.exists(gxx_bin2)) {
+                // Names like x86_64-pc-linux-gnu-g++
+                gccCompilers.CXX = gxx_bin2;
+            }
         }
         const gccKit: Kit = {
             name: version.detectedName,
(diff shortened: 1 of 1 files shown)


fourdim:
@bobbrow Would you be so kind to review it and run the workflow?

bobbrow:
> @bobbrow Would you be so kind to review it and run the workflow?

Approved.  It looks like there are two tests that need to be updated with this change:
```
  1) Kits scan test
       Detect a GCC cross compiler compiler file:
     AssertionError: expected { Object (C, CXX) } to not have property 'CXX'
  	at Context.<anonymous> (out/test/unit-tests/kit-scan.test.js:100:54)


  2) Kits scan test
       Detect an MinGW compiler file on linux:
     AssertionError: expected { Object (C, CXX) } to not have property 'CXX'
  	at Context.<anonymous> (out/test/unit-tests/kit-scan.test.js:136:54)
```

src/kit.ts
```diff
@@ -292,11 +292,33 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
...
+            const version2 = await getCompilerVersion('GCC', bin2, pr);
+            const version_is_match = version2 === null ? false : version2.fullVersion === version.fullVersion;
+            if (version_is_match && await fs.exists(bin2)) {
+                // Names like x86_64-pc-linux-gnu-gcc
+                gccCompilers.C = bin2;
+            } else {
+                // Names like x86_64-pc-linux-gnu-gcc-11.1.0
+                gccCompilers.C = bin;
+            }
+            if (version_is_match && await fs.exists(gxx_bin2)) {
```
bobbrow, line 318: I think this is the issue causing the tests to fail. `fname1 === fname2` because there was no substitution made.  As a result `version2` is the same as `version`.  No substitution is made for `gxx_fname2` for the same reason and `gxx_bin2` exists because it's not `g++`.

In this block you should make sure that a substitution was made before marking a C compiler as a CXX compiler.  You'll need to do the same in the clang code.

bobbrow (changes requested):

src/kit.ts
fourdim: Address the comments.

src/kit.ts
```diff
@@ -292,11 +292,33 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
...
+            gccCompilers.C = bin;
+            // Names like x86_64-pc-linux-gnu-g++-11.1.0
+            gccCompilers.CXX = gxx_bin1;
+        } else {
+            const fname2 = fname.replace(/gcc(-\d+(\.\d+(\.\d+)?)?)/, 'gcc');
+            const bin2 = path.join(path.dirname(bin), fname2);
+            const gxx_fname2 = fname2.replace(/gcc/, 'g++');
+            const gxx_bin2 = path.join(path.dirname(bin), gxx_fname2);
+            // Ensure the version is match
+            const version2 = await getCompilerVersion('GCC', bin2, pr);
```
bobbrow, line 309: I think the part I struggle with here is that `bin2` === `bin1` if no substitution is made, so `version_is_match` will be true because it's the same compiler.  Looking at the code, it seems like the correct result is produced because `gxx_bin2` is the same as `gxx_bin1` (it's basically the same as the code we're replacing); It's just not immediately clear from the code that that is what's happening.  Could you add a comment or two here explaining this case?

fourdim: Sure.
For the kits like:

```
x86_64-pc-linux-gnu-gcc-11.1.0
x86_64-pc-linux-gnu-gcc
x86_64-pc-linux-gnu-g++
```
It will find the `x86_64-pc-linux-gnu-gcc-11.1.0` first.

As it can not find `x86_64-pc-linux-gnu-gcc-11.1.0` which is `gxx_bin1`, it will goes to check the bin like `x86_64-pc-linux-gnu-gcc`. 

If exists and the version is match, it will use the `x86_64-pc-linux-gnu-gcc` instead.

And it will detect the `x86_64-pc-linux-gnu-g++`, if exists set, otherwise not.

---

For the kits like:

(which is not passed before)

```
x86_64-pc-linux-gnu-gcc
```

It will find the `x86_64-pc-linux-gnu-gcc` first.

As the `x86_64-pc-linux-gnu-g++` does not exist, it will goes to `bin2`'s check.

`bin2` is the same as `bin` which are all `x86_64-pc-linux-gnu-gcc`, hence, only the c compiler will be set.

---

For the kits like:

```
x86_64-pc-linux-gnu-gcc-11.1.0
```

It will find the `x86_64-pc-linux-gnu-gcc-11.1.0` first.

As the `x86_64-pc-linux-gnu-g++-11.1.0` does not exist, it will goes to `bin2`'s check.

`bin2` is `x86_64-pc-linux-gnu-gcc`, which does not exist. It will fall back to `bin` and then only the c compiler will be set.

bobbrow: Sorry if I was unclear.  I was asking if you could put a brief comment in the code about this.  Something like "if bin2 === bin1 because the regex did not make a replacement, then version_match will be true, but no C++ compiler will be found, so we will correctly skip setting the CXX compiler."

Thank you.

fourdim: Sorry for my misunderstanding. I will add comments on this later.

bobbrow: Thank you.

fourdim:
Hi, as I don't have commit access, would you please commit it on behalf of me?

PR #2311, opened 2022-01-03
Launch the target in the default terminal.

michallukowski:
Launch the selected target in the default terminal set in vscode instead of hard coded cmd.exe.


src/cmake-tools.ts
```diff
@@ -2169,11 +2169,6 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
             cwd: (user_config && user_config.cwd) || path.dirname(executable.path)
         };
 
-        if (process.platform === 'win32') {
```
bobbrow, line 2172: If I recall correctly, we had some trouble launching in PowerShell in some cases though I might be thinking about `tasks.json`.

Thank you for your contribution.  We'll need to do some testing to verify this and then we can approve it.

michallukowski: OK. Thanks for explaining. I have checked few terminals. Cmd and bash work fine but can't launching target in PowerShell and zsh. In these terminals invoking the command between double quotation marks causes problem with interpretation.

andreeis: Powershell works for me while testing this PR as long as I changed the below "PowerShell" from "powershell". @michallukowski you can also ensure a case insensitive comparison instead, in case VSCode will set vscode.env.shell to "powershell" or other capital variations in the future, but I don't think it's likely.

andreeis: Another terminal worth to look at and fix is wsl. You need to change for example "c:/folder/project/test.exe" to "/mnt/c/folder/project/test.exe".

PR #2312, opened 2022-01-05
Remove quotes from localizable portions of strings

bobbrow:
We're getting inconsistent localization of quotes across languages. I'm removing them from the localizable portions of the strings to try to reduce the noise.


src/cmake-tools.ts
```diff
@@ -1164,21 +1164,21 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
                 } catch (e: any) {
                     void vscode.window.showErrorMessage(localize('failed.to.create.parent.directory',
-                        'Tried to copy "{0}" to "{1}", but failed to create the parent directory "{2}": {3}',
-                        compdb_path, expanded_dest, pardir, e.toString()));
+                        'Tried to copy {0} to {1}, but failed to create the parent directory {2}: {3}',
+                        `"${compdb_path}"`, `"${expanded_dest}"`, `"${pardir}"`, e.toString()));
                 }
             } else {
-                log.debug(localize('cannot.copy.compile.commands', 'Cannot copy compile_commands.json because it does not exist at {0}', compdb_path));
+                log.debug(localize('cannot.copy.compile.commands', 'Cannot copy {1} because it does not exist at {0}', compdb_path, 'compile_commands.json'));
```
elahehrashedi, line 1181: compdb_path does this need to be changed too?

bobbrow: `comp_db` is already separated out and substituted as part of the `localize` call.  Or am I missing something?  I didn't add quotes here because it didn't already have them.

elahehrashedi (approved):

Issue #2222, opened 2021-11-02
Ninja not able to build single-threaded starting with 1.9.1


andreeis:
The fix for https://github.com/microsoft/vscode-cmake-tools/issues/1958 was meant for msbuild only, not as a general rule to skip over "-j" when parallelJobs is set to 1. This breaks ninja which cannot build single-threaded with CMT extension anymore.

mark-ulrich:
This was my fear when submitting the PR. So what is the best solution? Is it to add a new configuration option? I considered this when creating the fix, but I was/am not familiar enough with the codebase to do so. I would be happy to work on this if no one else is working on it at the moment.

bobbrow:
@mark-ulrich If you are willing to help out with this, we would appreciate it.

bobbrow:
The fix for this issue is available in 1.10 "pre-release".  You can help us validate it by:
1. Opening the extensions panel in VS Code
2. Selecting the CMake Tools extension
3. Clicking the "Switch to Pre-release Version" button

![image](https://user-images.githubusercontent.com/12818240/156218538-7bce4a1f-4fc9-43cc-afb0-7f423aa5980e.png)

Issue #696, opened 2019-05-31
Don't specify number of jobs when building with Ninja


rpavlik:
Or, at least default to not specifying.

https://github.com/vector-of-bool/vscode-cmake-tools/blob/db0ff13ef7bca95b814862b3c2b9fa75ca31d06f/src/driver.ts#L729

I have a RyZen 2700x (8 core, 2thread/core, so 16 logical cores). I use Linux and am frequently close to nearly-out-of-memory. I can use Ninja on its own just fine, but probably at least once a day, if I build within VS Code via this extension, X.org will stop responding. I can sometimes get in via a virtual console or ssh, which usually reveals that something prompted running out of memory (swap is somehow uselessly slow) or something else. I suspect it has to do with the fact that this extension calls `cmake --build` passing `-j 18` on my 16-thread machine.

When building with Ninja, even via `cmake --build`, it automatically defaults to "maximum parallelism", so passing `-j 18` is unnecessary to get parallel builds. Additionally, it's somehow more efficient than make, so specifying the "little bit more than the cores" value that works with make ends up overwhelming systems (I've done this on Windows too, even with more RAM). The line I pointed out above could be modified slightly to just not pass the jobs argument to ninja generators. There's already a setting for build args and build tool args, so users could override (presumably per-workspace) job counts if they wanted something less than max parallelism. (Ninja appears to honor the last `-j` argument on the command line, so I can work around this temporarily by setting `buildToolArgs` to something like `["-j", "14"]` in my user preferences, but that would break as soon as I set build tool args in a workspace for any reason - fortunately unlikely.)

I'm quite certain by now that this is the cause of the issue, as I've been digging into it for months and it always happens when building a CMake project via VSCode and this extension.

Hope this makes sense, let me know if I can help somehow.

slimtom95:
Agreed, and in my case, `-j 8` was passed on my 6 cores VM

MaxCreator:
hey, i was struggling to get make to run with` '-j 1'`. 
I found this issue and i feel like it's somehow related.
can i somehow prevent the auto parallel build with 'vscode-cmake-tools'

i set buildToolArgs to `["-j", "1"]` .. which works but executing a command with `'-- -j 6 -j 1'` doesn't look really nice and for someone unexperienced is diffecult to fix ("These are advanced options and should only be used if you know what you are doing"). 

is there any more elegant way to achive spawning make with `'-j 1'`?

[...]

PR #2320, opened 2022-01-11
Configure number of jobs (-j build args) for Ninja

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/2222 and https://github.com/microsoft/vscode-cmake-tools/issues/696


src/drivers/driver.ts
```diff
@@ -1623,21 +1623,29 @@ export abstract class CMakeDriver implements vscode.Disposable {
             const buildArgs: string[] = this.config.buildArgs.slice(0);
             const buildToolArgs: string[] = ['--'].concat(this.config.buildToolArgs);
 
-            // Only add '-j' argument if parallelJobs > 1
-            if (this.config.numJobs > 1) {
+            const configurationScope = this.workspaceFolder ? vscode.Uri.file(this.workspaceFolder) : null;
+            const parallelJobsSetting = vscode.workspace.getConfiguration("cmake", configurationScope).inspect<number|undefined>('parallelJobs');
+            let numJobs: number | undefined = (parallelJobsSetting?.globalValue || parallelJobsSetting?.workspaceValue || parallelJobsSetting?.workspaceFolderValue);
```
bobbrow, line 1628: This should be done in `config.ts` instead so that any place in the code that needs the setting can have the correct value.

src/config.ts
```diff
@@ -283,7 +286,7 @@ export class ConfigurationReader implements vscode.Disposable {
     get buildToolArgs(): string[] {
         return this.configData.buildToolArgs;
     }
-    get parallelJobs(): number {
+    get parallelJobs(): number | undefined {
         return this.configData.parallelJobs;
```
bobbrow, line 290: `this.configData.parallelJobs` will always be a number.  You should do the inspect of the setting here so that returning `undefined` is possible.

src/drivers/driver.ts
```diff
@@ -1623,21 +1623,29 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+            // for Ninja generator, don't '-j' argument if user didn't define number of jobs
```
bobbrow, line 1629: missing the word "add" after "don't".

andreeis (approved):

Issue #2329, opened 2022-01-16
CMakePresets.json bug with condition object


dslijepcevic:
There seems to be a bug when interpreting a conditional object in VSCode:

**CMakePresets.json**
```json
{
    "version": 3,
    "cmakeMinimumRequired": {
        "major": 3,
        "minor": 22,
        "patch": 0
    },
    "configurePresets": [
        {
            "name": "ninja-cond",
            "displayName": "ninja-cond",
            "description": "Build using Ninja (conditional)",
            "generator": "Ninja",
            "cacheVariables": {
                "FOO_BAR": "$env{FOO_BAR}"
            },
            "condition": {
                "type": "notEquals",
                "lhs": "$env{FOO_BAR}",
                "rhs": ""
            }
        }
    ]
}
```

As expected, `cmake --list-presets` does not list anything, while VSCode improperly interprets the condition, displays the preset as available and runs the configure step:

```
[proc] Executing command: cmake -DFOO_BAR= -H... -B... -G Ninja
```

System info: Visual Studion Code: 1.63.2, CMake Tools v1.9.2, WSL2, Windows 10

bobbrow:
Thank you for reporting this issue. I submitted a PR with the fix.

PR #2336, opened 2022-01-21
Test the existence of a property instead of the value when expanding preset conditions

bobbrow:
Addresses #2329

A bug existed in the code where the value of a property was inadvertently being tested using a falsy bool check instead of testing explicitly for `undefined` or `hasOwnProperty`. This resulted in fields that evaluated to an empty string being overwritten by their original unresolved values by mistake.


src/preset.ts
```diff
@@ -453,7 +453,7 @@ export function setCompilers(_kits: Kit[]) {
 function merge<T extends Object>(target: T, base: T) {
     Object.keys(base).forEach(key => {
         const field = key as keyof T;
-        if (!target[field]) {
+        if (!target.hasOwnProperty(field)) {
```
lygstate, line 456: better to use Object.hasOwn?

bobbrow: Fair point, though I don't think target will be `null`.  I can change it anyway.

bobbrow: Hmm... IntelliSense doesn't offer that API, so it might not be available in VS Code yet. It looks like browsers only just got it a few months ago.

lygstate: Then please use Object.prototype.hasOwnProperty.call  instead

andreeis (approved):

Issue #2289, opened 2021-12-14
Always rebuilds sources with autodetected clang and ninja on linux.


Ivan-Viktorov:
### Brief Issue Summary

On linux scan for kits founds /bin/clang++-X instead of /usr/bin/clang++-X, which breaks ninja (see https://github.com/ninja-build/ninja/issues/1330 https://github.com/ninja-build/ninja/issues/1400).

It would be nice, if scan for kits would prefer /usr/bin/compiler to /bin/compiler. 

Steps to reproduce:
- Get linux os, where /bin is symlink to usr/bin (fedora 35, for example).
- Install clang
- scan for kits
- select clang kit (it would point to /bin/clang*)
- try build a project
- try build project one more time, it will rebuild all sources, even none was modified.

If you change path of clang compiler from /bin/clang to /usr/bin/clang in .local/share/CMakeTools/cmake-tools-kits.json, ninja stops to rebuild all sources and rebuilds only modified, as expected.

### CMake Tools Diagnostics

```
<!-- Paste the `CMake: Log Diagnostics` output HERE -->
```

### Debug Log

```
<!-- Paste the debug log contents HERE -->
```

bobbrow:
Thank you for reporting this. So is the problem just the usage of a symlink?  In other words, if we do a `realpath` on it before setting it in the Kit, will that solve the problem?

Ivan-Viktorov:
>  In other words, if we do a realpath on it before setting it in the Kit, will that solve the problem?

I think yes. This should solve it.

teonik:
I stumbled upon the same problem. Manually changing clang's path in cmake-tools-kits.json, as Ivan-Viktorov suggested, seems to fix this issue.

nidefawl:
I was having the same problem. 

This should be fixed in vscode-cmake-tools. 
ninja, cmake and clang just keep pointing fingers at each other and provide no fix.

bobbrow:
`realpath` is actually not the solution here.  clang++ and clang ultimately point to the same binary and if I `realpath` it, you can't compile C++ anymore without additional switches.  I think the solution that will work better for this is to prioritize the paths we find that are first in the $PATH when we do our deduplication.

bobbrow:
The fix for this issue is available in 1.10 "pre-release".  You can help us validate it by:
1. Opening the extensions panel in VS Code
2. Selecting the CMake Tools extension
3. Clicking the "Switch to Pre-release Version" button

[...]

PR #2352, opened 2022-01-31
Scanning for Kits should use preferred paths to the compilers found

bobbrow:
Addresses: #2289 

Linux systems tend to have multiple layers of symlinks to the actual compiler binaries.  Clang uses the path as a hint to determine which language (C vs C++) to pick so we don't want to resolve symlinks to the actual binary.  Ninja fails to do incremental builds properly if the wrong symlink is used.

So instead of using realpath to resolve the symlink to its true binary location, this change ranks compilers based on their location in the $PATH and should satisfy the demands of both Clang and Ninja.


bobbrow:
This is ready for review now.

elahehrashedi (approved):

PR #2366, opened 2022-02-07
Dev/safreed/presetstoolset

sfreed141:
### This changes VS installation detection logic for CMakePresets

- Update CMakePresets to treat toolset.version to be interpreted as the msvc toolset version, not visual studio version. This is a breaking change but it is the correct interpretation. The latest VS version that matches the requested toolset version and host/target architecture will be used.
- Respect [CMAKE_GENERATOR_INSTANCE](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html) and use that when specified in "cacheVariables" (CMAKE_GENERATOR_INSTANCE only applies when using VS generators).
- Remove the reliance on kits for sourcing development environment with presets (now presets will query vswhere.exe and run vcvars*.bat itself).

## Notes
- If CMAKE_GENERATOR_INSTANCE is set in a toolchain file we will not respect it. This may or may not cause problems but this is the existing behavior anyways. Simple workaround would be to launch vscode from the desired developer prompt.
- If CMAKE_GENERATOR_INSTANCE isn't found we can still try to find a valid VS instance to use, however CMake will likely throw an error if it can't find the instance either. I considered setting this after we find our instance (and overriding any that the user supplies) but I don't think modifying the preset isn't very user friendly (and it wouldn't address the next caveat with toolchain files). If anything we could provide a toast asking the user for input here but I'll leave that for a future PR.
- Might be worthwhile to cache the results from vswhere.exe and vcvars*.bat, but it didn't seem that bad on my machine (with several VS instances installed).
- I'm not sure how cmake.exe is chosen but it doesn't come from the selected instance that presets will use. This is outside the scope of this PR but it might be nice to make that consistent.

### DIFF
--- a/src/installs/visual-studio.ts
+++ b/src/installs/visual-studio.ts
@@ -2,7 +2,13 @@
  * Module for querying MS Visual Studio
  */ /** */
 
+import * as util from '@cmt/util';
 import * as path from 'path';
+import { fs } from '@cmt/pr';
+import paths from '@cmt/paths';
+import { Environment, EnvironmentUtils } from '@cmt/environmentVariables';
+import * as iconv from 'iconv-lite';
+import * as codepages from '@cmt/code-pages';
 
 import * as logging from '../logging';
 import * as proc from '../proc';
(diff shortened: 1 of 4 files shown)


bobbrow:
> * Might be worthwhile to cache the results from vswhere.exe and vcvars*.bat, but it didn't seem that bad on my machine (with several VS instances installed).

We should already be caching the vswhere results.  The cache expires after a certain amount of time, but it's there.

src/installs/visual-studio.ts
```diff
@@ -107,3 +113,383 @@ export function compareVersions(lhs: string, rhs: string): number {
         }
     }
 }
+
```
bobbrow, line 116: Has any of this code that you're moving between files changed?  If so, would it be possible to keep the code in the old location and then do a second PR that moves it?  If that's too much trouble, can you just point me at what changed?

sfreed141: The initial refactor was a direct copy, besides a rename of `kitHostTargetArch` to `getHostTargetArchString`. That was all in the first commit (3a37c61).

The only two notable changes after that were in `varsForVSInstallation`
1. factoring the `getVcVarsBatScript` logic into its own function
2. adding -vcvars_ver to the args

bobbrow: Thanks. That's good to know. I have the correct diff set up in CodeFlow now.

src/installs/visual-studio.ts
```diff
@@ -107,3 +113,383 @@ export function compareVersions(lhs: string, rhs: string): number {
...
+    // instead of hard coding for win32 and x86.
+    // Currently, there is no need of a similar overwrite operation on hostArch,
+    // because CMake host target does not have the same name mismatch with VS.
+    targetArch = targetArchFromGeneratorPlatform(targetArch);
+
+    return (hostArch === targetArch) ? hostArch : `${hostArch}_${targetArch}`;
+}
+
+// Gets the MSVC toolsets installed for a given VS install.
+export async function EnumerateMSVCToolsets(vsInstallRoot: string): Promise<string[] | undefined> {
```
bobbrow, line 179: style:
* lower-case `e` for `enumerate` in the function name.
* `MSVC` -> `Msvc` (we try to do C# rules for acronyms)

src/installs/visual-studio.ts
```diff
@@ -107,3 +113,383 @@ export function compareVersions(lhs: string, rhs: string): number {
...
+        // Only the toolsets should be this directory (each in their own directories), but filter out anything else just in case.
+        // Sort in descending order, so if searching with a 1- or 2-component version (e.g. 14.27) we'll choose the latest version first
+        return dirContents.filter(item => item.isDirectory()).map(dir => dir.name).sort().reverse();
+    }
+
+    return undefined;
+}
+
+// Filters the given vsInstalls to those which have the given toolset.
+export function FilterVSInstallationsByMSVCToolset(vsInstalls: VSInstallation[], toolsetVersion: string): VSInstallation[] {
```
bobbrow, line 192: style:
* lower-case `f` for `filter` in function name
* `MSVC` -> `Msvc`

src/presetsController.ts
```diff
@@ -343,28 +342,22 @@ export class PresetsController {
                         // No selection was made
                         return false;
                     } else {
-                        if (chosen_kit.kit.name === SpecialKits.ScanForKits) {
```
xisui-MSFT, line 346: If we don't reuse this cache anymore, consider remove the "scan for compiler" command (which is an alias of "scan for kits" so removing it won't cause any lost of functions). If you would like to keep it, then I would suggest to hide it in presets mode like "scan for kits".

xisui-MSFT: Other changes don't seem to be consistent with this one...

sfreed141: Can you elaborate on what isn't consistent? I'm not very familiar with what this code is for. I just removed the `if` block here because all it appeared to do was `preset.setCompilers`, which is no longer needed (since I removed `kits` from preset.ts).

sfreed141: Thanks for the explanation, I'll revert the changes (sans removing `preset.setCompilers`)

xisui-MSFT: As Bob said, this is an option used to generate a preset using a selected compiler. By inconsistent, I mean:
- If you have a better way to look for compilers, yes you don't need this if statement, but you'll need to change the way a compiler is selected (chosen_kit) as well.
- If you still want to use kit for compiler selection, I can't see a reason why this if statement needs to be removed.

sfreed141: Gotcha, thanks for the extra info. Generating a preset from a kit still seems useful so I'll keep the if statement.

xisui-MSFT:
LGTM

src/installs/visual-studio.ts
```diff
@@ -107,3 +113,383 @@ export function compareVersions(lhs: string, rhs: string): number {
...
+}
+
+// Gets the MSVC toolsets installed for a given VS install.
+export async function enumerateMsvcToolsets(vsInstallRoot: string): Promise<string[] | undefined> {
+    const toolsetDir = path.join(vsInstallRoot, 'VC\\Tools\\MSVC');
+    if (await fs.exists(toolsetDir)) {
+        const dirContents = await fs.readdir(toolsetDir, { 'withFileTypes': true });
+        // Only the toolsets should be this directory (each in their own directories), but filter out anything else just in case.
+        // Sort in descending order, so if searching with a 1- or 2-component version (e.g. 14.27) we'll choose the latest version first
+        return dirContents.filter(item => item.isDirectory()).map(dir => dir.name).sort().reverse();
```
bobbrow, line 185: Instead of doing a string-based sort for numbers, can you use `util.parseVersion` to parse out the numeric versions and then do the comparison based on that?  Currently, I don't think we have any problems, but this will avoid potential issues around variable digit strings appearing in the wrong place in the sorted list. (e.g. "9" > "12")

Issue #2372, opened 2022-02-11
CI is broken as of 2/3/2022


bobbrow:
GitHub deployed new VM's that are causing a failure in the debugger test for the "successful-build" suite.

The last Ubuntu VM's that worked had version 20220123.1 and the failing VM's have version 20220131.1
macOS fails starting with version 20220201.3 and works with 20220124.1.
windows fails with 20220131.1 and works with 20220123.1.

I have not identified the change the causes the regression, but the place in the test that appears to be failing is:
```
        // Needed to get launch target result
        await new Promise(res => setTimeout(res, 3000));

        // Check that it is compiled as a new file
        expect(fs.existsSync(createdFileOnExecution)).to.be.true;
```
It is unclear whether the timeout is no longer sufficient or being honored.  I will disable this check and add logging around it to unblock PR's and do further investigation.

lygstate:
Does github stop creating files, that's wierd.

This can not mark done, as the issue still not fixed yet

PR #2373, opened 2022-02-11
workaround and logging for broken test

bobbrow:
Addresses #2372


lygstate:
Does github stop creating files, that's wierd.

PR #2374, opened 2022-02-12
Try fixes tests #2378

lygstate:
Addresses: #2378

### DIFF
--- a/test/extension-tests/successful-build/test/debugger.test.ts
+++ b/test/extension-tests/successful-build/test/debugger.test.ts
@@ -4,6 +4,7 @@ import { DefaultEnvironment, expect, getFirstSystemKit } from '@test/util';
 //import sinon = require('sinon');
 import * as fs from 'fs';
 import * as path from 'path';
+import { TerminalOptions } from 'vscode';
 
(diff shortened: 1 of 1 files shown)


src/cmake-tools.ts
```diff
@@ -2182,7 +2182,9 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         }
 
         const quoted = shlex.quote(executable.path);
-        this._launchTerminal.sendText(`${quoted} ${launchArgs}`);
+        const quotedCommand = `${quoted} ${launchArgs}\n`;
+        console.log(`execute command:${quotedCommand}`);
```
bobbrow, line 2186: Maybe also log the `cwd`.  My suspicion is that the file is not being written in the location we expect it to be.  We could also consider doing a recursive search for the file from a higher up directory.

bobbrow:
Crossing my fingers that #2414 fixes this.

bobbrow:
#2414 fixed the problem, so we don't need this PR anymore.  I'll submit a new one to clean up and re-enable the check that was failing.

PR #2377, opened 2022-02-14
Split vscode out of diagnostics parsing utils.

lygstate:
Diagnostics parsing is just parsing the error line number and column number from the error string
There is no need depends on vscode to do that.
So remove the depends on vscode from these error message parsing functions,
So these functions can be tested without vscode(by nodejs directly).
After this, the tests can be running in a faster way and easier debugging.

### DIFF
--- a/src/diagnostics/cmake.ts
+++ b/src/diagnostics/cmake.ts
@@ -5,9 +5,10 @@
 import { Logger } from '@cmt/logging';
 import { OutputConsumer } from '@cmt/proc';
 import * as util from '@cmt/util';
+import { oneLess } from '@cmt/basic/util';
 import * as vscode from 'vscode';
 
-import { FileDiagnostic, oneLess } from './util';
+import { FileDiagnostic } from './util';
 
 /**
  * Class which consumes output from CMake.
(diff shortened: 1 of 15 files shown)


src/diagnostics/cmake.ts
```diff
@@ -5,9 +5,10 @@
...
 import { OutputConsumer } from '@cmt/proc';
 import * as util from '@cmt/util';
+import { oneLess } from '@cmt/basic/util';
 import * as vscode from 'vscode';
 
-import { FileDiagnostic, oneLess } from './util';
+import { FileDiagnostic } from './util';
 
 /**
  * Class which consumes output from CMake.
```
elahehrashedi, line 14: Should these two lines be modified as well or not? 

Line 96                     `const vsdiag = new vscode.Diagnostic(new vscode.Range(lineno, 0, lineno, 9999), full, diagmap[level]);`

Line 157 `const related = new vscode.DiagnosticRelatedInformation(new vscode.Location(fileUri, new vscode.Range(lineNo, 0, lineNo, 999)), "In call to '${command}' here" );`

elahehrashedi (changes requested):

elahehrashedi:
Thank you for your contribution. Unfortunately, this is not the direction we want to take.

Issue #2382, opened 2022-02-15
Incorrect "cmake.additionalKits" setting breaks CMake extension: "command 'cmake.*' not found"


rustyx:
### Brief Issue Summary

After adding any folder<sup>1</sup> to `"cmake.additionalKits"` (e.g. `"cmake.additionalKits": [ "C:\\TEMP" ]`) and restarting VSCode:

`command 'cmake.build' not found`
`command 'cmake.selectConfigurePreset' not found`
`command 'cmake.selectActiveFolder' not found`
Or in other words:
`command 'cmake.*' not found`

### CMake Tools Diagnostics

`command 'cmake.logDiagnostics' not found` 🙂

### Debug Log

```
[main] Safe constructing new CMakeTools instance
[variant] Constructing VariantManager
[main] Constructing new CMakeTools instance
[main] Starting CMakeTools second-phase init
[variant] Loaded new set of variants
[main] CMakeTools instance initialization complete.
[kit] Reading kits file C:\Users\me\AppData\Local\CMakeTools\cmake-tools-kits.json
[kit] Successfully loaded 16 kits from C:\Users\me\AppData\Local\CMakeTools\cmake-tools-kits.json
[kit] Not reading non-existent kits file: c:\Projects\proj1\ctest\.vscode\cmake-kits.json
[kit] Reading kits file C:\TEMP
```

Ext host log:

```
[exthost] [error] Activating extension ms-vscode.cmake-tools failed due to an error:
[exthost] [error] Error: EISDIR: illegal operation on a directory, read
```

---

<sup>1</sup> I know it's wrong to specify a folder in `cmake.additionalKits`, the point of this bug report is that it would be nice if this was reported accordingly instead of silently crashing.

bobbrow:
Thanks for the bug report. I'm testing a fix for this.

PR #2384, opened 2022-02-15
Fix a crash caused by attempting to read a directory

bobbrow:
Addresses: #2382 

This also removes a call to `fs.exists` which has been marked as deprecated.


src/pr.ts
```diff
@@ -23,14 +23,12 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
  */
 export namespace fs {
 
-    export function exists(fspath: string): Promise<boolean> {
-        return new Promise<boolean>((resolve, _reject) => {
-            fs_.exists(fspath, res => resolve(res));
-        });
+    export async function exists(filePath: string): Promise<boolean> {
+        return (await tryStat(filePath))?.isFile() || false;
```
rustyx, line 27: Should `exists()` really check for `isFile()`? And if so, shouldn't `existsSync()` do the same?

bobbrow: No, it shouldn't. That's why the CI failed.  I just hadn't had the time to come back to this and was planning to today.  I wrote the code with a file in mind and that's where the mistake came from.  Thanks for the comment.

Issue #2392, opened 2022-02-22
Modifying CMakeLists.txt, changing/adding new *.cpp/hpp files caused "Configure project: Saving open files: to cycle indefinitely


H-G-Hristov:
This would be difficult to describe and reproduce:

1. Created some source files.
2. Added them to the CMake target
3. Added some file more file names (as the project is multiplatform) to the CMake target.
4. Then added the actual files by copy and pasting e.g FileMac.cpp to FileMac copy.cpp and then renamed it to MacLinux.cpp

I can reconfigure and build the project but this message remains:

![Screen Shot 2022-02-22 at 10 22 22 AM](https://user-images.githubusercontent.com/47526411/155091513-bc5c33a2-be3c-4439-bdf6-618ef2c3b05f.png)

bobbrow:
Unfortunately, we can't repro this right now.  If you have any additional steps you can share to help us out, we can investigate further.

H-G-Hristov:
The information above is all I have for now. It just happened. I guess it is one of these rare deadlocks. If I encounter it again I'll write back.

bobbrow:
I don't think it's a deadlock.  Just an orphaned progress bar.  I submitted a PR which will hopefully catch it.

H-G-Hristov:
I don't have anything particular in mind. I just wanted to share my observations about using the extension on macOS for your consideration. I have no particular suggestion/preference on how to resolve this situation. Perhaps more feedback is needed. 
In my experience the applications on macOS auto-save the changes. It feels like the do it immediately. So I enabled the auto-save feature for consistent user experience. On Windows I don't do it.

PR #2401, opened 2022-02-25
intercept rejected promises during the configure phase

bobbrow:
Possibly address: #2392

This change will ensure that promises are resolved with a success or failure code instead of rejected.


elahehrashedi (approved):
:shipit:

Issue #2232, opened 2021-11-08
Presets in CMakePresets.json should not inherit from presets in CMakeUserPresets.json.


lygstate:
### Brief Issue Summary
Shold warning or erroring when 
Presets in CMakePresets.json inherit from presets in CMakeUserPresets.json.

bobbrow:
Does the extension allow you to use invalid presets like this, or are you just saying you want us to add a warning for it?

lygstate:
> Does the extension allow you to use invalid presets like this, or are you just saying you want us to add a warning for it?

Hi, cmake report error for this, but vscode-cmake-tools didn't report that.

bobbrow:
Interesting. Ok, we'll need to debug it. A few weeks ago I was in the presets code and it looked like the inheritance for CMakePresets shouldn't know about CMakeUsersPresets, but we can look again.

PR #2413, opened 2022-03-03
inherit from user presets

elahehrashedi:
add warning message when Presets in CMakePresets.json inherit from presets in CMakeUserPresets.json
feature request: https://github.com/microsoft/vscode-cmake-tools/issues/2232


src/presetsController.ts
```diff
@@ -180,7 +180,11 @@ export class PresetsController {
         preset.setOriginalUserPresetsFile(folder, presetsFile);
     };
 
-    private async resetPresetsFile(file: string, setPresetsFile: SetPresetsFileFunc, setOriginalPresetsFile: SetPresetsFileFunc, fileExistCallback: (fileExists: boolean) => void) {
+    private async resetUserPresetsFile(file: string, setPresetsFile: SetPresetsFileFunc, setOriginalPresetsFile: SetPresetsFileFunc, fileExistCallback: (fileExists: boolean) => void) {
```
bobbrow, line 183: You don't need the split the function and add a boolean to track this state.  The `setPresetsFile` arg is a function that calls either `_setPresetsFile` or `_setUserPresetsFile`.  You should set the `isUserPreset` fields in that function instead.

bobbrow (changes requested):
see the comment in the PR

src/preset.ts
```diff
@@ -143,8 +144,14 @@ export function evaluateCondition(condition: Condition): boolean {
 }
 
 function evaluateInheritedPresetConditions(preset: Preset, allPresets: Preset[], references: Set<string>): boolean | undefined {
-    const evaluateParent = (parentName: string) => {
+    const evaluateParent = (parentName: string, isChildUserPreset?: boolean) => {
         const parent = getPresetByName(allPresets, parentName);
+        // If the child is not a user preset, the parent should not be a user preset.
+        // eslint-disable-next-line @typescript-eslint/tslint/config
```
bobbrow, line 150: If you do `if (parent && !isChildUserPreset && parent.isUserPreset)`, you should be able to fix this lint warning.

src/preset.ts
```diff
@@ -143,8 +144,14 @@ export function evaluateCondition(condition: Condition): boolean {
 }
 
 function evaluateInheritedPresetConditions(preset: Preset, allPresets: Preset[], references: Set<string>): boolean | undefined {
-    const evaluateParent = (parentName: string) => {
+    const evaluateParent = (parentName: string, isChildUserPreset?: boolean) => {
```
bobbrow, line 147: You shouldn't need to add a variable for this.  You have access to `preset.isUserPreset` in this function which is the same thing.

bobbrow:
Please also update the CHANGELOG.txt before we commit this.

bobbrow (approved):

PR #2422, opened 2022-03-11
Expand ctest arguments

khlebnikov:

- Use `expandString` for arguments passed to ctest


elahehrashedi (approved):
tested it. Seems fine.

Issue #2364, opened 2022-02-07
Kit env variables are not getting updated


elahehrashedi:
When changing the kit, the env variables defined in "`environmentVariables`" and "`environmentSetupScript`" defined in cmake-kits.json are not getting updated.

PR #2434, opened 2022-03-17
disposing the terminal when env variables are getting updated

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/2364


src/cmake-tools.ts
```diff
@@ -982,7 +982,11 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         return false;
     }
 
+    private isKitChanged: boolean = false;
```
bobbrow, line 985: This variable only serves the purpose of letting us know that a new terminal needs to be created so I think the name should encode that purpose instead of encoding what happened.  If the user never launches a terminal, then this will stay true forever and there's no obvious time when it should be changed to false.

Consider something along the lines of `launchTerminalStale` or `refreshLaunchEnvironment` instead.

src/cmake-tools.ts
```diff
@@ -2152,7 +2156,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         }
     });
 
-    private _createTerminal(options: vscode.TerminalOptions, executable: api.ExecutableTarget) {
+    private async createTerminal(executable: api.ExecutableTarget): Promise<vscode.Terminal> {
+        const user_config = this.workspaceContext.config.debugConfig;
+        const drv = await this.getCMakeDriverInstance();
```
bobbrow, line 2161: Did this code need to move? If you are moving it, I recommend putting it closer to where it's needed (line 2190) since an early return on line 2185 could mean that we don't need to initialize it at all.  It also means you don't need to check if `options` is defined or not.

bobbrow (changes requested):

bobbrow (approved):

elahehrashedi:
@bobbrow when creating the launch terminal for the first time, the code didn't set `this.refreshLaunchEnvironment = false;` 
because we didn't enter this for-loop `for (const [, terminal] of this._launchTerminals)` .

bobbrow (approved):

Issue #2288, opened 2021-12-13
Promise rejection: build Error: EMFILE: too many open files


nehaljwani:
On a project with a _huge_ number of build targets and with a capped `ulimit` on `nofile`, the CMake configure stage is marked as failed (although CMake itself succeeds). And on triggering the build step, it fails with:

```
[rollbar] Unhandled exception: Unhandled Promise rejection:
build Error: EMFILE: too many open files,
open '/path/to/builddir/.cmake/api/v1/reply/target-<redacted>o-Release<hash>.json' {}
```

The problem is that the following snippet:
https://github.com/microsoft/vscode-cmake-tools/blob/d2d98f80067738ae338c6e70dba7689012cd2e42/src/drivers/cmakefileapi/api_helpers.ts#L159-L161

Calls the following for each file with no consideration to limits and no retries:
https://github.com/microsoft/vscode-cmake-tools/blob/d2d98f80067738ae338c6e70dba7689012cd2e42/src/drivers/cmakefileapi/api_helpers.ts#L128

Yes, this problem can be worked around with increasing the `sysctl` for `fs.inotify.max_user_watches` or updating `/etc/security/limits.conf`, but that assumes that every developer has admin access or can get an admin to make this change for them, which is not always possible.

I found several discussions at https://stackoverflow.com/q/8965606/1005215 , perhaps this project can adopt one of them?

bobbrow:
Thanks for reporting this.  We'll see what we can do, but will also accept a PR if someone who has a repro wants to help out.

bobbrow:
This was fixed yesterday and is available in today's pre-release build, version 1.11.2.

PR #2439, opened 2022-03-18
Introduce FS access limiter for cmake api helpers

FrogTheFrog:
### This changes how the cmake api driver helpers use the async fs functions

- Add a limiter-like function to reduce the concurrent interaction with the file system.
- Set the said limit to maximum **50** concurrent read/writes or other access.

## Other Notes/Information

Might have went overboard with the function usage as it is required in place at the moment. However, for completeness sake added the it everywhere.

The project I'm working has ~5500 api files. I chose the value to be a third of the default initial per-configured limit on my system (which was around ~150 if I remember correctly).

### DIFF
--- a/src/drivers/cmakefileapi/api_helpers.ts
+++ b/src/drivers/cmakefileapi/api_helpers.ts
@@ -14,19 +14,23 @@ import { fs } from '@cmt/pr';
 import * as path from 'path';
 import * as nls from 'vscode-nls';
 import rollbar from '@cmt/rollbar';
+import * as pLimit from 'p-limit';
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 
 const log = logging.createLogger('cmakefileapi-parser');
 
+// Limits the concurrent async access to the file system to avoid errors such as "EMFILE: too many open files".
+const fsAccessLimiter = pLimit(50);
+
 export async function createQueryFileForApi(api_path: string): Promise<string> {
     const query_path = path.join(api_path, 'query', 'client-vscode');
     const query_file_path = path.join(query_path, 'query.json');
     const requests = { requests: [{ kind: 'cache', version: 2 }, { kind: 'codemodel', version: 2 }, { kind: 'toolchains', version: 1 }] };
     try {
-        await fs.mkdir_p(query_path);
-        await fs.writeFile(query_file_path, JSON.stringify(requests));
+        await fsAccessLimiter(() => fs.mkdir_p(query_path));
+        await fsAccessLimiter(() => fs.writeFile(query_file_path, JSON.stringify(requests)));
     } catch (e) {
         rollbar.exception(localize('failed.writing.to.file', 'Failed writing to file {0}', query_file_path), e);
         throw e;
@@ -36,11 +40,11 @@ export async function createQueryFileForApi(api_path: string): Promise<string> {
 
 export async function loadIndexFile(reply_path: string): Promise<index_api.Index.IndexFile | null> {
     log.debug(`Read reply folder: ${reply_path}`);
-    if (!await fs.exists(reply_path)) {
+    if (!await fsAccessLimiter(() => fs.exists(reply_path))) {
         return null;
     }
 
-    const files = await fs.readdir(reply_path);
+    const files = await fsAccessLimiter(() => fs.readdir(reply_path));
     log.debug(`Found index files: ${JSON.stringify(files)}`);
 
     const index_files = files.filter(filename => filename.startsWith('index-')).sort();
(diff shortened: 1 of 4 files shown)


src/drivers/cmakefileapi/api_helpers.ts
```diff
@@ -14,19 +14,23 @@ import { fs } from '@cmt/pr';
...
+const fsAccessLimiter = pLimit(50);
+
 export async function createQueryFileForApi(api_path: string): Promise<string> {
     const query_path = path.join(api_path, 'query', 'client-vscode');
     const query_file_path = path.join(query_path, 'query.json');
     const requests = { requests: [{ kind: 'cache', version: 2 }, { kind: 'codemodel', version: 2 }, { kind: 'toolchains', version: 1 }] };
     try {
-        await fs.mkdir_p(query_path);
-        await fs.writeFile(query_file_path, JSON.stringify(requests));
+        await fsAccessLimiter(() => fs.mkdir_p(query_path));
```
bobbrow, line 32: Since `fs` is actually our wrapper around NodeJS's `fs` would it make more sense to make this change in `pr.ts` instead?  Then all callers of the filesystem APIs will get this benefit and we won't have to remember to import this module everywhere.

bobbrow (changes requested):

src/drivers/cmakefileapi/api_helpers.ts
FrogTheFrog: If that is what you want, I can apply it to `pr.ts`. Does 50 sound like a sensible value to you or should we increase it?

bobbrow: My biggest concern is around `readdir`.  I'm not sure we need to limit file/directory enumeration, isn't it just file reads that are the problem?  Or does the OS hold on to file descriptors for the other operations too?

FrogTheFrog: But then again, in https://www.npmjs.com/package/rimraf docs, it seems to mention that `readdir` uses file handles, so I don't know what to think anymore.

FrogTheFrog: @bobbrow 

I did more extensive analysis for linux side at least:
* `fs.exists` - uses `fs.tryStat` - does not open FD
* `fs.existsSync` - not async, do not care
* `fs.readFile` - opens FD via `uv_fs_open`
* `fs.writeFile` - opens FD via `uv_fs_open`
* `fs.readdir` - uses `uv_fs_scandir` (yes, scandir) - does not open FD
* `fs.mkdir` - uses `uv_fs_mkdir` - does not open FD
* `fs.mkdtemp` - uses `uv_fs_mkdtemp` - does not open FD
* `fs.rename` - uses `uv_fs_rename` - does not open FD
* `fs.stat` - uses `uv_fs_stat` - does not open FD
* `fs.walk` - uses `fs.stat` and `fs.readdir` - does not open FD
* `fs.tryStat` - uses `fs.stat` - does not open FD
* `fs.readlink` - uses `uv_fs_readlink` - does not open FD
* `fs.unlink` - uses `uv_fs_unlink` - does not open FD
* `fs.appendFile` - uses `fs.writeFile` - opens FD via `uv_fs_open`
* `fs.mkdir_p` - uses `fs.exists`, `fs.stat` and `fs.mkdir` - does not open FD
* `fs.copyFile` - opens FD via `uv_fs_open`
* `fs.rmdir` - found no evidence that FD is opened, but the docs claim that it does it. At least it is handled by rimraf then, no need for a limit.

I will update the PR to limit only the functions which I'm 100% opens the FD.

bobbrow (approved):

src/drivers/cmakefileapi/api_helpers.ts
bobbrow: Awesome analysis. Thank you for being so thorough!

FrogTheFrog: Would you like me to squish the commits?

bobbrow: That won't be necessary.  We will do a squash merge.  Thanks for asking though.

bobbrow:
Hey @FrogTheFrog sorry to ask one more thing, but could you also add the following near the top of `CHANGELOG.md` as part of this?  We're trying to keep it up to date as commits go in.

```
## 1.11
Improvements:
- Fix build Error: EMFILE: too many open files. [#2288](https://github.com/microsoft/vscode-cmake-tools/issues/2288) [@FrogTheFrog](https://github.com/FrogTheFrog)
```

FrogTheFrog:
> Hey @FrogTheFrog sorry to ask one more thing, but could you also add the following near the top of `CHANGELOG.md` as part of this? We're trying to keep it up to date as commits go in.
> 
> ```
> ## 1.11
> Improvements:
> - Fix build Error: EMFILE: too many open files. [#2288](https://github.com/microsoft/vscode-cmake-tools/issues/2288) [@FrogTheFrog](https://github.com/FrogTheFrog)
> ```

Done

Issue #2389, opened 2022-02-18
formatting: update variables to follow new style


bobbrow:
We want to change the variable name styles from `variable_name` to `variableName`.  (remove underscores).

PR #2443, opened 2022-03-18
format cmake-tools.ts

elahehrashedi:
related to: https://github.com/microsoft/vscode-cmake-tools/issues/2389


src/cmakeTools.ts
```diff
@@ -4,7 +4,7 @@
 import { CMakeCache } from '@cmt/cache';
 import { CMakeExecutable, getCMakeExecutableInformation } from '@cmt/cmake/cmake-executable';
 import { CompilationDatabase } from '@cmt/compdb';
-import * as debugger_mod from '@cmt/debugger';
+import * as debuggerMode from '@cmt/debugger';
```
bobbrow, line 7: I think the `mod` is short for `module`, not `mode`.  Can it be removed?  Just call it `debugger`?

src/cmakeTools.ts
```diff
@@ -208,12 +199,12 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
             const expandedConfigurePreset = await preset.expandConfigurePreset(this.folder.uri.fsPath,
                 configurePreset,
                 lightNormalizePath(this.folder.uri.fsPath || '.'),
-                this.sourceDir,
+                this.srcDir,
```
bobbrow, line 202: If we have to rename this one, I'd like to stop abbreviating things.  `sourceDirectory` is better.

src/cmakeTools.ts
```diff
@@ -416,13 +404,13 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
     /**
      * Whether CTest is enabled
      */
-    get ctestEnabled() {
-        return this._ctestEnabled.value;
+    get cTestEnabled() {
```
bobbrow, line 407: I would rather it be `ctest` instead of `cTest`.  We do `cmake` instead of `cMake` everywhere.

src/cmakeTools.ts
```diff
@@ -1123,22 +1108,19 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
     static async createForDirectory(folder: vscode.WorkspaceFolder, ext: vscode.ExtensionContext): Promise<CMakeTools> {
         // Create a context for the directory
-        const dir_ctx = DirectoryContext.createForDirectory(folder, new StateManager(ext, folder));
-        return CMakeTools.create(ext, dir_ctx);
+        const dirContext = DirectoryContext.createForDirectory(folder, new StateManager(ext, folder));
+        return CMakeTools.create(ext, dirContext);
     }
 
-    private _activeKit: Kit | null = null;
-    get activeKit(): Kit | null {
```
bobbrow, line 1131: It's ok to keep this `get`.  The cases we wanted to remove are when there's a trivial `get` and `set`.

bobbrow: I don't know if we want to allow other modules to set the activeKit

src/cmakeTools.ts
```diff
@@ -1149,70 +1131,70 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+            if (await fs.exists(compdbPath)) {
+                compdbPaths.push(compdbPath);
                 if (this.workspaceContext.config.copyCompileCommands) {
                 // Now try to copy the compdb to the user-requested path
-                    const copy_dest = this.workspaceContext.config.copyCompileCommands;
-                    const expanded_dest = await expandString(copy_dest, opts);
-                    const pardir = path.dirname(expanded_dest);
+                    const copyDest = this.workspaceContext.config.copyCompileCommands;
+                    const expandedDest = await expandString(copyDest, opts);
+                    const pardir = path.dirname(expandedDest);
```
bobbrow, line 1147: Can you rename this to `parentDir`?

src/cmakeTools.ts
```diff
@@ -1252,58 +1234,58 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
-                    return this._doConfigure(type, progress, async consumer => {
-                        const IS_CONFIGURING_KEY = 'cmake:isConfiguring';
+                    return this.doConfigure(type, progress, async consumer => {
+                        const isConfiguringKey = 'cmake:isConfiguring';
                         if (drv) {
-                            let old_prog = 0;
-                            const prog_sub = drv.onProgress(pr => {
-                                const new_prog = 100 * (pr.progressCurrent - pr.progressMinimum) / (pr.progressMaximum - pr.progressMinimum);
-                                const increment = new_prog - old_prog;
+                            let oldProg = 0;
```
bobbrow, line 1244: Can we change `prog` -> `progress` so it doesn't get confused with `program`?

src/cmakeTools.ts
```diff
@@ -1470,22 +1452,22 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         return drv ? this.tasksBuildCommandDrv(drv) : null;
     }
 
-    private m_promise_build: Promise<number> = Promise.resolve(0);
+    private mPromiseBuild: Promise<number> = Promise.resolve(0);
```
bobbrow, line 1455: remove the `m`. It looks like C++ naming convention left over.  Might be better to call it `activeBuild` if we want to remove the type name 'promise' from the variable (generally desirable).

src/cmakeTools.ts
```diff
@@ -1495,79 +1477,79 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
             return 0;
         }
 
-        const config_retc = await this.ensureConfigured();
-        if (config_retc === null) {
+        const configRetc = await this.ensureConfigured();
```
bobbrow, line 1480: Let's replace `retc` (return code) with `result`.

src/cmakeTools.ts
```diff
@@ -1495,79 +1477,79 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
-            return config_retc;
+        } else if (configRetc !== 0) {
+            return configRetc;
         }
         drv = await this.getCMakeDriverInstance();
         if (!drv) {
             throw new Error(localize('driver.died.after.successful.configure', 'CMake driver died immediately after successful configure'));
         }
-        let targets = targets_;
+        let newTargets = targets;
```
bobbrow, line 1490: This gets immediately reassigned, so it's better to initialize it to undefined, I think.

bobbrow: Can you make a note and we'll fix it in the next PR since this one is just formatting?

src/cmakeTools.ts
```diff
@@ -1577,15 +1559,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
      * @param filePath The path to a file to try and compile
      */
     async tryCompileFile(filePath: string): Promise<vscode.Terminal | null> {
-        const config_retc = await this.ensureConfigured();
-        if (config_retc === null || config_retc !== 0) {
+        const configRetc = await this.ensureConfigured();
```
bobbrow, line 1562: `retc` -> `result`

src/cmakeTools.ts
```diff
@@ -1604,10 +1586,10 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         }
 
         if (!await fs.exists(drv.cachePath)) {
-            const do_conf = !!(await vscode.window.showErrorMessage(
+            const doConf = !!(await vscode.window.showErrorMessage(
```
bobbrow, line 1589: `doConfigure`

src/cmakeTools.ts
```diff
@@ -1706,27 +1688,27 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
      * Implementation of `cmake.cleanRebuild`
      */
     async cleanRebuild(): Promise<number> {
-        const clean_res = await this.clean();
-        if (clean_res !== 0) {
-            return clean_res;
+        const cleanRes = await this.clean();
```
bobbrow, line 1691: `res` -> `result`

src/cmakeTools.ts
```diff
@@ -1706,27 +1688,27 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
 
-    private readonly _ctestController = new CTestDriver(this.workspaceContext);
+    private readonly cTestController = new CTestDriver(this.workspaceContext);
     async ctest(): Promise<number> {
 
-        const build_retc = await this.build();
-        if (build_retc !== 0) {
-            this._ctestController.markAllCurrentTestsAsNotRun();
-            return build_retc;
+        const buildRetc = await this.build();
```
bobbrow, line 1701: `retc` -> `result`

src/cmakeTools.ts
```diff
@@ -2023,8 +2005,8 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
 
         const buildOnLaunch = this.workspaceContext.config.buildBeforeRun;
         if (buildOnLaunch || isReconfigurationNeeded) {
-            const rc_build = await this.build([chosen.name]);
-            if (rc_build !== 0) {
+            const rcBuild = await this.build([chosen.name]);
```
bobbrow, line 2008: `rc` (result code) -> `result`.  More specifically, call this one `buildResult` to match the naming of the others.

src/cmakeTools.ts
```diff
@@ -2333,11 +2315,13 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         await this.workspaceContext.state.reset();
     }
 
-    private _sourceDir = '';
+    // Don't get this from the driver. Source dir is required to evaluate presets.
+    // Presets contain generator info. Generator info is required for server api.
```
bobbrow, line 2319: This comment appears twice. I don't think we need both copies.

src/cmakeTools.ts
```diff
@@ -2333,11 +2315,13 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
+    private srcDir = '';
```
bobbrow, line 2320: Because we only have a `get` here, it's ok to leave this one unchanged as `_sourceDir`

src/cmakeTools.ts
```diff
@@ -130,24 +130,24 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
...
-                this._targetName.set('all');
+                this.targetName.set('all');
             }
         }
-        if (!useCMakePresets && this.targetName === this._targetsInPresetName) {
-            this._targetName.set('all');
+        if (!useCMakePresets && this.targetName.value === this.targetsInPresetName) {
+            this.targetName.set('all');
         }
         const oldValue = this._useCMakePresets;
```
bobbrow, line 143: I noticed you changed `this.useCMakePresets` -> `this._useCMakePresets` everywhere.  Are we doing that with other `get`s that have a `_` variable as the backing storage?  After this formatting change we may want to go back and double check for consistency.

bobbrow (changes requested):

src/cmakeTools.ts
elahehrashedi: debugger was a reserved word, couldn't use it.

src/cmakeTools.ts
elahehrashedi: I had to use sourceDir as it was in the API,  \src\api.ts
I assumed we don't change the API.
but I renamed the private one to _sourceDir to match the pattern

src/cmakeTools.ts
elahehrashedi: no we are not doing it with other gets, I fixed this. I have to make sure it didnt happen on other places.

bobbrow:
>         const user_config = this.workspaceContext.config.debugConfig;

This got reverted from the PR just merged in.

---
Refers to: src/cmakeTools.ts:2196 in 620439b. [](commit_id = 620439b94a1661128f70dcf8793951d3efe066c7, deletion_comment = False)

src/cmakeTools.ts
```diff
@@ -2182,15 +2167,15 @@ export class CMakeTools implements vscode.Disposable, api.CMakeToolsAPI {
         }
         const user_config = this.workspaceContext.config.debugConfig;
```
bobbrow, line 2168: >user_config [](http://example.com/codeflow?start=14&length=11)

update this one too

bobbrow (approved):

Issue #2122, opened 2021-09-13
Add support for targets in the CMake task provider


bobbrow:
To allow folks to create specific tasks for a set of targets, we should add a `targets` property to the task provider schema.  It would accept a list of targets that CMake should build as part of the task.

PR #2456, opened 2022-03-24
Add support for targets in the CMake task provider

elahehrashedi:
feature request: https://github.com/microsoft/vscode-cmake-tools/issues/2122


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 1.10.5
 Bug Fixes:
 - fix "CMake: compile active file" command. [#2438](https://github.com/microsoft/vscode-cmake-tools/issues/2438)
+- Add support for multiple targets in the CMake task provider. [#2122](https://github.com/microsoft/vscode-cmake-tools/issues/2122)
```
bobbrow, line 6: Let's move this to 1.11

src/cmakeTaskProvider.ts
```diff
@@ -39,23 +46,30 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
         const result: CMakeTask[] = [];
-        const taskName: string = "CMake: build";
+        this.updateDefaultTargets();
+        // Provide build task.
+        const taskName: string = CommandType.build;
         const definition: CMakeTaskDefinition = {
             type: CMakeTaskProvider.CMakeScriptType,
-            label: taskName,
-            command: "build"
+            label: CMakeTaskProvider.CMakeSourceStr + ": " + taskName,
```
bobbrow, line 62: I don't think this will have the necessary localization.  It looks like you only provide the build task now, but what about the others?  If so, we might want to use the same strings from package.nls.json.  I don't think the `:` separator is needed if we reuse those.

bobbrow: (and by "reuse", I mean copy the strings in here. We can't use those localizations directly)

package.json
```diff
@@ -537,6 +537,12 @@
             ],
```
bobbrow, line 537: I think we need to add the other commands to this enum.

bobbrow: (or are you doing that in a separate PR?)

src/cmakeTaskProvider.ts
```diff
@@ -39,23 +46,30 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
         };
-        const task = new vscode.Task(definition, vscode.TaskScope.Workspace, taskName, CMakeTaskProvider.CMakeSourceStr);
+        const task = new vscode.Task(definition, vscode.TaskScope.Workspace, taskName, CMakeTaskProvider.CMakeSourceStr,
+            new vscode.CustomExecution(async (resolvedDefinition: vscode.TaskDefinition): Promise<vscode.Pseudoterminal> =>
+                // When the task is executed, this callback will run. Here, we setup for running the task.
+                new CustomBuildTaskTerminal(resolvedDefinition.command, this.defaultTargets, resolvedDefinition.targets, resolvedDefinition.options, this.cmakeDriver)
+            ), []);
         task.group = vscode.TaskGroup.Build;
-        task.detail = "CMake template build task";
+        task.detail = localize('cmake.template.task', 'CMake template {0} task', CommandType.build);
```
bobbrow, line 72: `CommandType.build` needs to be localized

bobbrow (changes requested):

package.json
elahehrashedi: As they are not ready for users to use, I didn't add them here, I only added them to the code for our future development.
But I added them after your comment. Let me know if you want them to stay.

src/cmakeTaskProvider.ts
elahehrashedi: the reason I used the "Cmake:" at the beginning of the name is to be consistent with what other extensions are doing so that the user will know which extension is providing this task.
image
For example this is how the user will see the task in the lists of the provided tasks.

elahehrashedi: I will fix the localization though.

package.json
bobbrow: We can have them in the code, but not `package.json` or else people will get IntelliSense for a feature that is not available yet.

src/cmakeTaskProvider.ts
```diff
@@ -16,8 +16,19 @@ interface CMakeTaskDefinition extends vscode.TaskDefinition {
...
 }
 
+enum CommandType {
+    build = "build",
+    install = "install",
+    test = "test",
+    config = "configure"
+}
+
+const localizeCommandType = (cmd: CommandType): string => (cmd === CommandType.build) ? localize("build", "build") :
```
bobbrow, line 29: This would look better as a `switch (cmd) { ... }`

bobbrow (changes requested):

package.json
elahehrashedi: So I Will revert this change.

src/cmakeTaskProvider.ts
```diff
@@ -39,23 +66,30 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
             this.cmakeDriver ? [this.cmakeDriver.allTargetName] : [CMakeTaskProvider.allTargetName];
     }
 
     public async provideTasks(): Promise<CMakeTask[]> {
         // Create a CMake build task
         const result: CMakeTask[] = [];
-        const taskName: string = "CMake: build";
+        this.updateDefaultTargets();
+        // Provide build task.
+        const taskName: string = localizeCommandType(CommandType.build);
```
bobbrow, line 79: Conisder renaming `taskName` to `commandType` since it is not the entire task name (e.g. label)

src/cmakeTaskProvider.ts
```diff
@@ -39,23 +66,30 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
         const definition: CMakeTaskDefinition = {
             type: CMakeTaskProvider.CMakeScriptType,
-            label: taskName,
-            command: "build"
+            label: CMakeTaskProvider.CMakeSourceStr + ": " + taskName,
+            command: CommandType.build,
+            targets: this.defaultTargets
         };
-        const task = new vscode.Task(definition, vscode.TaskScope.Workspace, taskName, CMakeTaskProvider.CMakeSourceStr);
+        const task = new vscode.Task(definition, vscode.TaskScope.Workspace, taskName, CMakeTaskProvider.CMakeSourceStr,
```
bobbrow, line 86: is `taskName` here supposed to be `definition.label`?  Or do they rebuild the label with `taskName` and `CMakeSourceStr`?

PR #2476, opened 2022-04-03
Add c++23 standard

sweemer:
### This changes the handling of the C++ 23 standard

- Add `c++23` and `gnu++23` standard versions
- Bump `/std=c++latest` to C++ 23
- Reassign `c++2b` and `gnu++2b` to C++ 23

## Other Notes/Information

This is my first pull request for this repository, and I'm new to the code, so please go easy on me.

According to the [Visual Studio documentation](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170), C++23 is enabled with `/std=c++latest` for versions 17.0 and up, but how does this extension know which version of Visual Studio is installed? I would think that there should be a way to enforce consistency between the standard version associated with associated `/std=c++latest` and the version of Visual Studio installed - is there?


sweemer:
Thanks for the info - I have submitted a [pull request](https://github.com/microsoft/vscode-cpptools-api/pull/41) to the vscode-cpptools-api repository to add these enums.

src/cpptools.ts
```diff
@@ -65,7 +65,9 @@ interface TargetDefaults {
 
 function parseCppStandard(std: string, can_use_gnu: boolean): StandardVersion {
     const is_gnu = can_use_gnu && std.startsWith('gnu');
-    if (std.endsWith('++2a') || std.endsWith('++2b') || std.endsWith('++20') || std.endsWith('++latest')) {
+    if (std.endsWith('++23') || std.endsWith('++2b') || std.endsWith('++latest')) {
```
bobbrow, line 68: After updating `package.json` to get version `6.1.0` of the API, we should only allow c++23 if the API verison is v6 or higher.  There is a boolean parameter to this function `can_use_gnu` that is based on a similar version check.  Perhaps we should just pass the version into this function and create the local `can_use_gnu` variable here instead of in `parseCompileFlags`.

sweemer: I added the check for v6 or higher.

I think it's better to keep `can_use_gnu` inside of `parseCompileFlags` because it is also passed into `parseCStandard` and we wouldn't want to duplicate the logic in both functions.

Feel free to push changes directly to my branch if that's easier for you. Or I can make changes based on your suggestions - either way is fine with me.

bobbrow: This is fine.  I hadn't looked closely at the other function before making the suggestion.

bobbrow (approved):

src/cpptools.ts
bobbrow: Oh, it looks like `package.json` was not updated yet.  I'll make that change real quick.

sweemer: Thanks for all your help!

bobbrow: Thanks for taking the initiative to start this.

Issue #2477, opened 2022-04-04
Fix Partial Activation CMake VSCode Experiment Telemetry Issues


sinemakinci1:
Telemetry Issues with VSCode CMake Partial Activation Experiment:

We need to rename the Boolean for “ignoreExperiment” and add it to the control group for those who don't have CMakeLists. I know that it not firing for the control group by design since the Boolean is named “ignoreExperiment”, but after our conversation we want it to fire for both to ensure both groups add up to 100%. Essentially, we want a Boolean to determine if there is no cmake lists in the project whatsoever, so “ignoreExperiment” might be misleading since the control group will also have a small percentage of users without CMakeLists in the project and they are already ignoring the experiment.

Also, those who press 'esc' Key in control group prompt needs to fire partial telemetry, currently is only firing:

![image](https://user-images.githubusercontent.com/12818240/161611386-ee8b6f40-9e84-407c-81e1-4692259b784d.png)

PR #2494, opened 2022-04-11
Add telemetry

elahehrashedi:
fix for: https://github.com/microsoft/vscode-cmake-tools/issues/2477
new telemetry properties:

showCMakeListsExperiment: yes (experiment group), no(control group) 
hasCmakeLists: yes/no

### DIFF
--- a/src/cmakeTools.ts
+++ b/src/cmakeTools.ts
@@ -39,7 +39,7 @@ import { CMakeFileApiDriver } from '@cmt/drivers/cmakeFileApiDriver';
 import * as nls from 'vscode-nls';
 import { CMakeToolsFolder } from './folders';
 import { ConfigurationWebview } from './cache-view';
-import { updateFullFeatureSetForFolder, updateCMakeDriverInTaskProvider, enableFullFeatureSet, isActiveFolder, updateDefaultTargetsInTaskProvider, expShowCMakeLists } from './extension';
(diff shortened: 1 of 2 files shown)


src/cmakeTools.ts
```diff
@@ -574,11 +574,15 @@ export class CMakeTools implements api.CMakeToolsAPI {
...
 
-                    let showCMakeLists: boolean = await expShowCMakeLists();
+                    let showCMakeLists: boolean = await showCMakeListsExperiment();
                     const existingCmakeListsFiles: string[] | undefined = await util.getAllCMakeListsPaths(this.folder.uri);
-                    if (showCMakeLists) {
-                        showCMakeLists = existingCmakeListsFiles !== undefined && existingCmakeListsFiles.length > 0;
-                        telemetryProperties["ignoreExperiment"] = (!showCMakeLists).toString();
+
+                    telemetryProperties["showCMakeListsExperiment"] = (showCMakeLists).toString();
+                    if (existingCmakeListsFiles !== undefined && existingCmakeListsFiles.length > 0) {
```
bobbrow, line 581: A little bit simpler it can be done like this:
```ts
if (existingCmakeListsFiles === undefined || existingCmakeListsFiles.length == 0) {
    showCMakeLists = false;
}
telemetryProperties["hasCMakeLists"] = showCMakeLists.toString();
```

bobbrow (approved):

src/cmakeTools.ts
elahehrashedi: @bobbrow showCMakeLists is always false for the control group, no matter whether there is a cmakelists.txt available or not.

bobbrow: got it

Issue #2501, opened 2022-04-12
Warn about incorrect usage of $env{}


merlynomsft:
### Brief Issue Summary

I'm attempting to use prefix to append to the path

1. Create the following files in a folder and open in VS Code with vscode-cmake-tools

Appears in output:
[expand] Reached max string expansion recursion. Possible circular reference.

CMakePresets.json:
```
{
  "version": 3,
  "configurePresets": [
    {
      "name": "linux-debug",
      "displayName": "Linux Debug",
      "generator": "Ninja",
      "binaryDir": "${sourceDir}/out/build/${presetName}",
      "installDir": "${sourceDir}/out/install/${presetName}",
      "cacheVariables": {
        "CMAKE_BUILD_TYPE": "Debug"
      },
      "environment": {
        "PATH":  "$env{PATH}:/usr"
      },      
      "condition": {
        "type": "equals",
        "lhs": "${hostSystemName}",
        "rhs": "Linux"
      },
      "vendor": {
        "microsoft.com/VisualStudioRemoteSettings/CMake/1.0": {
          "sourceDir": "$env{HOME}/.vs/$ms{projectDirName}"
        }
      }
    }
  ]
}

```

CMakeLists.txt:
```
# CMakeList.txt : Top-level CMake project file, do global configuration
# and include sub-projects here.
#
cmake_minimum_required (VERSION 3.8)

project ("CMakeProject1")

# Include sub-projects.
add_subdirectory ("CMakeProject1")
```

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.66.1",
  "cmtVersion": "1.10.5",
  "configurations": [
    {
      "folder": "/mnt/c/workspace/CMakeProject1",
      "cmakeVersion": "3.16.3",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
[...]

PR #2503, opened 2022-04-14
avoid recursive expansion of env variables

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/2501


src/expand.ts
```diff
@@ -169,7 +169,14 @@ async function expandStringHelper(tmpl: string, opts: ExpansionOptions) {
...
         const full = mat[0];
         const varname = mat[1];
-        const repl = fixPaths(env[varname]) || '';
+        const repl: string = fixPaths(env[varname]) || '';
+        // Avoid replacing an env variable by itself, e.g. PATH:env{PATH}.
+        const env_re4 = RegExp(`\\$env\\{(${varValueRegexp})\\}`, "g");
+        mat = env_re4.exec(repl);
+        const varnameRepl = mat ? mat[1] : undefined;
+        if (varnameRepl && varnameRepl === varname) {
+            break;
```
bobbrow, line 178: I think we just want to log here.  We already have a failsafe that stops the recursion, just make sure we don't log the same warning 10 times.

bobbrow (changes requested):

src/expand.ts
elahehrashedi: @bobbrow we try to expand this 3 times, in `expandConfigurePreset` and `expandBuildPreset`, so although we won't see the error 10 times, but we will see it 3 times. IS that fien?

bobbrow: That's ok.

src/expand.ts
```diff
@@ -103,17 +103,21 @@ export async function expandString<T>(tmpl: string | T, opts: ExpansionOptions):
...
         result = expansion.result;
         didReplacement = expansion.didReplacement;
+        circularReference = expansion.circularReference;
         i++;
-    } while (i < MAX_RECURSION && opts.recursive && didReplacement);
+    } while (i < MAX_RECURSION && opts.recursive && didReplacement && !circularReference);
 
-    if (i === MAX_RECURSION) {
+    if (circularReference) {
+        log.warning(localize('circular.variable.reference', 'Circular variable reference found: {0}', circularReference));
```
bobbrow, line 119: This gives the string with the reference, like "${PATH}", but not the variable name that went with it.  Can we save `full` and `varname` down by line 183?

bobbrow (changes requested):

bobbrow (approved):

PR #2523, opened 2022-04-25
Also add a build preset when adding a multi-config configure preset

bobbrow:
Improving the experience for adding configure presets.  For multi-config generators, we were adding `CMAKE_BUILD_TYPE` which is ineffective.  Multi-config generators need a build preset instead.

Also, explicitly add 'cl.exe' as the `CMAKE_[C|CXX]_COMPILER` when creating a config based off a Visual Studio kit.

### DIFF
--- a/src/presetsController.ts
+++ b/src/presetsController.ts
@@ -266,6 +266,7 @@ export class PresetsController {
             return false;
         } else {
             let newPreset: preset.ConfigurePreset | undefined;
+            let isMultiConfigGenerator: boolean = false;
             switch (chosenItem.name) {
                 case SpecialOptions.CreateFromCompilers: {
                     // Check that we have kits
@@ -284,19 +285,19 @@ export class PresetsController {
                         if (kit.toolchainFile || kit.name === SpecialKits.Unspecified) {
                             continue;
                         }
-                        let found = false;
+                        let duplicate = false;
                         if (kit.visualStudio && !kit.compilers) {
                             for (const filteredKit of filteredKits) {
                                 if (filteredKit.preferredGenerator?.name === kit.preferredGenerator?.name &&
                                     filteredKit.preferredGenerator?.platform === kit.preferredGenerator?.platform &&
                                     filteredKit.preferredGenerator?.toolset === kit.preferredGenerator?.toolset) {
                                     // Found same generator in the filtered list
-                                    found = true;
+                                    duplicate = true;
                                     break;
                                 }
                             }
                         }
-                        if (!found) {
+                        if (!duplicate) {
                             filteredKits.push(kit);
                         }
                     }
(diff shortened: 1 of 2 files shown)


src/presetsController.ts
```diff
@@ -338,20 +339,25 @@ export class PresetsController {
                             return false;
                         } else {
                             log.debug(localize('user.selected.compiler', 'User selected compiler {0}', JSON.stringify(chosen_kit)));
+                            const generator = chosen_kit.kit.preferredGenerator?.name;
+                            const cacheVariables: { [key: string]: preset.CacheVarType | undefined } = {
+                                CMAKE_INSTALL_PREFIX: '${sourceDir}/out/install/${presetName}',
+                                CMAKE_C_COMPILER: chosen_kit.kit.compilers?.['C'] || chosen_kit.kit.visualStudio ? 'cl.exe' : undefined,
```
nasinlock, line 345: Can you help me understand why we will use the cl.exe when chosen_kit.kit.compilers.['C'] has a value? Is that a flag indicating the VS compiler?

elahehrashedi:
What if we add `"CMAKE_CONFIGURATION_TYPES": "Debug;Release"` to configure preset, and remove `"configuration`" from build preset?

with this presets: 
```
    "configurePresets": [
        {
            "name": "multi",
            "displayName": "Default multi config",
            "binaryDir": "${sourceDir}/build/default",
            "generator": "Ninja Multi-Config",
            "cacheVariables": {
                "CMAKE_C_COMPILER": "cl.exe",
                "CMAKE_CXX_COMPILER": "cl.exe",
                "CMAKE_CONFIGURATION_TYPES": "Debug;Release"
            }
        },
]
"buildPresets": [
        {
            "name": "multibuild",
            "displayName": "multibuild ninja",
            "configurePreset": "multi"        
        }
]
```
All defined Cmake build types, i.e. debug and release and ... will be built. note that there is no `"configuration"` item.

But in this branch, with these presets:
```
"configurePresets": [
        {
            "name": "testmulticonfig",
            "displayName": "Visual Studio Enterprise 2019 Release - amd64 script",
            "description": "Using compilers for Visual Studio 16 2019 (x64 architecture)",
            "generator": "Visual Studio 16 2019",
            "toolset": "host=x64",
            "architecture": "x64",
            "binaryDir": "${sourceDir}/out/build/${presetName}",
            "cacheVariables": {
                "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}",
                "CMAKE_C_COMPILER": "cl.exe",
                "CMAKE_CXX_COMPILER": "cl.exe"
            }
        }
    ],
    "buildPresets": [
        {
            "name": "testmulticonfig-debug",
            "displayName": "Visual Studio Enterprise 2019 Release - amd64 script - Debug",
            "configurePreset": "testmulticonfig",
            "configuration": "debug"
        }, 
]
```
This only builds the debug type. I tried removing "configuration" and adding "CMAKE_CONFIGURATION_TYPES" but it is still not building for all types.

src/presetsController.ts
bobbrow: if `chosen_kit.kit.compilers?.['C']` has a value, then the part after the `||` will not execute.  It's like `??` in C#.

bobbrow:
@elahehrashedi one of my goals was not to change the current behavior for adding a configure preset.  To get the same behavior as the single-config generator (e.g. Ninja), we just create what is necessary to have a Debug build.  The "Add Build Preset" command can still be used to create the other ones.  I didn't want to set `CMAKE_CONFIGURATION_TYPES` to something other than the default because I don't know how frequently this is modified in real world projects.  (we might need to add it to the schema for IntelliSense though...)

elahehrashedi (approved):

Issue #2516, opened 2022-04-21
Not finding the Visual C++ 2015 compiler from CMakePresets.json


bradenmcd:
### Brief Issue Summary

My `configurePreset` includes:
```json
"architecture": {
  "value": "x64",
  "strategy": "external"
},
"toolset": {
  "value": "v140",
  "strategy": "external"
}
```
...and I have Visual Studio 2015 installed. But when I try to build using this configuration, CMake configure fails to find the C++ compiler. 

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.66.2",
  "cmtVersion": "1.10.5",
  "configurations": [
    {
      "folder": "c:\\SRCDIR",
      "cmakeVersion": "3.21.4",
      "configured": true,
      "generator": "",
      "usesPresets": true,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
[main] Building folder: BUILDDIR 
[main] Saving open files before configure/build
[main] Configuring folder: BUILDDIR 
[main] Saving open files before configure/build
[driver] Start configure 
[driver] Running pre-configure checks and steps
[cmakefileapi-driver] Configuring using preset
[cmakefileapi-driver] Invoking CMake C:\Program Files\CMake\bin\cmake.EXE with arguments ["-DCMAKE_BUILD_TYPE=RelWithDebInfo","-DCMAKE_TOOLCHAIN_FILE=c:/SRCDIR/Windows-toolchain.cmake","-DCMAKE_INSTALL_PREFIX=C:/BUILDDIR/v140-x64-RelWithDebInfo/install","-Sc:/SRCDIR","-BC:/BUILDDIR/v140-x64-RelWithDebInfo","-G","Ninja"]
[...]

Issue #2423, opened 2022-03-12
Setting only the VS toolset name in CMakePresets.json does not select the correct toolset for Ninja generators


Djoulihen:
When specifying a toolset name in CMakePresets.json, the correct toolset version is selected when using Visual Studio generators but not Ninja Generators.

I have both VS 2017 and VS 2019 installed on my computer. I'm using the following preset:
```
{
    "name": "Ninja",
    "generator": "Ninja Multi-Config",
    "architecture": {
        "strategy": "external",
        "value": "x64"
    },
    "toolset": {
        "strategy": "external",
        "value": "v141,host=x64"
    },
    "cacheVariables": {
        "CMAKE_CXX_COMPILER": "cl"
    }
},
```

When running the cmake configuration, I get:
> [preset] Configure preset Ninja: No toolset version specified for cl.exe, using latest by default
> [preset] Using developer environment from Visual Studio (instance 59fddfac, version 16.11.31729.503, installed at "C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional")

And the toolset with version 14.29.30133 is used. This corresponds to v142, not v141. Using the same configuration with a Visual Studio Generator selects the correct toolset version (14.16.XXX for v141).

Pull request #2366 allows us to specify the toolset version directly as part of the toolset string, which can be used to select the correct version. If I use `"toolset": { "value": "v141,version=14.16.27023,host=x64", "strategy": "external" }` then the 14.16.27023 toolset is correctly selected. However, only specifying the toolset name should also be supported.

See #1965 for more context about this issue.

bobbrow:
Thanks for opening an issue to track this.  The current workaround is to specify the `version` in the string as well (a two-part version number is sufficient).

jakoch:
Just using a two-part version, like `14.32`, didn't work for me. Only passing the full version number worked.
So i ended up with:
```
"toolset": {
    "value": "v143,host=x64,version=14.32.31326",
    "strategy": "external"
},
```

For those running into the same issue:
You can find the version number by opening the `Developer Command Prompt` for the desired MSVC version
and running `echo %VCToolsVersion%`.

bobbrow:
The fix for this will be available in tomorrow's pre-release version.

PR #2524, opened 2022-04-25
Fix Ninja env for VS 2015 and map toolsets to versions

bobbrow:
Addresses: #2516 and #2423


elahehrashedi (approved):

elahehrashedi:
@bobbrow I updated the changelog, would you please take a quick look at it?

PR #2534, opened 2022-05-05
add partial activation telemetry

### LINKED ISSUE #2511: Add 'esc' key partial activation telemetry
### Brief Issue Summary

Those who press 'esc' Key in control group prompt needs to fire partial telemetry, currently is only firing:
image

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

_No response_

elahehrashedi:
indicate if a new CMakeLists.txt is being selected or if the menu is being canceled.
bug fix #2511 

[sinemakinci1](https://github.com/sinemakinci1) `cancel`, `updateSourceDirectory`, `invalidSourceDirectoryPath` will add up to 100%.

### DIFF
--- a/src/cmakeTools.ts
+++ b/src/cmakeTools.ts
@@ -623,7 +623,7 @@ export class CMakeTools implements api.CMakeToolsAPI {
 
                         let selectedFile: string | undefined;
                         if (!selection) {
-                            telemetryProperties["missingCMakeListsUserAction"] = "cancle";
+                            telemetryProperties["missingCMakeListsUserAction"] = "cancel";
                             break; // User canceled it.
                         } else if (selection.label === browse) {
                             const openOpts: vscode.OpenDialogOptions = {
--- a/src/cmakeTools.ts
+++ b/src/cmakeTools.ts
@@ -623,6 +623,7 @@ export class CMakeTools implements api.CMakeToolsAPI {
 
                         let selectedFile: string | undefined;
                         if (!selection) {
+                            telemetryProperties["missingCMakeListsUserAction"] = "cancle";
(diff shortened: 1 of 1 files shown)


src/cmakeTools.ts
```diff
@@ -623,6 +623,7 @@ export class CMakeTools implements api.CMakeToolsAPI {
 
                         let selectedFile: string | undefined;
                         if (!selection) {
+                            telemetryProperties["missingCMakeListsUserAction"] = "cancle";
```
bobbrow, line 626: cancle -> cancel

src/cmakeTools.ts
```diff
@@ -646,6 +647,7 @@ export class CMakeTools implements api.CMakeToolsAPI {
                                 // doesn't need to fire the settings change event (which would trigger unnecessarily
                                 // another immediate configure, which will be blocked anyway).
                                 config.updatePartial({ sourceDirectory: relPath }, false);
+                                telemetryProperties["missingCMakeListsUserAction"] = "updateSourceDirectory";
```
bobbrow, line 650: Do we also need to add something for `if (selectedFile)` (line 642) and `if (config)` (line 645) are not entered?

bobbrow (changes requested):

src/cmakeTools.ts
sinemakinci1: I think having the else case is a good idea for case of invalid SourceDirectory, to ensure all cases are accounted for

bobbrow (approved):

sinemakinci1 (approved):

Issue #2452, opened 2022-03-23
Support tasks: install, test, configure


elahehrashedi:
Currently, we are supporting "build" task. 
The plan is to add support for these tasks as well:

- install
- test
- configure

elahehrashedi:
Add support for additional Args as requested in https://github.com/microsoft/vscode-cmake-tools/issues/2470

PR #2539, opened 2022-05-11
support configure, install, and test tasks

elahehrashedi:
feature request: #2452


src/cmakeTools.ts
```diff
@@ -1799,8 +1799,7 @@ export class CMakeTools implements api.CMakeToolsAPI {
         }
         await this.setDefaultBuildTarget(target);
         const drv = await this.cmakeDriver;
-        const targets = await this.getDefaultBuildTargets();
-        this.updateDriverAndTargetsInTaskProvider(drv, targets);
+        this.updateDriverAndTargetsInTaskProvider(drv, [target]);
```
bobbrow, line 1802: Revert this for now.  I think that we actually shouldn't set anything here. Instead we should be updating the task provider when `configure` completes because we should have the list of available targets at that point.

src/cmakeTaskProvider.ts
```diff
@@ -118,64 +130,103 @@ class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.OutputConsu
...
         let cmakePath: string = "CMake.EXE";
         let args: string[] = [];
 
         if (this.cmakeDriver) {
-            buildCommand = await this.cmakeDriver.getCMakeBuildCommand(this.definedTargets ? this.definedTargets : this.defaultTargets);
-            if (buildCommand) {
-                cmakePath = buildCommand.command;
-                args = buildCommand.args ? buildCommand.args : [];
-                this.options.environment = buildCommand.build_env;
+            if (await this.cmakeDriver.checkNeedsReconfigure() || !isSubsetTarget(this.defaultTargets, this.definedTargets)) {
```
bobbrow, line 179: I don't think we should force a configure if the defined target is not a subset.  I think just checking for reconfigure is good enough.

bobbrow: (and then you can delete `isSubsetTarget`)

bobbrow (changes requested):

src/cmakeTaskProvider.ts
```diff
@@ -118,64 +122,101 @@ class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.OutputConsu
...
+                }
+            }
+            command = await this.cmakeDriver.getCMakeBuildCommand(this.definedTargets ? this.definedTargets : this.defaultTargets);
+            if (command) {
+                cmakePath = command.command;
+                args = command.args ? command.args : [];
+                this.options.environment = command.build_env;
             }
+        } else {
+            this.closeEmitter.fire(-1);
```
bobbrow, line 184: Should we return early here?

bobbrow: And write a message too.

bobbrow (approved):

PR #2541, opened 2022-05-12
Add "cmake.skipSelectKit"

ilg-ul:
### This changes visible behaviour

- add the boolean `cmake.skipSelectKit` option at the end of the list
- add the boolean `skipSelectKit` member to the `ExtensionConfigurationSettings` interface
- if this option is set, skip asking the user to select a kit ; the result is similar to the user selecting the _Undefined_ choice

## Other Notes/Information

- the option is defined in the `resource` scope, but it is only checked in the `workspaceContext.config`
- the option has an even emitter, but for now it is not used
- there are no i18n messages

### DIFF
--- a/src/kitsController.ts
+++ b/src/kitsController.ts
@@ -274,6 +274,11 @@ export class KitsController {
             return false;
         }
 
+        if (this.cmakeTools.workspaceContext.config.skipSelectKit) {
+            log.debug(localize('skip.select.kit', 'Skip selection of a kit'));
+            return false;
+        }
+
         const avail = this.availableKits;
         log.debug(localize('start.selection.of.kits', 'Start selection of kits. Found {0} kits.', avail.length));
 
--- a/package.json
+++ b/package.json
@@ -1945,6 +1945,12 @@
           "default": false,
           "description": "%cmake-tools.configuration.cmake.ignoreCMakeListsMissing.description%",
           "scope": "resource"
+        },
+        "cmake.skipSelectKit": {
+          "type": "boolean",
+          "default": false,
+          "description": "%cmake-tools.configuration.cmake.skipSelectKit.description%",
+          "scope": "resource"
         }
       }
     },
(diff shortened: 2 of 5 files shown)


src/kitsController.ts
```diff
@@ -274,6 +274,11 @@ export class KitsController {
             return false;
         }
 
+        if (this.cmakeTools.workspaceContext.config.skipSelectKit) {
```
bobbrow, line 277: Unfortunately, we can't do this here because it means the user can never select a kit even if they wanted to later.

I was thinking about this scenario more, and it's possible that you could achieve the same thing by using `"configureOnOpen": false` and updating some code in `async postWorkspaceOpen(info: CMakeToolsFolder)` in `extension.ts`.

bobbrow, line 277: Unfortunately, we can't do this here because it means the user can never select a kit even if they wanted to later.

I was thinking about this scenario more, and it's possible that you could achieve the same thing by using `"configureOnOpen": false` and updating some code in `async postWorkspaceOpen(info: CMakeToolsFolder)` in `extension.ts`.

ilg-ul: Sure, I'm open to any better suggestions. My specific goal is to simplify/streamline the user interaction when using the xPack extension (by skipping the unnecessary selections), but of course this should not interfere with the existing functionality for those who prefer to use kits & presets.

Could you provide some more details to help me rework the patch?

bobbrow:
I may have addressed the concerns for this PR with #2632.  Setting `cmake.configureOnOpen` to `false` should stop the extra UI from appearing until a user explicitly attempts to configure or build with CMake using the extension's commands.

ilg-ul:
Apart from the kit selection, does this commit silence the other unnecessary questions too?

bobbrow:
I expect that it will.  The other questions stem from the default value of `cmake.configureOnOpen` being `null` which effectively means "prompt."

ilg-ul:
Ok, I'll set `cmake.configureOnOpen` to `false` in my projects and see what happens.

bobbrow:
#2632 replaced this PR

Issue #1945, opened 2021-06-22
Remove the extra drop down for configurePresets


KUGA2:
### Brief Issue Summary

1. the target selection thing is missing (not discussed here. see #1872)
2. Selecting 3 presets (config, build, test) takes a lot of status bar space

### Expected:

1. target should be shown (like in non-preset-mode) (not discussed here. see #1872)
2. they should be combined into one button

Suggestion:

`<TOOLS> Presets | [Target] | <GEAR> Build | <BUG> | <PLAY> | <FLASK> Run Ctest`

Note: `<FOO>` are Icons, `|` seprarate clickable stuff

Presets should run > Select configure preset, Select build preset and Select test preset at once, (current one should be preselected) and not show the names in statusbar.
[Target] should be used for build and launch target

I know, some of this can be configured via `cmake.statusbar.advanced`, but some can not.

Another bonus would be to check if there is a config, build and test preset with the same name. Then use those on selection
```
> Select Presets
> Chose configuration presets
> The same preset name exist for build and test. Use those (**Yes** No)
Yes -> exit
No ->
> Chose build preset
> Chose test preset
```

### Platform and Versions

- **Operating System**:  Windows
- **CMake Version**: 3.20.4
- **VSCode Version**: 1.57
- **CMake Tools Extension Version**: 1.7.2
- **Compiler/Toolchain**: irrelevant

bobbrow:
For the first issue, there is more discussion here: #1872
We will use this issue to track the second request.

TheJCAB:
I have to agree with the OP sentiment. When working with presets, status bar real estate is now very tight when working on a vertical monitor, even though three things still seem sorely missing from it (I'm looking at cmake-tools 1.7.3):

- Which installation of the compiler, whether Visual Studio or other, to use. Used to be sort-of the "kit". The target architecture portion of the "kit" is now selectable from the preset, thankfully, but the actual compiler toolset to use, when multiple versions are installed on the computer (I currently have 3 Visual Studios to choose from), doesn't belong in the presets file (it's a factor of the computer itself) and should still be separately selectable "on the side".
- Which target to build: #1872 
- Which target to run/debug with the C/C++ tools extension.

That's way too much on the status bar, but all of it deserves a UI. Perhaps having the option of a sidebar pane with all of this...

I hope we can regularize all of this soon. In the meantime, presets are fairly new, so some inconvenience is to be expected :-).

esweet431:
[...]

Issue #1582, opened 2020-12-04
[Suggestion] Transfer the extension's settings from the status bar to the side bar


songyang-dev:
Hello, the current CMake extension populates the bottom status bar a lot. It would be nice if all of those settings went into the side bar.

![image](https://user-images.githubusercontent.com/10444460/101205395-8f4b9e00-363b-11eb-9452-abdce1733b94.png)

## Version
Version: 1.51.1 (user setup)
Commit: e5a624b788d92b8d34d1392e4c4d9789406efe8f
Date: 2020-11-10T23:34:32.027Z
Electron: 9.3.3
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Windows_NT x64 10.0.19041

bobbrow:
Yes, this is something we've been thinking about, but I don't believe an issue was created.  Thank you.  We'll use this one.

Note, we do have settings to control the status bar icons.  Take a look at `cmake.statusbar.visibility` and `cmake.statusbar.advanced` settings.

Zingam:
When designing the new UI please consider the following:

## Configuration & Build

Each workflow (CMake presets, Kits) should have their own UI design, which makes sense in the respective context. For example for CMake presets the order of UI elements should be:

1. Active folder
2. Configure preset
3. Build preset
4. Default Build Target
5. Build
6. Launch Target
7. Debug
8. Launch

I'm not sure about the Test Preset location. 

Additionaly in CMake 3.25.0 (Presets V6) there are::

> The [cmake-presets(7)](https://cmake.org/cmake/help/git-stage/manual/cmake-presets.7.html#manual:cmake-presets(7)) schema version has been bumped to 6.
>
>  The [cmake-presets(7)](https://cmake.org/cmake/help/git-stage/manual/cmake-presets.7.html#manual:cmake-presets(7)) format now supports a packagePresets field to specify presets for [cpack --preset](https://cmake.org/cmake/help/git-stage/manual/cpack.1.html#cmdoption-cpack-preset).
>
>  The [cmake-presets(7)](https://cmake.org/cmake/help/git-stage/manual/cmake-presets.7.html#manual:cmake-presets(7)) format now supports a workflowPresets field to specify presets for [cmake --workflow](https://cmake.org/cmake/help/git-stage/manual/cmake.1.html#cmdoption-cmake-workflow).

## Debugging

[...]

PR #2547, opened 2022-05-15
Assign names to status bar items

chausner:
## This change addresses item #1933

This assigns a name to each button on the status bar which allows users to individually show/hide buttons. Before this change, users can only show/hide all buttons at once.

Users have reported in https://github.com/microsoft/vscode-cmake-tools/issues/1945 that the status bar buttons take up a lot of space. Allowing users to hide individual buttons that they might not need (e.g. tests) could help.

In order to use the newer `createStatusBarItem` overload, I had to update `@types/vscode` to a newer version. I picked `1.63.0` which matches the minimum VS Code version specified in `package.json`:
```json
"engines": {
    "vscode": "^1.63.0"
  },
```
As part of this update, I had to make some minor changes to the mocking and tests code.

The names assigned to the buttons are shown in the context menu of the status bar:

image

They would need to be localized.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1229,24 +1229,24 @@
               "description": "%cmake-tools.configuration.cmake.debugConfig.additionalSOLibSearchPath.description%"
             },
             "externalConsole": {
-                "type": "boolean",
-                "description": "%cmake-tools.configuration.cmake.debugConfig.externalConsole.description%"
+              "type": "boolean",
+              "description": "%cmake-tools.configuration.cmake.debugConfig.externalConsole.description%"
             },
             "console": {
-                "type": "string",
-                "enum": [
-                  "internalConsole",
-                  "integratedTerminal",
-                  "externalTerminal",
-                  "newExternalWindow"
-                ],
-                "enumDescriptions": [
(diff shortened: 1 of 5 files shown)


bobbrow:
We do have settings in the extension that allow you to control the status bar such as how much space they use and which ones appear.  I'm concerned that adding all of these to that list will make it overly noisy.

The setting that controls this is `cmake.statusbar.advanced`.

chausner:
Oh, I did not know about the advanced settings for configuring the status bar. I agree that the many items make the context menu a bit crowded. On the other hand, some users like me might not know about the ability to hide some of the status bar items via the settings, and the context menu provides a more intuitive and consistent way to configure that. So I think there are pros and cons to both approaches, and either way is fine for me.

If you decide to leave it as it is, maybe it still makes sense to merge the update of `@types/vscode` to `1.63.0` so that it matches the minimum engine version.

bobbrow:
There is also issue #1582 which I think we still want to do.  This would move us in the direction of managing project state in the sidebar instead of the status bar.

elahehrashedi:
Thank you for your contribution. Unfortunately, this is not the direction we want to take, and we encourage users to use the settings to control the status bar.

Issue #2526, opened 2022-04-28
changes to included files not triggering reconfigure


BMBurstein:
### Brief Issue Summary

whenever I make a change to a `CMakeLists.txt` file, it triggers a reconfigure. But if I have e.g. `include(file.cmake)` in the `CMakeLists.txt` file and I change `file.cmake`, it doesn't trigger an automatic reconfigure. Is it possible to do this?

elahehrashedi:
This feature is not currently available, but we can add it in the future. We may not reach to implement this feature soon, but we appreciate any help from the community.

bobbrow:
This was fixed with PR #2548 and is available starting in version 1.12.8 (pre-release).

PR #2548, opened 2022-05-15
Use CMake driver to detect changes in CMake files

chausner:
## This change addresses item #2526

This leverages the CMake file API or server API drivers to obtain a list of CMake input files (`CMakeLists.txt` or included `.cmake` files) in order to detect when any of these were changed by the user to trigger a re-configuration.

I haven't been able to test the server API, unfortunately, as it always fails with
> connect ENOENT \\\\?\\pipe\\c:\\Users\\chris\\Desktop\\cmaketest2\\.vscode\\.cmserver-pipe

If somebody could give it a try, it would be much appreciated.

### DIFF
--- a/src/cmakeTools.ts
+++ b/src/cmakeTools.ts
@@ -896,14 +896,23 @@ export class CMakeTools implements api.CMakeToolsAPI {
             }
 
             const sourceDirectory = (this.sourceDir).toLowerCase();
-            let isCmakeListsFile: boolean = false;
+            
+            let isCmakeFile: boolean;
+            if (drv && drv.cmakeFiles.length > 0) {
+                // If CMake file information is available from the driver, use it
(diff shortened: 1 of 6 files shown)


chausner:
> I haven't been able to test the server API, unfortunately, as it always fails with

I found out it's been due to me using CMake 3.23.1 where server mode has been removed. Unfortunately, I haven't been able to test with the older CMake 3.19 that still supports server mode either because that version does not support the VS 2022 generator yet, and I do not have VS 2019 installed.

elahehrashedi:
Please add this to the CHANGELOG.md
- Triggering reconfigure after changes are made to included files. [#2526](https://github.com/microsoft/vscode-cmake-tools/issues/2526) [@chausner](https://github.com/chausner)

chausner:
> Please add this to the CHANGELOG.md

Done!

bobbrow:
> Unfortunately, I haven't been able to test with the older CMake 3.19 that still supports server mode either because that version does not support the VS 2022 generator yet, and I do not have VS 2019 installed.

We tested with CMake 3.19 and it works ok.

PR #2549, opened 2022-05-15
Fix clang version detection regexes

chausner:
This fixes some regexes used to determine the Clang compiler version for telemetry. The regexes had a trailing space which would prevent them from ever matching. I not only removed the space but also changed the regex so that it matches version strings that don't contain a dash.

See also https://regex101.com/r/lkqFHg/1.


bobbrow (approved):

elahehrashedi:
@andreeis what do you suggest?

Issue #2546, opened 2022-05-15
CMakeCache.txt removed when switching kit but cmake.buildDirectory depends on ${buildKitVendor}


david-fong:
### Brief Issue Summary

Severity (from my POV): minor bug

When I do the `select kit` action and change between kits, the extension deletes the `CMakeCache.txt` and `CMakeFiles.txt` files from the build directory. In my project's settings.json, I have set `cmake.buildDirectory` to `"${workspaceFolder}/build/${buildKitVendor}"`. due to the file removal, when I try to run `cmake --build` in the switched-from (the previous) kit's build directory, I get an error saying the cache file is gone.

```
[driver] Switching to kit: GCC 11.2.0 x86_64-linux-gnu
[driver] Removing /home/david/code/mine/okiidoku/cpp/build/Clang/CMakeCache.txt
[driver] Removing /home/david/code/mine/okiidoku/cpp/build/Clang/CMakeFiles
[proc] Executing command: /usr/bin/gcc -v
```

I think I can fix this by just running the cmake configure command manually in the switched-from kit's folder to recreate the cmake cache file (in which case, do I need to copy the exact command that the extension uses to create the same cmake cache file contents?).

EDIT: see my below comment for the description of the bug.

### Additional Information

I set up my project in this way because I want to be able to easily and quickly test my project locally with GCC, Clang, and (on my windows machine) MSVC, and also be able to compare the build artifact symbol maps and sizes. (Maybe there's a better way to setup my project to achieve this?)

elahehrashedi:
We should be able to detect a change in the build directory when changing the kit. We may not be able to get to this immediately, but we would accept a PR from the community.

david-fong:
Note: the code that removes the cache file is at [src/drivers/cmakeDriver.ts:395 (`_cleanPriorConfiguration()`)](https://github.com/microsoft/vscode-cmake-tools/blob/87c934a2fc6a34bc0b6f0ac1e03ff724c867f473/src/drivers/cmakeDriver.ts#L395).

I think the relevant call to that function is the one at [src/drivers/cmakeFileApiDriver.ts:180 (`doSetKit()`)](https://github.com/microsoft/vscode-cmake-tools/blob/87c934a2fc6a34bc0b6f0ac1e03ff724c867f473/src/drivers/cmakeFileApiDriver.ts#L180)

[...]

PR #2551, opened 2022-05-17
in setKit, check changed buildDirectory after instead of before refreshing expansions

david-fong:

- In `setKit`, call `_cleanPriorConfigurations` _after_ `_refreshExpansions`

## Other Notes/Information

This is just a draft / example. I have no idea if it is actually safe to move this check to go later. I also noticed similar code for `doSetConfigurePreset`, which might be a candidate for similar changes. What is `doSetConfigurePreset` for / for what user action does it get called?


bobbrow:
Sorry for the delay on this.  We'll take a look at it next week.

elahehrashedi:
@bobbrow I made a change to the PR.

src/drivers/cmakeDriver.ts
```diff
@@ -514,14 +514,16 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
-        const needs_clean = this.binaryDir === newBinaryDir && kitChangeNeedsClean(kit, this._kit);
-        await this.doSetKit(needs_clean, async () => {
+        const oldBinaryDir = this.binaryDir;
+        const needsCleanIfKitChange = kitChangeNeedsClean(kit, this._kit);
+        await this.doSetKit(async () => {
             await this._setKit(kit, preferredGenerators);
             await this._refreshExpansions();
+            const newBinaryDir = util.lightNormalizePath(await expand.expandString(this.config.buildDirectory, this.expansionOptions));
+            const needsCleanIfBuildDirectoryChange = (newBinaryDir !== oldBinaryDir);
+            if (needsCleanIfKitChange || needsCleanIfBuildDirectoryChange) {
```
bobbrow, line 524: I don't think that changing the build directory is the right condition to delete the cache.  What I expected to see here was "if the kit changed, but the build directory didn't, then remove the prior cache" because I think it makes sense to expect that a different Kit is a different compiler and would configure the project differently (basically the same check that was happening on deleted line 521).  The rest of the fix seems correct because it defers the previous check until after the kit has been set and the `newBinaryDir` can be computed correctly.

**TL;DR:**
I think line 524 should be:
```ts
  if (needsCleanIfKitChange && newBinaryDir === oldBinaryDir) {
```

david-fong: sorry about that. that was the old logic. I changed it in the first commit and I don't remember why. I think I was just confused.

david-fong: just pushed a commit for this.

bobbrow: No need to apologize. We appreciate you addressing this issue!

bobbrow (approved):

bobbrow:
Something went (surprisingly) wrong with the tests after this change.  We'll need to investigate.

bobbrow:
> Something went (surprisingly) wrong with the tests after this change. We'll need to investigate.

The good news is that it doesn't appear to be related to this change. I can repro it in `main`.  The bad news is that I don't have a root cause yet.

david-fong:
The linux and windows runs don't say much, but the MacOS output shows something wrong that looks related to these changes (I don't know what exactly though). When you say "main", do you mean vscode-cmake-tools/main? because my PR branch is the main branch of my fork.

david-fong:
Maybe it's an order of operations mistake. Before this PR, the call to `-`_cleanPriorConfiguration` was alwasy made _before_ calling the callback. But I moved it into the the callback and put it at the _end_ of the callback body, (now) _after_ these operations:

```
await this._setKit(kit, preferredGenerators);
await this._refreshExpansions();
```

david-fong:
Oh but if that's really the issue, there's a chicken-egg problem: I moved it to go after those things because refreshing expansions is what enables getting the new path, which would be required to compare the new and old paths. In this case it might make sense to do more rework like I suggested in #2546 and make refreshExpansions more like a pure function.

bobbrow:
I don't think there's anything wrong with your PR.  Other PR's are failing with the same issue (e.g. #2641).  A new version of VS Code was released this week and I suspect the test issues are related to that.  I'm going to commit this fix without the tests passing and work on the CI some more when I return from vacation.  With any luck the bug in VS Code will be addressed and fixed by then.  Otherwise, I'll report an issue when I return.

david-fong:
Thanks! enjoy your vacation :)

PR #2560, opened 2022-05-19
Added an example for msvc (launch.json)

Thaulino:
For copycats on windows

- add an example how to configure launch.json if using msvc

## The purpose of this change

I am using vs code + msvc and cmake tools, i always have to fix some issues if copying the minimal example from the docs. 
This pull-request solves it by supplying a minimal exampel for msvc, which can be copy and pasted.

## Additional info 

Thanks for vscode-cmake-tools - i appreciate the flexibility of this vs code plugin 👍


bobbrow (approved):

bobbrow:
Thank you for the improvement to the documentation!

PR #2561, opened 2022-05-19
Change project outline to respect `FOLDER` property

NekkoDroid:

- Change the project outline to respect `(CMAKE_)FOLDER` options of CMake

## Other Notes/Information

- the change completely replaces the old behavior of using relative paths for the targets
- possibly making old and new behavior user configurable might be worth exploring?

### DIFF
--- a/src/tree.ts
+++ b/src/tree.ts
@@ -433,9 +433,7 @@ class ProjectNode extends BaseNode {
         };
 
         for (const target of pr.targets) {
-            const srcdir = target.sourceDirectory || '';
-            const relpath = path.relative(pr.sourceDirectory, srcdir);
-            addToTree(tree, relpath, target);
+            addToTree(tree, target.folder || '', target);
         }
         collapseTreeInplace(tree);
 
--- a/src/tree.ts
+++ b/src/tree.ts
@@ -421,7 +421,7 @@ class ProjectNode extends BaseNode {
         return item;
     }
 
-    update(pr: codeModel.CodeModelProject, ctx: TreeUpdateContext) {
+    update(pr: codeModel.CodeModelProject, ctx: TreeUpdateContext, useFilesystemStructure: boolean) {
         if (pr.name !== this.name) {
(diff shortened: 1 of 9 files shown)


src/tree.ts
```diff
@@ -433,9 +433,7 @@ class ProjectNode extends BaseNode {
         };
 
         for (const target of pr.targets) {
-            const srcdir = target.sourceDirectory || '';
-            const relpath = path.relative(pr.sourceDirectory, srcdir);
-            addToTree(tree, relpath, target);
+            addToTree(tree, target.folder || '', target);
```
bobbrow, line 436: Should it be `target.folder || relpath`?  Or does the file api always have a value for folder even if CMAKE_FOLDER is not set?

NekkoDroid: When `CMAKE_FOLDER` is not set for a target `target.folder` doesn't have a value.

The reason I set it to empty instead of using the relative path is to not mix-n-match using the property and the relative path. So if one target doesn't have a value it gets put directly under the project (kinda how VS handles it) and if set it uses `CMAKE_FOLDER` for the path.

NekkoDroid:
I have been also working on adding an option to preserve the old behavior, but I am not sure if the implementation is ideal since it would propagate the option up to `extension.ts`

src/tree.ts
bobbrow: What happens if the developer doesn't use CMAKE_FOLDER at all?  It seems like this would change the behavior by passing '' for all targets.

NekkoDroid: When `CMAKE_FOLDER` isn't used at all it simply adds the target directly under the project node without any folders.

NekkoDroid: Old view without `CMAKE_FOLDER`:
Old Example
New view without `CMAKE_FOLDER`:
New Example
New view with one target before `CMAKE_FOLDER` is set and one after it's set to `Source`:
New Example2

NekkoDroid:
Imma put this actually on hold, since it seems that source files are also currently only organized how they are laid out on the filesystem and I will possibly explore to add `source_group`ing to the PR

Issue #2555, opened 2022-05-18
Add "clean" and "clean-rebuild" task


elahehrashedi:
Add  "clean" and "clean-rebuild" tasks to task provider

PR #2562, opened 2022-05-19
add clean and clean-rebuild task

elahehrashedi:
feature-request https://github.com/microsoft/vscode-cmake-tools/issues/2555

### DIFF
--- a/package.json
+++ b/package.json
@@ -539,7 +539,9 @@
               "build",
               "configure",
               "install",
-              "test"
+              "test",
+              "clean",
+              "clean rebuild"
             ],
             "description": "CMake build command"
           },
--- a/package.json
+++ b/package.json
@@ -541,7 +541,7 @@
               "install",
               "test",
               "clean",
-              "clean rebuild"
+              "cleanRebuild"
             ],
             "description": "CMake build command"
           },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -539,7 +539,9 @@
               "build",
               "configure",
               "install",
-              "test"
+              "test",
+              "clean",
+              "clean rebuild"
```
bobbrow, line 544: Since it's an enum, I would suggest not having a space in the name and use `cleanRebuild` instead.

src/cmakeTaskProvider.ts
```diff
@@ -45,6 +47,12 @@ const localizeCommandType = (cmd: CommandType): string => {
         case CommandType.config: {
             return localize("configure", "configure");
         }
+        case CommandType.clean: {
+            return localize("clean", "clean");
+        }
+        case CommandType.cleanRebuild: {
+            return localize("clean rebuild", "clean rebuild");
```
bobbrow, line 54: we usually put a dot in the id.  `clean.rebuild`

bobbrow (changes requested):
Please update the changelog as well.

bobbrow (approved):

PR #2563, opened 2022-05-20
Add support for CppTools API V6

elahehrashedi:
feature : https://github.com/microsoft/vscode-cmake-tools/issues/2508


Colengms (approved):

src/cpptools.ts
```diff
@@ -446,10 +454,27 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
...
+        let flags: string[] = [];
+        let extraDefinitions: string[] = [];
+        let standard: StandardVersion;
+        let targetArch: Architecture;
+        let intelliSenseMode: IntelliSenseMode;
+        let defines = (fileGroup.defines || target.defines || []);
+        if (!useFragments) {
+            // Send the intelliSenseMode and standard only for CppTools API v5 and below.
+            flags = getAsFlags(fileGroup.compileCommandFragments || target.compileCommandFragments);
+            ({ extraDefinitions, standard, targetArch } = parseCompileFlags(this.cpptoolsVersion, flags, lang));
```
bobbrow, line 473: You shouldn't need parentheses for this.  We didn't have them before.

src/cpptools.ts
```diff
@@ -460,25 +485,32 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
...
+            flags.push('--sysroot=' + shlex.quote(sysroot));
         }
 
         this.workspaceBrowseConfiguration = {
             browsePath: newBrowsePath,
-            standard,
             compilerPath: normalizedCompilerPath || undefined,
-            compilerArgs: flags || undefined
+            // Pass sysroot (without quote added) as the only compilerArgs for CppTools API V6 and above.
+            compilerArgs: !useFragments ? flags : sysroot ? [('--sysroot=' + sysroot)] : undefined,
```
bobbrow, line 496: Can you do the v6 adding of flags without quotes in the block for sysroot above (line 487)?  I think it will be less confusing.

src/cpptools.ts
```diff
@@ -489,7 +521,7 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
      * @param fileGroup The file group
      * @param options Index update options
      */
-    private updateFileGroup(sourceDir: string, fileGroup: codeModel.CodeModelFileGroup, options: codeModel.CodeModelParams, target: TargetDefaults, sysroot: string) {
+    private updateFileGroup(sourceDir: string, fileGroup: codeModel.CodeModelFileGroup, options: codeModel.CodeModelParams, target: TargetDefaults, sysroot: string | undefined) {
```
bobbrow, line 524: Instead of `| undefined`, prefer to use `sysroot?: string`

bobbrow (changes requested):

src/cpptools.ts
elahehrashedi: this is the error message: 
Parsing error: Declaration or statement expected. This '=' follows a block of statements, so if you intended to write a destructuring assignment, you might need to wrap the whole assignment in parentheses.
Declaration or statement expected.ts(1128)

bobbrow: Ok. It's ugly, but we can leave it for now.

src/cpptools.ts
```diff
@@ -460,25 +485,36 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
...
+            includePath: normalizedIncludePath,
             defines,
+            intelliSenseMode,
             standard,
-            includePath: normalizedIncludePath,
-            intelliSenseMode: getIntelliSenseMode(this.cpptoolsVersion, compilerPath, targetArchFromToolchains ?? targetArch),
+            // forcedInclude,
             compilerPath: normalizedCompilerPath || undefined,
-            compilerArgs: flags || undefined
+            compilerArgs: !useFragments ? flags : undefined,
```
elahehrashedi, line 515: change this to `flags`

bobbrow (approved):

src/cpptools.ts
```diff
@@ -201,37 +204,39 @@ export function parseCompileFlags(cptVersion: cpptools.Version, args: string[],
             const def = value.substring(2);
             extraDefinitions.push(def);
         } else if (value.startsWith('-std=') || lower.startsWith('-std:') || lower.startsWith('/std:')) {
```
michelleangela, line 206: It looks like the check `else if (value.startsWith('-std=') || lower.startsWith('-std:') || lower.startsWith('/std:'))` doesn't need to happen unless `extractStdFlag` is true. It would be more efficient then to do:

```
else if (extractStdFlag && (value.startsWith('-std=') || lower.startsWith('-std:') || lower.startsWith('/std:')))
```

src/cpptools.ts
```diff
@@ -424,14 +427,19 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
      * @param fileGroup The file group from the code model to create config data for
      * @param opts Index update options
      */
-    private buildConfigurationData(fileGroup: codeModel.CodeModelFileGroup, opts: codeModel.CodeModelParams, target: TargetDefaults, sysroot: string): cpptools.SourceFileConfiguration {
+    private buildConfigurationData(fileGroup: CodeModelFileGroup, opts: CodeModelParams, target: TargetDefaults, sysroot?: string): cpptools.SourceFileConfiguration {
+        // For CppTools V6 and above, build the compilerFragments data, otherwise build compilerArgs data
+        const useFragments: boolean = this.cpptoolsVersion >= cpptools.Version.v6;
```
bobbrow, line 432: Let's just check if fragments exist on the codemodel instead of making all the changes to the server codemodel.  If the fragments don't exist, then it must be a server codemodel object and we want to use flags instead.

src/drivers/cmakeServerClient.ts
```diff
@@ -220,51 +221,54 @@ export interface CodeModelParams {}
 
 export interface CodeModelRequest extends CookiedMessage, CodeModelParams { type: 'codemodel' }
 
-export interface CodeModelFileGroup {
+/*export interface CodeModelFileGroup {
```
bobbrow, line 224: revert all the changes to server codemodel, including the naming.

src/cmakeTools.ts
```diff
@@ -441,7 +442,7 @@ export class CMakeTools implements api.CMakeToolsAPI {
     get onCodeModelChanged() {
         return this._codeModelContent.changeEvent;
     }
-    private readonly _codeModelContent = new Property<codeModel.CodeModelContent | null>(null);
+    private readonly _codeModelContent = new Property<CodeModelContent | ServerCodeModelContent | null>(null);
```
bobbrow, line 445: This should be just `CodeModelContent`

src/tree.ts
```diff
@@ -1,7 +1,7 @@
 import * as path from 'path';
 import * as vscode from 'vscode';
 import * as nls from 'vscode-nls';
-
+import { ServerCodeModelContent } from '@cmt/drivers/cmakeServerClient';
```
bobbrow, line 4: remove uses of `ServerCodeModelContent` in this file.

src/drivers/cmakeDriver.ts
```diff
@@ -28,6 +28,7 @@ import { ConfigureArguments, VariantOption } from '@cmt/variant';
 import * as nls from 'vscode-nls';
 import { majorVersionSemver, minorVersionSemver, parseTargetTriple, TargetTriple } from '@cmt/triple';
 import * as preset from '@cmt/preset';
+import * as cms from '@cmt/drivers/cmakeServerClient';
```
bobbrow, line 31: remove `ServerCodeModelContent` in this file

src/drivers/cmakeServerDriver.ts
```diff
@@ -63,15 +64,55 @@ export class CMakeServerDriver extends CMakeDriver {
     private _prevConfigureEnv = 'null';
 
     // TODO: Refactor to make this assertion unecessary
-    private _codeModel!: null | cms.CodeModelContent;
-    get codeModel(): null | cms.CodeModelContent {
-        return this._codeModel;
-    }
-    set codeModel(v: null | cms.CodeModelContent) {
-        this._codeModel = v;
+    private codeModel!: CodeModelContent | null;
```
bobbrow, line 67: should it be initialized to null?

src/drivers/cmakeServerDriver.ts
```diff
@@ -63,15 +64,55 @@ export class CMakeServerDriver extends CMakeDriver {
...
+    private convertServerCodeModel(serverCodeModel: null | ServerCodeModelContent): CodeModelContent | null {
+        if (serverCodeModel) {
+            const codeModel: CodeModelContent = { configurations: [], toolchains: undefined};
```
bobbrow, line 70: >, toolchains: undefined [](http://example.com/codeflow?start=68&length=23)

let's not set `toolchains` if it's undefined

src/drivers/cmakeServerDriver.ts
```diff
@@ -63,15 +64,55 @@ export class CMakeServerDriver extends CMakeDriver {
...
+                            name: target.name,
+                            type: target.type,
+                            sourceDirectory: target.sourceDirectory,
+                            fullName: target.fullName,
+                            artifacts: target.artifacts,
+                            sysroot: target.sysroot,
+                            fileGroups: []
+                        };
+                        if (target.fileGroups) {
+                            newTarget.fileGroups = [];
```
bobbrow, line 91: >newTarget.fileGroups = []; [](http://example.com/codeflow?start=28&length=26)

you can remove this line

src/drivers/codeModel.ts
```diff
@@ -60,7 +60,7 @@ export interface CodeModelFileGroup {
     }[];
 
     /** Compiler flags */
-    compileFlags?: string;
```
bobbrow, line 63: I think we need to keep both flags and fragments and populate the right one based on the cpptools api.

src/drivers/cmakeServerDriver.ts
```diff
@@ -63,15 +64,55 @@ export class CMakeServerDriver extends CMakeDriver {
...
+                            for (const group of target.fileGroups) {
+                                const newGroup: CodeModelFileGroup = {
+                                    sources: group.sources,
+                                    language: group.language,
+                                    includePath: group.includePath,
+                                    defines: group.defines,
+                                    isGenerated: group.isGenerated,
+                                    compileCommandFragments: group.compileFlags ? [...shlex.split(group.compileFlags)] : []
```
bobbrow, line 99: we should populate the correct one (flags, or fragments) based on the API version.   So this code will need to know the API version somehow.  Maybe initialize the driver with it?  Or get it from extension.ts?

src/cpptools.ts
```diff
@@ -446,10 +454,27 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
...
+        };
+        let flags: string[] = [];
+        let extraDefinitions: string[] = [];
+        let standard: StandardVersion;
+        let targetArch: Architecture;
+        let intelliSenseMode: IntelliSenseMode;
+        let defines = (fileGroup.defines || target.defines || []);
+        if (!useFragments) {
+            // Send the intelliSenseMode and standard only for CppTools API v5 and below.
+            flags = getAsFlags(fileGroup.compileCommandFragments || target.compileCommandFragments);
```
bobbrow, line 472: this will need to change to get compileFlags instead (once the interface changes)

src/drivers/cmakeServerDriver.ts
```diff
@@ -184,13 +224,12 @@ export class CMakeServerDriver extends CMakeDriver {
...
         }, new Map<string, cache.Entry>());
-        this.codeModel = await client.codemodel();
+        this.codeModel = this.convertServerCodeModel(await client.codemodel());
 
         // Toolchain information is not available with CMake server.
-        this._codeModelChanged.fire({
-            configurations: this.codeModel.configurations,
-            toolchains: new Map<string, codeModel.CodeModelToolchain>()
-        });
+        // convert it into real codemodel, this is not the correct type
```
bobbrow, line 230: please remove this comment. The codemodel is converted now.

src/drivers/cmakeServerDriver.ts
```diff
@@ -19,6 +17,8 @@ import { errorToString } from '@cmt/util';
 import * as nls from 'vscode-nls';
 import * as ext from '@cmt/extension';
 import { BuildPreset, ConfigurePreset, TestPreset } from '@cmt/preset';
+import { ServerCodeModelContent } from '@cmt/drivers/cmakeServerClient';
```
bobbrow, line 20: There is still an import `cms` for `cmakeServerClient` (line 9).  Do you want to use that, or import each type in separately?  We shouldn't have both

src/drivers/codeModel.ts
```diff
@@ -9,7 +9,7 @@ export interface CodeModelTarget {
      *
      * (Source CMake Documentation cmake-file-api(7))
      */
-    readonly name: string;
+    name: string;
```
bobbrow, line 12: Can `readonly` be added back in?

test/unit-tests/cpptools.test.ts
```diff
@@ -201,26 +203,26 @@ suite('CppTools tests', () => {
...
+        provider.updateConfigurationData({ cache, codeModelContent: codeModel1, activeTarget: 'target1', activeBuildTypeVariant: 'Release', folder: here });
 
         // Update configuration with a 2nd workspace folder.
         const smokeFolder = path.join(here, '../smoke');
-        const sourceFile2 = path.join(smokeFolder, 'main.cpp');
-        const uri2 = vscode.Uri.file(sourceFile2);
-        const codeModelContent2: codeModel.CodeModelContent = {
+        const sourceFile3 = path.join(smokeFolder, 'main.cpp');
+        const uri3 = vscode.Uri.file(sourceFile3);
+        const codeModelTar3: codeModel.CodeModelContent = {
```
bobbrow, line 212: I think it should be `codeModel3` to match `codeModel1`

test/unit-tests/cpptools.test.ts
```diff
@@ -258,11 +260,121 @@ suite('CppTools tests', () => {
...
+            toolchains: new Map<string, codeModel.CodeModelToolchain>()
+        };
+
+        provider.updateConfigurationData({ cache, codeModelContent: codeModel1, activeTarget: 'target1', activeBuildTypeVariant: 'Release', folder: here });
+
+        // Update configuration with a 2nd workspace folder.
+        const smokeFolder = path.join(here, '../smoke');
+        const sourceFile3 = path.join(smokeFolder, 'main.cpp');
+        const uri3 = vscode.Uri.file(sourceFile3);
+        const codeModelTar3: codeModel.CodeModelContent = {
```
bobbrow, line 318: `codeModel3` here too

test/unit-tests/driver/driver-codemodel-tests.ts
```diff
@@ -130,7 +130,11 @@ export function makeCodeModelDriverTestsuite(driverName: string, driver_generato
             const codemodel_data = await generateCodeModelForConfiguredDriver();
             expect(codemodel_data).to.be.not.null;
 
-            const target = codemodel_data!.configurations[0].projects[0].targets.find(t => t.type === 'EXECUTABLE' && t.name === 'TestBuildProcess');
+            const target: CodeModelTarget | undefined = codemodel_data!.configurations[0].projects[0].targets.find((t: CodeModelTarget) => {
+                if (t.type === 'EXECUTABLE' && t.name === 'TestBuildProcess') {
```
bobbrow, line 134: You can revert the changes to this lambda to make it more concise

test/unit-tests/driver/driver-codemodel-tests.ts
```diff
@@ -171,25 +175,27 @@ export function makeCodeModelDriverTestsuite(driverName: string, driver_generato
...
             // Test main source file
             expect(compile_information!.sources).to.include('info.cpp');
             expect(compile_information!.sources).to.include('test2.cpp');
 
-            // compile flags for file groups
+            // compile flags or fragments for file groups
             if (process.platform === 'win32') {
-                expect(compile_information!.compileFlags?.trim()).to.eq('/DWIN32 /D_WINDOWS /W3 /GR /EHsc /MDd /Zi /Ob0 /Od /RTC1');
+                if (compile_information) {
+                    expect(compile_information.compileCommandFragments?.join(' ')).to.eq('/DWIN32 /D_WINDOWS /W3 /GR /EHsc /MDd /Zi /Ob0 /Od /RTC1');
```
bobbrow, line 189: We should trim the strings in the array before joining the arrays.  Or else we should trim the strings in `cmakeFileApi.ts` and `cmakeServerDriver.ts` when they set the `compileCommandFragments` in the code model.

test/unit-tests/driver/driver-codemodel-tests.ts
```diff
@@ -215,17 +221,18 @@ export function makeCodeModelDriverTestsuite(driverName: string, driver_generato
             // Language
             expect(target!.fileGroups![0].language).to.eq('C');
 
-            // compile flags for file groups
+            // compile flags or fragments for file groups
             if (process.platform === 'win32') {
-                expect(target!.fileGroups![0].compileFlags?.trim()).to.eq('/DWIN32 /D_WINDOWS /W3 /MDd /Zi /Ob0 /Od /RTC1');
+                expect(target!.fileGroups![0].compileCommandFragments?.join(' ')).to.eq('/DWIN32 /D_WINDOWS /W3 /MDd /Zi /Ob0 /Od /RTC1');
```
bobbrow, line 226: same comment about `trim` here.

test/unit-tests/driver/driver-codemodel-tests.ts
```diff
@@ -171,25 +175,27 @@ export function makeCodeModelDriverTestsuite(driverName: string, driver_generato
...
 
             // Test main source file
             expect(compile_information!.sources).to.include('info.cpp');
             expect(compile_information!.sources).to.include('test2.cpp');
 
-            // compile flags for file groups
+            // compile flags or fragments for file groups
             if (process.platform === 'win32') {
-                expect(compile_information!.compileFlags?.trim()).to.eq('/DWIN32 /D_WINDOWS /W3 /GR /EHsc /MDd /Zi /Ob0 /Od /RTC1');
+                if (compile_information) {
```
bobbrow, line 188: You should assert `compile_information!` like the test used to instead of changing this to `if`.

src/drivers/cmakeServerDriver.ts
elahehrashedi: removed the second one.

bobbrow (approved):
Ready to go after tests pass

PR #2567, opened 2022-05-24
Fix exception object type mismatch

chausner:
This fixes a type mismatch error that I often get after debugging the extension when starting the extension another time. For some reason, the TypeScript compiler does not complain on initial run, only on subsequent runs. I then have to close and reopen VS Code for it to work again.

The type error seems valid, though. `e` is of type `unknown` but `this.exception` requires an argument of type `Error`. I've seen multiple other places in the codebase where exceptions are casted to `Error` so it appears this is safe to do.


bobbrow:
Thanks. There are potentially other instances of this elsewhere too. We've seen them pop up as we edit files because of recent updates to TypeScript. They may be a linter rule that can help us find them all faster, but we haven't done a deep dive into this yet.

PR #2568, opened 2022-05-24
Properly log if CMake server fails to start

chausner:
This ensures that a message is logged if the CMake server fails to start properly, e.g. because the CMake version is 3.20 or newer where CMake server is no longer supported. In this case, the cmake process immediately exits after starting but the `close` event handler is only attached 1 second after process creation which is too late.

The suggested fix moves the event handler subscription out of `setTimeout`.


bobbrow:
Thank you for the update to this section of code.

PR #2619, opened 2022-06-24
Fix broken file watcher for presets

chausner:
## The purpose of this change

This fixes a bug introduced in #2544 where an exception would get thrown in `PresetsController.reapplyPresets` when called from the presets file watcher because the callbacks are invoked with the wrong `this` context. The fix is to pass them via arrow functions.

See also the bug report at https://github.com/microsoft/vscode-cmake-tools/pull/2544#issuecomment-1165257516.

### DIFF
--- a/src/presetsController.ts
+++ b/src/presetsController.ts
@@ -1086,10 +1086,13 @@ export class PresetsController {
             this._presetsWatcher.close().then(() => {}, () => {});
         }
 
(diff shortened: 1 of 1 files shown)


bobbrow (approved):

bobbrow:
Thanks for fixing this!

Issue #2622, opened 2022-06-26
[bug] Unable to detect `clang-cl.exe` as compilers installed in `C:\Program Files\LLVM\bin\`.


hwhsu1231:
There are two types of clang compilers on Windows: `clang.exe` and `clang-cl.exe`.

It seems that it can successfully detect both `clang.exe` and `clang-cl.exe` installed in **Visual Studio**. 

However, it can **ONLY** detect `clang.exe` installed in **LLVM**.

### Screenshots

![image](https://user-images.githubusercontent.com/71438617/175800453-d54c1d57-9026-445c-93cd-e235804c32c6.png)
![image](https://user-images.githubusercontent.com/71438617/175800422-1042a8ea-71c1-4c1a-9620-059aa8dfb50c.png)
![image](https://user-images.githubusercontent.com/71438617/175800425-a6960296-96d2-41ac-a742-357a61552a1e.png)

elahehrashedi:
We are working on a fix for this in this PR #2627

PR #2627, opened 2022-06-28
search for LLVM clang-cl compiler

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2622

a sample clang kit:
```
  {
    "name": "Clang 12.0.0 (MSVC CLI) for MSVC 16.11.32602.291 (Visual Studio Enterprise 2019 Release - x86)",
    "visualStudio": "dc3c7e8c",
    "visualStudioArchitecture": "x86",
    "preferredGenerator": {
      "name": "Visual Studio 16 2019",
      "platform": "x86",
      "toolset": "host=x86"
    },
    "compilers": {
      "C": "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Tools\\Llvm\\bin\\clang-cl.exe",
      "CXX": "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Tools\\Llvm\\bin\\clang-cl.exe"
    }
  }
```


src/kit.ts
```diff
@@ -796,21 +798,21 @@ async function scanDirForClangForMSVCKits(dir: string, vsInstalls: VSInstallatio
...
         const clangKits: Kit[] = [];
-        vsInstalls.forEach(vs => {
+        vsInstalls.forEach(async vs => {
             const install_name = vsDisplayName(vs);
             const vs_arch = (version.target && version.target.triple.includes('i686-pc')) ? 'x86' : 'amd64';
 
             const clangArch = (vs_arch === "amd64") ? "x64\\" : "";
-            const clangKitName = `Clang ${version.version} ${clang_cli} for MSVC ${vs.installationVersion} (${install_name} - ${vs_arch})`;
-            if (binPath.startsWith(`${vs.installationPath}\\VC\\Tools\\Llvm\\${clangArch}bin`) &&
-                util.checkFileExists(util.lightNormalizePath(binPath))) {
```
elahehrashedi, line 806: util.checkFileExists(util.lightNormalizePath(binPath) was never awaited for, so it was always true. I put an "await" on it, but it gets stuck searching for clang.exe files.

bobbrow: I think we should validate the path before adding the Kit. You can show me what you're seeing in our next meeting.

bobbrow: It could actually be that `.forEach` want a sync function.  Convert it to a regular for loop and I think it will be better.
```ts
  for (const vs of vsInstalls) {
    ...
  }
```

src/kit.ts
```diff
@@ -798,17 +800,31 @@ async function scanDirForClangForMSVCKits(dir: string, vsInstalls: VSInstallatio
...
-                        CXX: binPath
-                    }
-                });
+            const useClangVersion: boolean = versionLess(version.version, '14.0.0');
+            const clangKitName: string = (isClangCL ? `Clang-cl` : `Clang`) + (useClangVersion ? ` ${version.version}` : "") + ` ${clang_cli} for ${install_name} - ${vs_arch} (MSVC ${vs.installationVersion})`;
+            const clangCXX = isClangCL ? binPath : binPath.replace(/^clang/, 'clang++');
+            const installationVersion = /^(\d+)+./.exec(vs.installationVersion);
+            let generatorName: string | undefined;
+            if (installationVersion) {
+                generatorName = VsGenerators[installationVersion[1]];
```
bobbrow, line 809: Let's declare the variable here and then move the check on line 811 inside this block.  You can then remove the check from line 812 and just push the kit.

src/kit.ts
```diff
@@ -798,17 +800,31 @@ async function scanDirForClangForMSVCKits(dir: string, vsInstalls: VSInstallatio
             const vs_arch = (version.target && version.target.triple.includes('i686-pc')) ? 'x86' : 'amd64';
 
             const clangArch = (vs_arch === "amd64") ? "x64\\" : "";
-            const clangKitName = `Clang ${version.version} ${clang_cli} for MSVC ${vs.installationVersion} (${install_name} - ${vs_arch})`;
-            if (binPath.startsWith(`${vs.installationPath}\\VC\\Tools\\Llvm\\${clangArch}bin`) && await util.checkFileExists(util.lightNormalizePath(binPath))) {
```
bobbrow, line 802: We should keep the `checkFileExists` check.

src/kit.ts
```diff
@@ -798,17 +800,31 @@ async function scanDirForClangForMSVCKits(dir: string, vsInstalls: VSInstallatio
...
+            if (installationVersion) {
+                generatorName = VsGenerators[installationVersion[1]];
+            }
+            if (binPath.startsWith(`${vs.installationPath}\\VC\\Tools\\Llvm\\${clangArch}bin`)) {
+                if (generatorName) {
+                    clangKits.push({
+                        name: clangKitName,
+                        visualStudio: kitVSName(vs),
+                        visualStudioArchitecture: vs_arch,
+                        preferredGenerator: {
```
bobbrow, line 817: I think we only want to add preferred generator for MSVC CLI.

src/kit.ts
```diff
@@ -798,17 +800,31 @@ async function scanDirForClangForMSVCKits(dir: string, vsInstalls: VSInstallatio
...
-                    name: clangKitName,
-                    visualStudio: kitVSName(vs),
-                    visualStudioArchitecture: vs_arch,
-                    compilers: {
-                        C: binPath,
-                        CXX: binPath
-                    }
-                });
+            const useClangVersion: boolean = versionLess(version.version, '14.0.0');
+            const clangKitName: string = (isClangCL ? `Clang-cl` : `Clang`) + (useClangVersion ? ` ${version.version}` : "") + ` ${clang_cli} for ${install_name} - ${vs_arch} (MSVC ${vs.installationVersion})`;
```
bobbrow, line 804: I think we should keep the format of the kit name the same for now so that duplicate kits don't get generated.

src/kit.ts
```diff
@@ -796,21 +798,49 @@ async function scanDirForClangForMSVCKits(dir: string, vsInstalls: VSInstallatio
...
-            if (binPath.startsWith(`${vs.installationPath}\\VC\\Tools\\Llvm\\${clangArch}bin`) && await util.checkFileExists(util.lightNormalizePath(binPath))) {
-                clangKits.push({
-                    name: clangKitName,
-                    visualStudio: kitVSName(vs),
-                    visualStudioArchitecture: vs_arch,
-                    compilers: {
-                        C: binPath,
-                        CXX: binPath
+            const clangKitName: string = `Clang ${version.version} ${clang_cli} for MSVC ${vs.installationVersion} (${install_name} - ${vs_arch})`;
+            const clangCXX = isClangMsvcCli ? binPath : binPath.replace(/^clang/, 'clang++');
```
bobbrow: Let's not do this now.  It should be ok not to pick the clang++

src/kit.ts
```diff
@@ -796,21 +798,48 @@ async function scanDirForClangForMSVCKits(dir: string, vsInstalls: VSInstallatio
...
+                const generatorName: string | undefined = installationVersion ? VsGenerators[installationVersion[1]] : undefined;
+                if (generatorName) {
+                    if (await clangExists()) {
+                        clangKits.push({
+                            name: clangKitName,
+                            visualStudio: kitVSName(vs),
+                            visualStudioArchitecture: vs_arch,
+                            preferredGenerator: {
+                                name: generatorName,
+                                platform: vs_arch,
```
bobbrow, line 830: When this is 'x86' it needs to be converted to 'win32'

src/kit.ts
```diff
@@ -809,7 +809,7 @@ async function scanDirForClangForMSVCKits(dir: string, vsInstalls: VSInstallatio
                     clangKits.push({
                         name: clangKitName,
                         visualStudio: kitVSName(vs),
-                        visualStudioArchitecture: vs_arch,
+                        visualStudioArchitecture: generatorPlatformFromVSArch[vs_arch] as string || vs_arch,
```
bobbrow, line 812: This line is incorrect.  You only do it for the `platform`.  This should stay as `vs_arch`

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Bug Fixes:
 - CMakeUserPresets.json version not detected without CMakePresets.json. [#2469](https://github.com/microsoft/vscode-cmake-tools/issues/2469) [@chausner](https://github.com/chausner)
 - Clear output channel after auto-reconfigure. [#2628](https://github.com/microsoft/vscode-cmake-tools/issues/2628)
 - Don't delete CMakeCache.txt when switching kits if the buildDirectory also changes. [#2546](https://github.com/microsoft/vscode-cmake-tools/issues/2546) [@david-fong](https://github.com/david-fong)
+- Detect clang-cl.exe compilers. [#2622](https://github.com/microsoft/vscode-cmake-tools/issues/2622)
```
bobbrow, line 14: Add "not bundled with Visual Studio"

bobbrow (approved):

PR #2630, opened 2022-07-01
Enable no misused promises lint rule

bobbrow:
Noticed a missed await in a PR, and realized we didn't have this rule on.


test/unit-tests/config.test.ts
```diff
@@ -80,23 +80,49 @@ suite('Configuration', () => {
         expect(conf.parallelJobs).to.eq(4);
     });
 
-    test('Listen for config changes', () => {
+    test('Listen for config changes', async () => {
         const conf = createConfig({ parallelJobs: 22 });
         let jobs = conf.parallelJobs;
         expect(jobs).to.eq(22);
-        conf.onChange('parallelJobs', j => jobs = j);
```
elahehrashedi, line 87: I am trying to understand why do we need a promise here? why the previous approach not working?

src/kit.ts
```diff
@@ -796,14 +793,13 @@ async function scanDirForClangForMSVCKits(dir: string, vsInstalls: VSInstallatio
         }
 
         const clangKits: Kit[] = [];
-        vsInstalls.forEach(vs => {
+        for (const vs of vsInstalls) {
```
elahehrashedi, line 796: this may help with the problem I had in PR https://github.com/microsoft/vscode-cmake-tools/pull/2627. I will check it after merging this branch.

test/unit-tests/config.test.ts
bobbrow: The previous approach seems to work usually, but there is async going on that isn't being accounted for. I don't think it currently matters, but if it ever runs on another thread, this test will break unless we fix it.

Issue #2553, opened 2022-05-17
Passing environment variable to "cmake" tasks


TheophileBlard:
### Brief Issue Summary

Is it possible to pass environment variable to cmake tasks ?

I'm currently using the pre-release version of the extension, and want to use cmake tasks to install the binaries to predefined directories using [DESTDIR](https://cmake.org/cmake/help/latest/envvar/DESTDIR.html). For example, "shell" tasks provide the "env" keyword in "options", but for "cmake" tasks, only the "cwd" is available.

Thanks!

### Additional Information

CMakeTools version : `1.11.19`
VSCode verson: `1.67.0`
OS: Ubuntu 21.04 TLS

elahehrashedi:
You can define `configurePresets` in `CMakePresets.json`, and set the `CMAKE_INSTALL_PREFIX` there. 
Here is an example:
```
{
    "version": 2,
    "configurePresets": [
        {
            "name": "dev17-ninja-143",
            "displayName": "Ninja - v143,x64",
            "generator": "Ninja",
            "architecture": {
                "strategy": "external",
                "value": "x64"
            },
            "toolset": {
                "strategy": "external",
                "value": "host=x64,version=14.30"
            },
            "binaryDir": "${sourceDir}/out/build/${presetName}",
            "cacheVariables": {
                "CMAKE_BUILD_TYPE": "Debug",
                "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}"
            },
            "inherits": "base",
            "environment": {
                "PATH":  "$penv{PATH};D:\\ninja"
              }
        }
    ],
```
Will this approach solve your issue?
We can also look into adding more customizable features to cmake tasks in future.

elahehrashedi:
One option that can be added to the `configure` or `build` task can be the name of the preset (instead of always picking the default).
for example:
```json
        {
            "type": "cmake",
            "label": "CMake: build",
            "command": "build",
            "targets": [
                "all"
            ],
            **"preset": "presetName",**
            "group": "build",
            "problemMatcher": [],
            "detail": "CMake template build task"
        }
```

TheophileBlard:
Hi!

The `DESTDIR` variable is not interpreted at configure time but at install time, and it's not a CMake variable, but an environment variable that CMake uses to relocate the installation.
A typical use of this variable is: 
```sh
DESTDIR=/package/stage cmake --install .
```

[...]

PR #2636, opened 2022-07-05
Support for "preset" and "env" in task provider

elahehrashedi:
feature-request: https://github.com/microsoft/vscode-cmake-tools/issues/2553
bug fix: [CMake install does not run in terminal as a task](https://github.com/microsoft/vscode-cmake-tools/issues/2693)

sample:
```json
{
	"version": "2.0.0",
	"tasks": [
		{
			"type": "cmake",
			"label": "1CMake build with presets",
			"command": "build",
			"targets": [
				"main"
			],
			"preset": "${command:cmake.activeBuildPresetName}",
			"options": {
				"cwd": "${command:cmake.buildDirectory}"
			},
			"dependsOn": [
				"1CMake configure with presets"
			],
			"detail": "CMake build with presets"
		},
		{
			"type": "cmake",
			"label": "1CMake configure with presets",
			"command": "configure",
			"preset": "${command:cmake.activeConfigurePresetName}",
			"options": {
				"cwd": "${workspaceFolder}/"
			},
			"detail": "CMake configure with presets"
		},
		{
			"type": "cmake",
			"label": "2CMake build without presets",
			"command": "build",
			"targets": [
				"main"
			],
			"options": {
				"cwd": "${command:cmake.buildDirectory}"
			},
			"group": "build",
			"problemMatcher": [],
			"dependsOn": [
				"2CMake configure without presets"
			],
			"detail": "CMake build without presets"
		},
		{
			"type": "cmake",
			"label": "2CMake configure without presets",
			"command": "configure",
			"options": {
				"cwd": "${command:cmake.buildDirectory}"
			},
			"problemMatcher": [],
			"detail": "CMake configure without presets"
		}
	]
}
```


src/drivers/cmakeDriver.ts
```diff
@@ -112,6 +112,10 @@ export abstract class CMakeDriver implements vscode.Disposable {
      */
     abstract get targets(): api.Target[];
 
+    public get targetsName(): string[] {
```
bobbrow, line 115: Since `targets` is already available to use, I don't think we need to add this API.  You can just do it directly in the task provider.

src/cmakeTaskProvider.ts
```diff
@@ -65,45 +68,66 @@ export class CMakeTask extends vscode.Task {
...
+                case CommandType.build:
+                    preset = "${command:cmake.activeBuildPresetName}";
+                    break;
+                case CommandType.test:
+                    preset = "${command:cmake.activeTestPresetName}";
+                    break;
+                default:
+                    preset = undefined;
+            }
+            options.cwd = "${workspaceFolder}";
```
bobbrow, line 117: I believe that all of the drivers default to using the buildDirectory as the cwd for configure and test.  Oddly, it doesn't look like we set it for build.

I believe you can use `${command:cmake.buildDirectory}` to get it.

src/cmakeTaskProvider.ts
```diff
@@ -65,45 +68,66 @@ export class CMakeTask extends vscode.Task {
...
 
-    public async provideTask(commandType: CommandType): Promise<CMakeTask> {
+    public async provideTask(commandType: CommandType, cmakeDriver?: CMakeDriver, useCMakePresets?: boolean): Promise<CMakeTask> {
         const taskName: string = localizeCommandType(commandType);
+        let targets: string[] | undefined;
+        let preset: string | undefined;
+        const options: { cwd?: string ; environment?: Environment } = {};
+        if (!useCMakePresets) {
+            if (commandType === CommandType.build) {
+                if (cmakeDriver?.targetsName) {
```
bobbrow, line 97: Do we really want to list all the targets by default?  There could be hundreds!

src/cmakeTaskProvider.ts
```diff
@@ -182,29 +207,32 @@ class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.OutputConsu
...
+                this.options = {};
             }
-            command = await this.cmakeDriver.getCMakeBuildCommand(this.definedTargets ? this.definedTargets : this.defaultTargets);
-            if (command) {
-                cmakePath = command.command;
-                args = command.args ? command.args : [];
-                this.options.environment = command.build_env;
+            if (this.preset && this.preset.length !== 0) {
+                cmakePath = cmakeDriver.getCMakeCommand();
+                args = [ "--build", "--preset", this.preset];
```
bobbrow, line 219: Are we calling `expandString` on `this.preset` somewhere?  I didn't see it.

bobbrow, line 219: The environment needs to be set for the presets case too.

package.json
```diff
@@ -543,27 +543,48 @@
...
             "description": "Additional command options",
             "properties": {
               "cwd": {
                 "type": "string",
-                "description": "The current working directory of the executed program or script. If omitted Code's current workspace root is used."
+                "description": "%cmake-tools.taskDefinitions.properties.options.cwd.description%"
+              },
+              "environment": {
+                "type": "object",
+                "items": {
```
bobbrow, line 569: I don't think we should do environment as "name" "value" explicit pairs.  We only do that for debugConfig to match what VS Code defined for debugging.  It's a very verbose and unfortunate thing.  I think we just want normal string/string key/values in the object.

src/cmakeTaskProvider.ts
```diff
@@ -224,31 +252,51 @@ class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.OutputConsu
...
+        let result: number | undefined | null;
+        if (this.preset) {
+            const cmakeTools: CMakeTools | undefined = getCMakeToolsForActiveFolder();
+            const cmakeDriver: CMakeDriver | undefined = (await cmakeTools?.getCMakeDriverInstance()) || undefined;
+            const cmakePath: string = cmakeDriver?.getCMakeCommand() || "CMake.EXE";
+            const args: string[] = [ "--preset", this.preset];
+            const execResult = await proc.execute(cmakePath, args, this, this.options).result;
+            result = execResult?.retc;
+        } else {
+            result =  await vscode.commands.executeCommand('cmake.configure');
```
bobbrow, line 265: Since we are adding options to set the environment and cwd for the task, we need to figure out a way to make it apply to all the commands.  It will be very confusing to people if they set it and it only works for presets.

And now that I'm thinking about it more, I'm wondering if we should be treating presets differently here vs just calling `executeCommand` and letting the extension do the rest. Part of the value of the extension is that it can support presets even for versions of cmake that don't.

bobbrow (changes requested):
:clock1:

package.json
elahehrashedi:
```
	"options": {
		"environment": {
			"myenv1": "myenv1"
		}
	},
```
this is how we define the env in tasks.json. 

for `debugconfig` we define it like this:
`    "cmake.debugConfig": {"environment": [{"name":"myenv1","value": "myenv1"}]}`

src/cmakeTaskProvider.ts
elahehrashedi: I added the cwd for both preset and non-preset case for all tasks.
it's `${command:cmake.buildDirectory}` now.

src/cmakeTaskProvider.ts
elahehrashedi: I replaced this with default targets.

src/cmakeTaskProvider.ts
elahehrashedi: yes, it is already resolved in resolveTask. When the CustomBuildTaskTerminal is initiated, the resolved preset name is passed to it.

bobbrow (commented):
Not finished revieing, but I'll send this now.

bobbrow (changes requested):

src/ctest.ts
```diff
@@ -362,9 +362,11 @@ export class CTestDriver implements vscode.Disposable {
     private readonly _resultsChangedEmitter = new vscode.EventEmitter<BasicTestResults | null>();
     readonly onResultsChanged = this._resultsChangedEmitter.event;
 
-    async runCTest(driver: CMakeDriver): Promise<number> {
-        log.showChannel();
-        this._decorationManager.clearFailingTestDecorations();
+    public async runCTest(driver: CMakeDriver, customized: boolean = false, testPreset?: TestPreset): Promise<number|null> {
```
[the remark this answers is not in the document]
elahehrashedi: I remembered that we need that customized flag because there are two ways we customize the test, one is using a test preset, and one is without testpreset based on settings.

src/drivers/cmakeLegacyDriver.ts
```diff
@@ -108,10 +108,12 @@ export class CMakeLegacyDriver extends CMakeDriver {
             }).result;
             log.trace(res.stderr);
             log.trace(res.stdout);
-            if (res.retc === 0) {
-                this._needsReconfigure = false;
+            if (!taskCustomConfig) {
```
elahehrashedi, line 111: I need your opinion on this.

src/drivers/cmakeFileApiDriver.ts
```diff
@@ -221,9 +221,11 @@ export class CMakeFileApiDriver extends CMakeDriver {
         return 0;
     }
 
-    async doConfigure(args_: string[], outputConsumer?: proc.OutputConsumer, showCommandOnly?: boolean): Promise<number> {
-        const api_path = this.getCMakeFileApiPath();
-        await createQueryFileForApi(api_path);
+    async doConfigure(args_: string[], outputConsumer?: proc.OutputConsumer, showCommandOnly?: boolean, taskCustomConfig?: boolean): Promise<number> {
+        if (!taskCustomConfig) {
```
elahehrashedi, line 225: I need your opinion on this.

bobbrow (commented):
A few more comments on the stuff I reviewed this morning.  I still have a bit more to review.

bobbrow (commented):
Sorry, still not done reviewing. Here's a few comments that I had pending from today.

src/cmakeTaskProvider.ts
```diff
@@ -219,37 +339,86 @@ class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.OutputConsu
...
-        const result: number | undefined =  await vscode.commands.executeCommand('cmake.install');
-        this.closeEmitter.fire(result ? result : -1);
+        this.writeEmitter.fire(localize("install.started", "Install task started...") + endOfLine);
+        this.targetIsIgnored();
+        const cmakeTools: CMakeTools | undefined = this.getCMakeTools();
+        if (!cmakeTools) {
+            return;
+        }
+        const result: number | undefined =  await cmakeTools.runBuild(['install'], false, this);
+        if (result === undefined || result === null) {
```
bobbrow, line 355: `null` is not one of the type declarations for `result`. Do you need to check it?

src/cmakeTaskProvider.ts
```diff
@@ -178,33 +198,133 @@ class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.OutputConsu
...
+    }
+    private async isTaskCompatibleWithPresets(cmakeTools: CMakeTools): Promise<boolean> {
+        const useCMakePresets: boolean = cmakeTools?.useCMakePresets;
+        const presetDefined: boolean = this.preset !== undefined && this.preset !== null;
+        const isCompatible = (useCMakePresets && presetDefined) || (!useCMakePresets && !presetDefined);
+        if (isCompatible) {
+            return true;
+        }
+        const change: string = localize('change.preset.setting', "Change preset settings");
+        const ignore: string = localize('ignore', "Ignore");
```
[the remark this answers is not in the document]
elahehrashedi: The way this works is that:
1. we show this Error message but won't wait on what the user decides to do. 
2. we fail the task right away and show `The selected task is not compatible with preset setting` in the terminal, (while the user can see the error message and change the settings if they want to).

We are not awaiting for user to change the setting or not. After they changed it using the "change" button, they can re-run the task. 

So, in this design, the task will fail, but we help the user with the failure error by showing them the  "change" message.

src/cmakeTaskProvider.ts
```diff
@@ -178,33 +198,133 @@ class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.OutputConsu
...
-        let args: string[] = [];
+    private targetIsIgnored(): void {
+        if (this.targets && this.targets.length > 0) {
+            this.writeEmitter.fire(localize("target.is.ignored", "The defined targets in this task are being ignored.") + endOfLine);
+        }
+    }
+    private async isTaskCompatibleWithPresets(cmakeTools: CMakeTools): Promise<boolean> {
+        const useCMakePresets: boolean = cmakeTools?.useCMakePresets;
+        const presetDefined: boolean = this.preset !== undefined && this.preset !== null;
+        const isCompatible = (useCMakePresets && presetDefined) || (!useCMakePresets && !presetDefined);
```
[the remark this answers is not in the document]
elahehrashedi: OK, now I default to active preset if useCMakePresets is true.

src/cmakeTaskProvider.ts
```diff
@@ -178,33 +198,133 @@ class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.OutputConsu
...
-        let command: proc.BuildCommand | null;
-        let cmakePath: string = "CMake.EXE";
-        let args: string[] = [];
+    private targetIsIgnored(): void {
+        if (this.targets && this.targets.length > 0) {
+            this.writeEmitter.fire(localize("target.is.ignored", "The defined targets in this task are being ignored.") + endOfLine);
+        }
+    }
+    private async isTaskCompatibleWithPresets(cmakeTools: CMakeTools): Promise<boolean> {
+        const useCMakePresets: boolean = cmakeTools?.useCMakePresets;
```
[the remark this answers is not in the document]
elahehrashedi: In CMakeTools this has `true ` or `false ` value in it. We read it from the settings, but the default value is `false`.

src/cmakeTaskProvider.ts
```diff
@@ -178,33 +198,133 @@ class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.OutputConsu
...
+                }
+            }
+        });
+        this.writeEmitter.fire(localize('task.not.compatible.with.preset.setting', 'The selected task is not compatible with preset setting.') + endOfLine);
+        this.closeEmitter.fire(-1);
+        return false;
+    }
+
+    private getCMakeTools(): CMakeTools | undefined {
+        const cmakeTools: CMakeTools | undefined = getCMakeToolsForActiveFolder();
```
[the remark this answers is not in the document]
elahehrashedi: the **default build targets**, the **preset expanded info**, and `useCMakePreset` info are in `CMakeTools`.

the functions to generate the command is in `CMakeDriver`.

src/drivers/cmakeFileApiDriver.ts
```diff
@@ -257,14 +256,20 @@ export class CMakeFileApiDriver extends CMakeDriver {
...
+                log.debug("Configuring using tasks");
+            }
             log.debug('Invoking CMake', cmake, 'with arguments', JSON.stringify(args));
             const env = await this.getConfigureEnvironment();
             const res = await this.executeCommand(cmake, args, outputConsumer, { environment: env, cwd: this.binaryDir }).result;
             log.trace(res.stderr);
             log.trace(res.stdout);
             if (res.retc === 0) {
-                this._needsReconfigure = false;
+                if (!taskCustomConfig) {
```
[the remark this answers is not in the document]
elahehrashedi: a task can configure using a different config preset than the default config preset. 

If it is not a task related configure, it should be set to false,
if it is a task, it can be set to false if the config preset is the same as the default preset. we don't have the preset name info in this function, we can send it though.

src/drivers/cmakeDriver.ts
```diff
@@ -1203,6 +1203,33 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+        const expanded_configure_env = await this.getConfigureEnvironment();
+
+        // Expand all flags
+        const opts = this.expansionOptions;
+        const expanded_flags_promises = final_flags.map(
+            async (value: string) => expand.expandString(value, { ...opts, envOverride: expanded_configure_env }));
+        return Promise.all(expanded_flags_promises);
+    }
+
+    public async taskCustomConfigure(args: string[], taskConsumer?: proc.OutputConsumer): Promise<number> {
```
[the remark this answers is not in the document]
elahehrashedi: Oh, that is a great idea, I wish I would come up with it before, Let me try this approach.

src/cmakeTaskProvider.ts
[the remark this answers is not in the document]
bobbrow: Then let's change "ignore" to "OK" and they just acknowledge that we sent the error.

src/cmakeTaskProvider.ts
```diff
@@ -62,50 +66,65 @@ export class CMakeTask extends vscode.Task {
...
+            preset: preset
         };
         const task = new vscode.Task(definition, vscode.TaskScope.Workspace, taskName, CMakeTaskProvider.CMakeSourceStr,
             new vscode.CustomExecution(async (resolvedDefinition: vscode.TaskDefinition): Promise<vscode.Pseudoterminal> =>
                 // When the task is executed, this callback will run. Here, we setup for running the task.
-                new CustomBuildTaskTerminal(resolvedDefinition.command, this.defaultTargets, resolvedDefinition.targets, resolvedDefinition.options, this.cmakeDriver)
-            ), []);
-        task.group = commandType === CommandType.build ? vscode.TaskGroup.Build : undefined;
+                new CustomBuildTaskTerminal(resolvedDefinition.command, resolvedDefinition.targets, resolvedDefinition.preset, {})
+            ), '$gcc');
```
bobbrow, line 126: We don't want to set a default problem matcher.

src/drivers/cmakeDriver.ts
```diff
@@ -1223,7 +1247,8 @@ export abstract class CMakeDriver implements vscode.Disposable {
         try {
             // _beforeConfigureOrBuild needs to refresh expansions early because it reads various settings
             // (example: cmake.sourceDirectory).
-            await this._refreshExpansions();
+            const configurePreset: preset.ConfigurePreset | undefined | null = (trigger === ConfigureTrigger.taskProvider) ? taskPreset : this._configurePreset;
+            await this._refreshExpansions(taskPreset);
```
bobbrow, line 1251: Should this be `configurePreset` instead of `taskPreset`?

src/drivers/cmakeDriver.ts
```diff
@@ -1286,7 +1297,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
             if (this.useCMakePresets) {
                 telemetryProperties = {
                     CMakeExecutableVersion: cmakeVersion ? util.versionToString(cmakeVersion) : '',
-                    CMakeGenerator: this.getGeneratorNameForTelemetry(),
+                    CMakeGenerator: (trigger === ConfigureTrigger.taskProvider) ? taskPreset?.generator || "" : this.getGeneratorNameForTelemetry(),
```
bobbrow, line 1300: We don't want to skip the PII scrub in the generator name for telemetry. You will need to send the override preset to `getGeneratorNameForTelemetry`

src/drivers/cmakeFileApiDriver.ts
```diff
@@ -221,36 +221,47 @@ export class CMakeFileApiDriver extends CMakeDriver {
...
+                    args.push(typeof(configurePreset.toolset) === "string" ? configurePreset.toolset as string : configurePreset.toolset.value || "");
+                }
             }
-            if (gen.platform) {
-                args.push('-A');
-                args.push(gen.platform);
+        } else {
+            args.push(`-B${util.lightNormalizePath(this.binaryDir)}`);
+            const generator = this.generator;
+            if (!has_gen && generator) {
```
bobbrow, line 251: rather than have two copies of this code, I think it would be better to construct a `const generator` based off the data in the `configurePreset` and then have both paths run through the same code.

src/drivers/cmakeFileApiDriver.ts
```diff
@@ -259,13 +270,15 @@ export class CMakeFileApiDriver extends CMakeDriver {
         } else {
             log.debug(`Configuring using ${this.useCMakePresets ? 'preset' : 'kit'}`);
             log.debug('Invoking CMake', cmake, 'with arguments', JSON.stringify(args));
-            const env = await this.getConfigureEnvironment();
-            const res = await this.executeCommand(cmake, args, outputConsumer, { environment: env, cwd: this.binaryDir }).result;
+            const env = await this.getConfigureEnvironment(configurePreset);
+            const res = await this.executeCommand(cmake, args, outputConsumer, { environment: env, cwd: (configurePreset && configurePreset.binaryDir) ? configurePreset.binaryDir : this.binaryDir }).result;
```
bobbrow, line 274: Line 239 already has this logic for picking the binaryDir. Consider storing it in a variable and using that instead.

src/drivers/cmakeLegacyDriver.ts
```diff
@@ -85,14 +85,23 @@ export class CMakeLegacyDriver extends CMakeDriver {
         const args = Array.from(args_);
         args.push(util.lightNormalizePath(this.sourceDir));
 
-        const gen = this.generator;
-        if (gen) {
-            args.push(`-G${gen.name}`);
-            if (gen.toolset) {
-                args.push(`-T${gen.toolset}`);
+        if (configurePreset && configurePreset.generator) {
+            args.push('-G');
```
bobbrow, line 89: Same feedback as the file api driver. Construct `const gen` from the preset and use it here instead of having two copies of the code.

src/drivers/cmakeDriver.ts
elahehrashedi: when we use kits and there is no generator defined, we use `getGeneratorNameForTelemetry` to get the best generator, but when we are using the presets, if the generator entry is empty, we only log a message and won't find the best.

https://github.com/microsoft/vscode-cmake-tools/blob/9a9b1a51c582b0f93cc10e0d083915f55d4495db/src/drivers/cmakeDriver.ts#L459

I can send a specific message in telemetry instead of `""`, for example, `"no.generator"`, would that be good?

src/drivers/cmakeDriver.ts
elahehrashedi: It actually has the same results, but I wanted pass "undefined"  for all the scenarios that doesn't have a `presetOverride` to keep the behavior as before. To avoid confusion I moved the `configurePreset` a few lines below where it is actually being used.

src/drivers/cmakeLegacyDriver.ts
elahehrashedi: Thank you, the legacyDriver needed some other changes too.

src/preset.ts
```diff
@@ -1657,3 +1657,11 @@ export function configurePresetChangeNeedsClean(newPreset: ConfigurePreset, oldP
         return false;
     }
 }
+
+export function getValueStrategy(value: string | ValueStrategy): string | undefined {
```
bobbrow, line 1661: I don't think `strategy` should be in the name.  You don't get the strategy, you get the value.

src/drivers/cmakeDriver.ts
bobbrow: My concern is about sending `presetOverride?.generator` directly to telemetry.  That value needs to be scrubbed for PII first.

src/drivers/cmakeFileApiDriver.ts
```diff
@@ -221,36 +221,46 @@ export class CMakeFileApiDriver extends CMakeDriver {
...
-            args.push('-G');
-            args.push(gen.name);
-            if (gen.toolset) {
-                args.push('-T');
-                args.push(gen.toolset);
-            }
-            if (gen.platform) {
-                args.push('-A');
-                args.push(gen.platform);
+        const binaryDir = configurePreset?.binaryDir ? configurePreset.binaryDir : this.binaryDir;
```
bobbrow, line 236: It should be simplified to `configurePreset?.binaryDir ?? this.binaryDir`

(if `??` doesn't work, use `||` instead)

src/drivers/cmakeFileApiDriver.ts
```diff
@@ -276,17 +288,17 @@ export class CMakeFileApiDriver extends CMakeDriver {
         return true;
     }
 
-    private getCMakeFileApiPath() {
-        return path.join(this.binaryDir, '.cmake', 'api', 'v1');
+    private getCMakeFileApiPath(binaryDir?: string) {
+        return path.join(binaryDir ? binaryDir : this.binaryDir, '.cmake', 'api', 'v1');
```
bobbrow, line 292: preferred syntax: `binaryDir ?? this.binaryDir`

src/drivers/cmakeLegacyDriver.ts
```diff
@@ -76,25 +76,36 @@ export class CMakeLegacyDriver extends CMakeDriver {
         this._cacheWatcher.dispose();
     }
 
-    async doConfigure(args_: string[], outputConsumer?: proc.OutputConsumer, showCommandOnly?: boolean): Promise<number> {
+    async doConfigure(args_: string[], outputConsumer?: proc.OutputConsumer, showCommandOnly?: boolean, configurePreset?: ConfigurePreset | null): Promise<number> {
         // Ensure the binary directory exists
-        const binary_dir = this.binaryDir;
-        await fs.mkdir_p(binary_dir);
+        const binaryDir = configurePreset?.binaryDir ? configurePreset.binaryDir : this.binaryDir;
```
bobbrow, line 81: `configurePreset?.binaryDir ?? this.binaryDir`

src/drivers/cmakeLegacyDriver.ts
```diff
@@ -103,15 +114,15 @@ export class CMakeLegacyDriver extends CMakeDriver {
...
-                cwd: binary_dir
+                environment: await this.getConfigureEnvironment(configurePreset),
+                cwd: binaryDir
             }).result;
             log.trace(res.stderr);
             log.trace(res.stdout);
-            if (res.retc === 0) {
+            if (res.retc === 0 && !configurePreset) {
                 this._needsReconfigure = false;
+                await this._reloadPostConfigure();
```
bobbrow, line 124: reloadPostConfigure didn't used to be dependent on the return code.  I think it would be better to put it back, but guard with a `if (!configurePreset)`

```
if (res.retc === 0 && !configurePreset) {
  this._needsReconfigure = false;
}
if (!configurePreset) {
  await this._reloadPostConfigure();
}

Comment (elahehrashedi) [reply src/drivers/cmakeLegacyDriver.ts:124]: that's right

**Comment by bobbrow** [comment]
> 	"__metadata": {

This property should be removed from `package.json`

---
Refers to: package.json:2153 in 5602eb6. [](commit_id = 5602eb6898c42d9a5e4b781852b6961ced7df107, deletion_comment = False)

File: src/drivers/cmakeDriver.ts [hunk 28]
```diff
@@ -1286,7 +1290,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
             if (this.useCMakePresets) {
                 telemetryProperties = {
                     CMakeExecutableVersion: cmakeVersion ? util.versionToString(cmakeVersion) : '',
-                    CMakeGenerator: this.getGeneratorNameForTelemetry(),
+                    CMakeGenerator: (trigger === ConfigureTrigger.taskProvider) ? presetOverride?.generator || "no.generator" : this.getGeneratorNameForTelemetry(),
```
bobbrow, line 1293: `presetOverride?.generator` needs to be scrubbed for PII.  Use "" instead of "no.generator" otherwise.

bobbrow (changes requested):

src/drivers/cmakeDriver.ts
elahehrashedi: Oh Ok, now I got what you mean.

PR #2655, opened 2022-07-22
Speed up Windows tests

elahehrashedi:


test/extension-tests/successful-build/test/variable-substitution.test.ts
```diff
@@ -6,7 +6,7 @@ import { objectPairs, platformNormalizePath, makeHashString } from '@cmt/util';
 import { clearExistingKitConfigurationFile, DefaultEnvironment, expect, getFirstSystemKit } from '@test/util';
 import * as path from 'path';
 
-suite('Variable Substitution', () => {
+suite('Variable and Variant Substitution', () => {
```
bobbrow, line 9: It's still just variable substitution even though it's using variant syntax

bobbrow (approved):
The updated test appears to pass despite the other test being unable to find Clang reliably.  You can check this update in.

PR #2660, opened 2022-07-26
Use Ninja Generator to speed up tests

elahehrashedi:
image


test/unit-tests/driver/driver-test.ts
```diff
@@ -82,40 +81,38 @@ export function makeDriverTestsuite(driverName: string, driver_generator: (cmake
...
-        test(`All target for ${kitDefault.name}`, async () => {
+        test(`All target for ${ninjaKitDefault.name}`, async () => {
             const config = ConfigurationReader.create();
             const executable = await getCMakeExecutableInformation(cmakePath);
 
-            driver = await driver_generator(executable, config, kitDefault, defaultWorkspaceFolder, async () => {}, []);
+            driver = await driver_generator(executable, config, ninjaKitDefault, defaultWorkspaceFolder, async () => {}, []);
             const allTargetName = driver.allTargetName;
-            if (process.platform === 'win32') {
-                expect(allTargetName).to.eq('ALL_BUILD');
```
bobbrow, line 92: can we check this in the test that uses the VS generator?

bobbrow (approved):
Move the deleted check to one of the VS tests, then this is ok to commit.

PR #2662, opened 2022-07-27
fix the race between timeout and close in proc.execute

elahehrashedi:


src/proc.ts
```diff
@@ -164,94 +170,87 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
...
-                            stderr_line_acc = '';
-                            // Erase the first line from the list
-                            lines.splice(0, 1);
+            });
+            // The 'close' event is emitted after a process has ended and the stdio streams of a child process have been closed.
+            // This is distinct from the 'exit' event, since multiple processes might share the same stdio streams.
+            // The 'close' event will always emit after 'exit' was already emitted, or 'error' if the child failed to spawn.
+            child?.on('close', retc => {
+                if (retc !== 0) {
+                    log.warning(localize('process.close', 'The command: {0} closed with code: {1}', `${cmdstr}`, `${retc}`));
```
bobbrow, line 238: let's remove this log if we already logged for error/exit cases.

src/proc.ts
```diff
@@ -149,12 +153,14 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
             resolve({ retc: 0, stdout: '', stderr: '' });
         });
     } else {
+        // Since we won't be sending anything to this process, close its stdin.
+        spawn_opts.stdio = ['ignore', 'pipe', 'pipe'];
```
bobbrow, line 157: maybe let's remove this since this `execute` is a generic function, not just for querying compiler versions.

bobbrow (approved):

PR #2668, opened 2022-08-02
add stdio Spawn Option

elahehrashedi:


src/proc.ts
```diff
@@ -153,6 +153,8 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
             resolve({ retc: 0, stdout: '', stderr: '' });
         });
     } else {
+        // Since we won't be sending anything to this process, close its stdin.
+        spawn_opts.stdio = ['ignore', 'pipe', 'pipe'];
```
bobbrow, line 157: Do you know why 'pipe' is in the array twice?

elahehrashedi: corresponding to [stdin, stdout, and stderr]

bobbrow (approved):

PR #2681, opened 2022-08-08
markdown Description for LaunchBehavior

elahehrashedi:


package.nls.json
```diff
@@ -148,7 +148,7 @@
...
     "cmake-tools.configuration.cmake.launchBehavior.description": "Controls what happens with the launch terminal when you launch a target.",
-    "cmake-tools.configuration.cmake.launchBehavior.reuseTerminal.description": "The launch terminal instance is reused and the target will launch as soon as the terminal is idle.",
-    "cmake-tools.configuration.cmake.launchBehavior.breakAndReuseTerminal.description": "The launch terminal instance is reused and a break command is sent to terminate any active foreground process before launching the target.",
-    "cmake-tools.configuration.cmake.launchBehavior.newTerminal.description": "A new terminal instance is created and the target is launched in it. Existing terminals are not automatically cleaned up."
+    "cmake-tools.configuration.cmake.launchBehavior.reuseTerminal.markdownDescriptions": { "message":"The launch terminal instance is reused and the target will launch as soon as the terminal is idle.", "comment": [] },
```
bobbrow, line 151: If it doesn't need a comment, should these be left alone (line 151 & 153)?

elahehrashedi: They were a part of the same `markdownEnumDescriptions`.

bobbrow: do markdown descriptions have to have a comment?

elahehrashedi: it looks like it works without a comment (I tested it).
It didn't work if I define one in `"markdownEnumDescriptions"`  and the other two in `"enumDescriptions"`, 
but it works if I define all 3 in `"markdownEnumDescriptions"` and don't add `comment` to the other two.

bobbrow (approved):

PR #2683, opened 2022-08-09
Stabilize flaky test

bobbrow:


src/cmakeTools.ts
```diff
@@ -147,6 +147,7 @@ export class CMakeTools implements api.CMakeToolsAPI {
             const drv = await this.cmakeDriver;
             if (drv) {
                 log.debug(localize('disposing.driver', 'Disposing CMake driver'));
+                log.debug(new Error("getting callstack").stack || "");
```
elahehrashedi, line 150: Do we want to keep these new logs?

test/extension-tests/single-root-UI/test/envs-kits.test.ts
```diff
@@ -41,8 +41,6 @@ suite('Environment Variables in Variants', () => {
             await fs.rename(variantFileBackup, variantFile);
         }
 
-        await vscode.workspace.getConfiguration('cmake', vscode.workspace.workspaceFolders![0].uri).update('useCMakePresets', 'auto');
```
elahehrashedi, line 44: why were we changing this setting to 'auto'? Was it redundant from the get-go? because I can see we call `update('useCMakePresets', 'never')` or `update('useCMakePresets', 'always')` in the beginning of the tests anyways.

test/extension-tests/single-root-UI/test/envs-presets.test.ts
```diff
@@ -15,9 +15,11 @@ suite('Environment Variables in Presets', () => {
         testEnv = new DefaultEnvironment('test/extension-tests/single-root-UI/project-folder', build_loc, exe_res);
         testEnv.projectFolder.buildDirectory.clear();
 
-        await vscode.commands.executeCommand('cmake.setConfigurePreset', 'LinuxUser1');
+        await vscode.commands.executeCommand('cmake.setConfigurePreset', process.platform === 'win32' ? 'WindowsUser1' : 'LinuxUser1');
         await vscode.commands.executeCommand('cmake.setBuildPreset', '__defaultBuildPreset__');
         await vscode.commands.executeCommand('cmake.setTestPreset', '__defaultTestPreset__');
+
+        await vscode.workspace.getConfiguration('cmake', vscode.workspace.workspaceFolders![0].uri).update('useCMakePresets', 'always');
```
elahehrashedi, line 22: was this the main change? meaning that we needed to update the presets and then set the 'useCMakePresets' to 'always' so that the env variables are available on time for the tests?
Because I would normally think that the setup will run before the tests run and we will have the env variables available by that point Line 45).

src/cmakeTools.ts
bobbrow: no

test/extension-tests/single-root-UI/test/envs-kits.test.ts
bobbrow: It's best for tests to clean up after themselves so that the state going into each test is known. I'm removing this as a workaround for now because we can't currently await the work that happens after the setting is changed.

test/extension-tests/single-root-UI/test/envs-presets.test.ts
bobbrow: This is done because we can't currently await the work that happens after the setting change.  If that work is still running when we change the presets, the preset change will fail and that's what's currently causing the test failure.

bobbrow:
@elahehrashedi this is ready for review now.

elahehrashedi (approved):

PR #2695, opened 2022-08-17
resolving sourceDirectory relative paths in multi-root projects

elahehrashedi:
In a multi-root project, if we locate a `CMakeLists.txt` that is outside of the `workspaceFolder`, the relative path that is being created will be resolved incorrectly later on (the path looks something like `"${workspaceFolder}/../../root2"`). 
So I changed the code to keep the absolute path for` CMakeLists.txt` files that are located outside of the `workspaceFolder`, and only use relative-path when `CMakeLists.txt` is inside the `workspaceFolder`.

The main source of the issue is not this part of the code, but this change avoids the error.

### DIFF
--- a/src/cmakeTools.ts
+++ b/src/cmakeTools.ts
@@ -636,19 +636,21 @@ export class CMakeTools implements api.CMakeToolsAPI {
                             };
                             const cmakeListsFile = await vscode.window.showOpenDialog(openOpts);
                             if (cmakeListsFile) {
+                                // Keep the absolute path for CMakeLists.txt files that are located outside of the workspace folder.
                                 selectedFile = cmakeListsFile[0].fsPath;
                             }
                         } else {
-                            selectedFile = selection.fullPath;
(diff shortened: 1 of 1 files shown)


src/cmakeTools.ts
```diff
@@ -636,19 +636,21 @@ export class CMakeTools implements api.CMakeToolsAPI {
                             };
                             const cmakeListsFile = await vscode.window.showOpenDialog(openOpts);
                             if (cmakeListsFile) {
+                                // Keep the absolute path for CMakeLists.txt files that are located outside of the workspace folder.
                                 selectedFile = cmakeListsFile[0].fsPath;
                             }
                         } else {
-                            selectedFile = selection.fullPath;
+                            // Keep the relative path for CMakeLists.txt files that are located inside of the workspace folder.
```
bobbrow, line 643: Can you update this comment to remind that the relative path is stored in the label?

bobbrow (approved):
Update the comment and then check this in.

PR #2704, opened 2022-08-23
Update documents for tasks

elahehrashedi:
This note is added:

**Note**: If you are using a preset other than the active configure preset, you can change this settings in *settings.json* to avoid re-configure based on the active preset when editing.

```json
    "cmake.configureOnEdit": false
```

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -3,7 +3,7 @@ You can create a configure task from the VS Code command pallette by running the
 
 ![Configure a task](images/configure_task.png)
 
-By selecting "CMake: configure" template, if you are not using presets, this task will be generated in *tasks.json* file: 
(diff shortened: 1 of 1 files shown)


docs/tasks.md
```diff
@@ -3,7 +3,7 @@ You can create a configure task from the VS Code command pallette by running the
 
 ![Configure a task](images/configure_task.png)
 
-By selecting "CMake: configure" template, if you are not using presets, this task will be generated in tasks.json file: 
+By selecting "CMake: configure" template, if you are not using presets, this task will be generated in *tasks.json* file:
```
bobbrow, line 6: For file names we usually decorate with backtick instead of asterisk.

bobbrow (approved):
change the decorations around filenames and this is ok to submit

bobbrow (approved):

Issue #2714, opened 2022-08-29
Add "env" variables to "configure" and "install" tasks


elahehrashedi:
Add "env" variables to "configure" and "install" tasks

Issue #2706, opened 2022-08-24
enhancements on "Install" and "clean" task when using presets


elahehrashedi:
Currently, `install` and `clean` tasks when using presets, are operating on the active preset. 

There needs to be an entry in these tasks to target a specific build.
@sinemakinci1 @bobbrow , Both `install` and `clean` are a specific kind of build. Does it make sense to let them use the build preset name?
e.g.:
```json
		{
			"type": "cmake",
			"label": "CMake: clean",
			"command": "clean",
			"preset": "${command:cmake.activeBuildPresetName}",
			"detail": "CMake template clean task"
		}
```

PR #2716, opened 2022-08-29
Enhance config, install, clean tasks

elahehrashedi:
Add "`env`" variables to "`configure`" and "`install`" tasks: https://github.com/microsoft/vscode-cmake-tools/issues/2714
Add "`preset`" to "`Install`" and "`clean`" task: https://github.com/microsoft/vscode-cmake-tools/issues/2706

(Need to test)
Question: how will Cmake Server be changed?

### DIFF
--- a/src/cmakeTaskProvider.ts
+++ b/src/cmakeTaskProvider.ts
@@ -315,7 +315,7 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
                 const buildPreset: preset.BuildPreset | undefined = await cmakeTools?.expandBuildPresetbyName(this.preset);
                 if (!buildPreset) {
                     log.debug(localize("build.preset.not.found", 'Build preset not found.'));
-                    this.writeEmitter.fire(localize("build.failed", "Build preset {0} not found. Build failed.", this.preset) + endOfLine);
+                    this.writeEmitter.fire(localize("build.failed", "Build preset {0} not found. {1} failed.", this.preset, taskName) + endOfLine);
                     this.closeEmitter.fire(-1);
                     return;
                 }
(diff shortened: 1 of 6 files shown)


src/drivers/cmakeServerDriver.ts
```diff
@@ -154,7 +154,7 @@ export class CMakeServerDriver extends CMakeDriver {
         })();
     }
 
-    protected async doConfigure(args: string[], consumer?: proc.OutputConsumer, showCommandOnly?: boolean, configurePreset?: ConfigurePreset | null) {
+    protected async doConfigure(args: string[], consumer?: proc.OutputConsumer, showCommandOnly?: boolean, configurePreset?: ConfigurePreset | null, _options?: proc.ExecutionOptions) {
         await this._clientChangeInProgress;
         const cl = await this.getClient();
```
bobbrow, line 159: `this.getClient()` calls `this._startNewClient()`.  If you save the `options?.environment`, you can pass the env to `this.getConfigureEnvironment()` when the server is started.

bobbrow:
> Question: how will Cmake Server be changed?

I put a comment in the PR.  Overall, looks good.  I will do a final review this afternoon.

src/cmakeTaskProvider.ts
```diff
@@ -326,44 +335,27 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
             }
         } else {
             log.debug(localize("cmake.driver.not.found", 'CMake driver not found.'));
-            this.writeEmitter.fire(localize("build.failed", "Build failed.") + endOfLine);
+            this.writeEmitter.fire(localize("task.failed", "{0} failed.", taskName) + endOfLine);
```
bobbrow, line 338: Don't rename the localization ids for now.  They don't actually matter.  If you do rename them, the translations will all reset and we don't want that right before we release.

So please change them back to `"build.failed"`, etc instead of `"task.failed"`

src/drivers/cmakeServerDriver.ts
bobbrow: We talked about this and decided not to support CMake server.

bobbrow (approved):

PR #2718, opened 2022-08-30
Fix a high hitting crash in ctest processing

bobbrow:


elahehrashedi (approved):
May I ask what crash this was about?

bobbrow:
If the break on line 175 was not hit before [cursor + i] grew larger than the length of the array, then `untrimmedLine` on line 173 would be assigned `undefined` and line 174 would throw an exception.

PR #2719, opened 2022-08-30
Fix some of our higher hitting crashes

bobbrow:
Fixing a couple of high hitting crashes (by total count).  Not very many users have actually hit these, but when they do they hit them a lot.

This can be for 1.13.

### DIFF
--- a/test/unit-tests/ctest.test.ts
+++ b/test/unit-tests/ctest.test.ts
@@ -1,0 +1,24 @@
+import { readTestResultsFile } from "@cmt/ctest";
+import { expect, getTestResourceFilePath } from "@test/util";
+
+suite('CTest test', () => {
+    test('Parse XML test results', async () => {
+        const result = await readTestResultsFile(getTestResourceFilePath('TestResults.xml'));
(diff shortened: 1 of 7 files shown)


test/unit-tests/ctest.test.ts
```diff
@@ -0,0 +1,24 @@
+import { readTestResultsFile } from "@cmt/ctest";
+import { expect, getTestResourceFilePath } from "@test/util";
+
+suite('CTest test', () => {
```
elahehrashedi, line 4: is this test created to test the function "readTestResultsFile"  only ? The XML files are not being generated each time as a result of another test, right?

elahehrashedi (approved):

test/unit-tests/ctest.test.ts
bobbrow: Eventually we should add coverage for actual running of ctest.  I just wanted a baseline for the parsing for now because it was broken earlier.  We could replace the checked in xml file with one that is produced by ctest.

PR #2725, opened 2022-09-02
Add telemetry for CMake tasks

elahehrashedi:


bobbrow (changes requested):
We already have telemetry events named for configure/build.  For tasks, I think it would be better to have a new one that is the same for all tasks (e.g. "task") with "taskType" as a property.  We should also log "cleanRebuild" as a separate entity instead of "clean" and "rebuild" separately because it is difficult to pair those two events reliably in a query.

bobbrow (approved):

PR #2726, opened 2022-09-02
Fix a race condition in the single-root-UI test

bobbrow:
Fix a race condition in the single-root-UI test.

The problem is that when we update settings, there are event handlers that execute afterwards and we're not actually waiting for those to complete.  This means that the test can continue executing before the extension is using the effect of the settings change.  With this change, we track those outstanding events and provide a way to wait for those to complete.  The awaiter object lives in the extension which requires a special command to access it.  That command will only be available during testing.


bobbrow:
Ugh.  Now there's a problem with the configure-and-build test too. 😢

elahehrashedi (approved):

Issue #2735, opened 2022-09-08
Cmake presets: bulid presets aren't displayed


hinell:
### Brief Issue Summary

Build presets aren't displayed.

I've approximately the following `CMakePresets.json` settings:

```json

...
"buildPresets": [
        {
            "name": "common",
            "hidden": true,
            "displayName": "Common Build Configuration",
            "configurePreset": "common",
            "inherits": [],
            "description": "TODO: Add description",
            "inheritConfigureEnvironment": true,
            "environment": {},
            "jobs": 6,
            "cleanFirst": false,
            "verbose": false,
            "nativeToolOptions": []
        },
        {
            "name": "linux",
            "hidden": false,
            "configurePreset": "linux",
            "displayName": "Linux build",
            "inherits": ["common"],
            "description": "Linux basic build",
            "inheritConfigureEnvironment": true,
            "environment": {},
            "targets": ["all"],
            "configuration": "MinSizeRel",
            "condition": {
                "type": "inList",
                "string": "${hostSystemName}",
                "list": [
                    "Linux"
                ]
            },
            "cleanFirst": false
        },
        ...
       ]
....

```

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.71.0",
  "cmtVersion": "1.12.26",
  "configurations": [
    {
      "folder": "/media/all/ProgramsAndGames/pr/2020/c++.practice",
      "cmakeVersion": "3.24.1",
      "configured": true,
      "generator": "Ninja Multi-Config",
      "usesPresets": true,
      "compilers": {
        "C": "/usr/bin/clang",
        "CXX": "/usr/bin/clang++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

elahehrashedi:
Thank you for reporting this issue. We sent a PR to fix this issue: https://github.com/microsoft/vscode-cmake-tools/pull/2736

elahehrashedi:
[...]

PR #2736, opened 2022-09-08
Show presets with hidden parent preset

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2735

### DIFF
--- a/src/presetsController.ts
+++ b/src/presetsController.ts
@@ -590,7 +590,7 @@ export class PresetsController {
         interface PresetItem extends vscode.QuickPickItem {
             preset: string;
         }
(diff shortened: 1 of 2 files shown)


src/presetsController.ts
```diff
@@ -590,7 +590,7 @@ export class PresetsController {
         interface PresetItem extends vscode.QuickPickItem {
             preset: string;
         }
-        const presetsPool: preset.Preset[] = showHiddenPresets ? candidates : candidates.filter(_preset => !_preset.hidden && preset.evaluatePresetCondition(_preset, allPresets));
+        const presetsPool: preset.Preset[] = showHiddenPresets ? candidates : candidates.filter(_preset => !_preset.hidden && preset.evaluatePresetCondition(_preset, allPresets) !== undefined);
```
bobbrow, line 593: I don't believe this is the correct fix.  If the condition is false, we shouldn't return true.

bobbrow (changes requested):

Issue #2737, opened 2022-09-08
Variable substitution issues in workspace-specific cmake-kits.json


hellige:
### Brief Issue Summary

When using a project-specific kits file in `.vscode/cmake-kits.json`, variable substitution no longer seems to work normally. Given a kit definition like this:

```
  {
    "name": "GCC 11.2.0",
    "compilers": {
      "C": "${workspaceFolder}/../cpp-build-tools/gcc11.2.0-1",
      "CXX": "${workspaceFolder}/../cpp-build-tools/g++11.2.0-1"
    }
  },
```

when opening the project I now see the following lines in the output window:

```
[proc] Executing command: ${workspaceFolder}/../cpp-build-tools/gcc11.2.0-1 -v
[proc] The command: ${workspaceFolder}/../cpp-build-tools/gcc11.2.0-1 -v failed with error: Error: spawn ${workspaceFolder}/../cpp-build-tools/gcc11.2.0-1 ENOENT
...
[proc] The command: gcc11.2 -v failed with error: Error: spawn gcc11.2 ENOENT
[proc] The command: g++11.2 -v failed with error: Error: spawn g++11.2 ENOENT
```

Oddly, the actual cmake configure step still seems to work ok, and find the compiler at the correct path.

In addition (and this was actually the first thing I noticed), I get two new empty directories created _in my home dir_:

```
[~]% ls -ltr ~
...
drwxr-xr-x  2 mhellige mhellige            6 Sep  8 14:16  cpp-build-tools
drwxr-xr-x  2 mhellige mhellige            6 Sep  8 14:16 '${workspaceFolder}'
[~]%
```

I believe the `workspaceFolder` variable should be expanded in `cmake-kits.json`, and in any case the extension should **definitely not** be creating random directories in my home dir!

### CMake Tools Diagnostics

```{
  "os": "linux",
  "vscodeVersion": "1.71.0",
  "cmtVersion": "1.12.26",
  "configurations": [
    {
      "folder": "/home/mhellige/drw/test",
      "cmakeVersion": "3.12.1",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": false,
      "compilers": {
        "C": "/home/mhellige/drw/test/../cpp-build-tools/gcc11.2.0-1",
        "CXX": "/home/mhellige/drw/test/../cpp-build-tools/g++11.2.0-1"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
[...]

PR #2739, opened 2022-09-09
resolve variables in cmake-kits.json

elahehrashedi:
bug-fix: https://github.com/microsoft/vscode-cmake-tools/issues/2737

### DIFF
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -24,7 +24,7 @@ import { CMakeBuildConsumer } from './diagnostics/build';
 import { CMakeOutputConsumer } from './diagnostics/cmake';
 import { populateCollection } from './diagnostics/util';
 import { CMakeDriver, CMakePreconditionProblems } from '@cmt/drivers/cmakeDriver';
-import { expandString, ExpansionOptions } from './expand';
+import { expandArrayOfStrings, expandString, ExpansionOptions } from './expand';
 import { CMakeGenerator, Kit } from './kit';
 import { CMakeLegacyDriver } from '@cmt/drivers/cmakeLegacyDriver';
 import * as logging from './logging';
@@ -44,6 +44,7 @@ import { ConfigurationReader } from './config';
 import * as preset from '@cmt/preset';
 import * as util from '@cmt/util';
 import { Environment, EnvironmentUtils } from './environmentVariables';
+import paths from './paths';
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
@@ -2471,6 +2472,36 @@ export class CMakeProject implements api.CMakeToolsAPI {
 
     async selectEnvironments() {
         return null;
+    }
+
+    async getExpansionOptions(): Promise<ExpansionOptions> {
+        const workspaceFolder: string = this.workspaceContext.folder.uri.fsPath;
+        return {
+            vars: {
+                buildKit: this.activeKit?.name || "",
+                buildType: await this.currentBuildType() || "",
+                buildKitVendor: "",
+                buildKitTriple: "",
+                buildKitVersion: "",
+                buildKitHostOs: "",
+                buildKitTargetOs: "",
+                buildKitTargetArch: "",
+                buildKitVersionMajor: "",
+                buildKitVersionMinor: "",
+                generator: "",
+                userHome: paths.userHome,
+                workspaceFolder: workspaceFolder,
+                workspaceFolderBasename: path.basename(workspaceFolder),
+                workspaceHash: "",
+                workspaceRoot: this.workspaceContext.folder.uri.fsPath,
+                workspaceRootFolderName: path.basename(workspaceFolder)
+            }
+        };
+    }
+
+    async getExpandedAdditionalKitFiles(): Promise<string[]> {
+        const opts: ExpansionOptions = await this.getExpansionOptions();
+        return expandArrayOfStrings(this.workspaceContext.config.additionalKits, opts);
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/kit.ts
```diff
@@ -1108,29 +1108,38 @@ export async function descriptionForKit(kit: Kit, shortVsName: boolean = false):
     return localize('unspecified.let.cmake.guess', 'Unspecified (Let CMake guess what compilers and environment to use)');
 }
 
+async function expandKitVariable(variable: string, kitName: string): Promise<string> {
+    return expand.expandString(variable, {
```
bobbrow, line 1112: Splitting this out into a separate function means that the object passed in as the second argument has to be created every time it is called.  I think it would be better to create a temporary variable in `expandKitVariables` and just call `expand.expandstring` directly in there.

src/kit.ts
```diff
@@ -1108,29 +1108,38 @@ export async function descriptionForKit(kit: Kit, shortVsName: boolean = false):
...
-                workspaceHash: '${workspaceHash}',
-                workspaceRoot: '${workspaceRoot}',
-                workspaceRootFolderName: '${workspaceRootFolderName}'
-            }
-        });
+        kit.toolchainFile = await expandKitVariable(kit.toolchainFile, kit.name);
+    }
+    if (kit.compilers) {
+        for (const lang in kit.compilers) {
+            kit.compilers [lang] = await expandKitVariable(kit.compilers[lang], kit.name);
```
bobbrow, line 1141: nit: remote extra space before `[lang]`

bobbrow (changes requested):

src/kitsController.ts
```diff
@@ -99,7 +99,8 @@ export class KitsController {
             await KitsController.readUserKits(cmakeProject);
         }
 
-        const folderKitsFiles: string[] = [KitsController._workspaceKitsPath(cmakeProject.folder)].concat(await KitsController.expandAdditionalKitFiles(cmakeProject));
```
bobbrow, line 102: Is `expandAdditionalKitFiles` used anymore?  If not, please delete it.

src/kit.ts
```diff
@@ -1108,65 +1108,72 @@ export async function descriptionForKit(kit: Kit, shortVsName: boolean = false):
...
         return [];
     }
     const kits = kits_raw as Kit[];
-    log.info(localize('successfully.loaded.kits', 'Successfully loaded {0} kits from {1}', kits.length, filepath));
-    return Promise.all(dropNulls(kits).map(expandKitVariables));
+    log.info(localize('successfully.loaded.kits', 'Successfully loaded {0} kits from {1}', kits.length, filePath));
+
+    const expandedKits: Kit[] = [];
+    for (const kit of kits) {
+        opts = opts ? opts : {
```
bobbrow, line 1145: Don't do this inside the loop. It will end up creating the same object over and over.
Also, we don't generally set variables to themselves.  It's better to just use an `if` to conditionally set it.
```ts
if (!opts) {
  opts = ...;
}
```

src/cmakeProject.ts
```diff
@@ -2472,6 +2473,36 @@ export class CMakeProject implements api.CMakeToolsAPI {
     async selectEnvironments() {
         return null;
     }
+
+    async getExpansionOptions(): Promise<ExpansionOptions> {
+        const workspaceFolder: string = this.workspaceContext.folder.uri.fsPath;
+        return {
+            vars: {
+                buildKit: this.activeKit?.name || "",
```
bobbrow, line 2481: Please follow the pattern of the other ExpansionOptions creation function by replacing unsupported properties with themselves instead of empty string.

```ts
  vars: {
    buildKit: this.activeKit?.name || "${buildKit}",
    buildKitVendor: "${buildKitVendor}",
    ...
  }
```

src/cmakeProject.ts
```diff
@@ -2472,6 +2473,36 @@ export class CMakeProject implements api.CMakeToolsAPI {
...
+                buildType: await this.currentBuildType() || "",
```
bobbrow, line 2482: Is it only the `additionalKits` setting that uses these expansions?  If so, then we should not support substitution for `buildKit` or `buildType` because those are not stable.

src/expand.ts
```diff
@@ -249,3 +249,12 @@ async function expandStringHelper(input: string, opts: ExpansionOptions) {
     });
     return { result: finalString, didReplacement, circularReference };
 }
+
+export async function expandArrayOfStrings(inputs: string[], opts: ExpansionOptions): Promise<string[]> {
```
bobbrow, line 253: A common naming simplification for arrays is to pluralize the variable/type that the array holds.
i.e. this function name could be simplified to `expandStrings` (plural).

src/kit.ts
```diff
@@ -1108,65 +1108,72 @@ export async function descriptionForKit(kit: Kit, shortVsName: boolean = false):
...
         }
         return [];
     }
     const kits = kits_raw as Kit[];
-    log.info(localize('successfully.loaded.kits', 'Successfully loaded {0} kits from {1}', kits.length, filepath));
-    return Promise.all(dropNulls(kits).map(expandKitVariables));
+    log.info(localize('successfully.loaded.kits', 'Successfully loaded {0} kits from {1}', kits.length, filePath));
+
+    const expandedKits: Kit[] = [];
+    for (const kit of kits) {
```
bobbrow, line 1144: The previous implementation removed null kits before expanding variables.  I don't know the case where a kit would be null, but I think we should preserve the behavior just in case.

src/kit.ts
elahehrashedi: The "buildKit" had to be set for every kit. I modified the loop to make it more efficient.

src/kitsController.ts
elahehrashedi: right, I made the changes to eliminate this function in the first palce.

PR #2752, opened 2022-09-19
resolve "condition" in user presets

elahehrashedi:


src/preset.ts
```diff
@@ -368,14 +368,6 @@ export function setOriginalUserPresetsFile(folder: string, presets: PresetsFile
     originalUserPresetsFiles.set(folder, presets);
 }
 
-export function getPresetsFile(folder: string) {
```
bobbrow, line 371: Are these not used anywhere?

bobbrow (approved):

src/preset.ts
elahehrashedi: no, it might have been defined to replace "presetsFiles.get(folder)", but it wasn't used. 
I can use them, but I beleive it wouldn't make the code any more readable.

PR #2758, opened 2022-09-29
:bug: fix zh cn translation not showing

xgdgsc:
Need changing this to make zh cn translation appear.


bobbrow:
@Colengms can you check this?

Colengms (approved):
Lower case matches cpptools, so I believe this is a good fix.

Issue #1738, opened 2021-04-01
Disable "Build" button if configuration is already in progress


H-G-Hristov:
Clicking on "Build" while configuration is in progress results in an error message. Maybe it would make sense to just disable all actions that cannot be performed during this process and instead indicating what's going on with the option to cancel it?

andreeis:
Thank you for the idea, it would definitely be useful to implement something like that. We will monitor the community reactions to evaluate where this feature request would best fit in our backlog.

bobbrow:
I was also thinking that rather than disable it, we could just queue it up to execute after the configure completes.  I run into this frequently when I open a folder and want to immediately build it.

Danielmelody:
I guess this one would be solved In favor of #2759?

PR #2759, opened 2022-09-29
Support canceling configuration

Danielmelody:

* While performing configuration, the busy status will be set.
* After configuration being done, the busy status will be reset
* A user can cancel configuration by click the 'x' button, the same way canceling build.
image


Danielmelody:
I bundle configure and build status into one as we cannot configure and build at the same time.

bobbrow:
Thanks for the PR!  We should be able to review it in the next few days.  Can you please also add an entry to `CHANGELOG.md` and give yourself credit for the enhancement?

Danielmelody:
Sure.

src/drivers/cmakeDriver.ts
```diff
@@ -1603,7 +1603,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
      * The currently running process. We keep a handle on it so we can stop it
      * upon user request
      */
-    private _currentBuildProcess: proc.Subprocess | null = null;
+    protected _currentBuildOrConfigureProcess: proc.Subprocess | null = null;
```
elahehrashedi, line 1606: please rename `_currentBuildOrConfigureProcess` to `currentBuildOrConfigureProcess`.

elahehrashedi:
It looks like the code needs to be modified to be able to kill the configure process when CmakeServer is being used:
https://github.com/microsoft/vscode-cmake-tools/blob/2952f77a70891c6ccd9191afd2249374ea10649f/src/drivers/cmakeServerDriver.ts#L176

However, I am not sure how we can have access to the child process here.

elahehrashedi:
A design question about canceling configuration. When we run the command "Cmake: Configure", the "build" button turns into the "stop" which IMO is confusing, as we are planning to cancel the configure not the build. I was wondering if we can come up with a better UI design to cancel the configuration.
Also, when the configuration is running, it would make more sense if the "build" button gets disabled instead of turning into a "stop" button.

elahehrashedi:
Here, you can add the option "cancellable: true" to the config progress window:
https://github.com/microsoft/vscode-cmake-tools/blob/2952f77a70891c6ccd9191afd2249374ea10649f/src/cmakeProject.ts#L1295-L1299

Danielmelody:
@elahehrashedi 
> It looks like the code needs to be modified to be able to kill the configure process when CmakeServer is being used

I cannot figure out how to do that via the old cmake server api, how does the build process implement cancelation via C/S arch? 
A trivial way is to shutdown the server process while canceling stop or configuration, and launch a new one next time whenever necessary.

Danielmelody:
> I was wondering if we can come up with a better UI design to cancel the configuration. For example, we can add a cancel option to the configure progress window.

If there is another button indicating build status, while configuring project, I think the build button should be disabled.

Danielmelody:
I wonder this line should stop the configuration
https://github.com/microsoft/vscode-cmake-tools/blob/a96a4d28dc114b598a262052318d688d08b8c015/src/drivers/cmakeServerDriver.ts#L409
when https://github.com/microsoft/vscode-cmake-tools/blob/a96a4d28dc114b598a262052318d688d08b8c015/src/drivers/cmakeDriver.ts#L1731
being called? I can't test it as I have no older cmake installed.

elahehrashedi:
Thank you for your response.

"Disabling the buttons when the configure is in progress" is being tracked in another thread: https://github.com/microsoft/vscode-cmake-tools/issues/1738, if you are interested you can send a separate fix for that issue.

For the current issue, Please add the option "cancellable: true" to the config progress window, in order to stop the configuration process (Some other changes may be needed too). please make sure the "build" button is replaced with "stop" only when the build is in progress, not when the configure is in progress. 

As for the Cmake Server, that would be great if you can install the older versions of CMake for testing, otherwise, we will investigate it after you update this branch.

Danielmelody:
> For the current issue, Please add the option "cancellable: true" to the config progress window, in order to stop the configuration process (Some other changes may be needed too). please make sure the "build" button is replaced with "stop" only when the build is in progress, not when the configure is in progress.

I have done that for modify the previous commit, please review again.
As for the server api, I managed to installed an older version cmake, and add some logic to shutdown server directly.

elahehrashedi:
I tested your branch, The configuration doesn't have a cancel option, and the "build" button turns into a "stop" button while the configuration is run. Would you please apply these changes (Copied from the previous comments): 
> For the current issue, Please add the option "cancellable: true" to the config progress window, in order to stop the configuration process (Some other changes may be needed too). please make sure the "build" button is replaced with "stop" **only** when the build is in progress, **not** when the configure is in progress.

Danielmelody:
@elahehrashedi I really doubt it, I guess you did not get my latest update, I have add that cancel button (if you check the code changes), please make sure you are testing [Danielmeloldy/feat-cancel-configuration](https://github.com/Danielmelody/vscode-cmake-tools/tree/feat-cancel-configuration) and pull from remote.

src/cmakeProject.ts
```diff
@@ -1297,10 +1297,15 @@ export class CMakeProject implements api.CMakeToolsAPI {
         return vscode.window.withProgress(
             {
                 location: vscode.ProgressLocation.Notification,
-                title: localize('configuring.project', 'Configuring project')
+                title: localize('configuring.project', 'Configuring project'),
+                cancellable: true
```
Danielmelody, line 1301: @elahehrashedi Here

Zingam:
This is an interesting project. Is it making any progress?

Danielmelody:
@elahehrashedi @bobbrow Can you please review again? This PR has been there for such a long time and I 'm quite sure I have done what you suggested.

bobbrow:
Sorry for the delay.  We do plan to put this into 1.13, but the other feature has been taking longer than expected.  If @elahehrashedi hasn't finished reviewing and testing soon, then I will try to help out later this week (I split my time between several projects and am also a manager so I don't have as much time as I would like).

src/drivers/cmakeDriver.ts
```diff
@@ -1716,9 +1712,9 @@ export abstract class CMakeDriver implements vscode.Disposable {
                 }
             } else {
                 if (gen) {
-                    if (/(Unix|MinGW) Makefiles|Ninja/.test(gen) && targets.length === 1 && targets[0] !== 'clean') {
+                    if (/(Unix|MinGW) Makefiles|Ninja/.test(gen) && (targets.length === 1 && targets[0] === 'clean')) {
```
bobbrow, line 1715: why is this condition changing?  This and the next change look incorrect.

Danielmelody: I originally change this because of a warning. Seems I made a mistake here, please revert those two lines as I am on my flight now

bobbrow: Thanks. I reverted it.

elahehrashedi:
@bobbrow the "build" button in status bar is not turning into "stop" when configuration is running.

bobbrow:
Since the popup is cancellable, but not minimizable, I'll change the ProgressLocation to "Window" so that it doesn't annoy people.  It will start in the statusbar with a busy icon instead.

bobbrow (approved):

nasinlock (approved):

AbdulAbCellera:
This is good. Thanks! Sorry for spamming you, but I am happy because I no longer have to kill things from the terminal.

PR #2769, opened 2022-10-04
Command palette should run proper task

piomis:
### This changes visible behaviour of command palette and targets from CMake Outlline

- detect which task to run based on contents of targets list 
- config, install and test ignore rest of targets from list and call 'config', 'install', and 'test' task accordingly
- clean with  additional targets runs 'cleanRebuild' task, clean only calls 'clean' task

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Improvements:
 
 Bug Fixes:
 - Fix warning message that appears when using a default build preset with a multi-config generator. [#2353](https://github.com/microsoft/vscode-cmake-tools/issues/2353)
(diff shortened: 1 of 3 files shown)


piomis:

piomis:
@elahehrashedi this is additional pull-request to solve problem with wrong behaviour when useTask option is active.

piomis:

elahehrashedi:
Thank you for sending this PR. The setting "cmake.buildTask" is designed to run the build task only when the "Cmake: Build" command is run. For all the other commands such as install, clean, and run tests, the behavior should not change. I will close this PR and will send a PR to fix this issue:  https://github.com/microsoft/vscode-cmake-tools/pull/2770

PR #2772, opened 2022-10-05
Update kits documentation

jdeaton:
### This changes documentation.

- Update documentation for kits to correctly reference user-local kit file

## Other Notes/Information

It appears there is discrepancy between user-local kit filename (`cmake-tools-kits.json`) and the project-level filename (`.vscode/cmake-tools.json`). Was not clear if this was intended but maybe they should have the same name as they used to before the user-local filename was changed?


jdeaton:
The image in this section of documentation also shows the old user-local kit filename in the editor tab. Wasn't clear if I should create a new image showing the new user-local kit filename since the image appears in the user-local kit section or not. The name is consistent with the current `.vscode/cmake-kits.json` project-level filename.

elahehrashedi:
> The image in this section of documentation also shows the old user-local kit filename in the editor tab. Wasn't clear if I should create a new image showing the new user-local kit filename since the image appears in the user-local kit section or not. The name is consistent with the current `.vscode/cmake-kits.json` project-level filename.

I updated the image.

bobbrow:
No build is required for this change.  I'll merge it.

PR #2780, opened 2022-10-07
Calling build targets from CMake Project Outline always builds default target if useTasks option is set

### LINKED ISSUE #2778: Calling build targets from CMake Project Outline always builds default target if useTasks option is set bug
### Brief Issue Summary

After changes introduced in 1.12 it is not possible to start targets from CMake Outline when useTasks option is active.
Using above method will always just build target which is currently selected.

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

_No response_

piomis:
## This change fixes #2778

- updated minimum vscode package version to 1.60.0
- - fixed CleanRebuild command - Build phase hasn't used task even if buildTask setting was set to true
- Clean Rebuild - build phase should use proper build task if buildTasks option is set
- When build is requested following action is taken to select build task: if only one build task is present than it is used, if more tasks than one with default flag (if only one present), else user is asked to select build task which should be used from QuickPickMenu


piomis:
@elahehrashedi  here is my proposal. Please take a look if it makes sense. I haven't squashed commits to clearly present how this solution evolved :)

src/drivers/cmakeDriver.ts
```diff
@@ -1696,6 +1714,34 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+            interface TaskItem extends vscode.QuickPickItem {
+                task: vscode.Task
+            }
+            const choices = tasks.map((t): TaskItem => {
+                            return {
+                                label: t.name,
+                                task: t
+                            }
+            });
+            const sel = await vscode.window.showQuickPick(choices, { placeHolder: localize('select.active.target.tooltip', 'Select the default build target') });
```
piomis, line 1739: Incorrect message.

src/drivers/cmakeDriver.ts
```diff
@@ -1696,6 +1714,34 @@ export abstract class CMakeDriver implements vscode.Disposable {
         }
     }
 
+    private async _findBuildTask(): Promise<vscode.Task | undefined> {
```
elahehrashedi, line 1717: It looks like that `_findBuildTask` is an implementation of `vscode.commands.executeCommand("workbench.action.tasks.build");` , it fetches all of the tasks, choose the default is existed, otherwise shows a quickpick to choose from. 
The only difference is that `vscode.commands.executeCommand("workbench.action.tasks.build")` will show the template tasks besides the tasks defined in `tasks.json`.

src/cmakeTaskProvider.ts
```diff
@@ -161,6 +161,15 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
         }
         return undefined;
     }
+    public async cloneTaskAndAddTargets(task: CMakeTask, targets: string[]): Promise<CMakeTask | undefined> {
```
elahehrashedi, line 164: I can't see where this function is being called. Did you mean to call it in `_doCMakeBuild`?

piomis: Nope. It's remaining of my attempt to clone target and inject list of target into it.
Forgot to remove it from here.
Will do it.

src/cmakeTaskProvider.ts
```diff
@@ -326,11 +335,16 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
         if (generateLog) {
             telemetry.logEvent("task", {taskType: commandType, useCMakePresets: String(cmakeProject.useCMakePresets)});
         }
-        targets = await this.correctTargets(cmakeProject, commandType);
         const cmakeDriver: CMakeDriver | undefined = (await cmakeProject?.getCMakeDriverInstance()) || undefined;
         let cmakePath: string;
+        if(cmakeDriver)
```
elahehrashedi, line 340: This implementation doesn't seem to be a good practice. The task with a specific target is created, but the code is altering the targets during the execution time, without any warning. If the targets needs to be changed, it should be before the task is being executed. 

Besides that, I don't think it is good idea to choose a defined task and then alter the targets. The `_findBuildTask` could work like this:
- fetch all of the tasks in tasks.json
- find the task that has the targets matching the targets chosen from the project outline
- if there is no such task, show a quickpick of the template tasks to choose from. 
         - alter the targets if the template task to match the project outline target.

piomis: The question is what do you call 'template' task. I supose you mean these which are added by default by cmake Task Provider. 

E.g., I add task to tasks.json which does not have targets defined as it is tasks dedicated for all executable targets ( I add custom problem matcher I also seen cases when one was using tasks to get output in terminal which na n contrary to Output log support colours).

So in my case first condition to find tasks will never be met, second either as my taask is in tasks.json so wont appear on Quick pick list.

I agree that probably I could also eliminate in initial fetch all tasks which have target defined and it does not match the one requested.

package.json
```diff
@@ -2079,7 +2079,7 @@
     "@types/rimraf": "^3.0.0",
     "@types/sinon": "~9.0.10",
     "@types/tmp": "^0.2.0",
-    "@types/vscode": "1.56.0",
```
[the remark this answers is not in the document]
bobbrow: It is ok to upgrade if you need newer APIs from VS Code

src/cmakeTaskProvider.ts
elahehrashedi: > The question is what do you call 'template' task. I supose you mean these which are added by default by cmake Task Provider.

correct.

> In my case I add task to tasks.json which does not have targets defined as it is task dedicated for all executable targets ( I add custom problem matcher I also seen cases when one was using tasks to get output in terminal which na n contrary to Output log support colours). I supose in your definitione it is not template task.

The user need to know what are the specifications of the task that is running. So it needs to be a custom task they defined in the tasks.json or the template task we provide. 

> Thus first condition to find tasks will never be met, second either as my taask is in tasks.json so wont appear on Quick pick list.

Would you please elaborate on this? 

> I agree that probably I could also eliminate in initial fetch all tasks which have target defined and it does not match the one requested.

Correct, quickpick should show the tasks with the same target, or template task for that target. And importantly, don't alter a selected task when running it.

src/drivers/cmakeDriver.ts
elahehrashedi: That's right. However, it is still not delivering the results we want. Please read the later comments.

src/cmakeTaskProvider.ts
elahehrashedi: In template tasks, we only change the target, other characteristics stay the same.

piomis: I reimplemented approach a little bit.

I list all cmake tasks.
Then I search for ones target matching to requested one. If more than one found - I search for one with default flag. Else I let use pick from list.
If no tasks with matching targets, then I do similar query among ones with no targets defined.

In tasks itself I use same "correctTargets" method which was already present. So all already implemented guards are in power. I added only a capability to query driver for "requested target".

By the way - I noticed that method runTestsTask and runConfigTasks also calls correctTargets but they ignore returned value. Does it make then sense to call them?

src/cmakeTaskProvider.ts
piomis: Removed in latest commit

bobbrow:
I just published #2828 which should address the build from outline issues. It goes a little deeper than this PR does on that.  If you're ok with that approach, we could repurpose this PR to cover the new feature being added to select the build task.

Please also run `yarn lint` on your branch.  There are a number of errors that will need to be addressed.

piomis:
@bobbrow 
I've just pushed changes which should solve linter warnings.

WRT your proposal - is it something considered as final solution?
IMHO if someone sets useBuildTask option to true then I suppose one expects that building targets will really use tasks. E.g., in my case I have setup custom problem parser - with your proposal it won't work if target is built from Project Outline.

bobbrow:
> WRT your proposal - is it something considered as final solution? IMHO if someone sets useBuildTask option to true then I suppose one expects that building targets will really use tasks. E.g., in my case I have setup custom problem parser - with your proposal it won't work if target is built from Project Outline.

The way I interpreted #2778 is that the bug was that if you try to build a specific target from the outline and you have `cmake.buildTask` set, you can't because the build logic will override your decision and try to build with some other task.  My PR disables this behavior when the outline is in use and restores the correct build.  The theory is that if you're using the UI-heavy outline, you're picking something very specific.  If you have tasks configured, you can use the VS Code built-in experience for that.

Are you saying that you would like to still build with the task when building from the outline?  In my opinion, the UI is meant to help users skip configuring tasks.  The `cmake.buildTask` setting adds confusion and I would actually prefer to remove it at some point.

piomis:
#2778 is Indeed a bug. However buggy behaviour was not because of fact that task was used. But because task implementation always try to build target which is set as default.

IMHO consistent approach is more convenient approach. No matter what is source of build it should behave in same/similar way.
Additionally - my project has plenety of executabe, library and utility targets. Most convenient way to build them is ProjectOutline - it nicely lista all targets and I do not need to define individual tasks in tasks.json for each target or go through several QUICKpics. 

I would like also to be able to have logs of that builds parsed by problem parser or outputed to terminal window (to get coloured output).
Both that results are achivable only by tasks.
That is the reason for which build trigger source shouldn't be a condition to cross out task and switch to non-task build.

bobbrow:
I understand now.  I'll close my PR.  I do have a few concerns about some of the decisions made in this PR, but I'll add comments at the appropriate locations in the code instead of here.

src/drivers/cmakeDriver.ts
```diff
@@ -868,7 +868,23 @@ export abstract class CMakeDriver implements vscode.Disposable {
 
     private configRunning: boolean = false;
 
-    private buildRunning: boolean = false;
+    private _buildRunning: boolean = false;
+    get buildRunning(): boolean {
```
bobbrow, line 872: The `get` and `set` should be private.

src/drivers/cmakeDriver.ts
```diff
@@ -868,7 +868,23 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+        return this._buildRunning;
+    }
+
+    set buildRunning(running: boolean) {
+        if (!running) {
+            this._requestedTargets = undefined;
+        }
```
bobbrow, line 879: `this._buildRunning` should be updated in this function

src/drivers/cmakeDriver.ts
```diff
@@ -868,7 +868,23 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+    }
+    getRequestedTargets(): string[] {
+        if (this._requestedTargets) {
```
bobbrow, line 882: the requested targets should only be returned if `buildRunning` is true.

src/drivers/cmakeDriver.ts
```diff
@@ -1730,6 +1747,67 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+        }
+        const choices = tasks.map((t): TaskItem => ({ label: t.name, task: t }));
+        const sel = await vscode.window.showQuickPick(choices, { placeHolder: localize('select.build.task', 'Select build task') });
+        return sel ? sel.task : undefined;
+    }
+
+    private async _findBuildTask(targets?: string[]): Promise<vscode.Task | undefined> {
+
+        // Fetch all CMake task
+        const tasks = (await vscode.tasks.fetchTasks({ type: 'cmake' })).filter(task => ((task.group?.id === vscode.TaskGroup.Build.id)));
```
bobbrow, line 1762: Prior to the existence of `cmake` type tasks, people had to use `shell`.  Should these be included in the search?

src/drivers/cmakeDriver.ts
```diff
@@ -1730,6 +1747,67 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+
+        // There is only one found - stop searching
+        if (tasks.length === 1) {
+            return tasks[0];
+        }
+
+        // Get only tasks which target matches to one which is requested
+        const targetTasks = tasks.filter(task =>
+            (task.definition as CMakeTaskDefinition).targets === targets);
```
bobbrow, line 1771: I don't think we should try to match targets.  It would be better if we just replaced the targets with the one selected from the outline or command.  Consider the case where targets === ['all'].  The developer may be surprised that all the targets are being built.

src/drivers/cmakeDriver.ts
```diff
@@ -1742,12 +1820,33 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+                        const disposable = vscode.tasks.onDidEndTask(e => {
+                            if (e.execution.task.group?.id === vscode.TaskGroup.Build.id) {
+                                disposable.dispose();
+                                resolve({ retc: 0, stdout: '', stderr: '' });
+                            }
+                        });
+                    }) };
+                }
+                return { child: undefined, result: new Promise<api.ExecutionResult>((resolve) => {
+                    resolve({ retc: 0, stdout: '', stderr: '' });
```
bobbrow, line 1839: I think this should be an error and there should be come communication back to the developer that tells them we didn't find a task to build with.

src/drivers/cmakeDriver.ts
bobbrow: You should probably filter out tasks that don't use presets that match the current active ones either.  Those will fail if run.

src/drivers/cmakeDriver.ts
```diff
@@ -1730,6 +1747,67 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+    }
+
+    private async _findBuildTask(targets?: string[]): Promise<vscode.Task | undefined> {
+
+        // Fetch all CMake task
+        const tasks = (await vscode.tasks.fetchTasks({ type: 'cmake' })).filter(task => ((task.group?.id === vscode.TaskGroup.Build.id)));
+
+        // There is only one found - stop searching
+        if (tasks.length === 1) {
+            return tasks[0];
```
bobbrow, line 1766: I didn't think the extension currently picked any task that wasn't marked as default.  This is a change in policy.

src/drivers/cmakeDriver.ts
```diff
@@ -1730,6 +1747,67 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+                const defaultTargetTask = targetTasks.filter(task => task.group?.isDefault);
+                if (defaultTargetTask.length === 1) {
+                    return defaultTargetTask[0];
+                }
+                // None or more than one mark as default - show quick picker
+                return this._showTaskQuickPick(targetTasks);
+            }
+        }
+
+        // Get only tasks with no targets set (template tasks)
```
bobbrow, line 1788: What does "template task" mean in this context?  Is the vscode API returning our unconfigured tasks too?

src/drivers/cmakeDriver.ts
```diff
@@ -1742,12 +1820,33 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
-            const child = this.executeCommand(buildcmd.command, buildcmd.args, consumer, exeOpt);
-            this._currentBuildProcess = child;
-            await child.result;
-            this._currentBuildProcess = null;
-            return child;
+            if (useBuildTask) {
+                const task = await this._findBuildTask(targets);
+                if (task) {
+                    this._requestedTargets = targets;
+                    await vscode.tasks.executeTask(task);
```
bobbrow, line 1827: The other path has a way to cancel the build via the UI and block other operations (when `this._currentBuildProcess` is set).  Is there a way to get a handle to the task so that this path can do something similar?  I don't believe the current implementation works correctly, but it would be good to fix it while we're updating the code if possible.

test/helpers/vscodefake/memento.ts
```diff
@@ -12,6 +12,9 @@ export class TestMemento implements vscode.Memento {
         }
         return value;
     }
+    keys(): readonly string[] {
```
bobbrow, line 15: I'm not sure what the changes in this file are for.

src/drivers/cmakeDriver.ts
```diff
@@ -868,7 +868,23 @@ export abstract class CMakeDriver implements vscode.Disposable {
 
     private configRunning: boolean = false;
 
-    private buildRunning: boolean = false;
+    private _buildRunning: boolean = false;
```
bobbrow, line 871: Something I don't like about get/set properties that wrap another piece of data is that there's no protection against this code accidentally setting `_buildRunning` directly.  Consider wrapping this functionality in a small class that manages this state and the "requested targets".

bobbrow (changes requested):

test/helpers/vscodefake/memento.ts
piomis: API change in newer vscode (interface defines also 'keys').
Without this change tests wont compile.

src/drivers/cmakeDriver.ts
piomis: Any task with type cname which does not have target(s) defined.

src/drivers/cmakeDriver.ts
piomis: AFAIK it does.
Extension currently adds one task of each group (build, clean, install...).
These arent markes as 'default' and for example build task was used in case that no other tasks were present in tasks.json file.

src/drivers/cmakeDriver.ts
piomis: Agree. However I am not an expert (my domain is rather embedded C not typescript) so I used code which was used before while task was triggered (AFAIR it was execute methid in proc.ts).

src/drivers/cmakeDriver.ts
piomis: That was exactly was I did before (took default task and replace target). I was asked to change it.

I think that current proposal is sensible.
In first place I try finding tasks which targets are matching exactly to requested target. Because that means that there is specific task for this target.
When such task is not found than I try finding tasks without any target.
SO there will be no surprise because case when target X is requested and 'all' target is built should never happen.

src/drivers/cmakeDriver.ts
piomis: Current implementation does not work correctly. Even more - it does not wait for task processing to be finished (it startą asynchronous proces if task execution and exits). In this PR I have change it - currently extension will return Promise with task execution and will wait.
I will take a closer look for cancellation - maybe it works out of the box now :)

src/drivers/cmakeDriver.ts
piomis: > Prior to the existence of `cmake` type tasks, people had to use `shell`. Should these be included in the search?

No sure. Shell task could be created for many other reasons not related to cmake extension.
I believe that if one sets config.useBuildTask, one does it with intention to use cmake tasks.
Additionally from moment when cmake extension inttoduced its own task type, it should be depricated to use shell tas for cmake builds :)

src/drivers/cmakeDriver.ts
piomis: In last commit I've added possibility to cancel.

src/drivers/cmakeDriver.ts
piomis: `set` and `get` were removed.

src/drivers/cmakeDriver.ts
piomis: `get` and `set` were removed

src/drivers/cmakeDriver.ts
piomis: > You should probably filter out tasks that don't use presets that match the current active ones either. Those will fail if run.

I added also possibility to filter by preset (if build preset is set).

src/drivers/cmakeDriver.ts
piomis: I reverted `buildRunning` to what it was before.
I added new `CMakeBulldTaskRunner` class but I haven't moved `buildRunning` to it. Please let me know if you expect it there.

piomis: Have just done final refactor `CMakeBuildTaskRunner' is renamed to `CMakeBuildRunner` and tracks also Subprocess of non-task based builds.
buildRunning flag was moved into this class.

elahehrashedi (changes requested):

src/cmakeTaskProvider.ts
```diff
@@ -230,7 +237,13 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
         } else if (commandType === CommandType.clean) {
             targets = ['clean'];
         } else if (!shouldIgnore && !targetIsDefined && !cmakeProject.useCMakePresets) {
-            targets = [await cmakeProject.buildTargetName() || await cmakeProject.allTargetName];
+            const cmakeDriver: CMakeDriver | undefined = (await cmakeProject?.getCMakeDriverInstance()) || undefined;
```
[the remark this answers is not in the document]
piomis: I agree - but task execution interface does not allow to inject "targets".

src/cmakeBuildRunner.ts
```diff
@@ -0,0 +1,169 @@
...
+            this._buildRunning = true;
+        } else {
+            this._buildRunning = false;
+            this._targets = undefined;
+            this._buildPreset = null;
+        }
+    }
+
+    public async execute(buildPreset: preset.BuildPreset | null, targets?: string[]): Promise<void> {
+        const task = await this._findBuildTask(this._buildPreset, this._targets);
```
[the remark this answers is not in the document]
piomis: It shouldn't. I will fix.

src/cmakeBuildRunner.ts
```diff
@@ -0,0 +1,169 @@
...
+                const defaultTargetTask = targetTasks.filter(task => task.group?.isDefault);
+                if (defaultTargetTask.length === 1) {
+                    return defaultTargetTask[0];
+                }
+                // None or more than one mark as default - show quick picker
+                return this._showTaskQuickPick(targetTasks);
+            }
+        }
+
+        // Get only tasks with no targets set (template tasks)
```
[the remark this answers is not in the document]
piomis: In this context I use "template" task as name for tasks without "target" specified (so ones returned by provideTasks() but also ones added by user in tasks.json without any specific targets).

src/drivers/cmakeDriver.ts
piomis: @bobbrow I think I know what you mean.

I have just discovered that filtering wasn't implemented correctly.
I've fixed filtering but that revealed another thing: all tasks added by CMakeTaskProvider have targets set to ["all"] by default.

So question is what templetae tasks are really expected to do - I suppose that originally, they were planned as template for custom tasks. But at the end they are currently used as "build" tasks if tasks.json have no other cmake tasks.

src/cmakeBuildRunner.ts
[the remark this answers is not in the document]
piomis: I removed reference to "template" in comments.

src/drivers/cmakeDriver.ts
piomis: Alternatively CMakeTaskProvider May Just leave targets property as undefined instead of pushing ['all'] into it.
I dont think it is necessary to set it as later on when task is executed it can pull target which is set as default in CMakeProject.
But maybe there are some other reason to ha e it set to ['all']?

bobbrow: The reason was to demonstrate that the property exists. We aren't thinking about changing that behavior right now and many people have already generated tasks with this field set at this point.  So we'll need to handle that case.  My vote is to allow "all" to match any target.

This filter statement would need to change anyway because it's expecting the whole targets array to match and what we actually want to do is see if the current target (or "all") is in the array.

bobbrow: The algorithm sounds ok.  We don't have statistics on the typical configuration.  Target names are personal/private information and we don't track that.  We could log the frequency which the targets are set and how often "all" is encountered since that's not personally identifiable information, but we hadn't done that yet since we didn't see a need for the information at the time.  That said, most people use defaults and don't change them.

piomis: Uh... That won't be so easy.
I have just tried implementing mentioned algo.
The problem which arises is that tasks injected by CMakeTaskProvider do not neceserily have "targets" set to ['all'].
They have targets set to ['all'] only if default target (e.g. from status bar) is not set. If it is set then TaskProvider injects default target.
That brings some challenges:
 e.g., imagine tasks.json with single task inside without any targets (let's call it "universal" task). This definition lends as Prio "3".
But tasks added by CMakeTaskProvider have targets defined to "DefaultTarget". When from ProjectOutline I select target which matches defaultTarget then instead of "universal" task from tasks.json it is built with "template" task.

Maybe it would be better to let CMakeTaskProvider to set by default targets to ['all'] and when executing always query Cmakedriver/cmakeprojcect for requested/default targets?

piomis:
@elahehrashedi I have removed unused property - please re-run workflows checks.

I'm sorry that this PR looks a little bit abandoned - I was pretty busy. I plan to have closer look on what I wrote in comment https://github.com/microsoft/vscode-cmake-tools/pull/2780#discussion_r1020934395 in second half of December.
Is proposal from that comment acceptable?

> Maybe it would be better to let CMakeTaskProvider set by default targets to ['all'] and when executing always query Cmakedriver/cmakeprojcect for requested/default targets?


By the way a question - who should mark discussion as solved (there are currently few still open).

elahehrashedi:
@bobbrow in the commit `"before task finishes"` I have modified the code. I tested it, it's working. 

@piomis thank you for your contribution. I applied the changes directly to your branch (instead of a new branch) in order to keep your contribution to the code.

src/cmakeBuildRunner.ts
```diff
@@ -1,172 +1,71 @@
...
-            return;
-        }
-        this._currentBuildProcess = { child: undefined, result: new Promise<proc.ExecutionResult>((resolve) => {
-            resolve({ retc: 0, stdout: '', stderr: '' });
-        }) };
+    public async setBuildProcessForTask(taskExecutor: vscode.TaskExecution): Promise<void> {
+        this.taskExecutor = taskExecutor;
+        this.currentBuildProcess =  { child: undefined, result: new Promise<proc.ExecutionResult>(resolve => {
+            const disposable: vscode.Disposable = vscode.tasks.onDidEndTask((endEvent: vscode.TaskEndEvent) => {
+                if (endEvent.execution.task.group === vscode.TaskGroup.Build && endEvent.execution === this.taskExecutor) {
```
piomis: From what I have managed to find you cannot compare task.group with vscode.TaskGroup.Build
You need to compare "id" fields from these two.
This is because task.group is not exactly the same object.

src/cmakeTaskProvider.ts
```diff
@@ -162,6 +166,89 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
+            return resolvedTask;
+        }
+        return undefined;
+    }
+
+    public static async findBuildTask(presetName?: string, targets?: string[]): Promise<CMakeTask | undefined> {
+        // Fetch all CMake task from `tasks.json` files as well as tasks from task provider.
+        const allTasks: vscode.Task[] = await vscode.tasks.fetchTasks({ type: CMakeTaskProvider.CMakeScriptType });
+        const tasks: (CMakeTask | undefined)[] = allTasks.map((task: any) => {
+            if (!task.definition.label || !task.group || (task.group && task.group !== vscode.TaskGroup.Build)) {
```
piomis, line 188: From what I have managed to find you cannot compare task.group with vscode.TaskGroup.Build
You need to compare "id" fields from these two.
This is because task.group is not exactly the same object.

In my case it leads to situation that task defined in tasks.json is skipped,
Group of tasks from tasks.json is:
image

when vscode.TaskGroup.Build is:
image

src/cmakeTaskProvider.ts
```diff
@@ -162,6 +166,89 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
+            return undefined;
+        }
+
+        // Find tasks with a target that matches the input preset's target or the input targets
+        let matchingTargetTasks: CMakeTask[];
+
+        if (presetName) {
+            matchingTargetTasks = buildTasks.filter(task => task.definition.preset === presetName);
+        } else {
+            matchingTargetTasks = buildTasks.filter(task => {
```
piomis, line 221: This filter does not take into account tasks defined in tasks,json which are "generic" (have no target defined).
From my perspective it is crucial!

In my project I have following task in tasks.json which should be used as build task for every target:

```
{
    "version":"2.0.0",
    "tasks": [
        {
            "label": "CMake: Build",
            "type": "cmake",
            "command": "build",
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "problemMatcher": {
                "base": "$gcc",
                "source": "ArmClang",
                "fileLocation": "absolute"
            },
            "detail": "user defined"
        }
    ]
}
```


So, if there are no tasks with targets matching exactly it should also be checked if there are some tasks with targets being "unset".
I previously used following filter definition:
```
        // Get only tasks with no targets set
        const noTargetTasks = tasks.filter(task =>
            (task.definition as CMakeTaskDefinition).targets === undefined);
```

piomis:
@elahehrashedi  I wrote few comments which I believe must be resolved before merging (but I cannot mark them as "changes required").
Main concerns are:
- tasks from tasks.json which are mark as default will be automatically dropped due to comparison of task.group with vscode.TaskGroup.Build
- tasks from tasks.json without target defined will never be selected - and moreover won't even appear on QuickPick

src/cmakeBuildRunner.ts
```diff
@@ -0,0 +1,71 @@
...
+import * as vscode from 'vscode';
+import * as nls from 'vscode-nls';
+import * as proc from '@cmt/proc';
+import * as util from '@cmt/util';
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+
+export class CMakeBuildRunner {
+
+    private taskExecutor: vscode.TaskExecution | null = null;
```
bobbrow, line 14: Please use `undefined` instead of `null`  (applies to the whole file).  We will be trying to switch all `null` usages to `undefined` in the future.

piomis: Can do.
@elahehrashedi  please let me know if I should do it or you prefer to take over.

bobbrow: I think @elahehrashedi was planning to address some of your comments. She can fix this one too.

elahehrashedi:
[piomis](https://github.com/piomis) this change will consider the generic tasks with no target as well as the tasks defined for the specific target.

bobbrow (approved):

src/cmakeTaskProvider.ts
```diff
@@ -161,6 +171,91 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
+            });
+        }
+
+        if (matchingTargetTasks.length > 0) {
+            // One task is found.
+            if (matchingTargetTasks.length === 1) {
+                return matchingTargetTasks[0];
+            } else {
+                // Search for the matching default task.
+                const defaultTask: CMakeTask[] = matchingTargetTasks.filter(task => task.group?.isDefault);
```
piomis, line 239: You should change `task => task.group?.isDefault` into `task => task.isDefault`
This is because in temporary task created in line 210 you do not add full group but only this one field

src/cmakeTaskProvider.ts
```diff
@@ -161,6 +171,91 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
+                if (defaultTask.length === 1) {
+                    return defaultTask[0];
+                } else {
+                    // Search for the matching existing task.
+                    const existingTask: CMakeTask[] = matchingTargetTasks.filter(task => !task.isTemplate);
```
piomis, line 244: From what I see `isTemplate` property is not set for any task retuned in `buildTasks` array. I guess you need to set this property in similar way like you set `isDefault`

piomis:
@elahehrashedi  thank you.
Note that my comment: https://github.com/microsoft/vscode-cmake-tools/pull/2780#discussion_r1038154473 is still valid.
If task group in task.json contains "isDefault" than your filtering lamba won't work (as group direct comparison is not working).

I also added two comments - IMHO currently neither searching for default task nor filtering out template tasks will work.

piomis:
@elahehrashedi 
It looks that PR is automatically merged but like I wrote in above comment - IMHO it still has some important problems.

piomis:
@elahehrashedi @bobbrow 
Additional thing which is not working (it worked before) - stop ongoing build (via "X Stop" button on status bar). Currently it does not stop build however new button appeared on status bar "Building: xxxx" after clicking it "Cancel" button appears but it is not connected to same "stop()" handler as previous one. What leads to situation that BuildRunner status is desynchronized.

Should new Issues be reported and PRs created for these findings?

elahehrashedi:
[piomis](https://github.com/piomis) thank you for testing the PR. Would you please create a new report for the new issues?

piomis:
@elahehrashedi  Done.

For #2935 and for #2937  I've created pull-requests.
In regards to #2936 I am not sure when you intended isTemplate to be set - thus I am not sure about correct solution.

PR #2789, opened 2022-10-12
Add kit docs for environmentSetupScript

ernstp:

Updating the documentation was missed when the feature was added.

### DIFF
--- a/docs/kits.md
+++ b/docs/kits.md
@@ -134,6 +134,10 @@ The following additional options may be specified:
 
 > A JSON object of key-value pairs specifying additional environment variables to be defined when using this kit.
 
+`environmentSetupScript`
+
+> The absolute path to a script that modifies/adds environment variables for the kit. Uses `call` on Windows and `source` in `bash` otherwise.
+
 ## Next steps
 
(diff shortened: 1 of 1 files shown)


docs/kits.md
```diff
@@ -134,6 +134,10 @@ The following additional options may be specified:
 
 > A JSON object of key-value pairs specifying additional environment variables to be defined when using this kit.
 
+`environmentSetupScript`
+
+> A script that is evaluated to specify additional environment variables to be defined when using this kit. Uses `call` on Windows and `source` in `bash` otherwise.
```
elahehrashedi, line 139: @bobbrow do you think we can use this description from the kit schema:
`The absolute path to a script that modifies/adds environment variables for the Kit.`

elahehrashedi (approved):

docs/kits.md
ernstp: Done.

ernstp:
> If you would like to add a line to `CHANGELOG.md` to give yourself credit for your contribution, please go ahead and add that to the PR.

No need.

PR #2794, opened 2022-10-13
Add public API

benmcmorran:
### This changes the extension's exported API

- Add a new `vscode-cmake-tools` package in the `api` directory. This package defines the public API surface and will be published to NPM.
- Rename the existing `api.ts` file to `legacyApi.ts` to avoid confusion.
- Move code model types to the new API package. `codeModel.ts` re-exports these types to avoid source changes in other files and provide the flexibility to add additional internal fields to the types in the future without modifying the public API.

## Other Notes/Information

The initial API methods defined in this PR are based on needs from certain strategic partners, but more functionality can be added to the API in the future, including via community contributions.


api/api.ts
```diff
@@ -0,0 +1,244 @@
+/*---------------------------------------------------------------------------------------------
```
bobbrow, line 1: Just wanted to check... I was expecting we'd want to have a separate repo for this similar to the cpptools API.  I'm assuming that this is being added to the CMake Tools repo temporarily for convenience.  Is that correct?

api/api.ts
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * The version of the API.
+     */
+    readonly version: Version;
+
+    /**
+     * Shows the given UI element.
+     * @param element Element to show.
+     */
+    showUiElement(element: UiElement): Promise<void>;
```
bobbrow, line 43: nit: `Ui` -> `UI` (everywhere)

api/api.ts
```diff
@@ -0,0 +1,244 @@
...
+export interface CodeModelContent {
+    /** List of configurations provided by the selected generator */
+    configurations: CodeModelConfiguration[];
+
+    toolchains?: Map<string, CodeModelToolchain>;
+}
+
+/**
+ * Helper function to get the CMakeToolsApi from the CMake Tools extension.
+ * @param version The desired API version.
```
bobbrow, line 204: I know the cpptools API doesn't do this (yet), but we could rename it `desiredVersion` to make it more clear that this function may not give you the version you requested.

api/api.ts
```diff
@@ -0,0 +1,244 @@
...
+ * @param exactMatch If true, the version must match exactly. Otherwise, the
+ * function will attempt to return the requested version, but it may not match
+ * exactly.
+ * @returns The API object, or undefined if the API is not available.
+ */
+export async function getCMakeToolsApi(version: Version, exactMatch = true): Promise<CMakeToolsApi | undefined> {
```
bobbrow, line 210: I prefer `exactMatch` to be false by default.  Otherwise the behavior is different from the cpptools API.

api/api.ts
benmcmorran: I was originally thinking that the API type definitions can be hosted and published directly from this repo (it's currently structured as a separate package in the same repo), but if you'd rather the API live in a separate repo I'm not opposed.

bobbrow: Yes, I'd rather have it be separate.  But we can do that after you've implemented the CMake Tools side of things.

src/legacyApi.ts
```diff
@@ -0,0 +1,315 @@
+/**
+ * This module previously defined an exported API for the extension. While
```
bobbrow, line 2: Would you be able to help us clean this up?  I fear that if it doesn't happen as the new API comes in, this legacy will stick around forever.

api/api.ts
```diff
@@ -0,0 +1,239 @@
...
+ *--------------------------------------------------------------------------------------------*/
+'use strict';
+
+import * as vscode from 'vscode';
+
+/**
+ * API version information.
+ */
+export enum Version {
+    v0 = 0,         // 0.x.x
```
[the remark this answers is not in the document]
benmcmorran: Changed to 1.

api/api.ts
```diff
@@ -0,0 +1,244 @@
...
+
+    /**
+     * An event that fires when the selected launch target changes.
+     */
+    readonly onLaunchTargetNameChanged: vscode.Event<string>;
+
+    /**
+     * An event that fires when the active folder changes.
+     */
+    readonly onActiveFolderChanged: vscode.Event<vscode.WorkspaceFolder | undefined>;
```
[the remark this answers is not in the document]
benmcmorran: I've renamed the API calls to refer to projects and use URIs, but I didn't see any code that tracks the active project. Currently the implmentation internally is still using the URIs of workspace folders. Is that correct for now, or did I miss something?

api/api.ts
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * Hides the given UI element.
+     * @param element Element to hide.
+     */
+    hideUiElement(element: UiElement): Promise<void>;
+
+    /**
+     * An event that fires when the selected build target changes.
+     */
+    readonly onBuildTargetNameChanged: vscode.Event<string>;
```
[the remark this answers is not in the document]
benmcmorran: Renamed.

api/api.ts
```diff
@@ -0,0 +1,244 @@
...
+
+    /**
+     * An event that fires when the selected launch target changes.
+     */
+    readonly onLaunchTargetNameChanged: vscode.Event<string>;
```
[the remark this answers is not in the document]
benmcmorran: Renamed.

api/api.ts
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * An event that fires when the active folder changes.
+     */
+    readonly onActiveFolderChanged: vscode.Event<vscode.WorkspaceFolder | undefined>;
+
+    /**
+     * Gets the code model from the CMake File API, if it is available.
+     * @param folder The workspace folder containing the CMake project.
+     */
+    getFileApiCodeModel(folder: vscode.WorkspaceFolder): Promise<CodeModelContent | undefined>;
```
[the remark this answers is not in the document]
benmcmorran: Added a `getProject` top-level API call that returns a type that provides code model access. Old API functions still need to be added. The partner we're working with doesn't specifically need those old calls, but as long as it's cheap to add I should be able to add most of them back.

api/api.ts
```diff
@@ -0,0 +1,244 @@
...
+    | 'SHARED_LIBRARY'
+    | 'OBJECT_LIBRARY'
+    | 'EXECUTABLE'
+    | 'UTILITY'
+    | 'INTERFACE_LIBRARY';
+
+/**
+ * Describes a CMake target.
+ */
+export interface CodeModelTarget {
```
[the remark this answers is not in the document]
benmcmorran: Moved and renamed.

api/api.ts
benmcmorran: Renamed.

api/api.ts
benmcmorran: Renamed.

api/api.ts
benmcmorran: Changed default value.

src/legacyApi.ts
benmcmorran: I started doing that, but I'm concerned I might accidentally break something in the process. The extension appears to have dependencies on these types even when they aren't used for API access, and there are cases where functions are dynamically accessed by name (like command registration in setup) so language services will have a difficult time determining that there are no regressions. Do you think the tests are in a good enough state that passing tests after a refactor is enough to ensure quality?

api/api.ts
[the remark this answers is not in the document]
bobbrow: You are correct.  Support for multiple projects in a single workspace folder is currently under development.  It will we available soon.  I just didn't want us to add a method that would immediately be deprecated.

src/legacyApi.ts
bobbrow: I think most of the types survive.  They can move to TS modules where it makes sense, or we just have a new `types.ts` file where we can dump a bunch of stuff.  Just the old CMakeToolsAPI interface gets renamed to CMakeProject (or something) and we delete some of the functions/properties on it.

We currently only have `export class CMakeProject implements api.CMakeToolsAPI {` that I would consider to be a concern, but if we rename the interface as suggested above, not much actually changes.  You have CMakeToolsApiImpl which will be the real implementation of the new API and is the entry point to get a CMakeProject that they can work with.

benmcmorran: Legacy API is removed in the latest update. Most types went to `cache.ts`, `cmakeProject.ts`, and `cmakeDriver.ts`. I only refactored the types to remove `legacyApi.ts`, so any stale logic is still present.

api/api.ts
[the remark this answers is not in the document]
benmcmorran: Some of the old functionality is now added back.

benmcmorran:
@bobbrow This should be ready for re-review. If things look good, I can go ahead and create the new repository to host the API types, then once that is published come back here and take a dependency on it.

src/proc.ts
```diff
@@ -12,9 +12,9 @@ import rollbar from './rollbar';
 import * as util from './util';
 import * as vscode from 'vscode';
 import * as nls from 'vscode-nls';
-import { ExecutionResult } from './api';
+import { ExecutionResult } from '@cmt/cmakeProject';
 import { Environment, EnvironmentUtils } from './environmentVariables';
-export { ExecutionResult } from './api';
+export { ExecutionResult } from '@cmt/cmakeProject';
```
bobbrow, line 17: It feels like maybe `ExecutionResult` belongs in this module instead of `cmakeProject`.

src/cache.ts
```diff
@@ -15,19 +14,39 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
...
+    FilePath = 3,
+    Internal = 4,
+    Uninitialized = 5,
+    Static = 6,
+}
+
+/**
+ * Implements access to CMake cache entries.
+ */
+export class Entry {
```
bobbrow, line 32: Can we call this class `CacheEntry` instead?  It would solidify meaning in a number of places where we import the class name directly and just call it `Entry` instead of `cache.Entry`.  VS Code "rename" feature should make this easy on you.

api/api.ts
benmcmorran: Changed to pull from published NPM package in latest update.

src/cache.ts
benmcmorran: Renamed.

src/proc.ts
benmcmorran: I moved `ExecutionResult` to `proc.ts` and deleted `ExecutionOptions`. `proc.ts` had and existing different definition of `ExecutionOptions`, and it looks like the `cmakeProject.ts` version was only used by the now-removed API.

bobbrow (approved):

PR #2803, opened 2022-10-18
Handle build failure for task support

BIKA-C:
### This changes report when build has failed when using task

- Return error code when build has failed to indicate the task has faield
- Reveal the problem panel when build has failed like the java debugger extension

### DIFF
--- a/src/cmakeTaskProvider.ts
+++ b/src/cmakeTaskProvider.ts
@@ -373,6 +373,11 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
             const result: proc.ExecutionResult = await proc.execute(cmakePath, args, this, this.options).result;
             if (result.retc) {
                 this.writeEmitter.fire(localize("build.finished.with.error", "{0} finished with error(s).", taskName) + endOfLine);
+                if (doCloseEmitter) {
+                    this.closeEmitter.fire(result.retc);
+                }
+                vscode.commands.executeCommand("workbench.actions.view.problems")
+                return result.retc;
             } else if (result.stderr && !result.stdout) {
                 this.writeEmitter.fire(localize("build.finished.with.warnings", "{0} finished with warning(s).", taskName) + endOfLine);
             } else if (result.stdout && result.stdout.includes("warning")) {
--- a/src/cmakeTaskProvider.ts
+++ b/src/cmakeTaskProvider.ts
@@ -378,10 +378,12 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
                 }
                 vscode.commands.executeCommand("workbench.actions.view.problems")
                 return result.retc;
-            } else if (result.stderr && !result.stdout) {
+                
+              // to capture warning message. Previously the condition could not capture the case if both stderr and stdout have output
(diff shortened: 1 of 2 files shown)


bobbrow:
Thank you for the PR!  We will take a look in the next few days and let you know if we have any feedback.  Please also add a short description of the change in `CHANGELOG.md` and give yourself credit for your work.

BIKA-C:
@microsoft-github-policy-service agree company="default"

src/cmakeTaskProvider.ts
```diff
@@ -373,10 +373,17 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
             const result: proc.ExecutionResult = await proc.execute(cmakePath, args, this, this.options).result;
             if (result.retc) {
                 this.writeEmitter.fire(localize("build.finished.with.error", "{0} finished with error(s).", taskName) + endOfLine);
-            } else if (result.stderr && !result.stdout) {
-                this.writeEmitter.fire(localize("build.finished.with.warnings", "{0} finished with warning(s).", taskName) + endOfLine);
-            } else if (result.stdout && result.stdout.includes("warning")) {
+                if (doCloseEmitter) {
```
bobbrow, line 376: This shouldn't be done here.  There is a `this.closeEmitter.fire` call a few lines down where you should pass in `result.retc ?? 0` instead.

src/cmakeTaskProvider.ts
```diff
@@ -373,10 +373,17 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
...
+                    this.closeEmitter.fire(result.retc);
+                }
+                vscode.commands.executeCommand("workbench.actions.view.problems")
```
bobbrow, line 379: I don't think this is the correct approach.  To get the problems to focus, you need a problem matcher and then VS Code will do the right thing based on your settings in the task.  Add `"problemMatcher": "$msCompile"` to your task to make this work automatically.

src/cmakeTaskProvider.ts
```diff
@@ -373,10 +373,17 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
...
+                return result.retc;
```
bobbrow, line 380: Don't return early here.  Just update the `return 0;` down below to be `return result.retc ?? 0;`

src/cmakeTaskProvider.ts
```diff
@@ -373,10 +373,17 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
...
+                
+              // to capture warning message. Previously the condition could not capture the case if both stderr and stdout have output
+            } else if (result.stderr || (result.stdout && result.stdout.includes("warning"))) {
                 this.writeEmitter.fire(localize("build.finished.with.warnings", "{0} finished with warning(s).", taskName) + endOfLine);
+                vscode.commands.executeCommand("workbench.actions.view.problems")
```
bobbrow, line 385: Please revert these two lines. VS Code should do the right thing based on your task definition.

bobbrow (changes requested):

BIKA-C:
Yep yep. It's all fixed

bobbrow (approved):

PR #2805, opened 2022-10-18
When available, use the target from the toolchains file-api json

bobbrow:
## This change addresses item #2800 #1896

I think this should be all we need to do to enable passing the `--target` option to cpptools.  @Colengms do you have a sample project you could help me test this with?

### DIFF
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -485,12 +485,19 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
         }
 
         if (sysroot) {
-            if (!useFragments) {
-                // Send sysroot with quoting for CppTools API V5 and below.
-                flags.push('--sysroot=' + shlex.quote(sysroot));
-            } else {
+            if (useFragments) {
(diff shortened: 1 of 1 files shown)


src/cpptools.ts
```diff
@@ -485,12 +485,20 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
         }
 
         if (sysroot) {
-            if (!useFragments) {
+            if (useFragments) {
+                // Pass sysroot (without quote added) as the only compilerArgs for CppTools API V6 and above.
+                flags.push(`--sysroot=${sysroot}`);
```
elahehrashedi, line 490: didn't we want to add this to `compileCommandFragments` instead?

elahehrashedi (approved):

src/cpptools.ts
bobbrow: I'm just switching the order of the branches.  I'm not changing the implementation in this PR.  This works the way it is.

Issue #2810, opened 2022-10-20
clang (MSVC) kits should specify the 'ClangCL' toolset


fredemmott:
### Brief Issue Summary

Currently, the incorrect toolset 'host=amd64' is set (https://github.com/microsoft/vscode-cmake-tools/issues/2790) - as well as correcting 'amd64' to 'x64', `ClangCL` should be specified - e.g. `toolset: "ClangCL,host=x64"`

This makes `CMAKE_CXX_COMPILER_ID` and [`CMAKE_CXX_SIMULATE_ID`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_SIMULATE_ID.html) bepopulated correctly ('Clang' and 'MSVC' respectively, and the `MSVC` variable remains correctly set to 1)

For example:

```
[
  {
    "name": "Custom Clang-CL from MSVC 2022 x64",
    "visualStudio": "ee13335b",
    "visualStudioArchitecture": "x64",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "x64",
      "toolset": "ClangCL,host=x64"
    },
    "compilers": {
      "C": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\Llvm\\x64\\bin\\clang-cl.exe",
      "CXX": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\Llvm\\x64\\bin\\clang-cl.exe"
    }
  }
]
```

### Additional Information

Setting CMAKE_CXX_COMPILER_ID and CMAKE_CXX_SIMULATE_ID is useful to detect ClangCL within CMake; this is occasionally required when MSVC builds require a flag like `/await:strict` which errors with clang-cl.

bobbrow:
Thank you for opening this issue.  We should fix this in the same PR as the fix for #2790.

fredemmott:
Another related note is the `C` and `CXX` compiler specifications do not appear to do anything when using clang-cl - for example, if I specify:

```
    "compilers": {
      "C": "C:\\Program Files\\LLVM\\bin\\clang-cl.exe",
      "CXX": "C:\\Program Files\\LLVM\\bin\\clang-cl.exe",
    }
```

... or specify CMAKE_CXX_COMPILER by hand in the CMakeCache.txt, when using the visual studio generators, cmake will always use the Visual-Studio-bundled clang-cl instead of the one I tell it to.

[...]

Issue #2790, opened 2022-10-12
contains invalid field 'host=amd64'.


cjwijtmans:
### Brief Issue Summary

When selecting `clang 15.0.1 (MSVC CLI) for MSVC 17.4.33006.217` configuring cmake fails.
Selecting older clang or MSVC works.

```
[proc] Executing command: "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" --no-warn-unused-cli -DCMAKE_TOOLCHAIN_FILE:STRING=C:/usr/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET:STRING=x64-windows-static -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE "-DCMAKE_C_COMPILER:FILEPATH=C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64\bin\clang-cl.exe" "-DCMAKE_CXX_COMPILER:FILEPATH=C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64\bin\clang-cl.exe" -Sblah -Bblah/build -G "Visual Studio 17 2022" -T host=amd64 -A amd64
[cmake] Not searching for unused variables given on the command line.
[cmake] -- Selecting Windows SDK version 10.0.22621.0 to target Windows 10.0.19044.
[cmake] CMake Error at CMakeLists.txt:2 (project):
[cmake]   Generator
[cmake] 
[cmake]     Visual Studio 17 2022
[cmake] 
[cmake]   given toolset specification
[cmake] 
[cmake]     host=amd64
[cmake] 
[cmake]   that contains invalid field 'host=amd64'.
[cmake] 
[cmake] 
[cmake] -- Configuring incomplete, errors occurred!
```

### Additional Information

closed issue: https://github.com/microsoft/vscode-cmake-tools/issues/1352

elahehrashedi:
Thank you for reporting this issue. We will add a fix for this. Meanwhile, you can manually edit your kits and replace "amd64" with "x64" until we have a fix for this.

cjwijtmans:
Hello @elahehrashedi, I tried modifying it.

```json
  {
    "name": "Clang 15.0.1 (MSVC 17.4.33006.217) - amd64",
[...]

PR #2812, opened 2022-10-20
Generate correct ClangCL Kits

bobbrow:
Addresses #2790 and #2810


src/kitsController.ts
```diff
@@ -580,4 +552,49 @@ export class KitsController {
...
+
+    static isBetterMatch(newKit: Kit, existingKit?: Kit): boolean {
+        if (KitsController.isBetterClangCLDefinition(newKit, existingKit)) {
+            return true;
+        }
+        return KitsController.isBetterCompilerMatch(newKit.compilers, existingKit?.compilers);
+    }
+
+    static isBetterClangCLDefinition(newKit: Kit, existingKit?: Kit): boolean {
+        if (newKit.name.indexOf('MSVC CLI') >= 0) {
```
elahehrashedi, line 564: IS this part to fix the kits we created previously with "amd64" in their fields?

elahehrashedi (approved):

src/kitsController.ts
bobbrow: yes. If we don't do this, the old, incorrect kits will stay around forever.  Even after rescanning.

bobbrow: We might even want to be more aggressive in the future, detect the bad kits and automatically do the rescan.

PR #2817, opened 2022-10-25
Consider parallel jobs setting in generated defaultBuildPreset

maxmitti:
### This slightly modifies build preset handling when no CMake Build Preset is defined, but a Configuration Preset is used.

- Adhere to the config entry for parallel build jobs in the generated defaultBuildPreset.

## Other Notes/Information

I decided to pass the config class to the function which generates the default build preset, such that future changes according the defaultBuildPreset with regard to configuration values can be done easily without further cluttering the function signature.


maxmitti:

src/cmakeProject.ts
```diff
@@ -276,6 +276,7 @@ export class CMakeProject {
             buildPreset,
             lightNormalizePath(this.folder.uri.fsPath || '.'),
             this.sourceDir,
+            this.workspaceContext.config,
```
elahehrashedi, line 279: Can you just pass the `config.parallelJobs` instead of the whole `config`?

maxmitti: This was my rationale:

> I decided to pass the config class to the function which generates the default build preset, such that future changes according the defaultBuildPreset with regard to configuration values can be done easily without further cluttering the function signature.

But of course it can be changed.

bobbrow (approved):

nasinlock (approved):

PR #2828, opened 2022-10-28
Do not use the build task when building from the project outline.

bobbrow:
Addresses #2778


piomis:
Just my few cents.
Like I wrote in my PR which initially was intended to address #2778 - your proposal IMHO is just unblocking possibility to at least be able to build targets from project outline.
But IMHO one who decided to set option 'useBuildTask', does it from reason and expects that when any target build is executed as 'task' (e.g., to assign problem parser for build output).

bobbrow:
This isn't what 2778 wanted. Closing.

Issue #1374, opened 2020-07-17
Support Multiple CMakeLists.txt without Requiring a Multi-Root Workspace


MBetters:
### Brief Issue Summary

I maintain a repo in which there are multiple target architectures. For the most part, the executables being built for the different architectures have separate code, build dirs, and CMakeLists.txt files, because the processes running on the different archs do completely separate things (think a multi-component system made up of RTOS devices, beefier "cortex" boards, etc). There are some libs and include dirs shared commonly between these exec dirs, but not many. The ideal solution would be for CMake Tools to build these separate architectures independently. I tried a multi-root workspace + symlinks for the common code, but ran into issues with that (see https://github.com/microsoft/vscode-cmake-tools/issues/1371). Since symlinks are clunky and not great for developers who don't want to use Unix, ideally there'd be a way to just give CMake Tools a list of `CMakeLists.txt` files for it to build- something like `cmake.cmakelist_files`. Maybe such a setting could default to `[]` so that CMake Tools would carry on using  `${cmake.sourceDirectory}/CMakeLists.txt` like it normally does, but if `cmake.cmakelist_files` is set to multiple `CMakeLists.txt` files, it would build them all separately with different `build-*` dirs. Obviously it'd also be ideal that intellisense is supported from using the `compile_commands.json` from all the resulting `build-*` dirs, and auto-switches context depending on what file you're looking at.

A few potential ways to do this:

- Run multiple CMake Tools processes? Can VSCode even do that?
- Run CMake Configuration on all the CMakeLists.txt files listed in `cmake.cmakelist_files` (thus producing multiple `build-*` dirs), then merge all the `compile_commands.json` files into a single such file?

This request might sound a little far fetched, but I could see it being quite valuable and crucial to this extension's long-term success. Having to run CMake multiple times is quite common, not just for repos with multiple components. For example, building third-party libs that don't support `add_subdirectory(path/to/some-third-party-lib)`. It'd be AMAZING to have full intellisense support on such third-party libs.

Thanks for the awesome extension!

### Expected:

N/A

### Apparent Behavior:

N/A

### CMake Tools Log

N/A

### Developer Tools Log

N/A

### Platform and Versions

- **Operating System**: Win10 Remote-SSH'd into Ubuntu 20.04
- **CMake Version**: 3.16.3
- **VSCode Version**: 1.47.1
- **CMake Tools Extension Version**: 1.4.1
[...]

PR #2833, opened 2022-10-31
CMake multiproject support

bobbrow:
Addresses #1374


src/cmakeProject.ts
```diff
@@ -498,15 +529,18 @@ export class CMakeProject {
     });
 
     private readonly sourceDirSub = this.workspaceContext.config.onChange('sourceDirectory', async () => {
```
bobbrow, line 531: This doesn't look right.  I don't think the project should be listening for changes to `sourceDirectory`.  These changes should be listened for at the folder level.  The folder can then delete CMakeProjects that are no longer needed and add new ones when necessary.

src/cmakeProject.ts
```diff
@@ -545,7 +579,11 @@ export class CMakeProject {
         for (const sub of [this.generatorSub, this.preferredGeneratorsSub, this.communicationModeSub, this.sourceDirSub]) {
             sub.dispose();
         }
+        this.kitsController.dispose();
```
bobbrow, line 582: if the kitscontroller is shared by multiple projects in a workspaceFolder, then this would not be the correct place to dispose of it.

src/cmakeProject.ts
elahehrashedi: I added a listener at the folder level and missed removing this one. I think this was causing some errors.

bobbrow:
@elahehrashedi will resubmit. We found the issue we were looking for.

PR #2836, opened 2022-11-01
remove signal null warning

elahehrashedi:


src/proc.ts
```diff
@@ -211,7 +211,11 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
             });
             child?.on('exit', (code, signal) => {
                 if (code !== 0) {
-                    log.warning(localize({key: 'process.exit', comment: ['The space before and after all placeholders should be preserved.']}, 'The command: {0} exited with code: {1} and signal: {2}', `${cmdstr}`, `${code}`, `${signal}`));
+                    if (signal) {
```
bobbrow, line 214: If signal is 0, then it won't be printed.  Is that a valid signal?  If so, then you should check for signal !== null && signal !== undefined

bobbrow (approved):

Issue #2822, opened 2022-10-26
Clean Cmake "Active Folder"  commands


elahehrashedi:
- "CMake: Get Active Folder Path" and "CMake: Get Active Folder Name" don't need to be shown in the command palette. 
- cmake.selectActiveFolder should be modified to selectActiveProject

PR #2879, opened 2022-12-01
remove two commands + add telemetry

elahehrashedi:
I removed the two commands `CMake: Get active folder name` and `CMake: Get active folder path` from the command palette,
and added the new telemetry.

bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/2822

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1149,6 +1149,7 @@ export class ExtensionManager implements vscode.Disposable {
     }
 
     configure(folder?: vscode.WorkspaceFolder, showCommandOnly?: boolean) {
+        telemetry.logEvent("configure", { all: "false"});
         return this.runCMakeCommandForFolder(
             cmakeProject => cmakeProject.configureInternal(ConfigureTrigger.commandConfigure, [], showCommandOnly ? ConfigureType.ShowCommandOnly : ConfigureType.Normal),
             folder, undefined, true);
@@ -1159,6 +1160,7 @@ export class ExtensionManager implements vscode.Disposable {
     }
 
     configureAll() {
+        telemetry.logEvent("configure", { all: "true"});
         return this.runCMakeCommandForAll(cmakeProject => cmakeProject.configureInternal(ConfigureTrigger.commandCleanConfigureAll, [], ConfigureType.Normal), undefined, true);
     }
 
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -1224,26 +1229,27 @@ export class ExtensionManager implements vscode.Disposable {
...
 
     cleanAll() {
-        telemetry.logEvent("clean");
+        telemetry.logEvent("clean", { all: "true"});
         return this.buildAll(['clean']);
     }
 
     cleanRebuild(folder?: vscode.WorkspaceFolder) {
-        telemetry.logEvent("clean");
+        telemetry.logEvent("clean", { command: "cleanRebuild", all: "false"});
```
bobbrow, line 1242: Let's just rename this event to "cleanRebuild" instead of keeping it as "clean".

Same with the clean rebuild all.

src/extension.ts
```diff
@@ -1373,10 +1379,12 @@ export class ExtensionManager implements vscode.Disposable {
     }
 
     debugTarget(folder?: vscode.WorkspaceFolder, name?: string): Promise<vscode.DebugSession | null> {
+        telemetry.logEvent("debugTarget", { all: "false" });
```
bobbrow, line 1382: Can you call it "debug" instead of "debugTarget"?  Same for the "all" case.

src/extension.ts
```diff
@@ -1387,10 +1395,12 @@ export class ExtensionManager implements vscode.Disposable {
     }
 
     launchTarget(folder?: vscode.WorkspaceFolder, name?: string): Promise<vscode.Terminal | null> {
+        telemetry.logEvent("launchTarget", { all: "false" });
```
bobbrow, line 1398: Can you call this "launch" instead of "launchTarget"?  Same for the "all" case.

src/extension.ts
```diff
@@ -1401,6 +1411,7 @@ export class ExtensionManager implements vscode.Disposable {
     }
 
     selectLaunchTarget(folder?: vscode.WorkspaceFolder, name?: string) {
+        telemetry.logEvent("selectLaunchTarget");
```
bobbrow, line 1414: We don't need telemetry for this one.

bobbrow (changes requested):

bobbrow (approved):

PR #2887, opened 2022-12-03
Add arm64 to MSVC hosts

scaryrawr:
### This enables using native arm64 toolset

- Add arm64 as a valid host for MSVC

## The purpose of this change

Visual Studio 2022 17.4 Preview introduced native build tools on arm64. This change enables the vscode-cmake-tools to detect the native tools so we can build with it instead of emulating x86/x64 to run the non-native build tools.

image

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -588,7 +588,7 @@ function vsKitName(inst: VSInstallation, hostArch: string, targetArch?: string):
 /**
  * Possible msvc host architectures
  */
-export const MSVC_HOST_ARCHES: string[] = ['x86', 'x64'];
+export const MSVC_HOST_ARCHES: string[] = ['x86', 'x64', 'arm64'];
 
 /**
  * Gets the environment variables set by a shell script.
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -585,10 +585,12 @@ function vsKitName(inst: VSInstallation, hostArch: string, targetArch?: string):
     return `${vsDisplayName(inst)} - ${getHostTargetArchString(hostArch, targetArch, true)}`;
 }
 
+export type MsvcHostArches = 'x86' | 'x64' | 'ARM64';
+
 /**
  * Possible msvc host architectures
  */
-export const MSVC_HOST_ARCHES = ['x86', 'x64', 'ARM64'] as const;
+export const MSVC_HOST_ARCHES: MsvcHostArches[] = ['x86', 'x64'];
 
 /**
  * Gets the environment variables set by a shell script.
@@ -739,12 +741,25 @@ async function tryCreateNewVCEnvironment(inst: VSInstallation, hostArch: string,
  */
 export async function scanForVSKits(pr?: ProgressReporter): Promise<Kit[]> {
     const installs = await vsInstallations();
+
+    // Exclude ARM64 host checking on x86, x64, and unknown PROCESSOR_ARCHITECTURE
+    const hostArches: MsvcHostArches[] = (
+        // To determine ARM64 host, we check the processor id, which for Qualcomm's 8cx starts with
+        // ARMv8. If we check the PROCESSOR_ARCHITECTURE it's possible that we're emulating
+        // x86_64 on ARM64 and getting AMD64 for example. The processor id doesn't change
+        // even during x86_64 emulation.
+        process.env.PROCESSOR_IDENTIFIER?.startsWith('ARMv8')
+    ) ?
+        [...MSVC_HOST_ARCHES, 'ARM64'] :
+        MSVC_HOST_ARCHES;
+
     const prs = installs.map(async (inst): Promise<Kit[]> => {
         const ret = [] as Kit[];
         const targetArches: string[] = ['x86', 'x64', 'arm', 'arm64'];
         const version = util.tryParseVersion(inst.installationVersion);
 
         const sub_prs: Promise<Kit | null>[] = [];
+        hostArches.forEach(hostArch => {
             if (
                 hostArch === 'ARM64' &&
                 (
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -745,8 +745,13 @@ export async function scanForVSKits(pr?: ProgressReporter): Promise<Kit[]> {
         const version = util.tryParseVersion(inst.installationVersion);
 
         const sub_prs: Promise<Kit | null>[] = [];
-        MSVC_HOST_ARCHES.forEach(hostArch => {
-            if (hostArch === 'ARM64' && !(version && version.major >= 17 && version.minor >= 4)) {
+            if (
+                hostArch === 'ARM64' &&
+                (
+                    !version ||
+                    version.major < 17 || (version.major === 17 && version.minor < 4)
+                )
+            ) {
                 // ARM64 support as a host was added in Visual Studio 2022 17.4 and above,
                 // so we'll avoid checking it on anything lower.
                 return;
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -588,7 +588,7 @@ function vsKitName(inst: VSInstallation, hostArch: string, targetArch?: string):
 /**
  * Possible msvc host architectures
  */
-export const MSVC_HOST_ARCHES: string[] = ['x86', 'x64', 'arm64'];
+export const MSVC_HOST_ARCHES: string[] = ['x86', 'x64', 'ARM64'];
 
 /**
  * Gets the environment variables set by a shell script.
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -742,13 +742,11 @@ async function tryCreateNewVCEnvironment(inst: VSInstallation, hostArch: string,
 export async function scanForVSKits(pr?: ProgressReporter): Promise<Kit[]> {
     const installs = await vsInstallations();
 
-    // Exclude ARM64 host checking on x86, x64, and unknown PROCESSOR_ARCHITECTURE
+    // Exclude ARM64 host checking on x86 and x64
     const hostArches: MsvcHostArches[] = (
-        // To determine ARM64 host, we check the processor id, which for Qualcomm's 8cx starts with
-        // ARMv8. If we check the PROCESSOR_ARCHITECTURE it's possible that we're emulating
-        // x86_64 on ARM64 and getting AMD64 for example. The processor id doesn't change
-        // even during x86_64 emulation.
-        process.env.PROCESSOR_IDENTIFIER?.startsWith('ARMv8')
+        // To determine ARM64 host, we check the PROCESSOR_IDENTIFIER (please see: https://learn.microsoft.com/en-us/troubleshoot/windows-server/deployment/determine-the-type-of-processor)
+        // ARMv8 will be contained in the string on ARM64 devices.
+        process.env.PROCESSOR_IDENTIFIER?.includes('ARMv8')
     ) ?
         [...MSVC_HOST_ARCHES, 'ARM64'] :
         MSVC_HOST_ARCHES;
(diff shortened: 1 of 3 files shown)


scaryrawr:
Ah, it currently looks like cmake errors out at this time since it only expects x86 or amd64 for the host.

scaryrawr:
It looks like if I modify the `%LOCALAPPDATA%\CMakeTools\cmake-tools-kits.json` and remove the instances of:

```json
"toolset": "host=arm64"
```

We stop passing in `-T host=arm64` which works-ish for building and debugging... but not sure how to get this working correctly 😅. From [cmake documentation](https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2017%202022.html), it looks like only `host=x86` or `host=x64` are valid options, and not `host=arm64`.

elahehrashedi:
Is this screenshot from an arm64 machine where you scan for kits? 
if yes, do all 3 host compilers exists? the paths look something like this:

`C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.34.31921\bin\Hostx64`

scaryrawr:
> Is this screenshot from an arm64 machine where you scan for kits? if yes, do all 3 host compilers exists? the paths look something like this:
> 
> `C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.34.31921\bin\Hostx64`

It is a screenshot from an arm64 machine where all 3 host compilers exist.

There appears to be an issue with the cmake command that gets ran when attempting to configure with an arm64 host. `-T host=arm64` does not work as the only valid values are `x86` or `x64`, so with just the changes to scan it won't work for building yet.

elahehrashedi:
Can you check your hard drives to see if all 3 host compilers exist? and share the paths here? 
(I can't see the paths from the screenshot)

bobbrow:
Looking at the source code for CMake, it looks like ARM64 is all capitalized.  Can you try that instead of `arm64`?

https://github.com/Kitware/CMake/blob/af58b6d2ad9a000146e8db2b1c3c74b1a93456ef/Source/cmGlobalVisualStudio12Generator.cxx#L141

scaryrawr:
> Looking at the source code for CMake, it looks like ARM64 is all capitalized (and case-sensitive). Can you try that instead of `arm64`?
> 
> https://github.com/Kitware/CMake/blob/af58b6d2ad9a000146e8db2b1c3c74b1a93456ef/Source/cmGlobalVisualStudio12Generator.cxx#L141

Thanks for digging into this @bobbrow ! I had to do another check to get it back to lower case (otherwise I was missing `arm64` as both host and target), but it appears to be working and generate this as my `cmake-tools-kits.json`:

```json
[
  {
    "name": "Clang 15.0.1 (GNU CLI) for MSVC 17.5.33103.201 (Visual Studio Enterprise 2022 Preview - amd64)",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x64",
    "compilers": {
      "C": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\VC\\Tools\\Llvm\\x64\\bin\\clang.exe",
      "CXX": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\VC\\Tools\\Llvm\\x64\\bin\\clang.exe"
    }
  },
  {
    "name": "Clang 15.0.1 (GNU CLI) for MSVC 17.5.33103.201 (Visual Studio Enterprise 2022 Preview - x86)",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x86",
    "compilers": {
      "C": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\VC\\Tools\\Llvm\\bin\\clang.exe",
      "CXX": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\VC\\Tools\\Llvm\\bin\\clang.exe"
    }
  },
  {
    "name": "Clang 15.0.1 (MSVC CLI) for MSVC 17.5.33103.201 (Visual Studio Enterprise 2022 Preview - amd64)",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x64",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "x64",
      "toolset": "ClangCL,host=x64"
    },
    "compilers": {
      "C": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\VC\\Tools\\Llvm\\x64\\bin\\clang-cl.exe",
      "CXX": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\VC\\Tools\\Llvm\\x64\\bin\\clang-cl.exe"
    }
  },
  {
    "name": "Clang 15.0.1 (MSVC CLI) for MSVC 17.5.33103.201 (Visual Studio Enterprise 2022 Preview - x86)",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x86",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "win32",
      "toolset": "ClangCL,host=x86"
    },
    "compilers": {
      "C": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\VC\\Tools\\Llvm\\bin\\clang-cl.exe",
      "CXX": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\VC\\Tools\\Llvm\\bin\\clang-cl.exe"
    }
  },
  {
    "name": "Visual Studio Enterprise 2022 Preview - amd64",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x64",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "x64",
      "toolset": "host=x64"
    }
  },
  {
    "name": "Visual Studio Enterprise 2022 Preview - amd64_arm64",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x64",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "arm64",
      "toolset": "host=x64"
    }
  },
  {
    "name": "Visual Studio Enterprise 2022 Preview - amd64_x86",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x64",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "win32",
      "toolset": "host=x64"
    }
  },
  {
    "name": "Visual Studio Enterprise 2022 Preview - arm64",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "ARM64",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "arm64",
      "toolset": "host=ARM64"
    }
  },
  {
    "name": "Visual Studio Enterprise 2022 Preview - arm64_amd64",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "ARM64",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "x64",
      "toolset": "host=ARM64"
    }
  },
  {
    "name": "Visual Studio Enterprise 2022 Preview - arm64_x86",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "ARM64",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "win32",
      "toolset": "host=ARM64"
    }
  },
  {
    "name": "Visual Studio Enterprise 2022 Preview - x86",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x86",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "win32",
      "toolset": "host=x86"
    }
  },
  {
    "name": "Visual Studio Enterprise 2022 Preview - x86_amd64",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x86",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "x64",
      "toolset": "host=x86"
    }
  },
  {
    "name": "Visual Studio Enterprise 2022 Preview - x86_arm64",
    "visualStudio": "f64f9d0b",
    "visualStudioArchitecture": "x86",
    "preferredGenerator": {
      "name": "Visual Studio 17 2022",
      "platform": "arm64",
      "toolset": "host=x86"
    }
  }
]
```

scaryrawr:
Ah, the test failure appears to be because arm64 as a host was added in Visual Studio 2022 17.4 and running against 2019 where arm64 is an invalid option.

bobbrow:
The test appears to be failing for something else, but there are a lot of errors printing out during Kit scanning that should be addressed before we take this PR.  It should probably be addressed in `tryCreateNewVCEnvironment` in `kit.ts`.  It has access to the `VSInstallation` info, so I think you need to do a version check there before treating ARM64 as a valid host arch.

bobbrow:
> The test appears to be failing for something else, but there are a lot of errors printing out during Kit scanning that should be addressed before we take this PR. It should probably be addressed in `tryCreateNewVCEnvironment` in `kit.ts`. It has access to the `VSInstallation` info, so I think you need to do a version check there before treating ARM64 as a valid host arch.

The commit you just pushed works too, though I'd like to see if we can get a more accurate version number to test against.  I believe we have APIs for that.  Give me a minute.

bobbrow:
> The commit you just pushed works too, though I'd like to see if we can get a more accurate version number to test against. I believe we have APIs for that. Give me a minute.

There is a `parseVersion` function in `util.ts` but it expects a 3-part version number.  That function could be modified to accept a 2-part version number and then you could use it to get the version number instead of `parseInt` which is kind of a hack that was never cleaned up.

src/kit.ts
```diff
@@ -742,9 +742,16 @@ export async function scanForVSKits(pr?: ProgressReporter): Promise<Kit[]> {
     const prs = installs.map(async (inst): Promise<Kit[]> => {
         const ret = [] as Kit[];
         const targetArches: string[] = ['x86', 'x64', 'arm', 'arm64'];
+        const version = util.tryParseVersion(inst.installationVersion);
 
         const sub_prs: Promise<Kit | null>[] = [];
         MSVC_HOST_ARCHES.forEach(hostArch => {
+            if (hostArch === 'ARM64' && !(version && version.major >= 17 && version.minor >= 4)) {
```
bobbrow, line 749: This condition should also succeed if `version.major > 17` or else 18.0-18.3 will inevitably fail in the future.

src/util.ts
```diff
@@ -351,6 +351,20 @@ export function parseVersion(str: string): Version {
     };
 }
 
+export function tryParseVersion(str: string): Version | undefined {
```
bobbrow, line 354: To avoid multiple copies of the same code, I would prefer if this was written to call the `parseVersion` function instead.
e.g.
```ts
// warning: code written in browser, not tested
export function tryParseVersion(str: string): Version | undefined {
    try {
        return parseVersion(str);
    } catch {
        return undefined;
    }
}
```

bobbrow (changes requested):

src/kit.ts
bobbrow: Sorry, one last comment here, we should probably also disallow host=ARM64 if you're not on an ARM64 machine.  I guess we're not technically disabling x64 on x86 machines either, but I'm more concerned about adding unusable kits on x64 which has more users than ARM64 right now.

scaryrawr: Ah, the second point is interesting, as ARM64 can act as x86 and x64 host, but only ARM64 can be an arm64 host 🤣, so good point.

scaryrawr: I think I've fixed the conditional 🤔, the minor version for some reason makes it confusing 😅.

bobbrow (approved):
Looks great!  Thanks so much for helping out!  I'll follow up with an entry in the changelog to make sure you are recognized for your help.

Issue #2894, opened 2022-12-13
`command:cmake.buildType` always returns Debug when using CMakePresets


calebkiage:
### Brief Issue Summary

When using CMake presets, the `command:cmake.buildType` always resolves to Debug.
It can be tested with this task:
```jsonc
{
    "version": "2.0.0",
    "tasks": [
        {
            "type": "shell",
            "label": "test env",
            "command": "echo",
            "args": [
                "$env:Test"
            ],
            "options": {
                "env": {
                    "Test": "${command:cmake.buildType}"
                }
            },
            "problemMatcher": []
        }
    ]
}
```

TEST is always `Debug`

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.74.0",
  "cmtVersion": "1.13.28",
  "configurations": [
    {
      "folder": "c:\\Users\\calebmagiya\\Documents\\projects\\practice\\cpp\\sample",
      "cmakeVersion": "3.24.2",
      "configured": false,
      "generator": "Visual Studio 17 2022",
      "usesPresets": true,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [
      "file:///c%3A/Users/calebmagiya/Documents/projects/practice/cpp/sample/main.cpp"
    ],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": false
    }
  ]
}
```

### Debug Log

```shell
[main] Building folder: sample 
[main] Saving open files before configure/build
[main] Configuring folder: sample 
[main] Saving open files before configure/build
[driver] Start configure 
[driver] Running pre-configure checks and steps
[cmakefileapi-driver] Configuring using preset
[cmakefileapi-driver] Invoking CMake C:\Program Files\CMake\bin\cmake.EXE with arguments ["-SC:/Users/calebmagiya/Documents/projects/practice/cpp/sample","-BC:/Users/calebmagiya/Documents/projects/practice/cpp/sample/build","-G","Visual Studio 17 2022"]
[...]

PR #2895, opened 2022-12-13
return the preset buildType

elahehrashedi:
return the value of `"CMAKE_BUILD_TYPE"` in configure preset when running `cmake.buildType` command.
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/2894

### DIFF
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -1990,11 +1990,33 @@ export class CMakeProject {
      */
     async currentBuildType(): Promise<string | null> {
         let buildType: string | null = null ;
-        if (this.useCMakePresets && this.configurePreset && this.configurePreset.cacheVariables) {
-            const variable = this.configurePreset.cacheVariables["CMAKE_BUILD_TYPE"];
-            buildType = util.isString(variable) ? variable : null;
+        if (this.useCMakePresets) {
+            if (this.buildPreset) {
+                if (this.buildPreset.configuration) {
+                    // The `configuration` is set for multi-config generators, and is optional for single-config generators.
+                    // Get the first value for multi-config generators
+                    buildType = this.buildPreset.configuration.split(';')[0];
+                } else {
+                    try {
+                        const driver: CMakeDriver | null = await this.getCMakeDriverInstance();
+                        if (driver?.isMultiConfig) {
+                            // Get the value from cache for multi-config generators
+                            const cache: CMakeCache = await CMakeCache.fromPath(await this.cachePath);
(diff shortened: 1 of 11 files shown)


bobbrow (changes requested):

src/preset.ts
```diff
@@ -1653,3 +1645,12 @@ export function getValue(value: string | ValueStrategy): string | undefined {
         return value.value;
     }
 }
+
+export function getStringValueFromCacheVar(variable?: CacheVarType): string | null {
```
bobbrow, line 1649: please use `undefined` instead of `null`

bobbrow: Hmm.. There is a bit of legacy here.  Nevermind, leave it as `null` for now.

src/cmakeProject.ts
```diff
@@ -1989,7 +1989,37 @@ export class CMakeProject {
...
      */
     async currentBuildType(): Promise<string | null> {
-        return this.variantManager.activeVariantOptions.buildType || null;
+        let buildType: string | null = null ;
+        if (this.useCMakePresets) {
+            if (this.buildPreset) {
+                if (this.buildPreset.configuration) {
+                    // The `configuration` is set for multi-config generators, and is optional for single-config generators.
+                    // Get the first value for multi-config generators
+                    buildType = this.buildPreset.configuration.split(';')[0];
```
bobbrow, line 1998: According to the Presets spec, `this.buildPreset.configuration` should not be an array that needs to be split.  It should just be the string which has the `CMAKE_BUILD_TYPE` you are looking for.

src/cmakeProject.ts
```diff
@@ -1989,7 +1989,37 @@ export class CMakeProject {
...
+                    } catch (e: any) {
+                    }
+                }
+            }
+            if (!buildType && this.configurePreset && this.configurePreset.cacheVariables) {
+                // Single config generators set the build type in config preset.
+                buildType = preset.getStringValueFromCacheVar(this.configurePreset.cacheVariables["CMAKE_BUILD_TYPE"]);
+            }
+        }
+        if (!buildType) {
```
bobbrow, line 2019: This can just be an `else` instead of a new `if` since you only go here if you're not using presets.

bobbrow (changes requested):

src/cmakeProject.ts
bobbrow: Not fixed yet

bobbrow (approved):

Issue #2904, opened 2022-12-23
Running CTest rebuilds other build configuration in CMake multi generator configuration


jorisvanderpol:
### Brief Issue Summary

CMake Tools 1.12.27
Visual Studio Code 1.72.2

I have created a project with the following `CMakePresets.json`:

```
{
    "version": 6,
    "cmakeMinimumRequired": {
        "major": 3,
        "minor": 25,
        "patch": 1
    },
    "configurePresets": [
        {
            "name": "v140_x64",
            "displayName": "Visual Studio 2015",
            "generator": "Visual Studio 14 2015",
            "architecture": "x64",
            "environment": {
                "CMAKE_CONFIGURATION_TYPES": "Debug;Release"
            }
        }
    ],
    "buildPresets": [
        {
            "name": "v140_x64_debug",
            "displayName": "Build Debug",
            "configurePreset": "v140_x64",
            "configuration": "Debug"
            },
        {
            "name": "v140_x64_release",
            "displayName": "Build Release",
            "configurePreset": "v140_x64",
            "configuration": "Release"
        }
    ],
    "testPresets": [
        {
            "name": "v140_x64_debug",
            "displayName": "Test Debug",
            "configurePreset": "v140_x64",
            "configuration": "Debug"
        },
        {
            "name": "v140_x64_release",
            "displayName": "Test Release",
            "configurePreset": "v140_x64",
            "configuration": "Release"
        }
    ]
}
```

If I now select the `Build Debug` as the active build preset and the `Test Release` as the active test preset, it will build the `Debug` configuration and run the CTest tests with the `Release` configuration. This is confusing behavior. Removing the configuration from the test preset will not work (assuming it takes over the configuration of the build preset), as it produces an error (`Test not available without configuration`).

bobbrow:
I believe that this is a deficiency with the spec for test presets.  test presets are dependent on a configurePreset, not a buildPreset, so it wasn't considered when we implemented this.  Visual Studio appears to have the same problem.

It seems reasonable to me that we could filter out the test presets that don't apply to the active build preset when offering the list of presets to select since the build type is known.  We could probably also make the case where the build type is unknown work as well (assuming that it works on the command line... need to check).

MartinsSmirnovs:
@bobbrow 
Could it be possible to not to build at all when running CTests? There are cases when existing binary is not necessary for test case, only CMake configure.

bobbrow:
[...]

PR #2911, opened 2022-12-29
Compatibility between test and build presets is not enforced

bobbrow:
Addresses #2904


src/presetsController.ts
```diff
@@ -726,6 +726,20 @@ export class PresetsController {
         return selectedConfigurePreset;
     }
 
+    private async checkBuildPreset(): Promise<preset.BuildPreset | null> {
```
elahehrashedi, line 729: This seems more like a "pickBuildPreset"

src/presetsController.ts
```diff
@@ -802,18 +826,45 @@ export class PresetsController {
         }
     }
 
+    private checkCompatibility(configurePreset: preset.ConfigurePreset | null, buildPreset?: preset.BuildPreset | null, testPreset?: preset.TestPreset | null): {buildPresetCompatible: boolean; testPresetCompatible: boolean} {
```
elahehrashedi, line 829: can we rename this to "checkPresetCompatibility"?

elahehrashedi (approved):

src/presetsController.ts
bobbrow: The 3 inputs are presets, so I don't think we need to include "preset" in the name of the function.  I think it should be very clear what the subject of this function is.

src/presetsController.ts
bobbrow: It is named to match `checkConfigurePreset` above this.  `ensure` might be a better name and I think we've used it elsewhere, but I didn't think it was a big enough deal to change.  "pick" is not the best word choice because this function does not pick.  It will give you an option to pick if one is not selected already, but will do nothing if one is already "picked."  I can change it to "ensure" in both functions if you feel strongly about it.

PR #2913, opened 2022-12-29
Warn the user if the build directory is not unique for multi-project

bobbrow:
Missing implementation from #1374. There is a slight deviation from the proposal in that we won't block the user from using ${workspaceFolder} if they find a way to make it work themselves. But we will be pretty noisy about attempts to reuse the same buildDirectory for multiple projects in the same folder.

This also fixes some bugs with the extension not processing changes to buildDirectory and installPrefix immediately.

bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2916


test/extension-tests/successful-build/test/variable-substitution.test.ts
```diff
@@ -36,6 +36,7 @@ suite('Variable Substitution', () => {
             configureSettings: {
                 workspaceRoot: '${workspaceRoot}',
                 workspaceFolder: '${workspaceFolder}',
```
elahehrashedi, line 38: We can add it to the test.

elahehrashedi:
I am testing this branch. There is a bug, the status bar disappears after running build.

elahehrashedi:
Let's test it after https://github.com/microsoft/vscode-cmake-tools/pull/2926 is checked in.

elahehrashedi (approved):

Issue #2915, opened 2022-12-30
1.13.32/1.13.33 regression: Numerous Include Errors in Project


scottfurry:
### Brief Issue Summary

Version 1.13.31 works as expected.

Version 1.13.32 and 1.13.33 both show the same problem with ms-vscode.cpptools extension integration resulting in tremendous number of include errors on any open file in the editor. This is in spite of paths being set properly and working previously. Clearing cache files and/or reloading CMake settings for project does not resolve the issue.

CMake execution is normal. Project builds without error and can be executed as normal.
Issue appears to be disconnect between CMake discovered paths not being shared with ms-vscode.cpptools.

Reverting to version 1.13.31 of extension as a workaround.

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.74.2",
  "cmtVersion": "1.13.33",
  "configurations": [
    {
      "folder": "[PROJECT_HOME]",
      "cmakeVersion": "3.25.1",
      "configured": true,
      "generator": "Unix Makefiles",
      "usesPresets": false,
      "compilers": {
        "C": "/usr/lib/llvm/15/bin/clang",
        "CXX": "/usr/lib/llvm/15/bin/clang++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
[main] Safe constructing new CMakeProject instance
[variant] Constructing VariantManager
[main] Constructing new CMakeProject instance
[main] Starting CMake Tools second-phase init
[variant] Loaded new set of variants
[kit] Reading kits file [HOME]/.local/share/CMakeTools/cmake-tools-kits.json
[kit] Successfully loaded 3 kits from [HOME]/.local/share/CMakeTools/cmake-tools-kits.json
[kit] Not reading non-existent kits file: [PROJECT_HOME]/.vscode/cmake-kits.json
[main] Injecting new Kit into CMake driver
[main] CMakeProject instance initialization complete.
```

### Additional Information

NOTE - for log outputs above - paths edited
[...]

PR #2926, opened 2023-01-06
update codeModel in the Second-phase async init

elahehrashedi:
bug fix: https://github.com/microsoft/vscode-cmake-tools/issues/2915


src/extension.ts
```diff
@@ -174,6 +174,8 @@ export class ExtensionManager implements vscode.Disposable {
             }
             await this.initActiveProject();
         }
+        rollbar.takePromise('Post-folder-open', { folder: this.getActiveProject()?.folderPath }, this.postWorkspaceOpen(this.getActiveProject()));
```
bobbrow, line 177: It doesn't seem like this one is needed.  `postUpdateActiveProject` is called before we get here so `postWorkspaceOpen` will have already been called.

bobbrow (changes requested):
Please test without the change in `init`.  I don't think it's needed.

bobbrow:
> It doesn't seem like this one is needed. `postUpdateActiveProject` is called before we get here so `postWorkspaceOpen` will have already been called.

Or perhaps what we need to do instead is call `postWorkspaceOpen` on all projects, not just the active one.  That's what we used to do.  I see that the project outline does not initialize all the projects when you first open the workspace.  This is because we don't configure them all.  This could be considered a regression.

bobbrow:
`postWorkspaceOpen` should only ever be called once per project. I changed the logic and pushed a change which appears to work for my multi-project workspace.  Please confirm that this works for you as well.

elahehrashedi:
In my vc multi-project, this worked after a re-load, the first time that I opened the project, the browse path was still not available.
I have another simpler multi-project that this worked on.

bobbrow (approved):

Issue #2746, opened 2022-09-14
Re-open pop up in partial activation experiment in cmake-tools when user goes to configure


sinemakinci1:
### Brief Issue Summary

Currently, this command palette pop-up only opens up when users open a partially activated project (CMakeLists.txt is not in root folder), and when the user clicks out of it to do something else the feature gets flagged as ignored and it does not open up again in the same session.

This is a request to ship this feature we were experimenting with in the next milestone due to good experimentation results, but update the feature so that if the user is going to configure/run/debug/build, that the pop-up shows up again, and update the ignore telemetry accordingly.

PR #2931, opened 2023-01-09
Pop up "Choose CMakeLists.txt" when user goes to configure while feature set is partially activated.

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2746 (a bug related to this feature request).
Will re-base the branch later.


nasinlock (approved):

src/cmakeProject.ts
```diff
@@ -2674,10 +2670,7 @@ export class CMakeProject {
         if (path.basename(expandedSourceDirectory).toLocaleLowerCase() !== "cmakelists.txt") {
             expandedSourceDirectory = path.join(expandedSourceDirectory, "CMakeLists.txt");
         }
-
-        this.hasCMakeListsFile = await fs.exists(expandedSourceDirectory);
-
-        return this.hasCMakeListsFile;
+        return fs.exists(expandedSourceDirectory);
```
bobbrow, line 2673: It might be good for us to come back and make this more efficient in the future (or do it now).  For example, whenever we set `this._sourceDir` for the CMakeProject we could check the file exists and set `hasCMakeListsFile` at that time so we don't have to do any extra work in this function.  This function should probably have been calling `util.normalizeAndVerifySourceDir` instead of doing this custom work for the setting to populate `expandedSourceDirectory`.

bobbrow (approved):

PR #2942, opened 2023-01-11
Remove default value for mingwSearchDirs

bobbrow:
Remove the default path for `cmake.mingwSearchDirs` since the path is world-writable.


sean-mcmanus (approved):

sean-mcmanus:
@bobbrow Without the default value, the description isn't sufficient to understand the format of the setting, e.g. setting
```
"cmake.mingwSearchDirs": [
        "C:\\msys64\\mingw64\\bin"
    ]
```
doesn't work. The "\\bin" has to be removed. It may be helpful to others if the description included an example or something.

PR #2943, opened 2023-01-11
Fix configProvider activation for multi-root

bobbrow:
Addresses #2915.  Multi-root workspaces would not activate the configProvider unless all folders had CMake projects in them.


src/extension.ts
```diff
@@ -674,12 +674,15 @@ export class ExtensionManager implements vscode.Disposable {
                 }
                 // Inform cpptools that custom CppConfigurationProvider will be able to service the current workspace.
                 this.ensureCppToolsProviderRegistered();
-                if (cpptools.notifyReady && this.cpptoolsNumFoldersReady < this.projectController.numOfWorkspaceFolders) {
-                    ++this.cpptoolsNumFoldersReady;
-                    if (this.cpptoolsNumFoldersReady === this.projectController.numOfWorkspaceFolders) {
-                        // Notify cpptools that the provider is ready to provide IntelliSense configurations.
-                        cpptools.notifyReady(this.configProvider);
-                        this.configProvider.markAsReady();
+                if (cpptools.notifyReady) {
```
bobbrow, line 677: This was a long standing bug. These conditions need to be separated.  Also only affected multi-root.

bobbrow: The `else` case is for old versions of the config provider API (< 2.1) and we shouldn't have been going there for multi-root projects.

src/extension.ts
```diff
@@ -674,12 +674,15 @@ export class ExtensionManager implements vscode.Disposable {
...
+                    const projectCount = await this.projectController.getNumOfValidProjects();
+                    if (this.cpptoolsNumFoldersReady < projectCount) {
```
bobbrow, line 679: I'm seriously debating removing this check though.  If one project is ready, maybe we should just turn the config provider on immediately.

elahehrashedi: I also think we can just call `cpptools.notifyReady(this.configProvider)` if one project has a CMakeLists.txt `project.hasCMakeLists()`

elahehrashedi (approved):

elahehrashedi (approved):

PR #2944, opened 2023-01-13
Add an optional description field to cmake-kits.json file

TisziV:

- Add an optional description field to cmake-kits.json file.
- This description will appear in the kit selection menu instead of the default "Kit for toolcain file ...".

## The purpose of this change

In some projects there are many architectures to build for, and it is not always straightforward which kit should be used for which platform.
Hints helping with eg. product or platform names can be added to the description text to disambiguate.


TisziV:

src/kit.ts
```diff
@@ -1093,6 +1098,9 @@ export async function scanForKitsIfNeeded(project: CMakeProject): Promise<boolea
  */
 export async function descriptionForKit(kit: Kit, shortVsName: boolean = false): Promise<string> {
     if (kit.toolchainFile) {
+        if (kit.description) {
```
bobbrow, line 1101: This should be moved outside (and above) the branch for `kit.toolchainFile`

bobbrow (changes requested):

elahehrashedi (approved):

src/kit.ts
TisziV: Thanks for the comment.
Added it to later commit.
Also adjusted kits.md accordingly.

elahehrashedi:
@bobbrow adding description seems fine:

image

bobbrow (approved):
@elahehrashedi will fix the changelog and then this is good to go.

Issue #1963, opened 2021-06-29
CMakePresets.json schema is too restrictive


KUGA2:
### Brief Issue Summary

I cannot add the new CMake 3.21 feature "conditions" to the CMakePresets.json without breaking cmake-tools. Or any other feature that might come.

The plugin validates the CMakePresets.json and does not load it.
The same issue was seen here https://github.com/microsoft/vscode-cmake-tools/issues/1934, but this Issue is for a more general issue.

### Apparent Behaviour

Tooltip:
![image](https://user-images.githubusercontent.com/10581931/123763173-cfb68600-d8c3-11eb-9bad-5586e30cf3e1.png)

And no preset found:
![image](https://user-images.githubusercontent.com/10581931/123763309-ec52be00-d8c3-11eb-9177-1d9e5caa605f.png)

#### Code:

```
        {
            "name": "x86_64-windows-user-vs2019-Release",
            ...
            "condition": {
                "type": "equals",
                "lhs": "$env{VSCMD_ARG_TGT_ARCH}",
                "rhs": "x64"
            }
        },
```

### Expected:

It should leave the validation to cmake.exe. e.g. you could run cmake --list-presets and then offer those shown (this will also implicitly enable the condition feature).

```
$ set | grep VSCMD_ARG_TGT_ARCH
VSCMD_ARG_TGT_ARCH=x86

$ cmake --list-presets
Available configure presets:

  "x86-windows-user-vs2019-Release"
```

### CMake Tools Log

```
[presetController] Invalid kit contents CMakePresets.json (d:\git\xx\CMakePresets.json):
[presetController]  >> /configurePresets/1: should NOT have additional properties
[presetController]  >> /configurePresets/2: should NOT have additional properties
```

### Platform and Versions

- **Operating System**: Windows
- **CMake Version**: 3.21.0-rc1
- **VSCode Version**: 1.57.1
- **CMake Tools Extension Version**: private build. see https://github.com/microsoft/vscode-cmake-tools/issues/1934
- **Compiler/Toolchain**: VS2019

bobbrow:
It looks like we are going to need to make the schema check optional when the presets file version is higher than what we support.  Thanks for reporting this!

KUGA2:
Hi @bobbrow, why validate it at all? Just run `cmake --list-presets` and display those presets. Leave the verfication to cmake executable.

bobbrow:
[...]

PR #2945, opened 2023-01-13
Added a setting to ignore unknown presets features from versions we don't support yet

bobbrow:
Addresses #1963


src/presetsController.ts
```diff
@@ -1016,16 +1022,25 @@ export class PresetsController {
...
+                        logFunc(` >> ${err.dataPath}: ${err.message}`);
+                    }
                 }
+            };
+            if (validationErrorsAreWarnings) {
+                showErrors(log.warning);
+                return presetsFile;
+            } else {
+                showErrors(log.error);
+                log.error(localize('unsupported.presets.disable', 'If desired, unknown properties and macros can be ignored by using the {0} setting.', "'cmake.allowUnsupportedPresetsVersions'"));
```
elahehrashedi, line 1041: Maybe we can write something easier to understand, I am ok with the current explanation though.
localize('disable.unsupported.presets', 'In order to ignore the unknown properties and macros, modify the {0} setting.', "'cmake.allowUnsupportedPresetsVersions'")

elahehrashedi (approved):

elahehrashedi:
@bobbrow I made a small change in the description.

bobbrow:
> @bobbrow I made a small change in the description.

That's fine

src/presetsController.ts
elahehrashedi: I did it. Thanks

PR #2946, opened 2023-01-13
Check if "CMakeLists.txt" exists, currently or after renaming

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2986
Works for creating and deleting a CMakeLists.txt file

Added a "rename" situation as well.


nasinlock (approved):

nasinlock (approved):

src/cmakeProject.ts
```diff
@@ -1522,6 +1522,7 @@ export class CMakeProject {
         if (isCmakeFile) {
             // CMakeLists.txt change event: its creation or deletion are relevant,
             // so update full/partial feature set view for this folder.
+            this.cmakeListsExists = await fs.exists(path.join(this._sourceDir, "CMakeLists.txt"));
```
bobbrow, line 1525: It doesn't seem like this should be necessary since whenever `this._sourceDir` is set, `this.cmakeListsExists` should also be set.

src/cmakeProject.ts
```diff
@@ -2583,8 +2594,8 @@ export class CMakeProject {
         return expandStrings(this.workspaceContext.config.additionalKits, opts);
     }
 
-    async sendFileTypeTelemetry(textDocument: vscode.TextDocument): Promise<void> {
-        const filePath =  util.platformNormalizePath(textDocument.uri.fsPath);
+    async sendFileTypeTelemetry(textDocument: vscode.Uri): Promise<void> {
```
bobbrow, line 2597: This change seems unnecessary. Please revert.

src/cmakeProject.ts
```diff
@@ -1493,8 +1493,8 @@ export class CMakeProject {
     }
 
     // Reconfigure if the saved file is a cmake file.
-    async doCMakeFileSaveReconfigure(textDocument: vscode.TextDocument) {
-        const filePath = util.platformNormalizePath(textDocument.uri.fsPath);
+    async doCMakeFileChangeReconfigure(textDocument: vscode.Uri) {
```
bobbrow, line 1496: Changing the type of the argument seems unncessary. Please revert to `vscode.TextDocument`.

src/extension.ts
```diff
@@ -1425,7 +1425,7 @@ export class ExtensionManager implements vscode.Disposable {
             return false;
         }
         for (const project of projects) {
-            if (await project.hasCMakeLists()) {
+            if (project.hasCMakeLists()) {
```
bobbrow, line 1428: If the `await` is gone, consider removing the `async` specifier from this function.

bobbrow: You could also use `return projects.some(p => p.hasCMakeLists());` instead of a for-loop here too.

src/projectController.ts
```diff
@@ -117,7 +117,7 @@ export class ProjectController implements vscode.Disposable {
     async getNumOfValidProjects(): Promise<number> {
```
bobbrow, line 117: It looks like this doesn't have to be `async` anymore.

src/projectController.ts
```diff
@@ -441,11 +442,24 @@ export class ProjectController implements vscode.Disposable {
     }
 
     private async doSaveTextDocument(textDocument: vscode.TextDocument): Promise<void> {
+        await this.doCMakeFileChangeReconfigure(textDocument.uri);
+    }
+
+    private async onDidRenameFiles(renamedFileEvt: vscode.FileRenameEvent): Promise<void> {
```
bobbrow, line 448: this doesn't seem related to `hasCMakeLists`?

src/util.ts
```diff
@@ -864,9 +864,9 @@ export async function scheduleAsyncTask<T>(task: () => Promise<T>): Promise<T> {
     });
 }
 
-export function isFileInsideFolder(openEditor: vscode.TextDocument, folderPath: string): boolean {
+export function isFileInsideFolder(openEditor: vscode.Uri, folderPath: string): boolean {
```
bobbrow, line 867: I don't think this needs to change. Please revert.

bobbrow (changes requested):

src/util.ts
bobbrow: Oh, I see that the rename event is the cause for this signature change.  I think that should have been split out from this PR.

elahehrashedi: The rename event only checks for existance of CMakeLists.txt. That's why I added it to this PR.

```ts
            if (filePath.endsWith("cmakelists.txt")) {
                await this.doCMakeFileChangeReconfigure(file.newUri);
            }
```

elahehrashedi: I changed the title of the PR

src/cmakeProject.ts
elahehrashedi: This is related to checking if a renamed file is a "CMakeLists.txt" file. in "onDidRenameFiles" we had access to URI.

src/cmakeProject.ts
elahehrashedi: that's right!

src/cmakeProject.ts
elahehrashedi: This is related to checking if a renamed file is a "CMakeLists.txt" file. in "onDidRenameFiles" we had access to URI.

elahehrashedi:
@bobbrow I applied some renaming.

bobbrow (approved):

Issue #2952, opened 2023-01-20
CMake Tools 1.13.40 fails to pick up CMakeUserPresets.json (only on first start after updating?)


floooh:
### Brief Issue Summary

After updating to 1.13.40, CMake Tools fails to pick up my CMakeUserPresets.json file, ```CMake: Build``` does nothing no build or debug target is selected, and selecting one is not possible, in short: nothing works :)

This happened to me both on macOS and Windows (so now the second time on Windows I'm writing this ticket).

...ok, weird: after setting the logging output to debug as requested below, and restarting, it suddenly worked.

AHA: starting another time it doesn't work again, really weird.

Then starting again, it picks up the configuration and now it seems like it's consistently working. Very confusing.

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.74.3",
  "cmtVersion": "1.13.40",
  "configurations": [
    {
      "folder": "e:\\projects\\sokol-samples",
      "cmakeVersion": "3.25.1",
      "configured": false,
      "generator": "Visual Studio 17",
      "usesPresets": true,
      "compilers": {}
    },
    {
      "folder": "e:\\projects\\fips-imgui-dock",
      "cmakeVersion": "unknown",
      "configured": false,
      "generator": "unknown",
      "usesPresets": false,
      "compilers": {}
    },
    {
      "folder": "e:\\projects\\sokol",
      "cmakeVersion": "3.25.1",
      "configured": false,
      "generator": "",
      "usesPresets": true,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [
      "file:///e%3A/projects/sokol/tests/compile/sokol_gfx.c",
      "file:///e%3A/projects/sokol/tests/compile/sokol_gfx.cc",
      "file:///e%3A/projects/sokol/sokol_gfx.h"
    ],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": false
    },
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": null
    },
    {
      "communicationMode": "automatic",
[...]

PR #2961, opened 2023-01-23
CMake Tools fails to initialize the Active Project

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2952

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -534,9 +534,26 @@ export class ExtensionManager implements vscode.Disposable {
 
     private async initActiveProject(): Promise<CMakeProject | undefined> {
         let folder: vscode.WorkspaceFolder | undefined;
-        if (vscode.workspace.workspaceFolders && vscode.window.activeTextEditor && this.workspaceConfig.autoSelectActiveFolder) {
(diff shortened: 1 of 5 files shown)


src/extension.ts
```diff
@@ -534,9 +534,26 @@ export class ExtensionManager implements vscode.Disposable {
 
     private async initActiveProject(): Promise<CMakeProject | undefined> {
         let folder: vscode.WorkspaceFolder | undefined;
-        if (vscode.workspace.workspaceFolders && vscode.window.activeTextEditor && this.workspaceConfig.autoSelectActiveFolder) {
-            folder = vscode.workspace.getWorkspaceFolder(vscode.window.activeTextEditor.document.uri);
-            await this.updateActiveProject(folder, vscode.window.activeTextEditor);
```
bobbrow, line 539: It feels like this change is overly complex for what we need to do.  I would think you just need to check if `folder` has a value. If not, then set `folder` to `vscode.workspace.workspaceFolderes[0]` and `editor` to `undefined` like you did below.

It looks like you are reimplementing `getWorkspaceFolder` with that for loop.

bobbrow (changes requested):

PR #2966, opened 2023-01-23
Make sure a project is set on startup

bobbrow:
Addresses #2952


elahehrashedi (approved):

elahehrashedi:
@bobbrow I tested it. Will change the changelog.

dbenson2000:
I just upgraded to 1.13.41 and the 'No configure preset selected' issue is still occurring as reported in #2952
I'm using "cmake.useCMakePresets": "never"
Running CMake: delete cache and reconfigure doesn't help.

Issue #2963, opened 2023-01-23
Project using kits and variants doesn't work in 1.13.40


xisui-MSFT:
### Brief Issue Summary

Project using kits and variants doesn't work in 1.13.40

1. Create a CMake project without presets, in which case it should fall back to kits and variants.
2. Open project in VS Code and notice that the tool bar doesn't switch to kits and variants view, and none of the configure / build related commands work.
3. Multi-root with kits and variants doesn't work as well, not sure if that's also caused by this though.

elahehrashedi:
This is the project that repro the issue:
azure-sphere-samples/Samples/HelloWorld/HelloWorld_HighLevelApp at main · Azure/azure-sphere-samples (github.com)

elahehrashedi:
We are going to add a rename or delete event listener for CMakeLists.txt . We can add the presets files to the event listener as well (for the cases where `"cmake.useCMakePresets"` is set to `"auto"`.

Look at `onDidRenameFiles `events from here : https://github.com/microsoft/vscode-cmake-tools/pull/2946/files

xisui-MSFT:
Ah yeah I was using an old version of that sample... So the actual steps:
1. Remove presets file from the project so it uses kits/variants.
2. Install Azure Sphere SDK: https://aka.ms/AzureSphereSDKDownload/Windows
3. Install Azure Sphere extension in VS Code
4. Open the project

I believe the reason this issue repros is that the Azure Sphere extension calls the following commands on activation, and causes a deadlock:
- cmake.setKitByName
- cmake.launchTargetPath
- cmake.hideLaunchCommand
- cmake.hideDebugCommand

> We can add the presets files to the event listener as well (for the cases where "cmake.useCMakePresets" is set to "auto".

I remember I wrote this? I'm probably wrong though since that's a long time ago.

elahehrashedi:
By following the steps, I am able to see the kit menu in the status bar (in both situations where Azure Sphere extension is installed or not).

however, after installing your extension, I see this and I am not able to configure correctly:
![image](https://user-images.githubusercontent.com/16143955/214390304-cfda62b8-6756-465a-a77d-b5b10abd78ea.png)

elahehrashedi:
I used another project with a CMakeLists.txt as below and I am able to configure without setting the kits:
```txt
CMAKE_MINIMUM_REQUIRED(VERSION 3.20)
PROJECT(CMakeSample C)

[...]

PR #2971, opened 2023-01-24
check binary directories only if there are multiple sources

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2963


xisui-MSFT (approved):

src/projectController.ts
```diff
@@ -382,7 +385,9 @@ export class ProjectController implements vscode.Disposable {
                 const driver = await project.getCMakeDriverInstance();
                 await driver?.refreshSettings();
             }
-            await ProjectController.checkBuildDirectories(projects);
+            const workspaceContext = DirectoryContext.createForDirectory(folder, new StateManager(this.extensionContext, folder));
```
bobbrow, line 388: I would think we could get the config without creating a whole DirectoryContext.  Can we?

src/projectController.ts
```diff
@@ -212,12 +212,15 @@ export class ProjectController implements vscode.Disposable {
...
             projects.push(await CMakeProject.create(workspaceContext, source, isMultiProjectFolder));
         }
-        await ProjectController.checkBuildDirectories(projects);
+        await ProjectController.checkBuildDirectories(projects, sourceDirectory);
         return projects;
     }
 
     private static duplicateMessageShown = false;
-    private static async checkBuildDirectories(projects: CMakeProject[]) {
+    private static async checkBuildDirectories(projects: CMakeProject[], sourceDirectory: string[]) {
```
bobbrow, line 220: The setting is called `sourceDirectory` and in the `config` class it should be called that, but in code it would be better to name local variables as `sourceDirectories` instead so that we don't forget it's an array.

bobbrow:
Please add a changelog entry for this too.  We can release an update tomorrow morning.

bobbrow (approved):

PR #2973, opened 2023-01-25
Automatically configure CMake project directories when the kit or the configuration preset is changed

maxmitti:
### This changes visible behavior when selecting a configuration preset.

- Automatically configure after a different configuration preset has been selected
- Automatically configure after a different kit has been selected

## The purpose of this change

Remove the need of manually configuring every time a said change is made.

## Other Notes/Information

This is a very simple implementation of my feature request.
A few people use it successfully for a while now, however I am not sure if this is the proper way of doing it.
We only used it for single root projects so far.


bobbrow:
Thanks for the PR.  We'll evaluate if this is the best place to insert the reconfigure.  I know that there is already a check deeper in the code that will clean the binaryDir if it is reused across multiple presets.  I just want to make sure we don't double-configure.

Another thing that we'll want for this is a setting that turns this feature on and off.  Some developers have projects that take a very long time to configure and they may not want to auto-configure when changing the preset.  (for similar reasons we have a `configureOnOpen` setting).  I haven't evaluated whether we should have a similar feature for the non-presets case when changing the active Kit, so maybe we should add a `cmake.automaticReconfigure` setting.  The description can explain that it reconfigures the project when a Kit or configurePreset changes.

Zingam:
IMO this should be optional. Enabled by default is OK but I'd like to be able to turn it off.

maxmitti:
I added a `cmake.automaticReconfigure` setting to control this feature.
I phrased it only for preset selection, but it can always be adjusted to also cover kit changes.

I also added a German translation of the description, although I don’t really know yet how i18n is handled here.

maxmitti:
> I know that there is already a check deeper in the code that will clean the binaryDir if it is reused across multiple presets. I just want to make sure we don't double-configure.

I found two places where this is checked:

- https://github.com/microsoft/vscode-cmake-tools/blob/main/src/drivers/cmakeDriver.ts#L489
- https://github.com/microsoft/vscode-cmake-tools/blob/main/src/drivers/cmakeDriver.ts#L580

As far as I can tell, only old configuration files are cleared, but no reconfiguration happens in either case.

I actually tried to test the shared binaryDir with different presets and saw, that there has to be a generator change between the presets (https://github.com/microsoft/vscode-cmake-tools/blob/main/src/preset.ts#L1631).
I created two presets with same binaryDir and different generators, but to my surprise not even the old configuration gets cleared. When I configure the project afterwards, CMake gives an error that CMakeCache has a different generator specified.

i18n/deu/package.i18n.json
```diff
@@ -154,6 +154,7 @@
 	"cmake-tools.configuration.cmake.allowUnsupportedPresetsVersions.description": "Aktiviert die Verwendung von Voreinstellungsdateien, die Features aus den Versionen verwenden, die von der Erweiterung \"CMake Tools\" derzeit nicht unterstützt werden. Unbekannte Eigenschaften und Makros werden ignoriert.",
 	"cmake-tools.configuration.cmake.ignoreCMakeListsMissing.description": "`true` gibt an, dass der Benutzer von der Erweiterung nicht aufgefordert wird, eine CMakeLists.txt-Datei für die Konfiguration auszuwählen, wenn eine Datei im Arbeitsbereich, aber nicht im Stammordner gefunden wird.",
 	"cmake-tools.configuration.cmake.launchBehavior.description": "Steuert, was mit dem Startterminal beim Starten eines Ziels geschieht.",
+	"cmake-tools.configuration.cmake.automaticReconfigure.description": "Gibt an, ob nach Auswahl einer Konfigurationsvoreinstellung automatisch neu konfiguriert werden soll.",
```
elahehrashedi, line 157: please remove this translation. Translations are handled separately.

package.json
```diff
@@ -2007,7 +2007,13 @@
           "default": true,
           "description": "%cmake-tools.configuration.cmake.loadCompileCommands.description%",
           "scope": "resource"
-        }
+        },
+		"cmake.automaticReconfigure": {
```
elahehrashedi, line 2011: @bobbrow we have "configureOnOpen" and "configureOnEdit", can we name this "configureOnSwitch" or something similar?

package.nls.json
```diff
@@ -149,6 +149,7 @@
     "cmake-tools.configuration.cmake.allowUnsupportedPresetsVersions.description": "Enables the use of presets files that are using features from the versions that CMake Tools extension doesn't currently support. Unknown properties and macros will be ignored.",
     "cmake-tools.configuration.cmake.ignoreCMakeListsMissing.description": { "message": "If `true`, the extension will not ask the user to select a CMakeLists.txt file for configuration when one is found in the workspace but not in the root folder.", "comment": [ "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered." ] },
     "cmake-tools.configuration.cmake.launchBehavior.description": "Controls what happens with the launch terminal when you launch a target.",
+    "cmake-tools.configuration.cmake.automaticReconfigure.description": "Controls whether to automatically reconfigure on selection of a different configuration preset.",
```
elahehrashedi, line 152: "Automatically configure CMake project directories when the kit or the configuration preset is changed"

src/extension.ts
```diff
@@ -1531,6 +1531,10 @@ export class ExtensionManager implements vscode.Disposable {
         const testPreset = project.testPreset;
         this.statusBar.setTestPresetName(testPreset?.displayName || testPreset?.name || '');
 
+        if (this.workspaceConfig.automaticReconfigure && presetSelected) {
```
elahehrashedi, line 1534: I believe extension.ts is not the best place to apply the re-configure. A better option is in cmakeProject.ts where we change the config preset.

package.json
bobbrow: I couldn't think of a `configureOn` name that would be able to describe both presets and Kits changes, so I went a different way with the suggestion that was more generic.

maxmitti: So, at the moment we have two name proposals.
I will adjust the name once an agreement has been found.

src/extension.ts
maxmitti: Do you mean [here](https://github.com/maxmitti/vscode-cmake-tools/blob/auto_configure_after_preset/src/drivers/cmakeDriver.ts#L480-L493) or [here](https://github.com/maxmitti/vscode-cmake-tools/blob/auto_configure_after_preset/src/drivers/cmakeDriver.ts#L495-L510)?

bobbrow: I would add that we only want to reconfigure if the preset changed.  It's possible to select the same configurePreset and we would want to do nothing in that case.

We should also reconfigure if the Kit is changed.  You can take a look at the code in `selectKit` to see where we might want to do this.

package.json
```diff
@@ -2007,7 +2007,13 @@
...
+			"type": "boolean",
+			"default": true,
```
elahehrashedi, line 2013: please set the default to false

package.nls.json
elahehrashedi: We should also reconfigure if the Kit is changed.

src/extension.ts
elahehrashedi: it better be done in `selectConfigurePreset`, where the user is choosing a new preset. 
Also make sure we don't reconfigure if the user is choosing the same preset.

maxmitti: Hopefully I chose the right selectConfigurePreset :)

package.nls.json
maxmitti: I implemented it, but I don’t currently have a setup to test it.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.14
 Improvements:
+- Automatically configure after a configuration preset has been selected. [#2765](https://github.com/microsoft/vscode-cmake-tools/issues/2958) [@maxmitti](https://github.com/maxmitti)
```
elahehrashedi, line 5: Automatically configure CMake project directories when the kit or the configuration preset is changed

package.json
maxmitti: I don’t have much insights, why do you prefer to set it to false by default?
I thought setting it to true may benefit more users than it would harm.
Also, for those that don’t want the behavior, it is more likely to actively look in the settings for a possible solution, while other users may not actively check each available configuration option and thus be unaware.

package.json
```diff
@@ -2007,7 +2007,13 @@
           "default": true,
           "description": "%cmake-tools.configuration.cmake.loadCompileCommands.description%",
           "scope": "resource"
-        }
+        },
+		"cmake.automaticReconfigure": {
+			"type": "boolean",
+			"default": false,
+			"description": "%cmake-tools.configuration.cmake.automaticReconfigure.description%",
```
bobbrow, line 2014: Please fix the formatting of this file.  It looks like you inserted TABs.

src/kitsController.ts
```diff
@@ -5,7 +5,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import * as nls from 'vscode-nls';
 
-import CMakeProject from '@cmt/cmakeProject';
+import CMakeProject, { ConfigureTrigger, ConfigureType } from '@cmt/cmakeProject';
```
bobbrow, line 8: Can you fix this to: 
```ts
import { CMakeProject, ConfigureTrigger, ConfigureType } from '@cmt/cmakeProject';
```

package.json
bobbrow: I've been thinking about this more, and I think it would be ok for the default to be `true`.  We already reconfigure automatically when changing the variant from Debug -> Release.  Visual Studio also reconfigures automatically.

package.json
maxmitti: @bobbrow, can you please confirm that your resolving of the conversation means that the current name is ok?

bobbrow: yes, it is ok

package.json
maxmitti: So, I will change it back to true :)

bobbrow: I just did and push a few additional changes

bobbrow:
@maxmitti, thanks again for the PR.  I re-enabled the feature by default and added some logic to avoid a configuration if the Kit did not actually change.  This also fixed a bug with the Kit logic because the active kit was already changed by the time you did the check so the configure wasn't happening.  I also added logic to Variant changes.  It turns out we were already doing automatic configure for those and now that behavior will respond to the setting as well.

bobbrow (approved):

maxmitti:
@bobbrow thanks for your support. I’m glad we found a good solution.

Issue #2972, opened 2023-01-25
cmake tools removes path separators when running target


pedroVRVIU:
### Brief Issue Summary

Yesterday, everything was running fine in the land of VS Code.
Today: path separators are removed when running the target.

![image](https://user-images.githubusercontent.com/116630723/214460006-d5aa4884-b90d-4e12-a3e7-ac5f720e11bc.png)

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.74.3",
  "cmtVersion": "1.13.41",
  "configurations": [
    {
      "folder": "c:\\dev\\src\\vrviu\\JapanTeam\\kksk_congestion_control",
      "cmakeVersion": "3.24.3",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {
        "CXX": "C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "RelWithDebInfo",
    "buildTypesSeen": [
      "RelWithDebInfo"
    ],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 5,
    "executablesCount": 3,
    "librariesCount": 1,
    "targets": [
      {
        "name": "bwa",
        "type": "STATIC_LIBRARY"
      },
      {
        "name": "bwa_fuzz_tests",
        "type": "UTILITY"
      },
      {
        "name": "bwa_main_tests",
        "type": "EXECUTABLE"
      },
      {
        "name": "bwa_perf_tests",
        "type": "EXECUTABLE"
      },
      {
        "name": "bwa_unit_tests",
        "type": "EXECUTABLE"
      }
    ]
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

pedroVRVIU:
[...]

PR #2974, opened 2023-01-25
Update quote function to fix path separator regression

nasinlock:
## This change addresses item #2972

- Update quote function to add check for less than and greater than symbols in the existing character class check.
- Remove the white space check in the quote function.

## Other Notes/Information
N/A

### DIFF
--- a/src/shlex.ts
+++ b/src/shlex.ts
@@ -77,11 +77,7 @@ export function quote(str: string, opt?: ShlexOptions): string {
     if (str === '') {
         return '""';
     }
-    if (!/[\s]/g.test(str)) {
-        // Don't quote if the string doesn't have whitespace
-        return str;
-    }
-    if (/[^\w@%\-+=:,./|]/.test(str)) {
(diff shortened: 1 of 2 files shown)


src/shlex.ts
```diff
@@ -77,11 +77,7 @@ export function quote(str: string, opt?: ShlexOptions): string {
     if (str === '') {
         return '""';
     }
-    if (!/[\s]/g.test(str)) {
-        // Don't quote if the string doesn't have whitespace
-        return str;
-    }
-    if (/[^\w@%\-+=:,./|]/.test(str)) {
+    if (/[^\w@%\-+=:,./|><]/.test(str)) {
```
bobbrow, line 80: We can take this change since it shouldn't regress the other fix and will put us back in a state where it should work for the new bug report, but we'll need to update the changelog.  Can you add a line in the changelog for this?

I asked the customer for more information about their configuration so that hopefully we can reproduce the actual problem.

bobbrow (approved):

bobbrow (approved):

Issue #2976, opened 2023-01-26
Hidden CMake presets picked up at first configuration - project outline empty


vdilecce:
### Brief Issue Summary

VSCode: 1.74.3
CMake Tools: 1.13.42
CPPTools: 1.13.9

At the first configuration of a project, when the CMakePresets.json contains hidden presets, the extension picks the hidden preset as default (status bar), leaving the project outline in the Activity Bar empty (verified with build presets).

Selecting a proper (non-hidden) preset, configuring again, and restarting VSCode makes the targets appear in the project outline.

Verified on Windows and Linux.

### CMake Tools Diagnostics

```shell
At first run:

{
  "os": "win32",
  "vscodeVersion": "1.74.3",
  "cmtVersion": "1.13.42",
  "configurations": [
    {
      "folder": "XXXXXXXXXXXXXXXXX",
      "cmakeVersion": "3.25.0",
      "configured": true,
      "generator": "Visual Studio 17 2022",
      "usesPresets": true,
      "compilers": {
        "CXX": "C:/Program Files/Microsoft Visual Studio/2022/Professional/VC/Tools/MSVC/14.34.31933/bin/Hostx64/x64/cl.exe"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": false
    }
  ]
}


After configuring again with proper build preset and restarting VSCode:

{
  "os": "win32",
  "vscodeVersion": "1.74.3",
  "cmtVersion": "1.13.42",
  "configurations": [
    {
      "folder": "XXXXXXXXXXXXXXXXXXXXXXXXX",
      "cmakeVersion": "3.25.0",
      "configured": true,
      "generator": "Visual Studio 17 2022",
      "usesPresets": true,
      "compilers": {
        "CXX": "C:/Program Files/Microsoft Visual Studio/2022/Professional/VC/Tools/MSVC/14.34.31933/bin/Hostx64/x64/cl.exe"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
[...]

PR #2978, opened 2023-01-26
Validate config preset in initialization

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2976

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@ Improvements:
 
 Bug Fixes:
 - Compatibility between test and build presets was not enforced. [#2904](https://github.com/microsoft/vscode-cmake-tools/issues/2904)
-- Update quote function to fix path separator regression [#2974](https://github.com/microsoft/vscode-cmake-tools/pull/2974)
 - Validate config preset in initialization. [#2976](https://github.com/microsoft/vscode-cmake-tools/issues/2976)
 
 ## 1.13.43
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Improvements:
 Bug Fixes:
 - Compatibility between test and build presets was not enforced. [#2904](https://github.com/microsoft/vscode-cmake-tools/issues/2904)
 - Update quote function to fix path separator regression [#2974](https://github.com/microsoft/vscode-cmake-tools/pull/2974)
+- Validate config preset in initialization. [#2976](https://github.com/microsoft/vscode-cmake-tools/issues/2976)
 
 ## 1.13.42
 Bug Fixes:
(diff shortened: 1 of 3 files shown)


src/cmakeProject.ts
```diff
@@ -987,9 +987,14 @@ export class CMakeProject {
...
         if (this.useCMakePresets) {
-            const configurePreset = this.workspaceContext.state.configurePresetName;
-            if (configurePreset) {
-                await this.presetsController.setConfigurePreset(configurePreset);
+            const latestConfigPresetName = this.workspaceContext.state.configurePresetName;
+            if (latestConfigPresetName) {
+                // Check if the latest configurePresetName from the previous session is still valid.
+                const presets = await this.presetsController.getAllConfigPresets();
+                const latestConfigPreset: preset.ConfigurePreset | undefined = presets.find(preset => preset.name === latestConfigPresetName);
+                if (latestConfigPreset && !latestConfigPreset.hidden) {
```
nasinlock, line 995: It seems like if the latestConfigPresetName isn't valid anymore, we are left without a preset being set. Should we have a fall back behavior for the case where latestConfigPreset is 'undefined'?

elahehrashedi: This is the initialization step when we load the project, so if we don't initialize it, it will stay undefined and the user will have to pick a new preset.

nasinlock: Okay. So the user will be prompted to set a new preset rather than have the previous preset set as being enabled?

nasinlock (approved):

src/cmakeProject.ts
elahehrashedi: correct.

bobbrow (approved):

nasinlock (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Improvements:
 
 Bug Fixes:
 - Compatibility between test and build presets was not enforced. [#2904](https://github.com/microsoft/vscode-cmake-tools/issues/2904)
+- Validate config preset in initialization. [#2976](https://github.com/microsoft/vscode-cmake-tools/issues/2976)
```
bobbrow, line 9: It's not just config presets anymore.

src/presetsController.ts
```diff
@@ -607,11 +607,14 @@ export class PresetsController {
         return chosenPresets?.preset;
     }
 
-    async selectConfigurePreset(): Promise<boolean> {
+    async getAllConfigPresets(): Promise<preset.ConfigurePreset[]> {
```
bobbrow, line 610: I don't think we abbreviate `configure` in any of the function names.

bobbrow (changes requested):

bobbrow (approved):

Issue #2980, opened 2023-01-27
Not check targets and includes after configuration. (Since plugin version v1.13.40)


traversebitree:
### Brief Issue Summary

Since update to `v1.13.40`, "vscode-cpptools" not check whether target links and includes path exists.
For example
**Step 1 (Prepare)**:
![image](https://user-images.githubusercontent.com/30373236/215115983-7fdea988-24b0-486a-9aae-d7ec00b922d8.png)
![image](https://user-images.githubusercontent.com/30373236/215116316-2ae2f919-8757-4d33-8ea4-10ccef101d18.png)

**Step 2 (Bug appear)**:
![image](https://user-images.githubusercontent.com/30373236/215117055-5486f14c-f0a7-4481-9149-580ef2f550eb.png)
![image](https://user-images.githubusercontent.com/30373236/215117573-e802ef8c-cbef-4683-9ce7-f7a9bc5935aa.png)

**Step 3 (After build, it becomes normal)**:
![image](https://user-images.githubusercontent.com/30373236/215118000-78f34f72-fb59-44c1-8c4d-c4ce137271ff.png)
![image](https://user-images.githubusercontent.com/30373236/215118156-54bdbaee-475c-4693-a825-1d2049be610c.png)

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.74.3",
  "cmtVersion": "1.13.43",
  "configurations": [
    {
      "folder": "d:\\Workspace\\CPP\\CPPT001",
      "cmakeVersion": "3.25.2",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {
        "C": "gcc",
        "CXX": "g++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "Release",
    "buildTypesSeen": [
      "Release"
    ],
    "requests": [
      "file:///d%3A/Workspace/CPP/CPPT001/main.cpp"
    ],
    "responses": [
      {
        "uri": "file:///d%3A/Workspace/CPP/CPPT001/main.cpp",
        "configuration": {
          "includePath": [
            "d:/workspace/cpp/cppt001/extern/fmt_x64-mingw-static/include"
          ],
          "defines": [],
          "compilerPath": "c:/dev/tdm-gcc-64/bin/g++.exe",
          "compilerArgs": [],
          "compilerFragments": [
[...]

PR #2985, opened 2023-01-30
Fix problems with updating the code model

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2980
And possibly the reason behind having an empty project outline in the beginning: [project outline empty](https://github.com/microsoft/vscode-cmake-tools/issues/2976)

### DIFF
--- a/src/projectController.ts
+++ b/src/projectController.ts
@@ -270,7 +270,6 @@ export class ProjectController implements vscode.Disposable {
         let projects: CMakeProject[] | undefined = this.getProjectsForWorkspaceFolder(folder);
         if (projects) {
             rollbar.error(localize('same.folder.loaded.twice', 'The same workspace folder was loaded twice'), { wsUri: folder.uri.toString() });
-            return projects;
         } else {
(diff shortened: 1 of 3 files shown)


src/projectController.ts
```diff
@@ -197,7 +197,8 @@ export class ProjectController implements vscode.Disposable {
         this.folderToProjectsMap.clear();
         if (vscode.workspace.workspaceFolders) {
             for (const folder of vscode.workspace.workspaceFolders) {
-                await this.addFolder(folder);
+                const projects: CMakeProject[] = await this.addFolder(folder);
+                this.afterAddFolderEmitter.fire({ folder: folder, projects:  projects});
```
bobbrow, line 201: Should this always be fired after `this.addFolder`?  If so, then `this.addFolder` should do it so that we never make this mistake again.

nasinlock: I agree

nasinlock (approved):

bobbrow (approved):

PR #2987, opened 2023-01-31
Build selected build-target when starting to debug

Maddimax:

- When launching, not only build the selected launch target, but also the selected build target if the two differ.

## The purpose of this change

For some projects, you always want to build ALL. This allows to set ALL as the build target, but still launch the selected launch target.

### DIFF
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -2162,7 +2162,13 @@ export class CMakeProject {
 
         const buildOnLaunch = this.workspaceContext.config.buildBeforeRun;
         if (buildOnLaunch || isReconfigurationNeeded) {
-            const buildResult = await this.build([chosen.name]);
+            const buildTargets = await this.getDefaultBuildTargets() || [];
+            const allTargetName = await this.allTargetName;
+            if (!buildTargets.includes(allTargetName) && !buildTargets.includes(chosen.name)) {
+                buildTargets.push(chosen.name);
+            }
+
+            const buildResult = await this.build(buildTargets);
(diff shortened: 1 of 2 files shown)


Maddimax:

elahehrashedi:
We don't want to just add the build target to the list. A better approach would be to inform the user that the launch target doesn't match the build targets.

src/cmakeProject.ts
```diff
@@ -2162,7 +2162,10 @@ export class CMakeProject {
 
         const buildOnLaunch = this.workspaceContext.config.buildBeforeRun;
         if (buildOnLaunch || isReconfigurationNeeded) {
-            const buildResult = await this.build([chosen.name]);
+            const defaultBuildTargets = await this.getDefaultBuildTargets();
+            const allTargetsToBuild = new Set(defaultBuildTargets);
+            allTargetsToBuild.add(chosen.name);
```
elahehrashedi, line 2167: Check if the build target is ALL, we shouldn't add the launch targets to ALL.

bobbrow: No, you should check if it's `this.allTargetName`.  If that's the case, then you don't need to add the debug target to the list.

Maddimax: Done.

elahehrashedi: use `this.allTargetName` instead of `all`

Maddimax: Done.

elahehrashedi (approved):

PR #2991, opened 2023-02-01
Fix an issue with setKitByName ignoring the folder argument

bobbrow:
Addresses an issue with the `cmake.setKitByName` command reported by an extension author who used that command.


elahehrashedi:
I updated the changelog.

elahehrashedi (approved):

Issue #1932, opened 2021-06-15
Still getting build diagnostics with cmake.parseBuildDiagnostics = false


alexdesaint:
I have `"cmake.parseBuildDiagnostics": false` but still get error reported in the code as `GCC`.

![image](https://user-images.githubusercontent.com/33799602/122021992-7aa84980-cdc6-11eb-8422-3e00bbc0443e.png)

or in the error section:
![image](https://user-images.githubusercontent.com/33799602/122022678-1b970480-cdc7-11eb-80c4-b163225c9625.png)

- **Operating System**: Windows
- **CMake Version**: 3.20
- **VSCode Version**: 1.57
- **CMake Tools Extension Version**: 1.7.3
- **Compiler/Toolchain**: clang12

bobbrow:
Thanks for opening this issue.  This was mentioned in my comment in #1266, but a separate issue was never opened, so we lost track of it.

bobbrow:
Workaround: use `"cmake.enabledOutputParsers"` setting to hide the diagnostics you don't want to see.

PR #2992, opened 2023-02-02
implement cmake.parseBuildDiagnostics

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/1932

### DIFF
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -1651,7 +1651,7 @@ export class CMakeProject {
                             buildLogger.info(localize('build.finished.with.code', 'Build finished with exit code {0}', rc));
                         }
                         const fileDiags: FileDiagnostic[] | undefined = drv!.config.parseBuildDiagnostics ? consumer!.compileConsumer.resolveDiagnostics(drv!.binaryDir) : [];
-                        if (fileDiags && fileDiags.length >= 1) {
+                        if (fileDiags) {
                             populateCollection(collections.build, fileDiags);
                         }
                         await this.refreshCompileDatabase(drv!.expansionOptions);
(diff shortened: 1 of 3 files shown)


nasinlock (approved):

src/cmakeProject.ts
```diff
@@ -1650,7 +1650,7 @@ export class CMakeProject {
                         } else {
                             buildLogger.info(localize('build.finished.with.code', 'Build finished with exit code {0}', rc));
                         }
-                        const fileDiags = consumer!.compileConsumer.resolveDiagnostics(drv!.binaryDir);
+                        const fileDiags: FileDiagnostic[] | undefined = drv!.config.parseBuildDiagnostics ? consumer!.compileConsumer.resolveDiagnostics(drv!.binaryDir) : undefined;
```
bobbrow, line 1653: Can you set it to `[]` instead of undefined?  I think we should clear the list of diagnostics if the user changes the setting after getting some diagnostics listed in the Problems.

src/cmakeProject.ts
```diff
@@ -1650,8 +1650,8 @@ export class CMakeProject {
                         } else {
                             buildLogger.info(localize('build.finished.with.code', 'Build finished with exit code {0}', rc));
                         }
-                        const fileDiags = consumer!.compileConsumer.resolveDiagnostics(drv!.binaryDir);
-                        if (fileDiags) {
+                        const fileDiags: FileDiagnostic[] | undefined = drv!.config.parseBuildDiagnostics ? consumer!.compileConsumer.resolveDiagnostics(drv!.binaryDir) : [];
+                        if (fileDiags && fileDiags.length >= 1) {
```
bobbrow, line 1654: If you require the length to be greater than one, doesn't that block us from clearing the problems?  My point of setting it to `[]` was so that `populateCollection` would be called.

bobbrow (approved):

PR #2993, opened 2023-02-02
"Side Bar" feature implementation

elahehrashedi:
update: Hide buttons 
Animationhide

update: the stop button:

Animation

feature request: https://github.com/microsoft/vscode-cmake-tools/issues/1582


src/sideBar.ts
```diff
@@ -9,193 +18,238 @@ nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFo
...
+        return [];
+    }
+
+    async initialize(nodeType: NodeType): Promise<void> {
+        let icon: string;
+        if (!Node.cmakeProject) {
+            return;
+        }
+        switch (nodeType) {
+            case NodeType.Configure:
```
bobbrow, line 106: For each of the cases in this `switch` the code should go in the constructor of the matching class.

src/sideBar.ts
```diff
@@ -0,0 +1,737 @@
...
+    async initialize(): Promise<void> {
+        if (!BaseNode.cmakeProject) {
+            return;
+        }
+        this.command = {
+            title: localize('Configure', 'Configure'),
+            command: 'cmake.sideBar.configure',
+            arguments: [this, BaseNode.cmakeProject.workspaceFolder]
+        };
+        this.tooltip = "Configure";
```
bobbrow, line 261: Don't forget to localize tooltips. I think in most cases you can grab the command title which is already localized.

res/dark/launch-icon.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
bobbrow, line 1: This looks like it should be the debug-icon (to match cpptools).
Launch should be just the triangle with no bug on it.

src/sideBar.ts
```diff
@@ -0,0 +1,737 @@
...
+                this.command = {
+                    title: localize('change.preset', 'Change Preset'),
+                    command: 'cmake.sideBar.selectConfigurePreset',
+                    arguments: [this]
+                };
+                this.tooltip = 'Change Configure Preset';
+                this.contextValue = 'configPreset';
+                break;
+            case PresetType.Build:
+                this.label = BaseNode.cmakeProject.buildPreset?.name || noBuildPresetSelected;
```
bobbrow, line 522: We'll want to have a friendly name for the "default" build/test presets. In the GIF, it looks like the internal name for the test preset is showing.

bobbrow (dismissed):

src/sideBar.ts
```diff
@@ -0,0 +1,737 @@
...
+        }
+    }
+
+    clear(): Promise<any> {
+        BaseNode.updateActiveProject(undefined);
+        return this.refresh();
+    }
+
+    dispose(): void {
+        Disposable.from(...this.disposables).dispose();
```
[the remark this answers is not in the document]
elahehrashedi: This is how most getChildren are working, the first time that it's being called on the root, we create the nodes and then getChildren will be called on every node again.
I can create the nodes in the constructor and save them and when getChildren is called from the root, I return those nodes, but then I saved some extra nodes.

src/sideBar.ts
elahehrashedi: I changed it to [default]

elahehrashedi:
@bobbrow there are two updated gifs

package.json
```diff
@@ -396,6 +505,16 @@
         "when": "cmake:enableFullFeatureSet",
         "title": "%cmake-tools.command.cmake.outline.debugTarget.title%"
       },
+      {
+        "command": "cmake.projectStatus.debugTarget",
+        "when": "cmake:enableFullFeatureSet",
+        "title": "%cmake-tools.command.cmake.debugTarget.title%",
+        "category": "CMake",
+        "icon": {
```
bobbrow, line 513: Use the `$(debug-alt)` icon instead of the svg (we can delete the svg too)

package.json
```diff
@@ -408,11 +527,27 @@
         "when": "cmake:enableFullFeatureSet",
         "category": "CMake"
       },
+      {
+        "command": "cmake.projectStatus.setDebugTarget",
+        "when": "cmake:enableFullFeatureSet",
+        "title": "%cmake-tools.command.cmake.selectLaunchTarget.title%",
+        "icon": {
```
bobbrow, line 534: Use the `$(edit)` icon everywhere instead of the svg and delete the svg.

package.json
```diff
@@ -368,6 +467,16 @@
         "when": "cmake:enableFullFeatureSet",
         "category": "CMake"
       },
+      {
+        "command": "cmake.projectStatus.ctest",
+        "title": "%cmake-tools.command.cmake.ctest.title%",
+        "when": "cmake:enableFullFeatureSet",
+        "icon": {
```
bobbrow, line 474: We can use `$(beaker)` instead of the svg and you can delete the svg.

package.json
```diff
@@ -430,12 +565,31 @@
...
         "title": "%cmake-tools.command.cmake.stop.title%",
         "when": "cmake:enableFullFeatureSet",
         "category": "CMake"
       },
+      {
+        "command": "cmake.projectStatus.stop",
+        "title": "%cmake-tools.command.cmake.stop.title%",
+        "when": "cmake:enableFullFeatureSet",
+        "category": "CMake",
+        "icon": {
```
bobbrow, line 588: It looks like you can use `$(x)` for this one.  (We may be able to delete most or all of the svg's in our repo since you discovered this icon syntax is supported).  The reference page for all icons is here: https://code.visualstudio.com/api/references/icons-in-labels

src/sideBar.ts
```diff
@@ -0,0 +1,852 @@
...
+    private buildPreset?: BuildPreset;
+
+    async initialize(): Promise<void> {
+        if (!treeDataProvider.cmakeProject) {
+            return;
+        }
+        this.label = localize('Build', 'Build');
+        this.command = {
+            title: this.label,
+            command: 'cmake.projectStatus.build',
```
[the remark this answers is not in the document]
elahehrashedi: Let me test the presets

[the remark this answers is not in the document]
elahehrashedi: are you in multi-project mode? If yes, this should solve your issue: 
https://github.com/microsoft/vscode-cmake-tools/pull/3051
the targets are saved for only one project in one folder, not for all, which caused this issue for me.

elahehrashedi:
A reminder:
the file "sideBar.ts" needs to be renamed to "projectStatus.ts" at the end.

src/sideBar.ts
```diff
@@ -0,0 +1,852 @@
...
+        this.collapsibleState = vscode.TreeItemCollapsibleState.Expanded;
+        this.contextValue = "configure";
+        await this.InitializeChildren();
+    }
+
+    async InitializeChildren(): Promise<void> {
+        if (!treeDataProvider.cmakeProject) {
+            return;
+        }
+        if (!treeDataProvider.cmakeProject.useCMakePresets) {
```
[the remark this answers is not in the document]
bobbrow: I tried changing the settings and the UI appeared to update as expected.

src/proc.ts
```diff
@@ -142,7 +142,7 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
             // and doesn't have to be 100% correct.
             localize('executing.command', 'Executing command: {0}', cmdstr));
         if (options.environment) {
-            log.debug(localize('execution.environment', '  with environment: {0}', JSON.stringify(final_env)));
+            //log.debug(localize('execution.environment', '  with environment: {0}', JSON.stringify(final_env)));
```
bobbrow, line 145: Please restore this line

bobbrow:
I did some basic testing and things appear to be working with the exception of the issue that may be covered by #3051.  @benmcmorran @gcampbell-msft @snehara99, I believe this PR is ready for a final review and testing.

I didn't cover the "hide" button scenarios.  We expose a command that allows us to hide some of the buttons and I believe we continue to honor the `cmake.statusbar` setting (though we should probably replace/remove that in a followup PR).  To test the commands, you can hack them into package.json temporarily to make them appear in the Command Palette.

Add something like this to the "commands" section:
```
      {
        "command": "cmake.hideDebugCommand",
        "when": "cmake:enableFullFeatureSet",
        "title": "CMake: Hide Debug Command"
      },

```

src/proc.ts
bobbrow: nevermind. I did it.

elahehrashedi:
This is the setting in settings.json that hides the build, debug, and run button:

```json
    "cmake.statusbar.advanced": {
        "build": {"visibility": "hidden"},
        "debug": {"visibility": "hidden"},
        "launch": {"visibility": "hidden"}
    }
```

The setting is borrowed from status bar

Issue #2996, opened 2023-02-03
Update cmake.mingwSearchDirs documentation


bobbrow:
@bobbrow Without the default value, the description isn't sufficient to understand the format of the setting, e.g. setting
```
"cmake.mingwSearchDirs": [
        "C:\\msys64\\mingw64\\bin"
    ]
```
doesn't work. The "\\bin" has to be removed. It may be helpful to others if the description included an example or something.

_Originally posted by @sean-mcmanus in https://github.com/microsoft/vscode-cmake-tools/issues/2942#issuecomment-1414719064_

bobbrow:
Sure, we could do that or fix it so that it works with `bin` in the path.

sean-mcmanus:
@bobbrow Either seems fine to me.

PR #2998, opened 2023-02-03
update docs on cmake.mingwSearchDirs

elahehrashedi:
bugfix: https://github.com/microsoft/vscode-cmake-tools/issues/2996


bobbrow:
I think it's better to just not add 'bin' to the path.  Fix it here:

https://github.com/microsoft/vscode-cmake-tools/blob/f666577a9876272e2be6b358b7309d49e14c4835/src/kit.ts#L1212

Otherwise we also have to explain it in package.nls.json.  It's easier to just not append 'bin'.

bobbrow (changes requested):
Let's fix it to work instead of updating the description.

elahehrashedi:
> Let's fix it to work instead of updating the description.

The description needed to be updated to show that we removed the default value https://github.com/microsoft/vscode-cmake-tools/pull/2942

bobbrow (approved):

Issue #2936, opened 2023-01-10
Build task searching mechanism does not eliminate template tasks in search process and adds duplicate of original build template task to quick pick list 


piomis:
### Brief Issue Summary

This issue is continuation of issue https://github.com/microsoft/vscode-cmake-tools/issues/2778 and pull-request https://github.com/microsoft/vscode-cmake-tools/pull/2780
Newly merged implementation intention was to search among all tasks defined in tasks.json and at initial stage eliminate ones which are "template" tasks.

However currently these tasks are still included in search mechanism.
Additionally at the end all build tasks (ones from tasks.json + template tasks) should be displayed in QuickPick - it is working fine but additional "template" task is added to list. 

For details see https://github.com/microsoft/vscode-cmake-tools/pull/2780#discussion_r1066052934

elahehrashedi:
I will close this issue and we will follow the improvements from here: https://github.com/microsoft/vscode-cmake-tools/issues/2935

piomis:
@elahehrashedi 
Would it be sensible to re-open this issue? #2935 was closed with other two fixes (default task and cancellation) but not this particular one.

PR #3009, opened 2023-02-07
Expand  targets names while searching for matching build task

### LINKED ISSUE #2970: The tasks' targets are not resolved when "cmake.buildTask" is used
### Brief Issue Summary

When searching for matching tasks, we try to compare the targets.
https://github.com/microsoft/vscode-cmake-tools/blob/b6668bdb18d290216fe6be703eb1ba67fceaafa9/src/cmakeTaskProvider.ts#L223-L232

However, we don't resolve the targets such as `"${command:cmake.buildTargetName}"` before comparison.

piomis:
## This change fixes #2970


piomis:
@elahehrashedi  as requested I tried implementing expansion capability while searching for matching task.

src/cmakeTaskProvider.ts
```diff
@@ -189,15 +190,26 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
+        const project: CMakeProject | undefined = getActiveProject();
+        const cmakeDriver: CMakeDriver | undefined = (await project?.getCMakeDriverInstance()) || undefined;
+        const tasks: (CMakeTask | undefined)[] = await Promise.all(allTasks.map(async (task: any) => {
             if (!task.definition.label || !task.group || (task.group && task.group.id !== vscode.TaskGroup.Build.id)) {
                 return undefined;
             }
+
+            let taskTargets: string[];
+            if (cmakeDriver) {
+                taskTargets = await expand.expandStrings(task.definition.targets,  cmakeDriver.expansionOptions);
```
elahehrashedi, line 203: the function findBuildTask is being called from the CMkae Driver, it's better if we pass the expansion options as the input to findBuildTask, instead of fetching the active project and getting the driver instance.

elahehrashedi:
@piomis thank you for your contribution. I tested the branch and it seems to be working. Please apply the changes to simplify the code.

src/cmakeTaskProvider.ts
piomis: Done.
FYI - i defined `expansionOptions` as optional parameter (and code checks if it was passed before calling expandString) - just in case when expansion is not required at all.

src/cmakeTaskProvider.ts
```diff
@@ -186,18 +187,30 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
...
+
+        const tasks: (CMakeTask | undefined)[] = await Promise.all(allTasks.map(async (task: any) => {
             if (!task.definition.label || !task.group || (task.group && task.group.id !== vscode.TaskGroup.Build.id)) {
                 return undefined;
             }
+
+            let taskTargets: string[];
+            if (expansionOptions) {
+                taskTargets = await expand.expandStrings(task.definition.targets, expansionOptions);
+                if (task.definition.options?.cwd){
```
elahehrashedi, line 202: I added `cwd `to be expanded too

src/cmakeTaskProvider.ts
```diff
@@ -237,12 +250,12 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
             } else {
                 // Search for the matching default task.
                 const defaultTask: CMakeTask[] = matchingTargetTasks.filter(task => task.isDefault);
-                if (defaultTask.length === 1) {
+                if (defaultTask.length >= 1) {
                     return defaultTask[0];
                 } else {
                     // Search for the matching existing task.
                     const existingTask: CMakeTask[] = matchingTargetTasks.filter(task => !task.isTemplate);
```
elahehrashedi, line 257: I don't see a `task.isTemplate` in tasks after you fetch tasks using` vscode.tasks.fetchTasks`. You have to distinguish between the tasks that are fetched from the tasks.json file and the tasks that the task provider provides as a template.

piomis: @elahehrashedi
Yes. I mentioned this problem already.
https://github.com/microsoft/vscode-cmake-tools/issues/2936#issuecomment-1380908811
But it was like that in your refactor.

elahehrashedi: INstead of using `vscode.tasks.fetchTasks` we can write our own function to read the tasks in tasks.json to fetch the existing tasks, and use the function `provideTasks ` to get the template tasks.
Here is a sample: 

https://github.com/microsoft/vscode-cpptools/blob/a014fe80bdc03b4a9fa3b877866a11d13034f631/Extension/src/LanguageServer/cppBuildTaskProvider.ts#L218

If you need help implementing this let me know and I will modify your branch.

piomis: I'm not sure in manual parsing of Json is good solution (honestly speaking I am not convinced that ad hock task creation is good solution either).
First of all this will be very fragile in regards to format of Json.
Like I rised in comments - you have added is Template property previously but your code never sets it to true. If you point me what place was intended to have it set then I will add it.

By the way I found one comment from maintainers of VS code which suggests not to use ad hock tasks (https://github.com/microsoft/vscode/issues/59337#issuecomment-519846474) maybe an Original proposal with 'target' retrieval during task execution (instead of having target set by default when task is created). This would also let keep usage of 'shell' task.

By the way I have just discovered that ProblemMatchers stopped working. And this is also related to fact that it is not possible to programicaly define non standard problem matcher in Task constructor.

elahehrashedi (dismissed):

elahehrashedi:
@bobbrow this can be checked in.

bobbrow (approved):

PR #3014, opened 2023-02-09
print the time it takes to complete a build

### LINKED ISSUE #3008: Print out the build time in the Output window
### Discussed in https://github.com/microsoft/vscode-cmake-tools/discussions/3003

<div type='discussions-op-text'>

<sup>Originally posted by **developsessions** February  6, 2023</sup>
How can I display the build time of an cmake project in vs code? It's not possible to wrap the cmake.exe command right?
I am a Windows user.</div>

bobbrow:
Fix #3008

### DIFF
--- a/src/drivers/cmakeDriver.ts
+++ b/src/drivers/cmakeDriver.ts
@@ -1566,11 +1566,13 @@ export abstract class CMakeDriver implements vscode.Disposable {
         const timeStart: number = new Date().getTime();
         const child = await this._doCMakeBuild(targets, consumer, isBuildCommand);
         const timeEnd: number = new Date().getTime();
+        const duration: number = timeEnd - timeStart;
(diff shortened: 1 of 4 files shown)


elahehrashedi (approved):

src/drivers/cmakeDriver.ts
```diff
@@ -1566,11 +1566,13 @@ export abstract class CMakeDriver implements vscode.Disposable {
         const timeStart: number = new Date().getTime();
         const child = await this._doCMakeBuild(targets, consumer, isBuildCommand);
         const timeEnd: number = new Date().getTime();
+        const duration: number = timeEnd - timeStart;
```
elahehrashedi, line 1569: The only problem is that in the case of "cmake.buildTask": true" the user may need to select a task (if there is no default task) in `_doCMakeBuild` that will be calculated as the build time.

PR #3028, opened 2023-03-02
Fix exportCompileCommandsFile scope access.

sean-mcmanus:
Fix for the warning in the Extension Host logging:

> Accessing a window scoped configuration for a resource is not expected. To associate 'cmake.exportCompileCommandsFile' to a resource, define its scope to 'resource' in configuration contributions in 'package.json'.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1920,7 +1920,8 @@
         "cmake.exportCompileCommandsFile": {
           "type": "boolean",
           "default": true,
-          "description": "%cmake-tools.configuration.cmake.exportCompileCommandsFile.description%"
+          "description": "%cmake-tools.configuration.cmake.exportCompileCommandsFile.description%",
+          "scope": "resource"
         },
         "cmake.useCMakePresets": {
           "type": "string",
--- a/src/drivers/cmakeDriver.ts
+++ b/src/drivers/cmakeDriver.ts
@@ -1496,13 +1496,14 @@ export abstract class CMakeDriver implements vscode.Disposable {
         }
 
         const configurationScope = this.workspaceFolder ? vscode.Uri.file(this.workspaceFolder) : null;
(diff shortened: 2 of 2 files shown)


bobbrow:
It seems the better fix would be to make `exportCompileCommandsFile` have resource scoping.  It looks like it is missing the property in `package.json`.  I'm not sure why it wasn't set (and also why the code you're changing reads from the configuration directly instead of using `this.config`).

```json
        "cmake.exportCompileCommandsFile": {
          "type": "boolean",
          "default": true,
          "description": "%cmake-tools.configuration.cmake.exportCompileCommandsFile.description%"
        },
```

sean-mcmanus:
> It seems the better fix would be to make `exportCompileCommandsFile` have resource scoping. It looks like it is missing the property in `package.json`. I'm not sure why it wasn't set and also why the code you're changing reads from the configuration directly instead of using `this.config`. (I imagine it's because the `ConfigurationReader` is missing these two settings).
> ```

It doesn't seems like making it resource scope makes sense. Keeping it workspace scope keeps the current functionality. Giving users the ability to set this variable differently for different workspace folders seems like it would be best to wait for a user asking for it with an example of why it makes sense for their project?

Regarding this.config, I just followed the existing code example. I could do something different.

bobbrow:
> It doesn't seem like making it resource scope makes sense. Keeping it workspace scope keeps the current functionality.

Allowing resource scoping does not remove the current functionality.  It's the easier change.  Please add the attribute to package.json.

sean-mcmanus:
> Allowing resource scoping does not remove the current functionality. It's the easier change. Please add the attribute to package.json.

Sure, that seems fine to me.

bobbrow (approved):

PR #3051, opened 2023-03-10
Save the state for multiple projects in the same folder

elahehrashedi:
The state of multiple projects in the same folder needs to be saved with a different name in the workspaceContext,  I added the name of the folder to distinguish between the two.

This better be checked in before PR #2993


elahehrashedi:
@bobbrow  I have tested this branch on multi-PR

src/state.ts
```diff
@@ -21,134 +21,181 @@ export class StateManager {
...
-    get ignoreCMakeListsMissing(): boolean {
-        return this._get<boolean>('ignoreCMakeListsMissing') || false;
+    getIgnoreCMakeListsMissing(folderName: string): boolean {
+        return this.isMultiProject ? (this._get<boolean>(`${folderName} ignoreCMakeListsMissing`) || false) :
+            (this._get<boolean>('ignoreCMakeListsMissing') || false);
     }
 
-    async setIgnoreCMakeListsMissing(v: boolean) {
-        await this._update('ignoreCMakeListsMissing', v);
+    async setIgnoreCMakeListsMissing(folderName: string, v: boolean) {
```
bobbrow, line 37: `folderName` cannot be trusted to be unique across multiple projects. It would be better to pass in some other information here instead. You could also consider passing in the `CMakeProject` instance and pick the data you want here instead of extracting the information first. It signals to the caller that the information should be coming from the CMake Project object (though I'm not sure if all the callers have access to that or not).

bobbrow (changes requested):

src/state.ts
```diff
@@ -21,134 +21,181 @@ export class StateManager {
         return this.extensionContext.globalState.update(this.folder.uri.fsPath + key, value);
     }
 
+    setIsMultiProject(isMultiProject: boolean) {
```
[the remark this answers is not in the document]
elahehrashedi: I removed the isMultiProject from the state, CMakeProject instances save this variable which shows if they are a part of a multi-project or single-project. Hence, I beleive CMakeProject is a more reliable place to check for this variable.

src/state.ts
elahehrashedi: `setIgnoreCMakeListsMissing` is only called in `reset` which is only called on CMakeProject. I think it is reliable in this case.
In this new design, we are saving a different value for `ignoreCMakeListsMissing` for every project. previously we were only saving one variable for all projects.

snehara99: marking resolved due to `ignoreCmakeListsMissing` being unique for each project

snehara99 (approved):
tested

PR #3064, opened 2023-03-16
Test Explorer - Part 3

xisui-MSFT:
Enabled debugging tests, using selected launch configuration in launch.json or code-workspace file.

`${cmake.testProgram}` and `${cmake.testArgs}` are resolved in debug handler instead of using commands (which has format like `${command:cmake.launchTargetPath}`), since both of them require test name as a parameter, and having one launch configuration for each test is obviously not a good idea.

If there are multiple launch configs, one must be chosen from a quick pick menu. We may remember this chosen config, but since there's currently a status bar / side panel rework, this should be done after.

debugCTest


gcampbell-msft (approved):

benmcmorran:
@xisui-MSFT this PR also ended up with merge conflicts.

xisui-MSFT:
> @xisui-MSFT this PR also ended up with merge conflicts.

Fixed, and added one more line to CHANGELOG.

benmcmorran (approved):

PR #3056, opened 2023-03-14
Add default compiler search paths for msys2 default installation paths, fix mingw bugs

philippewarren:
### This changes visible behavior

The following changes are proposed:

- MSYS2 default installation folders (mingw64, mingw32, ucrt64, clang64, clang32 and clangarm64) are now scanned by default
- MinGW path is now correctly added when using a MinGW kit (was not added correctly)
- MinGW path is now prepended instead of appended (to allow using a MinGW environment even if there is another one in the standard PATH)
- The configuration option `cmake.mingwSearchDirs` is now `cmake.additionalCompilerSearchDirs`
- The value of `cmake.mingwSearchDirs`/`cmake.additionalCompilerSearchDirs` is now correctly used when scanning, no matter how the scan is performed (was only used when the scan was performed from the command `CMake: Scan for kits`)
- The compiler name now includes the name of the parent directory of the parent directory of the compiler (i.e. for a compiler `C:\msys64\mingw64\bin\gcc.exe`, the name will include `(mingw64)` to prevent removal of different compilers with the exact same name, but different MinGW environments.

## The purpose of this change

Fixes bugs with MinGW environments.
Allows to more easily use MSYS2-installed MinGW environments.
Allows to use alternative MSYS2 MinGW environments (clang64, ucrt64, etc.)

## Other Notes/Information

Feel free to suggest any other improvements ideas concerning MSYS2 and/or MinGW support.

PR #3067, opened 2023-03-21
Add trust to kit scanning paths

benmcmorran:
This PR adds a notion of trust to each path used when scanning for kits. Paths from the `PATH` environment variable, in Program Files (on Windows), and from `settings.json` are implicitly trusted. Other paths are not trusted. Kits in untrusted paths will still be scanned, but the executables will not be invoked until the user has trusted them. If untrusted kits are found, the user is prompted to trust them with this notification. Clicking the "Trust" button will automatically edit `settings.json` to add the paths and scan again. 

image

This is a security follow-up from #3056. FYI @philippewarren

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -14,7 +14,7 @@ import CMakeProject from './cmakeProject';
 import * as expand from './expand';
 import { VSInstallation, vsInstallations, getHostTargetArchString, varsForVSInstallation, generatorPlatformFromVSArch } from './installs/visualStudio';
 import * as logging from './logging';
-import paths from './paths';
+import paths, { PathWithTrust } from './paths';
 import { fs } from './pr';
 import * as proc from './proc';
 import { loadSchema } from './schema';
@@ -146,6 +146,11 @@ export interface Kit extends KitDetect {
      * If `true`, keep this kit around even if it seems out-of-date
      */
     keep?: boolean;
+
+    /**
+     * If `true`, this kit comes from a trusted path.
+     */
+    isTrusted: boolean;
 }
 
 interface CompilerVersion {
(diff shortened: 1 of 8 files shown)


src/kit.ts
```diff
@@ -1186,7 +1238,11 @@ export async function readKitsFile(filePath: string, workspaceFolder?: string, e
         }
         return [];
     }
-    const kits = kits_raw as Kit[];
+    const kits = (kits_raw as Kit[]).map(kit => {
+        // Serialized kits are trusted for backwards compatibility if not otherwise specified.
```
gcampbell-msft, line 1242: Is this because if they're serialized we're assuming they've already been ran? Or why is this assumption being made?

gcampbell-msft (dismissed):
Everything looks reasonable to me, I like the idea even without the added security safety, because it helps keep the user aware of what is being ran. 👍

src/kit.ts
benmcmorran: CMake Tools has a cache of previously discovered kits. If we don't trust this cache, then all previously discovered kits will be considered untrusted until kits are scanned again, which potentially breaks user workflows.

gcampbell-msft: Makes sense.

src/kit.ts
```diff
@@ -347,8 +353,8 @@ export async function kitIfCompiler(bin: string, pr?: ProgressReporter): Promise
     const clang_cl_res = clang_cl_regex.exec(fname);
 
     if (gcc_res) {
-        const version = await getCompilerVersion('GCC', bin, pr);
-        if (version === null) {
+        const version = isTrusted ? await getCompilerVersion('GCC', bin, pr) : null;
+        if (isTrusted && version === null) {
```
snehara99, line 357: why is the `isTrusted` check needed here? `version` should be null if `isTrusted === false` right?

snehara99 (approved):

gcampbell-msft (approved):

PR #3112, opened 2023-04-14

csigs:


benmcmorran:
Abandoning this PR because I think it's gone stale. Looking at the internal runs of the localization pipeline I see new localized strings, but here it looks like things are just getting reverted to English.

bobbrow:
Something could be out of sync since I believe this file was deleted as part of the new status panel, but then status.ts was added back to the repository.  They may need to retranslate everything before this PR will clear up (it will likely be back tomorrow).

Issue #3121, opened 2023-04-20
Unable to use the extension when `cmake.sourceDirectory` is a subdirectory


FtZPetruska:
### Brief Issue Summary

Starting version 1.14.29, the extension fails to work when the source directory is not the root of the workspace.

Whether the project is already configured or not, the `CMake/Build` tab in `OUTPUT` reads:
```
Unhandled exception: Failed to open CMake cache file on code model update Folder is not found in Test Explorer: /users/ganyu/git/dsda-doom/prboom2 {}
```

For reference, the workspace path is `/users/ganyu/git/dsda-doom`, and the top-level CMakeLists.txt is located in the subdirectory `prboom2`, and set through this workspace setting:
```
"cmake.sourceDirectory": "${workspaceFolder}/prboom2"
```

When this error happens, the CPU usage skyrockets, and no command from the status bar or the command palette have any effect.

It should be noted that this issue does not happen when then CMakeLists.txt is at the root of the directory.

### CMake Tools Diagnostics

```shell
{
  "os": "darwin",
  "vscodeVersion": "1.77.3",
  "cmtVersion": "1.14.29",
  "configurations": [
    {
      "folder": "/Users/ganyu/git/dsda-doom",
      "cmakeVersion": "unknown",
      "configured": false,
      "generator": "unknown",
      "usesPresets": false,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": null
    }
  ]
}
```

### Debug Log

```shell
[main] Starting new CMake driver
[main] Starting CMake driver
[main] Using generator from user configuration: Ninja
[cmakefileapi-driver] Creating instance of CMakeFileApiDriver
[driver] Initializating base driver using kit
[kit] Testing Clang binary: /usr/bin/clang
[proc] Executing command: /usr/bin/clang -v
[kit] Detected Clang compiler: /usr/bin/clang
[driver] CMakeDriver Kit set to Clang 14.0.0 arm64-apple-darwin21.6.0
[...]

PR #3122, opened 2023-04-20
Fix a regression when initializing with sourceDir in a subfolder

bobbrow:
Fixes: #3121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # What's New?
 
-## 1.14
+## 1.14.30
+Bug Fixes:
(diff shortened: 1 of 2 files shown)


xisui-MSFT (approved):

bobbrow:
Don't accept this yet.  I'm seeing instability in VS Code while debugging this with the standard case.

bobbrow:
Ok. It seems to be an unrelated issue with one of my other extensions.  But the test explorer gets stuck for me when I try to run a simple set of tests.  The tests work in the Terminal.  @xisui-MSFT you might want to help me investigate that.

benmcmorran (approved):

benmcmorran:
Bypassing non-Linux tests. Bob's manual testing was on Windows and this issue is currently blocking important scenarios in the official release.

Issue #3118, opened 2023-04-19
Compiling files with "warning" in the filename causes the extension to report "build finished with warning(s)"


Longhanks:
### Bug Summary and Steps to Reproduce

**Bug Summary**: When using a build task to compile a C++ project where a file is called e.g. `warnings.cpp`, the extension prints the build summary "build finished with warning(s)." despite there no being any warnings.

I think the issue stems from this line: https://github.com/microsoft/vscode-cmake-tools/blob/db625d2b897cceb3141eec9b7f8b7e71e6b796ae/src/cmakeTaskProvider.ts#L495 which parses the stdout output of `cmake --build`. Parsing e.g. `[1/2] Building CXX object CMakeFiles\test.dir\warnings.cpp.obj` as a warning is a false positive: There is no warning, merely the *filename* contains "warning".

**Steps to reproduce:**

[test.zip](https://github.com/microsoft/vscode-cpptools/files/11263235/test.zip)

1. Download attached sample project and VS Code configuration
  - `test/CMakeLists.txt`
  - `test/CMakePresets.json`
  - `test/.vscode/tasks.json`
  - `test/warnings.cpp`
3. Use VS Code with vscode-cpptools and vscode-cmake-tools to open the directory.
4. Run command "CMake: Select build preset" and select x64-debug
5. Run command "CMake: Configure"
6. Run command "Tasks: Run Build task"

**Expected behavior:**

Windows:
```
 *  Executing task: build 

build task started....
"C:\Program Files\CMake\bin\cmake.EXE" --build C:/Users/aschulz/Desktop/test/out/build/x64-Debug-win --target test
[1/2] Building CXX object CMakeFiles\test.dir\warnings.cpp.obj
[2/2] Linking CXX executable test.exe
build finished successfully.
```

Linux:
```
 *  Executing task: build 

build task started....
/usr/local/bin/cmake --build /home/aschulz/Desktop/test/out/build/x64-Debug-linux --target test
[ 50%] Building CXX object CMakeFiles/test.dir/warnings.cpp.o
[100%] Linking CXX executable test
[100%] Built target test
build finished successfully.
```

**Actual output:**

Windows:
```
 *  Executing task: build 

build task started....
"C:\Program Files\CMake\bin\cmake.EXE" --build C:/Users/aschulz/Desktop/test/out/build/x64-Debug-win --target test
[1/2] Building CXX object CMakeFiles\test.dir\warnings.cpp.obj
[...]

PR #3142, opened 2023-04-26
fixing warning identification in outputs when building through tasks

snehara99:
When running a build using tasks, if "warning" was a part of the target file path (which is part of the output), it would identify the output as "finished with warning(s)" because we would check for "warning" appearing anywhere in the output to identify actual compiler warnings. I changed the check from looking for "warning" to looking for ": warning" to be more specific. This addresses issue [3118](https://github.com/microsoft/vscode-cmake-tools/issues/3118).

I tested various compilers [here](https://godbolt.org/) to see how warning messages were structured:
image


benmcmorran (changes requested):
Code changes look good to me. Can you update the 1.15 section of `CHANGELOG.md` to include this bug fix?

benmcmorran (approved):

Issue #3154, opened 2023-05-01
Starting a test run from the Test Explorer with "No Test Preset Selected" just hangs instead of asking for the preset


TheJCAB:
### Brief Issue Summary

Make sure there's no test preset selected (says so in the status bar), open the Test Explorer and click the play button to start a test run.

It'll do a build, and then... nothing. It'll say it's running the tests forever.

After a lot of looking around, I found the little "stop" button in the test Explorer. (aside: I was expecting there to be a "cancel" button in the notification panel, as is done with other types of tasks, but there was none).

After stopping the run, the notification panel changed to an error with the message: "An error occurred attempting to run tests: Test preset is not set".

Arguably, this error should happen immediately, not wait for the user to cancel the run.

Also arguably, the extension should ask the user for a preset. The picker menus in VSCode can be very easy to miss, but I checked specifically for it and there's none when this happens.

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.77.3",
  "cmtVersion": "1.14.30",
  "configurations": [
    {
      "folder": "x:\\TTD4\\CoreTtd.Tests",
      "cmakeVersion": "3.20.5",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {
        "C": "cl.exe",
        "CXX": "cl.exe"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "Debug",
    "buildTypesSeen": [
      "Debug"
    ],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 33,
    "executablesCount": 25,
    "librariesCount": 8,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
[cmakefileapi-parser] Read reply folder: X:\TTD4\CoreTtd.Tests\build\ARM64-Debug-Win32\.cmake\api\v1\reply
[...]

PR #3163, opened 2023-05-05
Fix test explorer hang

xisui-MSFT:
Fix issue #3154, and show a test preset picker when it's not set when running tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # What's New?
 
+## 1.15.0
+Bug Fixes:
+- Handle multiple test results in one test run. [#3160](https://github.com/microsoft/vscode-cmake-tools/pull/3160)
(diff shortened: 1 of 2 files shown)


benmcmorran (changes requested):
The code changes look good to me. Can you add an entry to the changelog for this bug fix and for the fix in #3160? You'll need to add a new section at the top of the changelog for version 1.15 since we released version 1.14.31 earlier this week.

benmcmorran (approved):

bobbrow:
I am getting prompted to select a test preset every time I open my folder.  Is this being called too early?

xisui-MSFT:
The fix executes the extension command and doesn't call a function. So the preset should be saved. I'll try to repro.

Thanks,
Xinyu Sui
________________________________

PR #3175, opened 2023-05-13
Clear test explorer during CMake driver creation

xisui-MSFT:
If configuring on open, refreshing test explorer requires selecting test preset, which makes test explorer refresh in CMake driver creation.

Refreshing test explorer during driver creation doesn't seem to make sense - if user need auto-reconfigure, then test explorer will be refreshed during configure; and if not, they probably don't want something that can potentially require configuring their project automatically invoked.


bobbrow (approved):

benmcmorran (approved):

bobbrow:
Yes, I don't think the current behavior is desireable.  We don't ask to select a build preset until the customer tries to build and I don't think we should ask to select a test preset until we try to test.  Not sure if the Test Explorer lets you know when it is focused, but if not you can set a placeholder item in the test explorer that prompts you to select a test preset. Then it can refresh when the test preset is selected.  The other thing about my case is that I don't even have test presets.  The drop down is asking me to select the [[Default]] generated preset which it could do without my input.

xisui-MSFT:
Looks good to me.

PR #3183, opened 2023-05-17
added ability to select either C or C++ with the Quick Start commmand.

REDl3east:
### This changes the Quick Start command to include an option to specify either C or C++.

- Made the Quick Start relevant source files given the language selected. .cpp for C++ and .c for C.
- Added LANGUAGES to the generated CMakeLists.txt file.
- Generating similar source files to C++, except it uses stdio.h instead of iostream.

## The purpose of this change

This pull request is for the convenience of C programmers.

### DIFF
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -2439,6 +2439,21 @@ export class CMakeProject {
             return -1;
         }
 
+        const targetLang = (await vscode.window.showQuickPick([
+            {
+                label: 'C++',
+                description: localize('create.cpp', 'Create a C++ project')
+            },
+            {
+                label: 'C',
+                description: localize('create.c', 'Create a C project')
+            }
+        ]));
+
+        if (!targetLang) {
+            return -1;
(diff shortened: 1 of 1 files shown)


REDl3east:

src/cmakeProject.ts
```diff
@@ -2452,16 +2467,19 @@ export class CMakeProject {
         }
 
         const type = targetType.label;
+        const lang = targetLang.label;
+        const langName = lang === "C++" ? "CXX" : "C";
```
benmcmorran, line 2471: Given that [CMake's default behavior is to enable both `C` and `CXX`](https://cmake.org/cmake/help/latest/command/enable_language.html), please either change this to `C CXX` when `lang` is `C++` or add a new `C and C++` option to the quick pick.

benmcmorran (approved):

benmcmorran:
Thanks again! These changes should be available in the pre-release channel tomorrow.

PR #3199, opened 2023-06-05
Dev/snehara/set sidebar default false

snehara99:
set sidebar to false by default


benmcmorran:
Approved. Bypassing checks so that we can get this hotfix out quickly.

PR #3207, opened 2023-06-13
Don't ask for test preset if it's not set when refreshing test explorer

xisui-MSFT:
As we discussed in #3175, a placeholder message will be displayed if test explorer is initialized, and no test preset has been selected:

image


src/ctest.ts
```diff
@@ -602,14 +604,40 @@ export class CTestDriver implements vscode.Disposable {
...
+        let presetMayChange = false;
+        for (const test of tests) {
+            if (test.id === testPresetRequired) {
+                const folder = test.parent ? test.parent.id : test.id;
+                const project = await this.projectController?.getProjectForFolder(folder);
+                if (!project) {
+                    log.error(localize('no.project.found', 'No project found for folder {0}', folder));
+                    return false;
+                }
+                await vscode.commands.executeCommand('cmake.selectTestPreset', project.workspaceFolder);
```
bobbrow, line 618: Doesn't this prompt for a selection?  I thought this method just wants to "check" presets, not set them.  It might be cleaner to check the result of calling this method and then prompt to select the test preset if it fails.

src/drivers/cmakeDriver.ts
```diff
@@ -485,78 +485,113 @@ export abstract class CMakeDriver implements vscode.Disposable {
      * Change the current configure preset. This lets the driver reload, if necessary.
      * @param configurePreset The new configure preset
      */
-    async setConfigurePreset(configurePreset: preset.ConfigurePreset): Promise<void> {
-        if (!this.useCMakePresets) {
-            log.info(localize('skip.set.config.preset', 'Using kits, skip setting configure preset: {0}', configurePreset.name));
-            return;
-        }
+    async setConfigurePreset(configurePreset: preset.ConfigurePreset | null): Promise<void> {
```
bobbrow, line 488: We should prefer APIs that use `undefined` over `null`.  We'd been starting to remove `null` in the codebase and hoped to get all references to it out.  The syntax is cleaner too. `configurePreset?: preset.ConfigurePreset`

src/presetsController.ts
```diff
@@ -699,6 +699,8 @@ export class PresetsController {
         );
 
         this._isChangingPresets = false;
+
+        // await vscode.commands.executeCommand('cmake.refreshTests', this.workspaceFolder);
```
bobbrow, line 703: Delete this if it's not needed.

src/ctest.ts
```diff
@@ -602,14 +604,40 @@ export class CTestDriver implements vscode.Disposable {
...
+                presetMayChange = true;
```
bobbrow, line 619: After this, you probably want to break out of the loop or it will run for all the tests.

src/drivers/cmakeDriver.ts
```diff
@@ -485,78 +485,113 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
-            await this._setConfigurePreset(configurePreset);
-        });
+            const newBinaryDir = configurePreset.binaryDir;
+            const needs_clean = this.binaryDir === newBinaryDir && preset.configurePresetChangeNeedsClean(configurePreset, this._configurePreset);
+            await this.doSetConfigurePreset(needs_clean, async () => {
+                await this._setConfigurePreset(configurePreset);
+            });
+        } else {
+            if (!this.useCMakePresets) {
+                log.info(localize('skip.unset.config.preset', 'Using kits, skip unsetting configure preset'));
```
bobbrow, line 504: Is this called in real scenarios?  It seems like a bug if we're trying to set a preset while using kits.

src/ctest.ts
```diff
@@ -602,14 +604,40 @@ export class CTestDriver implements vscode.Disposable {
...
+
+    private async runTestHandler(request: vscode.TestRunRequest, useCMakePresets: boolean, cancellation: vscode.CancellationToken) {
         if (!testExplorer) {
             return;
         }
 
         const requestedTests = request.include || this.testItemCollectionToArray(testExplorer.items);
         const tests = this.uniqueTests(requestedTests);
 
+        if (!await this.checkTestPreset(tests)) {
```
bobbrow, line 637: `useCMakePresets` is passed in to this function, I don't think you should check the test preset if you're not using presets.

src/ctest.ts
```diff
@@ -619,6 +647,11 @@ export class CTestDriver implements vscode.Disposable {
 
     private async debugCTestHelper(tests: vscode.TestItem[], run: vscode.TestRun, cancellation: vscode.CancellationToken): Promise<number> {
         let returnCode: number = 0;
+
+        if (!await this.checkTestPreset(tests)) {
```
bobbrow, line 651: Is there a way to know at this point whether presets are being used?  We shouldn't check this if you're not using presets.

xisui-MSFT: `useCMakePresets` in runCTest* was added by Elah. Now that you say this, I found it has some problems. There's only one test explorer instance, and all folders share it. So userCMakePresets shouldn't be passed in, or there will be issues when trying to run tests in a preset folder and kit folder at the same time. I'll fix this.

As for this function itself, it checks if there are tests with magic name `_test_preset_required_`. And it shouldn't be added to the test list if the folder is not using CMake presets. I'll use a better name, like "testsRequireTestPreset", and add a comment.

src/ctest.ts
xisui-MSFT: See the reply to the comment below.

src/drivers/cmakeDriver.ts
xisui-MSFT: The original code has this, so I duplicated it. Technically, I guess this can probably happen when user invokes the command directly, bypassing the UI?

src/drivers/cmakeDriver.ts
xisui-MSFT: Agreed. I'll change those touched in this PR.

src/ctest.ts
xisui-MSFT: See my reply to the comment below, this function relies on the test id to get the folder name and other info. Separating it is rather complicated. I'll try to find a proper name for it and add some comments.

src/ctest.ts
xisui-MSFT: I'd like to ask the user to select test presets for all folders that that user want to run tests. Do you think picking one each time a better option?

src/drivers/cmakeDriver.ts
bobbrow: I suppose another extension could theoretically execute the command even though it's disabled in the UI, but I would think we should just exit early as opposed to logging this behavior.

src/ctest.ts
bobbrow: Does `test` represent a folder, or an individual test?  If it's a workspace folder, you might want to rename it.  If it's a single test, then the current implementation will ask you to set the preset for every single test in the folder which I don't think makes sense.

src/ctest.ts
bobbrow: It looks like you select presets for every test in every folder.  I think you need to track whether you've already set the preset or not so that you don't prompt the user N times.

src/ctest.ts
xisui-MSFT: When a test preset is not selected, as shown in the screenshot above, there will only be one "test" under each folder. This "test" has a special ID. If the test ID matches this special ID, user will be prompted to select a test preset for this folder.

bobbrow: I'm confused. Is it a test, or a node in the test explorer with tests under it?

xisui-MSFT: When no presets are selected, we can't know the correct test list. So the structure will be
```
- Folder 1
  - Select a test preset to discover tests <- this is the special "test"
- Folder 2
  - ...
```

bobbrow: So `test` is the placeholder for the folder?  Can you say that in a comment when we enter this `if` block?

xisui-MSFT: It's a placeholder entry for the list of test in the folder. But I think you know what I mean. Sure I'll add a comment.

src/drivers/cmakeDriver.ts
xisui-MSFT: Turns out this requires too many changes. We should change them all together in a separate PR.

src/ctest.ts
```diff
@@ -602,23 +603,63 @@ export class CTestDriver implements vscode.Disposable {
...
-    private async runTestHandler(request: vscode.TestRunRequest, cancellation: vscode.CancellationToken, useCMakePresets: boolean) {
+    /**
+     * This function checks if tests require test presets already have a test preset selected.
+     * Check is done by looking for magic test item testPresetRequired. When test preset is not selected, there will
+     * be one and only one such test item under that folder.
+     * When test preset is not selected, this function will prompt for test preset selection. Changing test preset triggers
+     * test explorer refresh.
+     *
+     * Returns false if any test preset wasn't selected already. This means either test explorer is going to be refreshed,
+     * or user cancelled the selectoin. So we shouldn't proceed in most cases.
```
bobbrow, line 614: [sp] selection

benmcmorran (approved):

gcampbell-msft (approved):

PR #3223, opened 2023-06-28
Update package.json to have experimental setting

sinemakinci1:
Adding experimental setting tag so this setting gets picked up by experimentation platform

### DIFF
--- a/package.json
+++ b/package.json
@@ -2374,7 +2374,10 @@
           "type": "boolean",
           "default": false,
           "description": "%cmake-tools.configuration.cmake.useProjectStatusView.description%",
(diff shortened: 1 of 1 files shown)


snehara99 (approved):
looks good to me

gcampbell-msft (approved):

benmcmorran:
@sinemakinci1 @snehara99 Closing this PR because we already shipped the experimental setting in a 1.14 patch release, and I believe we are not planning to continue experimentation in 1.15 (which is what merging this PR would enable). If that understanding isn't correct, please reactivate the PR.

PR #3225, opened 2023-06-30
attempt specific permissions

gcampbell-msft:

- 
- 
-

### DIFF
--- a/.github/actions/locker/action.yml
+++ b/.github/actions/locker/action.yml
@@ -29,5 +29,5 @@ inputs:
   readonly:
     description: If true, changes are not applied.
 runs:
-  using: 'node12'
+  using: 'node16'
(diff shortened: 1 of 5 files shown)


bobbrow:
I ran the Action/bot manually with this branch and it still failed. https://github.com/microsoft/vscode-cmake-tools/actions/runs/5447019987/jobs/9908535054

gcampbell-msft:
> I ran the Action/bot manually with this branch and it still failed. https://github.com/microsoft/vscode-cmake-tools/actions/runs/5447019987/jobs/9908535054

I also tried this and failed. I got another response on the ticket and they suggested trying to update from node12 to node16, but I'm still noticing a failure. I'm wondering if we need to update our dependencies on the @actions npm packages?

Issue #3206, opened 2023-06-12
`cmake.parallelJobs` ignored in workspace's `.vscode/settings.json`


diablodale:
### Brief Issue Summary

When `cmake.parallelJobs` is set the the user's main settings.json and is also set in workspace's `.vscode/settings.json`, then the workspace value is ignored and only the main user's setting is used. (I thought this worked in the past but noticed it does not today).

#### user's settings.json
```json
"cmake.parallelJobs": 14,
```

#### workspace .vscode/settings.json
```json
"cmake.parallelJobs": 1,
```

#### Result
a cmake build in vscode of...
```
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.exe" --build c:/njs/depthai-core/build --config Debug --target all -j 14 --
[build] [14/50   2% :: 0.140] cmd.exe /C "cmd.exe /C "...
```

#### Workaround 
Don't use the workspace settings file for this cmt setting. Instead, set the desired parallelJobs in the main user settings.json

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.79.0",
  "cmtVersion": "1.14.33",
  "configurations": [
    {
      "folder": "c:\\njs\\depthai-core",
      "cmakeVersion": "3.26.4",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": false,
      "compilers": {
        "C": "C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe",
        "CXX": "C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "Debug",
    "buildTypesSeen": [
      "Debug"
    ],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 144,
    "executablesCount": 112,
    "librariesCount": 4,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": false
    }
  ]
}
```

benmcmorran:
Thanks for the report! We'll try to reproduce this issue and let you know if we have any additional questions.

[...]

PR #3233, opened 2023-07-10
fix precedence order of cmake.parallelJobs setting

gcampbell-msft:
## This change addresses item [#3206](https://github.com/microsoft/vscode-cmake-tools/issues/3206)

This fixes the precedence order of the `cmake.parallelJobs` setting.


gcampbell-msft:

benmcmorran (dismissed):
Needs a changelog update.

snehara99 (approved):
seems good to me! Do we have a note somewhere on the preference order that users can see?

gcampbell-msft:
> seems good to me! Do we have a note somewhere on the preference order that users can see?

Not directly from us, but from VSCode itself, there is a proper settings order: https://code.visualstudio.com/docs/getstarted/settings#_settings-precedence

Issue #2897, opened 2022-12-13
Incorrect schema version constraint applied to CMakeUserPresets.json


craigscott-crascit:
### Brief Issue Summary

If a user preset defined in `CMakeUserPresets.json` inherits from a base preset defined in `CMakePresets.json`, VS Code incorrectly requires the `version` field in `CMakeUserPresets.json` to be no older than the version needed for the base preset.

To reproduce the bug:

* In `CMakePresets.json`, set `version` to 3.
* In `CMakePresets.json`, define a configure preset and add the `toolchainFile` field (this requires schema version 3 or later).
* In `CMakeUserPresets.json`, set `version` to 2.
* In `CMakeUserPresets.json`, define a configure preset that inherits from the one defined above. Don't add anything to this new configure preset, it just needs to inherit from the other one.

CMake itself accepts the above arrangement, which you can verify using `cmake --list-presets`. However, VS Code complains that the `toolchainFile` field requires version 3 and refuses to accept it. But that version constraint is satisfied in the context where `toolchainFile` is used (`CMakePresets.json`). The `CMakeUserPresets.json` file is not required to also be version 3 in order to inherit the base preset which does use something from a version 3 schema.

Potentially related: https://gitlab.kitware.com/cmake/cmake/-/issues/22428

elahehrashedi:
using `cmake --list-presets`, I see that CMake doesn't accept the above arrangement either. 
Can you share your preset settings, and the results of  `cmake --list-presets` ?

craigscott-crascit:
I can confirm that CMake _does_ accept it. Here are some minimal files that you can use. I confirmed that `cmake --list-presets` works with both the latest CMake release (3.25.1) and an older feature release 3.23.3 which still supports presets version 3.

<details><summary>CMakePresets.json</summary>

```json
{
  "version": 3,
  "configurePresets": [
    {
      "name": "default",
      "generator": "Ninja",
      "binaryDir": "${sourceDir}/build",
      "toolchainFile": "${sourceDir}/toolchain.cmake"
    }
  ]
}
```
</details>

<details><summary>CMakeUserPresets.json</summary>

```json
{
  "version": 2,
  "configurePresets": [
    {
      "name": "my_thing",
      "inherits": "default"
    }
  ]
}
```
</details>

<details><summary>CMakeLists.txt</summary>

```cmake
cmake_minimum_required(VERSION 3.21)
project(vscode_bug_2897 LANGUAGES NONE)
```
[...]

PR #3234, opened 2023-07-10
check for incorrect versioning and settings before expanding

gcampbell-msft:
## This change addresses item [#2897](https://github.com/microsoft/vscode-cmake-tools/issues/2897)

- Modify where we check for invalid settings in less than or equal to version 2. 

## The purpose of this change
This will fix the problem where we basically don't allow different versions in userpresets and presets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Bug Fixes:
 
 Improvements:
 - Decreased the number of cases where we reconfigure erroneously upon usage of `cmake.getLaunchTargetPath` [#2878](https://github.com/microsoft/vscode-cmake-tools/issues/2878)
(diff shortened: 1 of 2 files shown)


gcampbell-msft:

benmcmorran (approved):
Needs a changelog update.

Also just double-checking, are there any other cases where we error out based on version number that would also need to move to per-file processing?

gcampbell-msft:
> Needs a changelog update.
> 
> Also just double-checking, are there any other cases where we error out based on version number that would also need to move to per-file processing?

I'll double check before merging.

gcampbell-msft:
> Needs a changelog update.
> 
> Also just double-checking, are there any other cases where we error out based on version number that would also need to move to per-file processing?

Not that I see

snehara99 (approved):

PR #3238, opened 2023-07-14
Improve getLaunchTargetPath

### LINKED ISSUE #2878: Any reference to ${command:cmake.getLaunchTargetPath} will reconfigure after a cmake configure task
### Brief Issue Summary

Running a configure task such as
``` json
{
    "label": "cmake configure",
    "type": "cmake",
    "command": "configure",
    "preset": "${command:cmake.activeConfigurePresetName}",
},
```
to completion, build with
``` json
{
  "label": "cmake build",
  "type": "cmake",
  "command": "build",
  "preset": "${command:cmake.activeConfigurePresetName}",
  "targets": [
      "${command:cmake.buildTargetName}"
  ],
}
```
a follow up with
``` json
{
    "label": "cmake run target",
    "type": "shell",
    "command": "${command:cmake.getLaunchTargetPath}",
    "dependsOn": "cmake build",
},
```
will always reconfigure.

It seems that the `command:cmake.<vars>` state are not saved/fetched when using a cmake configure task. Such that the extension itself need to load it by doing a configure for itself.  
Deleting the build folder, configuring, building and running the `cmake run target` task directly after this does _not_ make it reconfigure a second time, which I guess perhaps is also wrong?
If one does `Ctrl+shift+p` -> `Developer: Reload window` after the project is fully configured and built, the `cmake run target` task will make the cmake tools extension reconfigure again.

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

_No response_

gcampbell-msft:
### This decreases the number of cases where we erroneously reconfigure upon `cmake.getLaunchTargetPath`.

- Pass in the default current preset name (if present) in order to compare against the tasks preset name.
- In the build task, also check if we can configure using our task provider in order to make sure the `needsReconfigure` gets set correctly for any future `getLaunchTargetPath`. 

## The purpose of this change

To resolve #2878. This will improve the configure task such that it will also set the `needsReconfigure` flag to false when the requested preset matches the default. 

## Other Notes/Information

However, this  still cannot guarantee that this will always work in all cases, due to the fact that tasks can configure/build/run items from any preset, which may or may not match the default. Additionally, the build task directly invokes cmake.exe, so if the build tasks causes cmake to configure, the extension will not be aware of this, so there are still cases where the configure will happen twice.

### DIFF
--- a/src/drivers/cmakeFileApiDriver.ts
+++ b/src/drivers/cmakeFileApiDriver.ts
@@ -198,7 +198,7 @@ export class CMakeFileApiDriver extends CMakeDriver {
         return 0;
     }
 
-    async doConfigure(args_: string[], outputConsumer?: proc.OutputConsumer, showCommandOnly?: boolean, configurePreset?: ConfigurePreset | null, options?: proc.ExecutionOptions, debuggerInformation?: DebuggerInformation): Promise<number> {
+    async doConfigure(args_: string[], outputConsumer?: proc.OutputConsumer, showCommandOnly?: boolean, defaultConfigurePresetName?: string, configurePreset?: ConfigurePreset | null, options?: proc.ExecutionOptions, debuggerInformation?: DebuggerInformation): Promise<number> {
         const binaryDir = configurePreset?.binaryDir ?? this.binaryDir;
         const api_path = this.getCMakeFileApiPath(binaryDir);
         await createQueryFileForApi(api_path);
@@ -291,7 +291,7 @@ export class CMakeFileApiDriver extends CMakeDriver {
             log.trace(result.stderr);
             log.trace(result.stdout);
             if (result.retc === 0) {
-                if (!configurePreset) {
+                if (!configurePreset || (configurePreset && defaultConfigurePresetName && configurePreset.name === defaultConfigurePresetName)) {
                     this._needsReconfigure = false;
                 }
                 await this.updateCodeModel(binaryDir);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Features:
 Bug Fixes:
 - When using CMake presets, the Project Status View now shows the build target along with the build preset. [PR #3241](https://github.com/microsoft/vscode-cmake-tools/pull/3241)
 
+Improvements:
+- Decreased the number of cases where we reconfigure erroneously upon usage of `cmake.getLaunchTargetPath` [#2878](https://github.com/microsoft/vscode-cmake-tools/issues/2878)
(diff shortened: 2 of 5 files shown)


gcampbell-msft:
@bobbrow @benmcmorran This fix could possibly increase the number of cases where we don't erroneously configure twice. However, I wonder if a better way to do this is to remove the check for `needsReconfigure` in the `cmake.getLaunchTargetPath`, since we already don't ensure that the target exists.  

I'm curious on your feedback, thanks.

gcampbell-msft:

src/drivers/cmakeFileApiDriver.ts
```diff
@@ -291,7 +291,7 @@ export class CMakeFileApiDriver extends CMakeDriver {
             log.trace(result.stderr);
             log.trace(result.stdout);
             if (result.retc === 0) {
-                if (!configurePreset) {
+                if (!configurePreset || (configurePreset && defaultConfigurePresetName && configurePreset.name === defaultConfigurePresetName)) {
```
benmcmorran, line 294: Just curious based on the `defaultConfigurePresetName && configurePreset.name === defaultConfigurePresetName` check, is there a known case where `configurePreset` is truthy but `configurePreset.name` is falsy?

benmcmorran (dismissed):

benmcmorran:
This also needs a changelog update.

src/drivers/cmakeFileApiDriver.ts
gcampbell-msft: I don't believe so. My understanding is that there MUST be a name for a CMake Preset, because the name should be the unique identifier of it compared to other presets. If we were using `displayName` then I think there could be a case, but not for `name`. Thanks!

bobbrow: I don't remember why there would need to be a condition around setting `this._needsReconfigure`. It used to be set unconditionally here until #2636.

bobbrow: And if you make a change to the condition here, I would think you'd want to do it for the other drivers as well.

gcampbell-msft: @bobbrow I had looked through that PR as well while implementing this. The reason it's not set unconditionally is basically that it's using that variable primarily for extension state, thus the commands, and auto-configure, etc. However, tasks can be configuring for any preset, which might not match what is chosen in the UI. 

I'll check about changing the condition for other drivers.

snehara99 (approved):

PR #3241, opened 2023-07-17
allows build target to display when using presets

snehara99:
The build target shows in the Project Status View in the side panel now and mimics the same behaviour of the option in the status bar. Below are before and after screenshots when using CMake Presets.

Before (can't select build target):
image

After (can select build target):
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 Features:
 - Added support for the CMake Debugger. [#3093](https://github.com/microsoft/vscode-cmake-tools/issues/3093)
 
-## 1.14.35
 Bug Fixes:
 - When using CMake presets, the Project Status View now shows the build target along with the build preset. [PR #3241](https://github.com/microsoft/vscode-cmake-tools/pull/3241)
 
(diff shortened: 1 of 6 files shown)


gcampbell-msft:
@snehara99 Before approving, could you help me understand exactly what the change is by putting some screenshots in of before and after? 

The changes seem fine, but I want to make sure I understand it well before I approve. Thanks!

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 Features:
 - Added support for the CMake Debugger. [#3093](https://github.com/microsoft/vscode-cmake-tools/issues/3093)
 
+## 1.14.35
```
benmcmorran, line 7: I think you can merge this with the 1.15 section. I don't think we're planning to do another patch release before 1.15.

benmcmorran (dismissed):

CHANGELOG.md
snehara99: sounds good

gcampbell-msft (approved):

Issue #3146, opened 2023-04-27
Ctest Label Filter No Longer Available (1.14.29)


sjg10:
### Brief Issue Summary

Before the Test Explorer #3032 I would set my `ctestArgs` as follows:
```
    "cmake.ctestArgs": [
        "-L",
        "UnitTest"
    ]
```
This would ensure that only my tests labelled as UnitTest would be executed by "Run Ctest".

The TestExplorer runs all tests and does not seem to have a setting for filtering by label. Instead the arguments are passed to each test, so for a test we expect to filter TestExplorer attempts the following:
```
[proc] Executing command: /usr/bin/ctest -j10 -C Debug -T test --output-on-failure -L UnitTest -R ^ExampleIntegrationTest$
[ctest] Cannot find file: /home/.../build/DartConfiguration.tcl
[ctest]    Site: 
[ctest]    Build name: (empty)
[ctest] Test project /home/.../build
[ctest] Cannot find file: /home/.../build/DartConfiguration.tcl
[ctest] No tests were found!!!
[ctest] CTest finished with return code 0
```

This test is filtered but then appears as a failure in TestExplorer. Ideally there should be a way to configure TestExplorer to respect filter labels, or it should at least not mark this test as having failed.

gcampbell-msft:
@sjg10 Thanks for letting us know! We will mark them for our next release and get them on our backlog!

Sturla22:
I have a similar problem with the test explorer ignoring filters in CMakePresets.json
This means I cannot use my process of using the `CMake: Run Tests` command to run the tests I'm currently working on. I also tried adding my filter to the filter window in test explorer, but `CMake: Run Tests` ignores it and runs all tests.

khlebnikov:
I have a very similar problem, but perhaps a bit more involved.  My label filter is set in `settings.json` to:

```
    "cmake.ctestArgs": [
        "-L",
        "^${command:cmake.buildTargetName}$"
    ],
``` 

This way, I can select a particular test target to build (plus we have meta-targets that contain certain subsets of our thousands of tests) and run CTest on it without needing to remember updating multiple places in the workflow.  Please consider this scenario when making changes as well.

gcampbell-msft:
#3243 Should enable this again. Please open another issue if this does not resolve your issue! Thanks @xisui-MSFT!

PR #3243, opened 2023-07-19
Support CTest labels

xisui-MSFT:
Support CTest labels. This is implemented using test tags in VS Code testing API, i.e., filter in the search bar using `@cmake-tools.CTest:<label>`.

#3146 

image


src/ctest.ts
```diff
@@ -105,7 +105,7 @@ interface CTestInfo {
         backtrace: number;
         command: string[];
         name: string;
-        properties: { name: string; value: string }[];
+        properties: { name: string; value: string | string[] }[];
```
gcampbell-msft, line 108: Couldn't we simplify the logic farther down by simply making this just `string[]`? Then later we don't need the if condition.

gcampbell-msft (approved):
Besides my one comment about switching to just `string[]` instead of `string | string[]`, this lgtm.

src/ctest.ts
xisui-MSFT: It can be a string though?

xisui-MSFT: String[] doesn't include string

gcampbell-msft: Right, but wouldn't it make it simpler if this is just an array and if there is only one string it's just a single element array? 

Or is this user facing and we're trying to make this easier for the user to enter a simple string?

xisui-MSFT: This is output from "CTest --show-only=json-v1". User don't need to change it directly.

gcampbell-msft: Gotcha, if it's not user facing then I would say we could just have `string[]`, but if you think it's better to have `string | string[]`, that's good with me!

Approved.

xisui-MSFT: We don't really have a choice since we are just parsing the output of CTest, and that field can be either a string or an array of strings.

gcampbell-msft: Makes sense to me. Thanks for the clarification!

PR #3249, opened 2023-07-19
fix per folder provide browse configuration

### LINKED ISSUE #3155: Per-folder browse configurations incorrectly contain paths from other workspace folders
The [C/C++ extension API](https://github.com/microsoft/vscode-cmake-api) provides an way (`canProvideBrowseConfigurationsPerFolder` field) for a configuration provider to indicate that it can provide a separate 'browse configuration' per workspace folder in multi-root scenarios.  CMake Tools is providing a value of true for this, and is returning results from calls to `provideFolderBrowseConfiguration`.

We're seeing that CMake Tools incorrectly includes paths from workspace folder A in the browse paths for workspace folder B. 

This causes a problem with exclusions.  It's possible that workspace folder A may have excluded some subdirectories of its browse paths.  When those same paths get associated with another folder, those exclusions are bypassed (not defined by the other workspace folder), resulting in files being parsed into the browse database that should have been excluded.

gcampbell-msft:
We were returning the most recent stored browse configuration, which may be incorrect. Rather, we should return null if we don't have the browseConfiguration for the requested URI.

### This changes the implementation of `provideFolderBrowseConfiguration`

- Return `null` instead of a default / last stored `workspaceBrowseConfiguration`

## The purpose of this change

We were returning what effectively was the default or last stored `workspaceBrowseConfiguration` if we didn't have stored a browseConfiguration for that URI. This is incorrect as that default or last stored has no guarantee of actually applying. We instead return `null`.

Closes #3155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Features:
 
 Bug Fixes:
 - When using CMake presets, the Project Status View now shows the build target along with the build preset. [PR #3241](https://github.com/microsoft/vscode-cmake-tools/pull/3241)
+- Fix per-folder browse configurations returning incorrect information [#3155](https://github.com/microsoft/vscode-cmake-tools/issues/3155)
 
 Improvements:
 - Decreased the number of cases where we reconfigure erroneously upon usage of `cmake.getLaunchTargetPath` [#2878](https://github.com/microsoft/vscode-cmake-tools/issues/2878)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ Features:
 
 Bug Fixes:
 - When using CMake presets, the Project Status View now shows the build target along with the build preset. [PR #3241](https://github.com/microsoft/vscode-cmake-tools/pull/3241)
-- Fix per-folder browse configurations returning incorrect information [#3155](https://github.com/microsoft/vscode-cmake-tools/issues/3155)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Features:
 
 Bug Fixes:
 - When using CMake presets, the Project Status View now shows the build target along with the build preset. [PR #3241](https://github.com/microsoft/vscode-cmake-tools/pull/3241)
+- Fix per-folder browse configurations returning incorrect information [#3155](https://github.com/microsoft/vscode-cmake-tools/issues/3155)
```
sean-mcmanus, line 9: Our cpptools changelog always end changelog lines with a period, i.e. before the markdown link. I see that style followed in the line above but not below, so I'm not sure if you guys follow that same style or not.

sean-mcmanus (commented):
I don't understand the code base, but if it fixes the bug then that sounds good to me. I assume you got a repro from Colen.

snehara99:
I don't think the checks like that the return could be `null`. Does this check need to be bypassed or is there another option?

CHANGELOG.md
gcampbell-msft: Updated.

gcampbell-msft:
> I don't think the checks like that the return could be `null`. Does this check need to be bypassed or is there another option?

@snehara99 THis has been fixed, ping for review. Thanks!

snehara99 (approved):

PR #3252, opened 2023-07-21
localize descriptions and make names better

gcampbell-msft:
This improves the names of the properties available for the cmake debugger in launch.json. 

Also, localizes them.


package.nls.json
```diff
@@ -194,6 +194,10 @@
     },
     "cmake-tools.configuration.cmake.launchBehavior.description": "Controls what happens with the launch terminal when you launch a target.",
     "cmake-tools.configuration.cmake.automaticReconfigure.description": "Automatically configure CMake project directories when the kit or the configuration preset is changed.",
+    "cmake-tools.debugger.pipeName.description": "The named piped that the debugger will communicate with CMake with.",
```
benmcmorran, line 197: Awkward grammar. What about "Name of the pipe (on Windows) or domain socket (on Unix) to use for debugger communication."? (taken from the CMake docs)

benmcmorran (approved):

package.nls.json
gcampbell-msft: Changed. 👍

benmcmorran (approved):

package.json
```diff
@@ -722,27 +722,24 @@
             "label": "%cmake-tools.debugger.label%",
             "configurationAttributes": {
                 "launch": {
-                    "required": [
-                        "debuggerPipeName"
-                    ],
                     "properties": {
-                        "debuggerPipeName": {
+                        "pipeName": {
```
bobbrow, line 726: Is anyone using this feature yet?  You may want to include temporary support for the old names (with a log message to ask them to migrate) so that they don't get broken unexpectedly.  If this feature is only in pre-release, there's less of an impact, but it could smooth the experience out for early adopters.

gcampbell-msft: There are a small number of people, we have gotten some issues posted about it. @benmcmorran and I had talked about this and felt like since it's still in pre-release, we aren't making any gaurantees, so we wanted to do this before we have lots of users and especially before we get it into the official release.

PR #3253, opened 2023-07-22
Update kits.md naming of cmake-tools-kits.json

r-downing:
Example config filename seems to be wrong.

- Updates the `docs/kits.md` naming from `cmake-kits.json` to `cmake-tools-kits.json`

## The purpose of this change

I believe the example naming was incorrect - when I tried creating a project-level `.vscode/cmake-kits.json` it was not detected, and vscode showed errors about not being able to load the json schema. Adding the "tools-" makes it work as expected.

### DIFF
--- a/docs/kits.md
+++ b/docs/kits.md
@@ -32,9 +32,9 @@ You can manually edit this file to define new global kits, however the contents 
 
 ### Project kits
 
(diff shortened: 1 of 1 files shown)


r-downing:
On second thought - `.vscode/cmake-kits.json` is working but getting the json schema error, while `.vscode/cmake-tools-kits.json` schema is recognized, but it doesn't actually work. May have a separate issue with the json schema detection.

Issue #1842, opened 2021-05-06
CMake cache empty string variables are ignored in presets


andreeis:
Opening from comment in issue https://github.com/microsoft/vscode-cmake-tools/issues/1815:

> Another related issue here is that cache entries which are just an empty string also don't get created, they are ignored and not added to the cmake configure command line.

MathiasMagnus:
@bobbrow This is actually a blocker, because the variable [`CMAKE_MSVC_DEBUG_INFORMATION_FORMAT`](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_DEBUG_INFORMATION_FORMAT.html) is meant to be set to an empty variable if we want to avoid CMake populating the `<DebugInformationFormat>` field of the MSBuild project files with `ProgramDatabase`. Not being able to set this variable to an empty string (when CMake 3.25+ is used, even when policy [CMP0141](https://cmake.org/cmake/help/latest/policy/CMP0141.html) is set to OLD) causes configuration errors when trying to compile HIP code using AMD's HIP extension. (Their platform toolset updates parts of the vcxproj schema which don't correspond to anything CMake understands at the moment. The escape hatch is telling CMake not to insert anything into `try_compile()` commands, just go with the project defaults. This is the part that breaks.)

andreeis:
Fix is merged in and should be included in the 1.15 release.

PR #3264, opened 2023-07-27
Fix for 1842: don't ignore empty string cache variables

andreeis:
Fix for request https://github.com/microsoft/vscode-cmake-tools/issues/1842: don't ignore empty string cache variables when configuring from presets.
I checked and when using settings and kits, such empty cache variables are already seen and passed to the cmake configure command, no extra fix was needed.


gcampbell-msft (approved):

andreeis (commented):
Approving myself since now I only fixed a merge conflict in changelog

snehara99 (approved):

Issue #1914, opened 2021-06-02
Variable substitution stopped working in version 1.7


davidmneto:
### Brief Issue Summary

I am having an issue similar to the one described in https://github.com/microsoft/vscode-cmake-tools/issues/1526. I am trying to set the **cmake.sourceDirectory** using the variable **${buildKit}** as follows:

`    "cmake.sourceDirectory": "${workspaceFolder}/project_1/${buildKit}",`

I cannot configure the cmake project and I receive the error shown below.

I was able to later solve this by installing the previous version of cmaketools version. 1.6.0, and the substitution worked fine. I believe something broke in the transition to 1.7 

### CMake Tools Log

```
[expand] Invalid variable reference ${buildKit} in string: ${workspaceFolder}/project_1/${buildKit}
[rollbar] "sourceDirectory" is not a directory {"sourceDirectory":"/workspaces/copateneida/project_1/${buildKit}"}
```

### Platform and Versions

- **Operating System**: Ubuntu 20.04
- **CMake Version**: 1.7.3

bobbrow:
Are you using `CMakePresets.json` to configure your project?  Or are you using Kits and Variants?

davidmneto:
I believe I am using kits. I have configured the kits in a separate cmake-kits.json file. And I am setting the cmake.sourceDirectory in the settings.json

bobbrow:
I see what happened.  [This commit](https://github.com/microsoft/vscode-cmake-tools/commit/f69cfabfbe4cac7f46b9d0bc2aa529a68aaa39b2) disabled using certain variables in `cmake.sourceDirectory`.  @andreeis can explain further.  Is there a way we can make this work again?

https://github.com/microsoft/vscode-cmake-tools/blob/c1f11cb574bfb049e07d452e31ea950c6bba0533/src/extension.ts#L437-L438

andreeis:
Yes I can fix this. The original mistake on my part was that I was focused on scenarios where we want to identify if a project is or not a CMake project and that needs to happen before that annoying popup that we had about setting a kit. So I thought any kit information that is missing shouldn't block our analysis of having or not having a CMakeLists.txt (via "cmake.sourceDirectory"). But I didn't realize that if we do have a kit set then we need to expand any variables kit related. Quick fix.

lygstate:
Hwo to fix?

ToBoMi:
[...]

PR #3268, opened 2023-07-27
Fix1914 - Variable substitution stopped working with release 1.7

andreeis:
Fix1914 - Variable substitution stopped working with release 1.7, related to sourceDirectory and kit info variables.
https://github.com/microsoft/vscode-cmake-tools/issues/1914

Will add explanations later, do more cleanup and also ensure more variables are considered. For now, just kit name and vendor, two examples that are different and exist in different data structures. Creating the PR to see tests outcome.


src/presetsController.ts
```diff
@@ -42,12 +43,26 @@ export class PresetsController {
         const presetsController = new PresetsController(project, kitsController, isMultiProject);
         const expandSourceDir = async (dir: string) => {
             const workspaceFolder = project.workspaceFolder.uri.fsPath;
+            const kit = project.getActiveKit();
```
gcampbell-msft, line 46: This part worries me, if we're in the presetsController then don't we not care about kits? Since it's either kits or Presets? Not both?

src/drivers/cmakeDriver.ts
```diff
@@ -390,12 +391,28 @@ export abstract class CMakeDriver implements vscode.Disposable {
         return { vars, variantVars };
     }
 
-    static sourceDirExpansionOptions(workspaceFolderFspath: string | null): expand.ExpansionOptions {
+    static async sourceDirExpansionOptions(workspaceFolderFspath: string | null, kit?: Kit | null, generatorName?: string | null): Promise<expand.ExpansionOptions> {
         const ws_root = util.lightNormalizePath(workspaceFolderFspath || '.');
 
         // Fill in default replacements
+        if (!kit) {
```
gcampbell-msft, line 398: Same worry here, I'm not sure we want to be using kits every time the cmakeDriver uses this method, since couldn't a preset use this method?

src/presetsController.ts
andreeis: I didn't step back and realize this obvious fact. I just noticed errors and complains in my repro and just went head first into making those go away. I'll rethink a bit and get back with an update. Very good essential point.

andreeis:
The last PR update does the following:
- error when presets are on and user still has kit variables to expand. we don't expand even if, with presets off, we have all info available saved in the project state
- fix wrong assumption in the previous approach: we can't expand the generator name variable in sourceDir. Generator deduction depends on configure which depends on sourceDir. So we error if sourceDir contains generator variable.

PR #3269, opened 2023-07-27
source-map dependency isn't needed, runtime loc wasn't working

gcampbell-msft:
This should fix our runtime loc. We weren't pulling in the vscode-nls-dev/lib/webpack-loader. In order for this to work, I had to update the source-map dependency.


gcampbell-msft:
Blocking comment because I should test this vsix. I may have noticed a bug or two due to these changes.

PR #3270, opened 2023-07-29
Update project outline on configure

vlavati:
### This changes visible behavior of project outline

- the project outline should contain projects/targets for all projects in multiroot workspaces
- it should be possible to run/debug target from project outline in multiroot workspace

## Other Notes/Information
On the image below there is a project outline for the "multi-root-UI". It contains projects for project-folder-2/subroot1, project-folder-2/subroot2, and project-folder-2.

изображение

If commands (Build/Launch) are called from context menu for the targets in multiroot workspace, it is needed to provide source directory of the project to choose correct project to execute cmake commands. (By default there is used active project)


vlavati:

src/extension.ts
```diff
@@ -1044,32 +1030,30 @@ export class ExtensionManager implements vscode.Disposable {
...
-                for (const project of projects) {
-                    // Choose the active project.
-                    if (activeProject?.folderPath === project.folderPath) {
-                        return project;
-                    }
-                }
-                // Choose the first project folder.
-                return projects[0];
+                // Choose project by corresponding source directory
+                return projects.find(project => sourceDir && (path.normalize(sourceDir) === path.normalize(project.folderPath))) ??
```
gcampbell-msft, line 1042: Assuming this is all one large return statement, could you correct the indentation?

src/projectOutline.ts
```diff
@@ -546,7 +561,15 @@ export class ProjectOutline implements vscode.TreeDataProvider<BaseNode> {
         }
 
         const updates: BaseNode[] = [];
-        existing.updateCodeModel(model, { ...ctx, nodesToUpdate: updates, folder });
+        existing.updateCodeModel(
```
gcampbell-msft, line 564: Am I reading this right and this still has 3 parameters even though the method signature only has two parameters now?

gcampbell-msft (changes requested):
Overall it looks pretty good, I just have a couple of questions and changes to be made.

src/projectOutline.ts
vlavati: It is called WorkspaceFolderNode::updateCodeModel, which has three parameters now.

src/extension.ts
vlavati:
```
                // Choose project by corresponding source directory
                return projects.find(project => sourceDir && (path.normalize(sourceDir) === path.normalize(project.folderPath)))
                    // Choose project by folder of active project
                    ?? projects.find(project => activeProject?.folderPath === project.folderPath)
                    // Fallback to first project
                    ?? projects[0];
```
Can I indent this statement that way?

vlavati:
Thanks for review, I hope it will be useful.

gcampbell-msft:
@vlavati Thanks for the PR! I plan to test on Monday to confirm no breakages and then approve. Thanks!

gcampbell-msft:
@benmcmorran Do you think the Windows test failure is due to these changes? It seems to be failing reliably but the failure itself is a "Resource is busy" error, which doesn't immediately strike me as something wrong with the code. Thanks!

benmcmorran:
@gcampbell-msft I just queued a test run for https://github.com/microsoft/vscode-cmake-tools/pull/3336. If that one passes then I suspect changes in this PR are causing the failure.

benmcmorran:
@gcampbell-msft Saw the same failure in the other PR, so I don't think we should consider it blocking here (but we should independently investigate why it started failing).

gcampbell-msft:
@vlavati Hi! I've tested and all of the changes look good! One final thing before merging, could you please add a `CHANGELOG.md` entry? 

Once that is done we can merge!

vlavati:
@gcampbell-msft , @benmcmorran Regarding the failed tests on Windows, I suppose that it is because of the Terminal which blocks a current working directory. Some tests run "await cmakeProject.launchTarget();", which starts a terminal and a compiled target. But I did not find a way to check the exit status or wait until the Terminal is closed.

gcampbell-msft:
Merging. I'm bypassing the failing Windows tests because we're seeing the same failure on other PRs and it's not related to this change.

gcampbell-msft (approved):

PR #3277, opened 2023-08-03
Add script mode and ability to connect to externally launched CMake process.

gcampbell-msft:
### This updates the options that are available in launch.json for the `cmake` debug type.

- Add required `cmakeDebugType` to the launch.json that defines whether the user is trying to debug under the three available modes: `configure`, `external`, `script`.
- Add the ability to debug externally launched CMake processes, or to debug a CMake script.
- Adds an activation event when a debug session of type `cmake` is initiated. 
- Remove the check for whether `cmakeDebugType` is `external` regarding dapLog. We can simply assign it unconditionally, knowing that it will be ignored in the `external` mode. In `script` and `configure` it will be used as normal.
- Adds the ability to run and debug CMake script files when there is no launch.json present through the "Run and Debug" button.
- Adds initial snippets that help people create launch.json configurations.

## The purpose of this change is to add the ability to debug externally launched CMake processes or debug Cmake scripts.


gcampbell-msft:
NOTE: This should not be merged before we release 1.15.

Neumann-A:
I got:
image
with the script stuff. 

External debugging worked as expected. 

Configure debugging still cannot be stopped if the pipe never connects. This only happens if you trick code to use a different CMake which only reacts like CMake without the debugger; e.g. set `"cmake.cmakePath"` to a batch file with the following:
```
@ECHO OFF
SETLOCAL EnableDelayedExpansion

set call_cmake=1
REM Call CMake always except if the debugger should be attached
FOR %%a IN (%*) do (
    set "_arg_=%%a"
    if !_arg_!==--debugger set call_cmake=0
)

@ECHO OFF
if !call_cmake!==1 cmake.exe %* && Exit /B %ERRORLEVEL%
```

docs/debug.md
```diff
@@ -0,0 +1,48 @@
...
+  * required
+  * optional
+    * `pipeName` - Name of the pipe (on Windows) or domain socket (on Unix) to use for debugger communication.
+    * `clean` - Clean prior to configuring.
+    * `configureAll` - Configure for all projects.
+    * `dapLog` - Where the debugger DAP log should be logged.
+* `external`
+  * required
+    * `pipeName` - Name of the pipe (on Windows) or domain socket (on Unix) to use for debugger communication.
+  * optional
```
Neumann-A, line 38: can this get an optional ``command` : []` field/array? So that VS code executes the command I would normally use to start the cmake pipe externally?

gcampbell-msft:
> I got: image with the script stuff.
> 
> External debugging worked as expected.
> 
> Configure debugging still cannot be stopped if the pipe never connects. This only happens if you trick code to use a different CMake which only reacts like CMake without the debugger; e.g. set `"cmake.cmakePath"` to a batch file with the following:
> 
> ```
> @ECHO OFF
> SETLOCAL EnableDelayedExpansion
> 
> set call_cmake=1
> REM Call CMake always except if the debugger should be attached
> FOR %%a IN (%*) do (
>     set "_arg_=%%a"
>     if !_arg_!==--debugger set call_cmake=0
> )
> 
> @ECHO OFF
> if !call_cmake!==1 cmake.exe %* && Exit /B %ERRORLEVEL%
> ```

This may be because I haven't done any handling of variables. To confirm, could you attempt it again by putting either a relative or full path?

docs/debug.md
gcampbell-msft: To help me think through the possible design implications, could you go into more detail regarding your use case for the `external` mode? 

What kind of external processes are you expecting? It is a CMake invocation? Or is it invoking something else that internally invokes CMake?

Thanks!

Neumann-A: > It is a CMake invocation? Or is it invoking something else that internally invokes CMake?

Could be both of those. 

> What are you expecting or hoping to use in the "command" field?

For example I could write a (CMake) script (not the debug target) which does the following:
- FetchContent vcpkg
- bootstrap vcpkg
- vcpkg install (with some additional options to pass on the debugger pipe info)
- vcpkg internally calls `cmake -P` which could be debugged
- vcpkg internally calls `cmake` in configure mode which could be debugged

The intention of having a `command` field is to automate the process. E.g. 
`"command" : ["cmake", "-P", "my-internal-cmake-caller-script.cmake", "-DDEBUGGER_PIPE_NAME=<name_of_debugger_pipe>"] `
`DEBUGGER_PIPE_NAME` is internally passed on to some process calling cmake. (doesn't really matter what it is)

But if "command" is added "workingDir" probably also makes sense to define where the command is executed.

Neumann-A:
> Could you please try it again?

If you give me a new `vsix`, then yes. One problem I have is that code always tries to reinstall the current release extension.

docs/debug.md
Neumann-A: That being said I am also happy with the current approach. It is just that it takes more manual labor for startup.

gcampbell-msft: Understood. I wonder if using a `preLaunchTask` is what you want. This way the pre launch task can be fully arbitrary. Do you think using that would suit your needs?

Neumann-A: Don't know about that. If there is a possibility to detach then probably yes if it needs to return than probably no. Need to read up on that first.

gcampbell-msft: I just tested and found that if you set up a tasks.json task with the following settings as part of that task: 

```
"isBackground": true,
        // This task is run before some debug tasks.
        // Problem is, it's a watch script, and since it never exits, VSCode
        // complains. All this is needed so VSCode just lets it run.
        "problemMatcher": [
            {
            "pattern": [
                {
                "regexp": ".",
                "file": 1,
                "location": 2,
                "message": 3
                }
            ],
            "background": {
                "activeOnStart": true,
                "beginsPattern": ".",
                "endsPattern": ".",
            }
            }
        ]
```
then you can successfully have a preLaunch task.

Neumann-A: Hmm I need some delay or some retry count to connect the debugger to the pipe since the debugger wants to start before the prelaunch task reached the point where the pipe is opened.

Brian-Ruff-CardinalPeak:
This PR was a timely discovery for me; I need to debug a cmake issue with several subprocess invocations of cmake.  Is there an existing mechanism to get .vsix files for this branch?  I was able to attach the debugger to the initial cmake process via configure/reconfigure with CMake debugger, but am not yet having success with an external to attach to a subprocess.  I pulled the branch and was able to create a .vsix - I haven't done much with VSCode extensions at all.

Brian-Ruff-CardinalPeak:
One item I noted, it seems like an extension can provide seed/snippets via the Add Configuration UI, though I'm not familiar with the mechanism/API.

docs/debug.md
Neumann-A: OK I think there is nothing to do here. I can just define a task and let it run via VS Code directly instead of manually typing it. Never used task since I mainly use CMake presets. Would be nice if there is a way to somehow get the pipename from a task.json via some variable.

gcampbell-msft:

docs/debug.md
```diff
@@ -0,0 +1,48 @@
+# CMake Debugging
+
+Starting with CMake 3.27, debugging CMake is now supported in CMake Tools!
+
+The following documentation will help you understand the various ways you can debug CMake.
```
benmcmorran, line 5: General comment for this document -- consider rewording "debug CMake" to "debug CMake scripts and cache generation" or similar to avoid confusion with debugging actual the CMake executable, which is also technically possible but requires a C++ debugger.

docs/debug.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Debugging from CMake Tools UI entry points
+
+The most common reason to debug CMake is to debug CMake Configuration! There are many ways that you can accomplish this:
+
+* Commands
+  * CMake: Configure with CMake Debugger
+  * CMake: Delete Cache and Reconfigure with CMake Debugger
+* Folder Explorer
+  * Right click on CMakeLists.txt context menu entry.
```
benmcmorran, line 15: "context menu entry" feel redundant right after "right click". Can we instead tell the user what to actually click on?

docs/debug.md
```diff
@@ -0,0 +1,48 @@
...
+* Project Outline
+  * Right click on CMakeLists.txt context menu entry.
+  * Expand the "..." in the project outline. There is an entry to use the Debugger!
+
+## Debugging from CMake Tools launch.json
```
benmcmorran, line 20: I'd cut "CMake Tools" from this header. `launch.json` is a general VS Code file, not a CMake Tools file.

docs/debug.md
```diff
@@ -0,0 +1,48 @@
...
+
+CMake Tools now provides a new debug type `cmake`!
```
benmcmorran, line 22: Here and above at the beginning of the document, I don't think we need to say "now". That implies a point-in-time comparison with an older version of the extension, which would make sense in an announcement like a blog post, but this is persistent documentation.

docs/debug.md
```diff
@@ -0,0 +1,48 @@
...
+
+The `cmake` debug type supports three different types of `cmakeDebugType`: `configure`, `external`, `script`. They each come with their own settings that can be used to modify and control the debug session!
+
+Listed below are the settings that are available for each configuration based on `cmakeDebugType`:
+
+* `configure`
+  * required
```
benmcmorran, line 29: For clarity consider `* required: none`.

docs/debug.md
```diff
@@ -0,0 +1,48 @@
...
+    * `clean` - Clean prior to configuring.
+    * `configureAll` - Configure for all projects.
+    * `dapLog` - Where the debugger DAP log should be logged.
+* `external`
+  * required
+    * `pipeName` - Name of the pipe (on Windows) or domain socket (on Unix) to use for debugger communication.
+  * optional
+* `script`
+  * required
+    * `scriptPath` - The path to the script to debug.
```
benmcmorran, line 41:
```suggestion
    * `scriptPath` - The path to the CMake script to debug.
```

I'd make the same edit to all references to "script" in this section of the docs. We should be explicitly that by "script" we mean a CMake script, not the more general notion of an arbitrary Bash/Python/PowerShell/whatever script.

package.json
```diff
@@ -737,14 +777,140 @@
...
+                },
+                {
+                    "label": "%cmake-tools.debugger.script.snippet.label%",
+                    "description": "%cmake-tools.debugger.script.snippet.description%",
+                    "body": {
+                        "type": "cmake",
+                        "request": "launch",
+                        "name": "%cmake-tools.debugger.script.snippet.body.name%",
+                        "cmakeDebugType": "script",
+                        "scriptPath": "^\"\\${workspaceFolder}/<...>\""
```
benmcmorran, line 899: What's the `^` for? Does this act as a regex? If it's not, we can probably hardcode something like `<script>.cmake` even though it won't exist just to make the example clearer.

src/debug/debugConfigurationProvider.ts
```diff
@@ -0,0 +1,58 @@
+import * as vscode from "vscode";
+import * as nls from "vscode-nls";
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+export class DebugConfigurationProvider implements vscode.DebugConfigurationProvider {
+    resolveDebugConfiguration(_folder: vscode.WorkspaceFolder | undefined, debugConfiguration: vscode.DebugConfiguration, _token?: vscode.CancellationToken | undefined): vscode.ProviderResult<vscode.DebugConfiguration> {
+        if (!debugConfiguration.type && !debugConfiguration.request && !debugConfiguration.name) {
```
benmcmorran, line 9: Does this let you use the active editor as a CMake script playground without manually writing a launch configuration? That's pretty cool.

src/debug/debuggerScriptDriver.ts
```diff
@@ -0,0 +1,64 @@
...
+
+            const result = await child.result;
+            if (result.retc === 0) {
+                cmakeLogger.info(localize('run.script.successful', "CMake script: \"{0}\" completed successfully.", scriptPath));
+            } else {
+                cmakeLogger.info(localize('run.script.failed', "CMake script: \"{0}\" completed unsuccessfully.", scriptPath));
+                throw new Error("HEY");
+            }
+        } else {
+            cmakeLogger.error(localize('run.script.cmakeDebugger.not.supported', "Cannot debug a script with this version of CMake, ensure your version of CMake supports debugging"));
```
benmcmorran, line 61: May want to explicitly refer to version 3.27 or later in this error message.

benmcmorran (approved):

docs/debug.md
```diff
@@ -0,0 +1,48 @@
+# CMake Debugging
```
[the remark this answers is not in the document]
gcampbell-msft: Making the change.

docs/debug.md
```diff
@@ -0,0 +1,48 @@
...
+
+Starting with CMake 3.27, debugging CMake is now supported in CMake Tools!
```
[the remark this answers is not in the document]
gcampbell-msft: Removing all occurrences of '!'.

docs/debug.md
gcampbell-msft: Changing.

docs/debug.md
gcampbell-msft: Modifying.

docs/debug.md
gcampbell-msft: Removing.

docs/debug.md
gcampbell-msft: That makes sense. Changing.

docs/debug.md
```diff
@@ -0,0 +1,48 @@
...
+
+Listed below are the settings that are available for each configuration based on `cmakeDebugType`:
+
+* `configure`
+  * required
+  * optional
+    * `pipeName` - Name of the pipe (on Windows) or domain socket (on Unix) to use for debugger communication.
+    * `clean` - Clean prior to configuring.
+    * `configureAll` - Configure for all projects.
+    * `dapLog` - Where the debugger DAP log should be logged.
```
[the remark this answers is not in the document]
gcampbell-msft: Changing.

docs/debug.md
gcampbell-msft: Changing.

docs/debug.md
```diff
@@ -0,0 +1,48 @@
...
+  * Right click on CMakeLists.txt context menu entry.
+  * Expand the "..." in the project outline. There is an entry to use the Debugger!
+
+## Debugging from CMake Tools launch.json
+
+CMake Tools now provides a new debug type `cmake`!
+
+The `cmake` debug type supports three different types of `cmakeDebugType`: `configure`, `external`, `script`. They each come with their own settings that can be used to modify and control the debug session!
+
+Listed below are the settings that are available for each configuration based on `cmakeDebugType`:
```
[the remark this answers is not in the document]
gcampbell-msft: That makes sense, I'll include examples for cmake script debugging and external connections, I don't see it being common that users use a launch.json for debugging configuration.

package.json
```diff
@@ -723,9 +726,46 @@
...
+                            "type": "array",
+                            "items": {
+                                "type": "object",
+                                "properties": {
+                                    "name": {
+                                        "type": "string",
+                                        "description": "%cmake-tools.debugger.name%"
+                                    },
+                                    "value": {
+                                        "type": "string",
```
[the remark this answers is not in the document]
gcampbell-msft: What do you think?

I can see the reason for it, maybe someone wants to ensure that an environment variable isn't set before configuring, or something like that, but I'm not sure how common that would be. 

Couldn't someone just set it to an empty string?

package.json
```diff
@@ -723,9 +726,46 @@
             "configurationAttributes": {
                 "launch": {
                     "properties": {
+                        "scriptPath": {
+                            "type": "string",
+                            "descripttion": "%cmake-tools.debugger.scriptPath.description%",
+                            "default": ""
```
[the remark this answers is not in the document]
gcampbell-msft: I think that makes sense, I'll modify.

package.json
gcampbell-msft: I don't think so. I copied it from somewhere and it seemed to simply allow us to insert it without resolving the ${workspaceFolder} to an actual value. 

When I created an automatic debug configuration using this, this is what appeared: 
image

src/debug/debugConfigurationProvider.ts
gcampbell-msft: Yes! That's exactly right. This allows you to click the "Run and Debug" button when there are no launch configurations. 

DebugCMakeScript

Now that you mention it, it actually gives me another idea. This entry point isn't available, to my knowledge, if you do have launch configurations. So, I could add either a better entry point in the debug dropdown, or even a context menu entry point on files that is present only in .cmake files. I might look into this before merging.

gcampbell-msft: CMake_AutoDebugConfig

package.json
gcampbell-msft: @benmcmorran ping for review.

Specifically, https://github.com/microsoft/vscode-cmake-tools/pull/3277#discussion_r1298817459

package.json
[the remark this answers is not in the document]
benmcmorran: Depending on the platform, empty string and unset may be distinct states for an environment variable. Bash example:

```bash
$ if [[ -z ${foo+x} ]]; then echo true; else echo false; fi
true
$ export foo="bar"
$ if [[ -z ${foo+x} ]]; then echo true; else echo false; fi
false
$ export foo=""
$ if [[ -z ${foo+x} ]]; then echo true; else echo false; fi
false
$ unset foo
$ if [[ -z ${foo+x} ]]; then echo true; else echo false; fi
true
```

Neumann-A:
@gcampbell-msft Is this waiting on something?

gcampbell-msft:
> @gcampbell-msft Is this waiting on something?

We are pending another code review. Our team hopes to get on this in the next week or two, thank you for the ping!

package.json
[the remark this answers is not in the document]
gcampbell-msft: I think that's a good idea, but I think it may require a couple of changes elsewhere in the code, so we'll leave it for either a different PR or if users request it.

benmcmorran (dismissed):

benmcmorran (dismissed):

benmcmorran (approved):

gcampbell-msft:
Merging without tests because they passed prior to the changelog changes, and the changelog doesn't affect tests. 
The Windows tests fail due to other changes, they are not due to this PR.

PR #3292, opened 2023-08-10
Github Action to Close Stale Issues

snehara99:
This bot should:
- mark issues as "stale" if there has been no activity on it for 60 days with a message.
- remove the "stale" label from open issues if there is new activity.
- close issues as "not_planned" that have been marked as "stale" after 7 days with a message.

Currently the bot is set to run every hour since the max number of operations per run is 50 (to not exceed Github's API call rate limit). After a week or two, I will create another PR that updates this bot to running once a day.


snehara99:
I would appreciate it if people could double-check that I used `env` variables properly.

snehara99:
Also, this doesn't require a `CHANGELOG.md` update, right?

bobbrow:
If you remove "schedule" from the yml, you can still manually run the bot to test from the "Actions" tab it because you have "workflow_dispatch."  When you've confirmed it runs properly, you can enable the daily schedule again.  I'm not sure that you want to run it hourly.

snehara99:
> If you remove "schedule" from the yml, you can still manually run the bot to test from the "Actions" tab it because you have "workflow_dispatch." When you've confirmed it runs properly, you can enable the daily schedule again. I'm not sure that you want to run it hourly.

Good idea, thanks!

bobbrow:
In 67 days this is going to close a lot of issues.  Are you sure that's what you want?  Previously we only allowed automatic closing of issues marked as "question" or "more info needed" to avoid closing bugs and feature requests by mistake.

gcampbell-msft:
> In 67 days this is going to close a lot of issues. Are you sure that's what you want? Previously we only allowed automatic closing of issues marked as "question" or "more info needed" to avoid closing bugs and feature requests by mistake.

@snehara99 This is a good point. Maybe we only want to mark stale on items that have "more-info-needed"? This would mean that it's stale due to customers not giving feedback, as compared to situations where bugs don't have action because we haven't responded?

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    env:
+      stale-days: 60
+      close-days: 7
+      stale-label: stale
+    steps:
+    - uses: actions/stale@v4.1.1
+      with:
+        stale-pr-label: $stale-label
+        exempt-issue-labels: stale-exempt
```
gcampbell-msft, line 21: Is this up-to-date? This field is looking for labels to see if we can exempt them from marking them as stale, I don't think we currently use a `stale-exempt` label.

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,30 @@
...
+  stale:
+    runs-on: ubuntu-latest
+    env:
+      stale-days: 60
+      close-days: 7
+      stale-label: stale
+    steps:
+    - uses: actions/stale@v4.1.1
+      with:
+        stale-pr-label: $stale-label
```
gcampbell-msft, line 20: Do we want to mark PR's stable as well? If not, I doubt we need to define this variable.

snehara99:
> > In 67 days this is going to close a lot of issues. Are you sure that's what you want? Previously we only allowed automatic closing of issues marked as "question" or "more info needed" to avoid closing bugs and feature requests by mistake.
> 
> @snehara99 This is a good point. Maybe we only want to mark stale on items that have "more-info-needed"? This would mean that it's stale due to customers not giving feedback, as compared to situations where bugs don't have action because we haven't responded?
> 
> However, I do like the idea of closing items that we've avoided for more than 2 months. This indicates that we either don't have time or don't plan to work on it. If the issue is persistent enough users can re-create an issue or we can investigate closed bugs. @bobbrow Thoughts?

@bobbrow  @gcampbell-msft  I do think this warrants a discussion. Maybe at our next sync. However I thought we already had a bot that closed stale "more info needed" issues. Maybe that bot is a relic and doesn't work anymore? I'll investigate

.github/workflows/close-stale-issues.yml
snehara99: You're right, It should be `stale-issue-label`. fixing

.github/workflows/close-stale-issues.yml
snehara99: we don't currently have a `stale-exempt` label, but I thought we could start using it. Is there another label you'd recommend?

bobbrow:
Yes, we had bots that did this, but I disabled them because they stopped working.

gcampbell-msft:
@snehara99 Am I reading this right that the "old stale" bot will mark as stale after year and then 14 days later close it?

bobbrow:
The old bots only closed issues marked as "more info needed" and "question" and these were manually assigned tags.  I think it's risky to let a bot add the tags and do the automated closing.  You have issues older than 1 year in here that I don't think you want to close wholesale.

.github/workflows/close-stale-issues.yml
snehara99: fixed

snehara99:
> @snehara99 Am I reading this right that the "old stale" bot will mark as stale after year and then 14 days later close it?

now it's 30 days later but yup

snehara99:
> The old bots only closed issues marked as "more info needed" and "question" and these were manually assigned tags. I think it's risky to let a bot add the tags and do the automated closing. You have issues older than 1 year in here that I don't think you want to close wholesale.

After the each of labelling is done every month for the oldest issues, I'm going to review each `stale-old` labelled issue manually and exempt them if needed. I'll keep up this routine for the next year. For all future issues we'll add an exempt label if we want.

.github/workflows/close-old-issues.yml
```diff
@@ -0,0 +1,29 @@
...
+      readonly:
+        description: "readonly: Specify true or 1 to prevent changes from being commited to GitHub"
+        default: false
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    env:
+      stale-days: 720
+      close-days: 30
+      stale-label: stale-old
```
bobbrow, line 14: should this be `stale` instead of `stale-old`?

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,33 @@
...
+        stale-issue-label: $stale-label
+        exempt-issue-labels: $stale-exempt-label
+        only-label: $only-label
+        days-before-pr-stale: -1
+        days-before-pr-close: -1
+        days-before-stale: $stale-days
+        stale-issue-message: "This issue is now marked as '$stale-label' due to there being no activity on it for the past $stale-days days and being labelled '$only-label'. Unless the '$stale-label' label is removed or the issue is commented on, this will be closed in $close-days days. If you would like to make this issue exempt from getting stale, please remove the 'more info needed' label or add the '$stale-exempt-label' label"
+        days-before-close: $close-days
+        close-issue-reason: not_planned
+        close-issue-message: "This issue is now closed due to there being no activity on it for the past $close-days days since being marked as 'stale'."
```
bobbrow, line 31: Replace `'stale'` with `'$stale-issue-label'`?

bobbrow:
> @bobbrow could you also confirm that I have it set up correctly to manually run this bot? thanks!

I think so.  Go to the Actions tab and run to run it in read-only mode.

I still have concerns about automatically closing anything over 2 years old, but I suppose you can reopen issues and mark them as exempt if you get complaints.

snehara99:
> > @bobbrow could you also confirm that I have it set up correctly to manually run this bot? thanks!
> 
> I think so. Go to the Actions tab and try to run it in read-only mode.
> 
> I still have concerns about automatically closing anything over 2 years old, but I suppose you can reopen issues and mark them as exempt if you get complaints.

I removed the auto-close so this shouldn't be an issue anymore :) It just auto marks as stale-old, and i'll review them manually

.github/workflows/close-stale-issues.yml
snehara99: fixed

benmcmorran (approved):

PR #3304, opened 2023-08-19
Add an option to use a custom compile commands filename

TingPing:

- Add a `compileCommandsFilename` config option.

## The purpose of this change

I use a project that generates a custom compile commands database so this allows
this extension to load that file.

## Other Notes/Information

Tested on a real project including changing the value at runtime.


TingPing:

TingPing:
Closing this as I ended up using a different solution. I'll leave the branch around if it's useful to others.

PR #3344, opened 2023-09-15
add workflow on release that closes fixed-pending-release issues

gcampbell-msft:
This adds a workflow that is invoked when a new release is published. It will search through all issues and it will close those that are marked `fixed-pending-release` as well as leave a comment indicating that it was closed.


.github/workflows/closed-fixed-pending-release.yml
```diff
@@ -0,0 +1,14 @@
...
+on:
+    release:
+        types: [published]
+
+jobs:
+    close_fixed_pending_release:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Close issues marked 'fixed-pending-release'
+              uses: gcampbell-msft/fixed-pending-release@0.0.12
```
benmcmorran, line 11: Is it possible to host the action directly in this repository, or move it to the `microsoft` org? I'd like to avoid permissions issues if we need to modify this action in the future.

gcampbell-msft: You mean the actual GitHub Action that's published to the marketplace? 

I don't think we'd want to host it directly in this repository, since it's possible for it to be used in other repositories such as the makefile extension. What is the process / requirements for moving it to the `microsoft` org?

benmcmorran: You'd need to follow [this OSS release process](https://docs.opensource.microsoft.com/releasing/). If the code is small enough, it will likely qualify for expedited approval, or no formal approval at all.

gcampbell-msft: For now, we'll leave it as is. Once we confirm that it suits our needs and works well, we'll investigate this. Thanks.

Issue #3321, opened 2023-08-31
Error for pre-release v1.16.3: TypeError: Cannot destructure property 'folder' of 'undefined' as it is undefined. {}


tophoo:
### Brief Issue Summary

When using `cmake.getLaunchTargetFilename` inside an `input` context of a task, e.g.:
```
{
    "inputs": [
        {
            "command": "cmake.getLaunchTargetFilename",
            "type": "command",
            "id": "cmake-target-filename"
        }
    ],
    "tasks": [
        {
            "command": "echo ${input:cmake-target-filename}",
            "label": "run task with bug",
            "type": "shell",
            "problemMatcher": []
        },
    ]
}
```
the error
```
Unhandled exception: Unhandled Promise rejection: getLaunchTargetFilename TypeError: Cannot destructure property 'folder' of 'undefined' as it is undefined. {}
```
occurs. With v1.15 everything works fine. The error was probably introduced with #2801. We rely on getting various cmake commands as input for other tasks.

gcampbell-msft:
@tophoo Thanks for reporting the issue! We will attempt to reproduce and investigate this issue.

PR #3348, opened 2023-09-18
Fix3321 - Use empty to prevent destructure of undefined

vlavati:
## This change addresses item #3321

- if command is used without arguments, the empty object is used to use destructuring without exception

## The purpose of this change

This change fixes the bug.

## The reproducing
task


vlavati:
Hi, @gcampbell-msft . It is updated. Could you review? Thanks in advance.

gcampbell-msft (approved):

PR #3354, opened 2023-09-24
Impl3137 - SidePanel: Add buttons for project

vlavati:

- add configure,build,clean buttons on SidePanel for Project item
- add clean icons


src/extension.ts
```diff
@@ -1101,11 +1101,11 @@ export class ExtensionManager implements vscode.Disposable {
         return this.runCMakeCommandForAll(cmakeProject => cmakeProject.cleanConfigureWithDebugger(ConfigureTrigger.commandCleanConfigureAllWithDebugger, debuggerInformation), undefined, true);
     }
 
-    configure(folder?: vscode.WorkspaceFolder, showCommandOnly?: boolean) {
+    configure(folder?: vscode.WorkspaceFolder, sourceDir?: string, showCommandOnly?: boolean) {
```
gcampbell-msft, line 1104: This technically gets exposed as a public VS Code command, that can programmatically be invoked, in which case, this change would break our current "contract". 

To fix this, please move the `sourceDir` parameter to be the last parameter, this will not break the contract.

src/extension.ts
```diff
@@ -1876,6 +1876,9 @@ async function setup(context: vscode.ExtensionContext, progress?: ProgressHandle
         vscode.commands.registerCommand('cmake.outline.editCacheUI', () => runCommand('editCacheUI')),
         vscode.commands.registerCommand('cmake.outline.cleanRebuildAll', () => runCommand('cleanRebuildAll')),
         // Commands for outline items
+        vscode.commands.registerCommand('cmake.outline.configure', (what: ProjectNode) => runCommand('configure', what.folder, what.sourceDirectory, false)),
```
gcampbell-msft, line 1879: With the other change I requested, these parameters will need to be flipped.

gcampbell-msft:
@vlavati Where did you grab the icons that you are adding to the code from? We need to ensure we know where they came from and ensure that we have legal rights to use those in the product.

gcampbell-msft:
@vlavati Additionally, please make sure to make a CHANGELOG change to mark the change there and give yourself credit. Please follow the pattern of the changelog entries already present.

gcampbell-msft (changes requested):

vlavati:
> @vlavati Where did you grab the icons that you are adding to the code from? We need to ensure we know where they came from and ensure that we have legal rights to use those in the product.

Hi, @gcampbell-msft ! The added icons are drawn by me. So I hope they are under the same license agreement like the code. I tried to use the same colors as other icons have.

src/extension.ts
vlavati: done

src/extension.ts
vlavati: done

gcampbell-msft (approved):

gcampbell-msft:
> > @vlavati Where did you grab the icons that you are adding to the code from? We need to ensure we know where they came from and ensure that we have legal rights to use those in the product.
> 
> Hi, @gcampbell-msft ! The added icons are drawn by me. So I hope they are under the same license agreement like the code. I tried to use the same colors as other icons have.

Perfect! This should be safe. I've approved the code. I will do a couple more tests to make sure everything works and the icons look good in all themes. Pending no issues, we'll take this change!

gcampbell-msft:
@snehara99 Requesting your review before merging.

gcampbell-msft:
Windows tests are currently known to fail, merging.

Issue #3311, opened 2023-08-22
Add multiple directories to the workspace and all directories will be configured even if I only want to configure one directory


traversebitree:
## Brief
Add multiple directories to the **workspace** and **all** directories will be configured even if I only want to configure **one** directory.

### Step2. Add multiple folders to the workspace (untitled).  
  ![image](https://github.com/microsoft/vscode-cmake-tools/assets/30373236/5c4861a7-3b38-402f-ae36-3146417f8216)
  Then like this:  
  ![image](https://github.com/microsoft/vscode-cmake-tools/assets/30373236/78bd170f-b35f-4a05-99c4-b39f538379ed)
**Note**: both are top directories, with no subdirectory or include.
### Step3. Write some list files:  
  ![image](https://github.com/microsoft/vscode-cmake-tools/assets/30373236/e9b38516-e35e-4e8b-85f2-159cfddae498)
Code in `CMakeLists.txt` of TEST01:
```cmake
cmake_minimum_required(VERSION 3.25)
project(TEST01)
list(APPEND CMAKE_MESSAGE_INDENT "[TEST01]  ")
message(STATUS "hello from test01.")
```
Code in `CMakeLists.txt` of TEST02:
```cmake
cmake_minimum_required(VERSION 3.25)
project(TEST02)
list(APPEND CMAKE_MESSAGE_INDENT "[TEST02]  ")
message(STATUS "hello from test02.")
```
#### Try1. `Configure All Projects` button
![image](https://github.com/microsoft/vscode-cmake-tools/assets/30373236/4d6d41d5-8ca1-4746-8925-f21ee8578d64)
Outputs here:
```sh
[main] Configuring project: TEST01 
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.EXE" --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -SD:/Workspace/CPP/PROFESSIONAL-CMAKE-TEST/TEST01 -Bd:/Workspace/CPP/PROFESSIONAL-CMAKE-TEST/TEST01/build -G "Visual Studio 17 2022" -T host=x64 -A x64
[cmake] Not searching for unused variables given on the command line.
[cmake] -- Selecting Windows SDK version 10.0.22000.0 to target Windows 10.0.22621.
[cmake] -- [TEST01]  hello from test01.
[cmake] -- Configuring done (0.0s)
[cmake] -- Generating done (0.0s)
[...]

PR #3372, opened 2023-10-10
Configure only one project for CMakelists.txt node in ProjectOutline

vlavati:

- configure only project, that corresponds to CMakeLists.txt file node, instead of all projects

## The purpose of this change

The issue #3311 describes case with few projects in the ProjectOutline and it is possible to configure all projects, but there is no way to configure only a project that corresponds to the selected CMakeLists.txt. This PR changes the behavior of "configure" button to configure only one.


src/extension.ts
```diff
@@ -1876,7 +1876,13 @@ async function setup(context: vscode.ExtensionContext, progress?: ProgressHandle
         vscode.commands.registerCommand('cmake.outline.editCacheUI', () => runCommand('editCacheUI')),
         vscode.commands.registerCommand('cmake.outline.cleanRebuildAll', () => runCommand('cleanRebuildAll')),
         // Commands for outline items
-        vscode.commands.registerCommand('cmake.outline.configure', (what: ProjectNode) => runCommand('configure', what.folder, false, what.sourceDirectory)),
+        vscode.commands.registerCommand('cmake.outline.configure', async (what: object) => {
```
gcampbell-msft, line 1879: It'd be good if we could restrict this type more than just `object`, something that only allows for `ProjectNode` or `SourceFileNode`.

gcampbell-msft (changes requested):
Everything looks really good except for my single comment. Thanks!

src/extension.ts
vlavati: Thanks for review.
I've changed type annotation. Is it better?

gcampbell-msft: Yes! Everything looks good. Please add a Changelog entry and then I will approve and merge!

vlavati: Done

gcampbell-msft (approved):

Issue #3310, opened 2023-08-21
Support for $penv{} in include in CMake(User)Presets.json


jogerh:
### Brief Issue Summary

https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#includes describes that $penv{} can be used in include statements. This can be quite useful, particularly for centralizing user settings. One useful way of using this can be:

`{
  "version": 7,
  "include": [
    "$penv{USERPROFILE}\MyPresets.json",
  ]
}
`

This works with CMake, but it seems like it does not play well when configuring from VSCode with CMake Tools.

gcampbell-msft:
@jogerh Could you provide some information about how it doesn't seem to play well? 

A couple of things that could be helpful for us to have are the following:
1) CMake Tools Diagnostics
2) Repro project that shows the issue. Detail what you are expecting versus what happens. 

Thanks!

benmcmorran:
Relabeling as an enhancement because this is asking for a part of CMakePresets v7 support, which we don't yet formally support.

PR #3377, opened 2023-10-10
Support $penv{} in include paths

benmcmorran:
Addresses #3310

This ended up as more of a new feature than a bug fix. We don't formally support all of CMakePresets v6 or v7 right now, but this is a simple change to enable the include path part of the v7 changes.

### DIFF
--- a/src/presetsController.ts
+++ b/src/presetsController.ts
@@ -1029,7 +1029,7 @@ export class PresetsController {
             const includePath = presetsFile.version >= 7 ?
                 // Version 7 and later support $penv{} expansions in include paths
                 substituteAll(rawInclude, getParentEnvSubstitutions(rawInclude, new Map<string, string>())).result :
-                presetsFile.include[i];
+                rawInclude;
             const fullIncludePath = path.normalize(path.resolve(path.dirname(file), includePath));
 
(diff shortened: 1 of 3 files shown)


src/presetsController.ts
```diff
@@ -1025,7 +1025,12 @@ export class PresetsController {
 
         // Merge the includes in reverse order so that the final presets order is correct
         for (let i = presetsFile.include.length - 1; i >= 0; i--) {
-            const fullIncludePath = path.normalize(path.resolve(path.dirname(file), presetsFile.include[i]));
+            const rawInclude = presetsFile.include[i];
+            const includePath = presetsFile.version >= 7 ?
+                // Version 7 and later support $penv{} expansions in include paths
+                substituteAll(rawInclude, getParentEnvSubstitutions(rawInclude, new Map<string, string>())).result :
+                presetsFile.include[i];
```
gcampbell-msft, line 1032: Nit: This could just refer to `rawInclude` couldn't it? This would avoid duplicate references.

gcampbell-msft (dismissed):
One nit comment, other than that, everything looks good.

gcampbell-msft (approved):

PR #3387, opened 2023-10-16
updated changelog; added selected mingw kit before path rather than a…

snehara99:
### This changes how the selected MinGW kit is combined with the 'PATH' environment variable.

- removing the stripping of other MinGW kits from the 'PATH' environment variable
- adding the selected MinGW kit before the 'PATH' environment variable rather than after (top-most).

## The purpose of this change

This should make sure that the selected MinGW kit is used since it is added before the other MinGW kits on the 'PATH' environment variable. Before, if not all MinGW kits were stripped properly from the 'PATH' environment variable, conflicts could occur since the selected MinGW kit was added after the 'PATH' environment variable.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -989,9 +989,8 @@ export async function effectiveKitEnvironment(kit: Kit, opts?: expand.ExpansionO
                 path_list.push(mingwPath);
             }
             if (env.hasOwnProperty('PATH')) {
-                const current_path_list = (env['PATH']?.split(';') ?? []);
-                // since mingwPath is at the front of path_list, we shouldn't need to remove other mingw from current_path_list
-                path_list.push(current_path_list.join(';'));
+                // since mingwPath is at the front of path_list, we shouldn't need to remove other mingw from env['PATH']
+                path_list.push(env['PATH'] ?? '');
                 env['PATH'] = path_list.join(';');
             }
         }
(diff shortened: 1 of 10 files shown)


gcampbell-msft (approved):

src/kit.ts
```diff
@@ -989,12 +989,9 @@ export async function effectiveKitEnvironment(kit: Kit, opts?: expand.ExpansionO
                 path_list.push(mingwPath);
             }
             if (env.hasOwnProperty('PATH')) {
-                let current_path_list = (env['PATH']?.split(';') ?? []);
-                if (mingwPath) {
-                    // Remove other mingw from PATH as it will cause conflict
-                    current_path_list = current_path_list.filter(p => !isMingw(p));
-                }
-                path_list.unshift(current_path_list.join(';'));
+                const current_path_list = (env['PATH']?.split(';') ?? []);
```
benmcmorran, line 992: Given that we aren't filtering `current_path_list` anymore, we shouldn't need to `split()` and `join()`. We can just directly push `env['PATH']` to `path_list`.

benmcmorran (approved):

src/kit.ts
snehara99: fixed

benmcmorran (approved):

PR #3392, opened 2023-10-18
increased operation limit

snehara99:
The operation limit prevented going through all of our issues, so I increased it


benmcmorran (approved):

snehara99:
@benmcmorran @gcampbell-msft I made it so the the stale-old issues can only be closed after a minimum of 14 days too. In the workflow dispatch for close-days, if you type anything less than 14, it will default to -1 (never).

gcampbell-msft (approved):

PR #3397, opened 2023-10-19
Stale Bot Upgrades

snehara99:
## The following changes are proposed:

- increased operation limits to 500
- automatically scheduled stale-old bot daily, but an hour after the stale bot
- stale-old bot can add and remove stale-old labels automatically but can never close issues automatically
- now scheduled at 12am and 1am PDT

## The purpose of this change

Currently if someone comments on an issue with a 'stale-old' label, the label will never be removed unless someone manually removes it or manually runs the stale-old bot. This fix allows the stale-old bot to run daily and remove the label if the issue is updated without fear of automatically closing the issue. I scheduled the stale-old bot one hour after the stale bot to prevent operation limit issues.

### DIFF
--- a/.github/workflows/close-old-issues.yml
+++ b/.github/workflows/close-old-issues.yml
@@ -17,8 +17,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       stale-days: ${{ github.event.inputs.inactive-days || 720 }}
-      close-days: ${{ github.event.inputs.stale-label-days >= 14 && github.event.inputs.stale-label-days || -1 }}
-      stale-label: stale-old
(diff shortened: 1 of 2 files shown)


gcampbell-msft (approved):

.github/workflows/close-old-issues.yml
```diff
@@ -1,12 +1,16 @@
 name: Close Old Issues
 on:
+  schedule:
+    - cron: "0 1 * * *"
```
lukka, line 4: this is 6pm PDT, I think it would be good to run it overnight in PDT, e.g.  8am UTC

snehara99: good suggestion, I'll fix that

snehara99: this is fixed!

gcampbell-msft (approved):

Issue #3318, opened 2023-08-30
Environment PATH in `CMakePresets.json` not be use when set compiler is clang


WittonBell:
### Brief Issue Summary

VSCode+MinGW, MinGW is not in system path, and set it in  `CMakePresets.json`  environment PATH.
Select `clang Debug` configuration, will report error:
```
[main] 正在配置项目: t 
[proc] 执行命令: H:/msys64/mingw64/bin/cmake.exe -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_INSTALL_PREFIX=D:/Work/t/out/install/clang-debug -SD:/Work/t -BD:/Work/t/out/build/clang-debug -G Ninja
[cmake] -- Path Is: C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\x86C:\Program Files\Microsoft Visual ....
[cmake] -- The C compiler identification is Clang 16.0.5
[cmake] -- The CXX compiler identification is Clang 16.0.5
[cmake] CMake Error at CMakeLists.txt:6 (project):
[cmake]   The CMAKE_C_COMPILER:
[cmake] 
[cmake]     clang
[cmake] 
[cmake]   is not a full path and was not found in the PATH.
[cmake] 
[cmake]   Tell CMake where to find the compiler by setting either the environment
[cmake]   variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
[cmake]   the compiler, or to the compiler name if it is in the PATH.
[cmake] 
[cmake] 
[cmake] CMake Error at CMakeLists.txt:6 (project):
[cmake]   The CMAKE_CXX_COMPILER:
[cmake] 
[cmake]     clang++
[cmake] 
[cmake]   is not a full path and was not found in the PATH.
[cmake] 
[cmake]   Tell CMake where to find the compiler by setting either the environment
[cmake]   variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
[cmake]   to the compiler, or to the compiler name if it is in the PATH.
[cmake] 
[cmake] 
[cmake] -- Configuring incomplete, errors occurred!
```

select `gcc Debug` configuration is work correctly:
```
[...]

PR #3400, opened 2023-10-20
Dev/gcampbell/attempt fix3318

snehara99:
## This change addresses item #3318 

This expands penv before `where.exe` is called so that `where.exe` can be found incase it was only found in the expanded penv path.

The original issue was that `where.exe` couldn't be called to find clang since the penv wasn't expanded by that time yet.
## Other Notes/Information
 
This is Garrett's fix. I tested with gcc as well as clang from the original issue to confirm that this works. The code seems clean enough to me because penv should be parsed at that point.

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -921,6 +921,8 @@ async function expandConfigurePresetHelper(folder: string, preset: ConfigurePres
     inheritedEnv = EnvironmentUtils.mergePreserveNull([process.env, inheritedEnv]);
 
     let compilerEnv = EnvironmentUtils.createPreserveNull();
+    const expansionOpts: ExpansionOptions = await getExpansionOptions(workspaceFolder, sourceDir, preset);
+    expansionOpts.envOverride = inheritedEnv;
 
     // [Windows Only] If CMAKE_CXX_COMPILER or CMAKE_C_COMPILER is set as cl, clang, clang-cl, clang-cpp and clang++,
     // but they are not on PATH, then set the env automatically.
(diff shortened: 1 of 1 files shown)


gcampbell-msft (approved):

src/preset.ts
```diff
@@ -931,6 +935,11 @@ async function expandConfigurePresetHelper(folder: string, preset: ConfigurePres
             // The env variables for the supported compilers are the same.
             const compilerName: string | undefined = util.isSupportedCompiler(cxxCompiler) || util.isSupportedCompiler(cCompiler);
             if (compilerName) {
+                for (const key in preset.environment) {
+                    if (preset.environment[key]) {
+                        preset.environment[key] = await expandString(preset.environment[key]!, expansionOpts);
```
benmcmorran, line 940: I see that we also expand the environment in existing code on line 689. Is it safe to potentially expand an environment variable twice? Consider this case:

```
${dollar}{sourceDir}
```

The correct expansion is literally `${sourceDir}`, but I'm concerned that with this change it would end up expanding to the source directory itself.

snehara99: That's a good point. Maybe we can keep a copy of the unexpanded environment variable so that we can reset it after using it for `where.exe`

snehara99: @benmcmorran I believe I fixed the issue by expanding `preset.environment` into `expandedPreset.environment` instead overwriting `preset.environment`. There are instances of this elsewhere in that file. I also had to include `name` as a parameter, but I made sure that all the references to that function includes `name` now.

gcampbell-msft (approved):

PR #3424, opened 2023-11-06
options telemetry

snehara99:
this adds telemetry for the status bar and project status view options settings

this is how the telemetry would look if someone changed `options.statusBarVisibility` to `visible` but left everything else default.
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -2461,14 +2461,14 @@
                     "compact",
                     "icon",
                     "hidden"
+                  ],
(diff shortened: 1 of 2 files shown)


gcampbell-msft (dismissed):
Please debug this and ensure that the data desired is being sent to this telemetry call, but other than that, this looks good to me 👍

gcampbell-msft (approved):

Issue #3447, opened 2023-11-16
[regression] cannot setup msvc enviroments with specified architecture when using inherit (cmake-presets)


ArcticLampyrid:
### Brief Issue Summary

Cannot setup msvc enviroments with specified architecture when using inherit

CMakePresets.json:
```json
{
    "version": 3,
    "configurePresets": [
        {
            "name": "windows-base",
            "hidden": true,
            "generator": "Ninja",
            "binaryDir": "${sourceDir}/out/build/${presetName}",
            "installDir": "${sourceDir}/out/install/${presetName}",
            "cacheVariables": {
                "CMAKE_C_COMPILER": "cl.exe",
                "CMAKE_CXX_COMPILER": "cl.exe"
            },
            "condition": {
                "type": "equals",
                "lhs": "${hostSystemName}",
                "rhs": "Windows"
            }
        },
        {
            "name": "x64-debug",
            "displayName": "x64 Debug",
            "inherits": "windows-base",
            "architecture": {
                "value": "x64",
                "strategy": "external"
            },
            "cacheVariables": {
                "CMAKE_BUILD_TYPE": "Debug"
            }
        }
    ]
}
```

MSVC x64 should be setup for `x64-debug`, but actually MSVC x86 do.

Issue #3440, opened 2023-11-15
envs for cl.exe is not set automatically when using CMakePresets.json (v1.16.29+)


ArcticLampyrid:
### Brief Issue Summary

Environment variables for cl.exe is not set automatically when using CMakePresets.json (v1.16.29+)  
The same `CMakePresets.json` works well for v1.15.31.

traversebitree:
I also have the same problem.

BA7LYA:
Same.
I `git pull`ed the `2b14b606` to my local vcpkg submodule of my project in this morning, and vscode told me that:
```
[cmake] -- Running vcpkg install - done
[cmake] -- The CXX compiler identification is unknown
[cmake] CMake Error at CMakeLists.txt:5 (project):
[cmake]   The CMAKE_CXX_COMPILER:
[cmake] 
[cmake]     cl
[cmake] 
[cmake]   is not a full path and was not found in the PATH.
[cmake] 
[cmake]   Tell CMake where to find the compiler by setting either the environment
[cmake]   variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
[cmake]   to the compiler, or to the compiler name if it is in the PATH.
```
Info:
```
OS : windows-10.0.19045
arch : x64
vcpkg : Menifest mode
vscode : 1.84.2(user setup)
msvc compiler : 17.8.0
cmake version : 3.27.7
```

traversebitree:
I wonder to know how to find `cl` using `CMakePreset.json` without the help of Vscode?
Or, how to use `vswhere.exe` to locate `cl` using `CMakePreset.json`?

benmcmorran:
Thanks everyone for the prompt reporting! Given that it sounds like this is a regression in today's release, we'll investigate this with high priority and get a fix out as soon as possible. In the meantime, you should be able to work around the issue by manually switching the version of CMake Tools back to 1.15.31.

![image](https://github.com/microsoft/vscode-cmake-tools/assets/6981284/a4b3d02c-525f-43c8-93ef-d2231e73da5a)

gcampbell-msft:
@ArcticLampyrid @traversebitree @BA7LYA   Could you please send us your repro? Or at least your `CMakePresets.json`? We are investigating this in hopes of getting out a fix as soon as possible and a repro would be highly beneficial to diagnose the issue.

traversebitree:
```cpp
{
    "version": 6,
    "cmakeMinimumRequired": {
[...]

PR #3449, opened 2023-11-16
Be smarter about where we attempt magic PATH.

gcampbell-msft:
We attempt to, if we can't find their compiler for cl, clang, etc. to try and provide paths for them from VS. However, we did this in a location that got invoked for each level of expanding presets (preset inheritance). This was causing issues.
Instead, we can just do this at the very end before we return the preset.

This fixes: #3447.

Overview of regression and why it didn't work:

Our handling of preset inheritance is recursive. In 1.15, this bug was not present, because when we invoked where.exe, we were using the `preset.environment` which had not been merged with any parent environments. This caused it to, at the lowest stage, not find cl.exe and search for VS locations. https://github.com/microsoft/vscode-cmake-tools/blob/v1.15.31/src/preset.ts#L935C21-L935C78

However, in our recent attempt at fixing some issues, we were using the `inheritedEnv` as the environment for invoking `where.exe` (https://github.com/microsoft/vscode-cmake-tools/blob/1d1e79108b1d963436d6a8d0b75f654bc2495144/src/preset.ts#L938C1-L938C117), which had been merged with the parent. The flow looked like this. 
- The child preset enters `expandConfigurePresetHelper`.
  - We find that this child preset inherits from a parent. 
- The parent preset enters `expandConfigurePresetHelper`. It searches for the cl.exe and doesn't find it, and then, if an arch isn't provided, uses a default of 'x86' and adds the x86 cl.exe to the PATH. 
- Then, the child preset resumes and merges the parent preset env with the inheritedEnv. Then, when the child preset calls `where.exe` for cl, it DOES find it, even though it doesn't match the arch. 

SO, instead of calling where.exe multiple times, we should do it after the presets have all been merged and expanded. After all of that, we then call where.exe in cl, clang, etc. cases so that we can try to do some magic for the user. 

@snehara99 and I have pair tested the regression from #3447, #3440, as well as the original issue #3318 are all still fixed with this change.

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -691,6 +691,8 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
         }
     }
 
+    // TODO: We should try to find visual studio installations here.
+
     expansionOpts.envOverride = expandedPreset.environment;
 
     if (preset.__file && preset.__file.version >= 3) {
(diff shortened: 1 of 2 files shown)


src/preset.ts
```diff
@@ -691,6 +691,8 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
         }
     }
 
+    // TODO: We should try to find visual studio installations here.
```
benmcmorran, line 694: Why can't we do this in this PR? Is the change too risky?

gcampbell-msft: Oh, we do do this in this PR, I just forgot to remove the TODO, we do it right before returning.

benmcmorran (approved):

Issue #3446, opened 2023-11-15
Commands with folder arguments broken


xisui-MSFT:
### Brief Issue Summary

Commands with folder arguments, like build, are now broken.

Repro steps:
1. Create an extension (A), that depends on CMakeTools.
2. In extension (A), do something like:
    ```
    async test() {
        if (!vscode.workspace.workspaceFolders) return;
        for (const folder of vscode.workspace.workspaceFolders) {
            await vscode.commands.executeCommand("cmake.build", folder);
        }
    }
    ```
3. Build and install extension (A), open a multiroot workspace containing multiple CMake projects, and run `test`.

Expected:
All folders are built.

Actual:
Only the first folder is built.

Note:
I think this command has been broken for `build` for several versions, but some other commands triggers `build` worked, so we didn't hit this issue. Now none of them work, and this breaks our extension.

benmcmorran:
Are you able to use [the API](https://github.com/microsoft/vscode-cmake-tools-api) instead? I'm not sure if the commands have ever been an officially supported extensibility point.

xisui-MSFT:
@benmcmorran Since we wrote our extension before this API, we didn't have this option. While we can try that API, I do hope this issue can be fixed so our extensions that are already published won't be broken.

gcampbell-msft:
@xisui-MSFT Is this specific to the latest CMake Tools version?

gcampbell-msft:
@xisui-MSFT I just followed your repro steps and wasn't able to reproduce the issue. I'm seeing that all of the folders are built. Is there any way that you could provide a repro project and/or a video of what you're seeing? Thanks.

xisui-MSFT:
Actually this commit https://github.com/vlavati/vscode-cmake-tools/commit/3cfda1ad18cb4cb5a0ba0818c68fc989859c41c2. Why are object parameters required?

gcampbell-msft:
This has been fixed.

PR #3453, opened 2023-11-17
Fix compatibility issue in commands

xisui-MSFT:
## This change addresses item #3446

Now these commands accept both folder and {folder, targetName} object.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1363,7 +1363,7 @@ export class ExtensionManager implements vscode.Disposable {
             folder = undefined;
             targetName = undefined;
         } else if (util.isString(args) || util.isWorkspaceFolder(args)) {
-            folder = folder;
+            folder = args;
             targetName = undefined;
         } else {
             folder = args.folder;
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -66,6 +66,7 @@ export let extensionManager: ExtensionManager | null = null;
 
 type RunCMakeCommand = (project: CMakeProject) => Thenable<any>;
 type QueryCMakeProject = (project: CMakeProject) => Thenable<string | string[] | null>;
+type FolderTargetNameArgsType = vscode.WorkspaceFolder | string | { folder?: vscode.WorkspaceFolder | string; targetName?: string };
 
 interface Diagnostics {
     os: string;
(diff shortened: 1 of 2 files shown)


gcampbell-msft:
@xisui-MSFT Thanks for the suggestion. However, I think we should simply correct what happened in the original PR. Rather than using an object, it should be two separate parameters for the methods. 

For example:

`launchTargetPath(folder?: vscode.WorkspaceFolder | string, targetName?: string)`

This means that you can still pass them as regular parameters and it keeps the code simple. It does slightly affect the original implementatition, but the structure with the first parameter being `folder` had been there much longer and thus might break more people than the three months that it's been an object. 

Could you make those changes?

xisui-MSFT:
@gcampbell-msft I don't think regular parameters are supported in `args` field in `inputs`, as proposed in the original PR. When `args` provided is an array, it's passed to the command as a single array argument; and when it's provided as an object, it's passed to the command as a single object argument. I think this is why the original PR uses object instead of normal parameters.

gcampbell-msft:
> @gcampbell-msft I don't think regular parameters are supported in `args` field in `inputs`, as proposed in the original PR. When `args` provided is an array, it's passed to the command as a single array argument; and when it's provided as an object, it's passed to the command as a single object argument. I think this is why the original PR uses object instead of normal parameters.

Ah, I just tested and you are correct. I'll review the current state of the PR now.

src/extension.ts
```diff
@@ -66,6 +66,7 @@ export let extensionManager: ExtensionManager | null = null;
 
 type RunCMakeCommand = (project: CMakeProject) => Thenable<any>;
 type QueryCMakeProject = (project: CMakeProject) => Thenable<string | string[] | null>;
+type FolderTargetNameArgsType = vscode.WorkspaceFolder | string | { folder?: vscode.WorkspaceFolder | string; targetName?: string };
```
gcampbell-msft, line 69:
```suggestion
type FolderTargetNamedArgsType = vscode.WorkspaceFolder | string | { folder?: vscode.WorkspaceFolder | string; targetName?: string };
```

src/util.ts
```diff
@@ -738,7 +738,7 @@ export function isNullOrUndefined(x?: any): x is null | undefined {
     return (x === null || x === undefined);
 }
 
-export function isWorkspaceFolder(x?: any): boolean {
+export function isWorkspaceFolder(x?: any): x is vscode.WorkspaceFolder {
```
gcampbell-msft, line 741: Instead of this, should we leave it as boolean and simply add a check in the method itself for `x is vscode.WorkspaceFolder`?

gcampbell-msft: If this is possible I think I'd rather keep the generic `boolean`

gcampbell-msft (approved):
Approved but with a couple of changes that I think would be nice, but aren't required before merging.

src/util.ts
xisui-MSFT: It's returning a boolean. The difference is when `x is vscode.WorkspaceFolder` is used as return type, it lets the interpreter know its type when it returns true. I.E, when `x is vscode.WorkspaceFolder` is returned:
```
if (isWorkspaceFolder(x)) {
    // x is now a WorkspaceFolder
}
```

When returning boolean,
```
if (isWorkspaceFolder(x)) {
    // x is still any
}
```

src/extension.ts
xisui-MSFT: It means this type contains `Folder` and `TargetName`. It's not meant to be an adjective. But I'm open to better naming.

src/util.ts
gcampbell-msft: Ah, that sounds great! Thanks for the explanation.

src/extension.ts
gcampbell-msft: Ahh, that seems reasonable to me, I just read it wrong.

Issue #3214, opened 2023-06-20
copyCompileCommands to the original path makes the file empty


tocic:
### Brief Issue Summary

I'm using `cmake.copyCompileCommands` as a workaround to support _clangd_ (see https://github.com/microsoft/vscode-cmake-tools/issues/654):

```json
"cmake.copyCompileCommands": "${workspaceFolder}/build/compile_commands.json"
```

But if the `${workspaceFolder}/build` directory is the original `compile_commands.json`'s directory then the extension makes the file empty. I expect it to either not copy the file or overwrite the file with its original contents.

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.79.2",
  "cmtVersion": "1.14.33",
  "configurations": [
    {
      "folder": "/home/tocic/test",
      "cmakeVersion": "3.26.4",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": false,
      "compilers": {
        "C": "/usr/bin/gcc",
        "CXX": "/usr/bin/g++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": false
    }
  ]
}
```

### Debug Log

```shell
[main] Copying compile_commands.json from /home/tocic/test/build/compile_commands.json to /home/tocic/test/build/compile_commands.json
[compdb] Error parsing compilation database "/home/tocic/test/build/compile_commands.json": 
	SyntaxError: Unexpected end of JSON input
	at JSON.parse (<anonymous>)
	at CompilationDatabase.fromFilePaths (/home/tocic/.vscode-oss/extensions/ms-vscode.cmake-tools-1.14.33-universal/dist/main.js:37178:38)
	at async CMakeProject.refreshCompileDatabase (/home/tocic/.vscode-oss/extensions/ms-vscode.cmake-tools-1.14.33-universal/dist/main.js:34944:27)
	at async /home/tocic/.vscode-oss/extensions/ms-vscode.cmake-tools-1.14.33-universal/dist/main.js:35056:33
[...]

PR #3500, opened 2023-12-18
prevent compilation db corruption (#3214)

parniere:
- if copyCompileCommands is equal to default db path,
    we need to avoid the copy otherwise it would  copy on itself


parniere:

gcampbell-msft (dismissed):

gcampbell-msft:
@parniere Also, if you could, it would be great to make note of your changes (and take credit for yourself!) in the CHANGELOG.md. If you could make this change before merging, that would be great!

gcampbell-msft (approved):

parniere:
@gcampbell-msft, Hi Garrett, I've added an entry into the changelog by modifying my initial commit.
I tried to follow the existing format.

Cheers
Olivier

PR #3524, opened 2024-01-04
Improve our initialization and interaction with cpptools

gcampbell-msft:
### This aims to improve our initialization and interaction with the cpptools API.

- Remove the unconditional `updateCodeModel()` at the end of `postWorkspaceOpen`
- Only call the initial `notifyReady` if we actually have a code model, which will either be because we configured or because we have a cache. 
- Only call `.set` on the target name in `startNewCMakeDriver` if the name is different than what we already have set, this will avoid duplicate udpates.


gcampbell-msft:
@snehara99 @benmcmorran @lukka FYI, this is a draft PR for now until I learn more about how the cpptools API interaction should work, once I confirm that I can confirm that these changes are good. Thanks.

src/extension.ts
```diff
@@ -543,8 +543,6 @@ export class ExtensionManager implements vscode.Disposable {
                 }
             }
         }
-
-        this.updateCodeModel(project);
```
bobbrow, line 547: Can you double check that the `ConfigureTrigger.configureWithCache` case continues to work if you remove this? I walked through the code and I think it will still fire the `onCodeModelChanged` event, but I didn't test it explicitly.

The scenario is `cmake.configureOnOpen` set to false and the user has manually configured at least once with CMake Tools.  In this scenario the file-api data is available and we can update the code model.  That should trigger the event that then calls `this.updateCodeModel`.

gcampbell-msft: I do believe I've tested this case and confirmed it works, but I'll make a note to test this again explicitly to ensure it does.

gcampbell-msft: Just tested again and confirmed this works.

Issue #2889, opened 2022-12-07
I keep getting "Not all open documents were saved. Would you like to continue anyway?"


sean-mcmanus:
### Brief Issue Summary

I keep getting "Not all open documents were saved. Would you like to continue anyway?" after updating. I'd like some way to stop getting that and default to Yes.

![image](https://user-images.githubusercontent.com/19859882/206055836-e264d7f8-77e7-40da-89b6-9567f3c48c65.png)

elahehrashedi:
Do you want to skip saving files before configure? can you use this setting? `"cmake.saveBeforeBuild": false` ?
Or are your files already saved and you are still getting this error?

Manewu:
I have the same problem. After pressing "Build" the changed files are saved automatically because "cmake.saveBeforeBuild" is true, but  then I still get this message. I think that behavior is relatively new, maybe came in with the latest Update (7.12.2022)?

sean-mcmanus:
I get the error when I'm not doing a build, so it seems like the `"cmake.saveBeforeBuild": false` setting may not work, but I could try. I don't care if the file is saved or not -- in my case I'm typically changing a source file, so it won't impact configuring.

Yeah, I only started noticing this a week or so ago.

elbrandt:
I have the same problem, and it started appearing around a week or two ago as well.
My hypothesis was that it has something to do with format-on-save, but I have not tested this hypothesis.

my-name-is-already-taken:
I have observed the same problem. For me, it only appears when I have a diff tab open that contains unsaved changes. To reproduce:
- Open a Folder that contains a git repository
- Modify some file
- Save the file
- Modify the file again, without saving it this time
- Open a diff tab via the source control panel
- Compile the project

Manewu:
Yes, I can confirm that, changed file has to be open normally and in a diff tab at the same time. Good to know:-).

elbrandt:
FYI, I am experiencing this with no diff windows open.

TheJCAB:
It's happening to me too. Diff windows open are a pretty normal thing for me, so likely related.

Note that just hitting build again does start the build (because the files were indeed saved), even if the message is still there. Seems like some sort of data race between the file save and the build proper.

Annoying but not blocking. But annoying :-).

notskm:
> It's happening to me too. Diff windows open are a pretty normal thing for me, so likely related.

[...]

PR #3527, opened 2024-01-05
add option to disable asking if we should still build when not all do…

gcampbell-msft:
Add an option to disable the popup that asks if the user wants to continue building even if there were issues saving all the files. 

This helps mitigate the raciness that seems to be happening for some people. If the popup is disabled, then the build will continue by default. 

This will address #2889.


gcampbell-msft:
@snehara99 @benmcmorran @lukka Ping. Thanks!

snehara99 (approved):

PR #3528, opened 2024-01-09
Update the api.ts for the getActiveFolderPath method.

gcampbell-msft:
This will need to also update and pull down the 1.1 version of the API, so this is pending on a new version being published of https://www.npmjs.com/package/vscode-cmake-tools?activeTab=readme. 

These changes are thanks to @kemaweyan, who started this idea here: #3520. This makes the changes in our formal API. 

We add the API method implementation and we update the dependency on the public API interface.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -39,6 +39,10 @@ export class CMakeToolsApiImpl implements api.CMakeToolsApi {
         return project ? new CMakeProjectWrapper(project) : undefined;
     }
 
+    getActiveFolderPath(): string {
+        return this.manager.activeFolderPath();
+    }
+
     private async setUIElementVisibility(element: api.UIElement, visible: boolean): Promise<void> {
         switch (element) {
             case api.UIElement.StatusBarDebugButton:


gcampbell-msft:
Make sure to update the api version in the class.

gcampbell-msft:
Npm package was updated for the API, as well as the dependency here, so this is ready to review. I confirmed that the API works by using the VS Code API Explorer (@benmcmorran)

src/api.ts
```diff
@@ -12,7 +12,7 @@ import { assertNever } from '@cmt/util';
 export class CMakeToolsApiImpl implements api.CMakeToolsApi {
     constructor(private readonly manager: ExtensionManager) {}
 
-    version: api.Version = api.Version.v1;
+    version: api.Version = api.Version.v2;
```
benmcmorran, line 15: It's not a huge deal given what I expect is pretty low usage of the API, but we can do better by dynamically selecting which version of the API this claims to be based on the version that was requested from `getApi()` call.

As is, the extension will always return a v2 API object, even if the caller requested v1. Because v2 is backwards-compatible with v1, there should be no harm in simply overriding the API version in the returned object to be v1 when v1 is requested. That will allow any existing exact match queries for the v1 version of the API to continue to succeed even after the 1.17 release.

benmcmorran (approved):

src/api.ts
gcampbell-msft: I like this, good idea, I'll modify.

gcampbell-msft: Made the changes, let me know if you think I should do it in a different way @benmcmorran . Thanks!

gcampbell-msft: @benmcmorran Ping.

snehara99 (approved):

PR #3537, opened 2024-01-16
Allow overriding of CMakePresets cache variables and arguments

gcampbell-msft:
### This changes the behavior of the settings provided by the CMake extension for arguments and environments.

- Allow various fields that modify configuration, build, and testing, to override CMakePresets environments and arguments.
- Specifically, this includes the following fields that are now supported when using presets, while doing various actions
    - cmake.environment (any cmake command)
    - cmake.configureArgs (configure)
    - cmake.configureEnvironment (configure)
    - cmake.buildArgs (build)
    - cmake.buildEnvironment (build) 
    - cmake.buildToolsArgs (build, but specific to the build tool)
    - cmake.ctestArgs (test)
    - cmake.testEnvironment (test)
- With the above changes, we added UI notifications to ensure users are aware of when any of the above settings are possibly overriding settings from their Preset. In general, the UI will appear if they have any setting (from the above list) affecting their configurations. 
- Also, we added log statements that will appear in the output that will notify the user as well. 

## The purpose of this change

The general purpose is to allow for the settings we provide via VS Code, to be used during CMake Presets so that people can temporarily override various settings and cache variables in order to test and investigate their configuration and project set-up. 

## Screenshots of various functionality.
Project status view:
If there are settings set in any of the above list, the affected configuration will have a UI string to notify the user:
image

When there are settings that are affecting the configuration, a setting icon will appear that will open the settings to the possible settings that may be affecting the configuration: 
image.
This will open the settings as such: 
image.

Additionally, if the settings are removed, then the "override settings applied" text will be removed from the UI. 

Status bar:
image
An `*` is added to the status bar if the configure preset, build preset, or test preset icons are visible and there are settings (from the above list) set. 

Output pane:
There will be a simple text output that indicates that there are settings defined that may affect or override your preset configuration. This will happen for things like configure, build, test, launch, etc. 
Here is an example: 
image


gcampbell-msft:
@snehara99 @benmcmorran @lukka Ping for review. I'd especially like review on the various strings I've modified and added and whether the wording is good, or if I should rework some of them. This includes the strings for tooltips for the command I added, the output text in the output pane, and the text in the project status view. 

Other than that, it should be a normal PR review. Thanks!

src/ctest.ts
```diff
@@ -301,29 +301,11 @@ export class CTestDriver implements vscode.Disposable {
...
-                const args = [];
-                for (const value of this.ws.config.ctestArgs) {
-                    args.push(await expandString(value, opts));
-                }
-                ctestArgs = [`-j${jobs}`, '-C', configuration].concat(defaultArgs, args);
+            const ctestArgs = await this.getCTestArgs(driver, customizedTask, testPreset);
+
+            if (ctestArgs === undefined) {
+                log.error(localize('test.preset.not.set', 'Test preset is not set'));
+                return -3;
```
lukka, line 308: this seems like a change in the prior behavior: the error `test.preset.not.set` is filed with `ctestArgs` is `undefined`, but previously it was based solely on the condition `!driver.testPreset`.

src/projectStatus.ts
```diff
@@ -180,6 +192,27 @@ class TreeDataProvider implements vscode.TreeDataProvider<Node>, vscode.Disposab
         await this.refresh();
     }
 
+    public async refreshConfigNode(): Promise<any> {
+        if (this.configNode) {
```
lukka, line 196: probably we could diminish code duplication by doing:

```ts
public async refreshNode(node: Node): Promise<any>
{ 
  if (node} { 
    await node.refresh(); 
    this.Data.Fire(node);
  }
}
```

.eslintrc.js
```diff
@@ -92,7 +92,7 @@ module.exports = {
         "no-trailing-spaces": "error",
         "no-undef-init": "error",
         "no-unsafe-finally": "error",
-        "no-unused-expressions": "error",
+        "no-unused-expressions": "off",
```
lukka, line 95: is this change really needed?

src/ctest.ts
```diff
@@ -247,29 +247,29 @@ export class CTestDriver implements vscode.Disposable {
...
     private async getCTestArgs(driver: CMakeDriver, customizedTask: boolean = false, testPreset?: TestPreset): Promise<string[] | undefined> {
         let ctestArgs: string[];
+        const opts = driver.expansionOptions;
+        const initialArgs = await Promise.all(this.ws.config.ctestDefaultArgs.map(async (value) => expandString(value, driver.expansionOptions)));
+        const additionalArgs = await Promise.all(this.ws.config.ctestArgs.map(async (value) => expandString(value, driver.expansionOptions)));
+
+        ctestArgs = initialArgs.slice(0);
+
         if (customizedTask && testPreset) {
-            ctestArgs = ['-T', 'test'].concat(testArgs(testPreset));
```
lukka, line 251: where are `[-T, test]` now defined? Are they still being used?

src/drivers/cmakeDriver.ts
```diff
@@ -1714,9 +1764,19 @@ export abstract class CMakeDriver implements vscode.Disposable {
         } else {
             buildPreset.__targets = buildPreset.targets;
         }
-        const args = preset.buildArgs(buildPreset);
+        let args = preset.buildArgs(buildPreset, this.config.buildArgs);
+        args = args.concat('--', ...this.config.buildToolArgs);
```
lukka, line 1768: in general, how the user provided arguments are going to be separated in strings? e.g. if the user provides "-G Ninja -DVAR=1 -DANOTHERVAR="value" -- -j 2", how this single string is going to be split in the proper arguments for the process being spawn? Or is it up to the user to provide the arguments already properly split?

package.nls.json
```diff
@@ -83,9 +86,9 @@
     "cmake-tools.configuration.cmake.generator.description": "The CMake generator to use.",
     "cmake-tools.configuration.cmake.toolset.description": "The CMake toolset to use when configuring.",
     "cmake-tools.configuration.cmake.platform.description": "The CMake platform to use when configuring.",
-    "cmake-tools.configuration.cmake.configureArgs.description": "Additional arguments to pass to CMake when configuring.",
-    "cmake-tools.configuration.cmake.buildArgs.description": "Additional arguments to pass to CMake when building.",
-    "cmake-tools.configuration.cmake.buildToolArgs.description": "Additional arguments to pass to the underlying build tool when building.",
+    "cmake-tools.configuration.cmake.configureArgs.description": "Additional arguments to pass to CMake when configuring. When using CMake Presets, these are to be used as temporary overrides.",
```
benmcmorran, line 89: As currently written, this could be interpreted as the value of `configureArgs` replacing any existing arguments ("override") when I think what actually happens is `configureArgs` is appended to the existing arguments. What if we reword to something like this?

> Additional arguments to pass to CMake when configuring. When using CMake Presets, these arguments are temporarily appended to the arguments provided by the active configure preset.

.eslintrc.js
gcampbell-msft: Yes, things like `getStatusBar()?.updateBuildPresetButton();` were being failed during build, but that seems unreasonable, when this is a good way to succinctly write this code.

Thanks.

package.nls.json
gcampbell-msft: I think that makes sense, I'll make similar changes for all of the affected strings.

src/drivers/cmakeDriver.ts
gcampbell-msft: We are leaving it up to the user to provide the arguments already properly split. However, thanks for calling this out, because I'm noticing that I didn't change this code correctly, making a fix to ensure that we only add `--` once before the buildTool arguments.

src/ctest.ts
gcampbell-msft: They are still being used, however, we are falling back to allowing the user to have control over it. They are now coming from the `ctestDefaultArgs` (see line 251 in the new code). In package.json, the default values for `ctestDefaultArgs` are `-T, test, --output-on-failure`. So, instead of hardcoding this in, we are now allowing the user to modify the default arguments that are always appended to the front of the command.

src/projectStatus.ts
gcampbell-msft: Good idea for sharing code, making the change.

src/ctest.ts
gcampbell-msft: Good catch, making the change.

package.json
```diff
@@ -1451,7 +1484,12 @@
         },
         {
           "command": "cmake.projectStatus.selectConfigurePreset",
-          "when": "view == cmake.projectStatus && cmake:enableFullFeatureSet && viewItem == 'configPreset'",
+          "when": "view == cmake.projectStatus && cmake:enableFullFeatureSet && viewItem =~ /configPreset/",
```
snehara99, line 1487: do you need quotes around the `/configPreset/`

package.json
```diff
@@ -1466,9 +1504,14 @@
         },
         {
           "command": "cmake.projectStatus.selectBuildPreset",
-          "when": "view == cmake.projectStatus && cmake:enableFullFeatureSet && viewItem == 'buildPreset'",
+          "when": "view == cmake.projectStatus && cmake:enableFullFeatureSet && viewItem =~ /buildPreset/",
```
snehara99, line 1507: here too for the quotes (single quotes)

package.json
```diff
@@ -1481,9 +1524,14 @@
         },
         {
           "command": "cmake.projectStatus.selectTestPreset",
-          "when": "view == cmake.projectStatus && cmake:enableFullFeatureSet && viewItem == 'testPreset'",
+          "when": "view == cmake.projectStatus && cmake:enableFullFeatureSet && viewItem =~ /testPreset/",
```
snehara99, line 1527: here too sorry

snehara99 (dismissed):

package.json
gcampbell-msft: No. The `=~` operator asks for regex, so the right hand side needs to be regex, you can see the documentation that I followed here: https://code.visualstudio.com/api/references/when-clause-contexts#match-operator

package.json
gcampbell-msft: No. The `=~` operator asks for regex, so the right hand side needs to be regex, you can see the documentation that I followed here: https://code.visualstudio.com/api/references/when-clause-contexts#match-operator

src/drivers/cmakeDriver.ts
```diff
@@ -276,6 +276,8 @@ export abstract class CMakeDriver implements vscode.Disposable {
         let envs;
         if (this.useCMakePresets) {
             envs = EnvironmentUtils.create(configurePreset ? configurePreset.environment : this._configurePreset?.environment);
+            envs = EnvironmentUtils.merge([envs, await this.computeExpandedEnvironment(this.config.environment, envs)]);
```
benmcmorran, line 279: Does this mean we'll (perhaps unintentionally) support preset style ${} macros in the settings.json specified environment?

benmcmorran (dismissed):

benmcmorran (approved):

PR #3542, opened 2024-01-18
add pinned cmake commands view

moyo1997:


snehara99 (dismissed):

snehara99:
Nbd, but for UI changes it would be easier to review if you add a screenshot

package.nls.json
```diff
@@ -197,7 +197,14 @@
...
     "cmake-tools.configuration.views.cmake.folders.description": "Folders",
     "cmake-tools.configuration.views.cmake.projectStatus.description": "Project Status",
+    "cmake-tools.configuration.views.cmake.pinnedCommandsWelcome.description":{
+        "message": "No CMake commands have been pinned yet.\n[Add Command](command:cmake.pinnedCommands.add)",
+        "comment": [
+            "The text in parentheses () should not be localized or altered in any way. Also the square brackets and parentheses themselves [] () should not be altered or as well. However, the text inside the square brackets [] should be localized."
+        ]
+    }, 
     "cmake-tools.configuration.views.cmake.outline.description": "Project Outline",
+    "cmake-tools.configuration.views.cmake.pinnedCommands.description": "CMake: Pinned Commands",
```
gcampbell-msft, line 207: We shouldn't have the `CMake` part here. This is covered by the "category" in the package.json. 

See why with this screenshot, this applies to all of the added descriptions. Also, is this meant to be a title? 
image

src/cmake/cmakeExecutable.ts
```diff
@@ -66,5 +67,5 @@ export async function getCMakeExecutableInformation(path: string): Promise<CMake
 }
 
 export async function setCMakeDebuggerAvailableContext(value: boolean): Promise<void> {
-    await util.setContextValue("vscode-cmake-tools.cmakeDebuggerAvailable", value);
+    await setContextAndStore("vscode-cmake-tools.cmakeDebuggerAvailable", value);
```
gcampbell-msft, line 70: This is a nit, as it's not something that you actually made changes to, but I think it might be nice if we change the usage of `vscode-cmake-tools.cmakeDebuggerAvailable`, to match the pattern everywhere else `cmake:cmakeDebuggerAvailable`.

src/util.ts
```diff
@@ -479,6 +480,64 @@ export function thisExtensionPackage(): PackageJSON {
...
+    const localizedStrings = fs.readFileSync(localizedFilePath, "utf8");
+    return JSON.parse(localizedStrings);
+}
+
+interface CommandPalette {
+    command: string;
+    when: string | null;
+}
+
+function evaluateExpression(expression: string | null, context: {[key: string]: any}): boolean {
```
gcampbell-msft, line 498: Cool! Should we have unit tests for this to ensure that we're using this correctly to make sure we always understand the context right? 

Or, did you happen to pull this from the vscode core implementation, so we know that it's right? IF so, we should leave a comment with a link.

src/util.ts
```diff
@@ -479,6 +480,64 @@ export function thisExtensionPackage(): PackageJSON {
     };
 }
 
+export async function GetExtensionLocalizedPackageJson(): Promise<{[key: string]: any}> {
```
gcampbell-msft, line 483: Cool again! I honestly don't know if I would have thought of this. 

I assume that we need this in order to make sure the strings we output are actually localized? I do wonder if there is any way that VSCode allows for you to grab this without having to do this way, but either way, if it works, this is cool!

gcampbell-msft (changes requested):
Overall, the changes of implementation look good and I've tested the feature and it seems pretty cool and works well! Just some code and design changes I'd like to see if possible before merging. 

One other thing, when you test, when you open VS Code for the first time, does the Pinned Commands welcome view appear within the same CMake Activity bar view with the Project Status and the Project Outline? If not, I think that it should. 

Great job 👍

gcampbell-msft:
Also, there seem to be some build failures that may be due to flaky issues with the windows test, but I requeued to try again.

src/util.ts
moyo1997: From my research it doesnt look like there's a way. the loc file is really meant to localize the package.json itself

src/pinnedCommands.ts
```diff
@@ -0,0 +1,184 @@
...
+        this.commandName = command;
+        this.isVisible = isVisible;
+    }
+
+    getTreeItem(): vscode.TreeItem {
+        return this;
+    }
+
+    async runThisCommand() {
+        await vscode.commands.executeCommand(this.commandName);
```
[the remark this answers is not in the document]
moyo1997: Good call. Testing, there are no lags and the tree can still be interacted with even when the command is still running.

src/extension.ts
```diff
@@ -235,6 +246,44 @@ export class ExtensionManager implements vscode.Disposable {
...
+
+    public GetWorkspaceConfig() {
+        return this.workspaceConfig;
+    }
+
+    public UpdateContextValues(key: string, value: string) {
+        this.contextValues[key] = value;
+
+        // contextvalues have changed so update active extension commands.
+        if (this.extensionActiveCommandsInfo && (!this.extensionActiveCommandsInfo.contextUsed.hasOwnProperty(key) || this.extensionActiveCommandsInfo.contextUsed[key] !== value)) {
```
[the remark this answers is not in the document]
moyo1997: yup exactly that

src/config.ts
```diff
@@ -333,6 +334,9 @@ export class ConfigurationReader implements vscode.Disposable {
     get enableOutputParsers(): string[] | null {
         return this.configData.enabledOutputParsers;
     }
+    get pinnedCommandsList(): string[] | null {
```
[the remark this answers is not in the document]
moyo1997: yeah I see your point. refactored.

src/cmake/cmakeExecutable.ts
moyo1997: refactored!

package.nls.json
moyo1997: that's interesting. I only see one CMake: prefix when i launch the extension. But removed the "CMake:" as suggested to match consistency with the other views in our extension.

src/extension.ts
```diff
@@ -235,6 +246,44 @@ export class ExtensionManager implements vscode.Disposable {
             telemetryProperties['autoSelectActiveFolder'] = `${this.workspaceConfig.autoSelectActiveFolder}`;
         }
         telemetry.sendOpenTelemetry(telemetryProperties);
+
+        // do these last
+        this.extensionLocalizedStrings = await util.GetExtensionLocalizedPackageJson();
+        this.SetExtensionActiveCommands();
+    }
+
+    public GetWorkspaceConfig() {
```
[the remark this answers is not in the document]
moyo1997: refactored!

src/util.ts
moyo1997: Did not take this from vscode core implementation. Doing some research, theirs is a lot more robust I found the files that help vscode with this logic. It's very robust and is overkill for ours currently, but will be including the recycled file in case we ever include more complex conditions in our package.json when clauses.

gcampbell-msft (approved):

PR #3545, opened 2024-01-19
Add a setting to enable/disable the Test Explorer Integration

gcampbell-msft:
### This changes our integration with the Test Explorer. 

- Add a setting that allows users to enable or disable our integration with the Test Explorer.
- This keeps the ability to run tests, however, the integration with the Test Explorer should now be allowed to be disabled. 

image

If we disable it, it automatically revokes the integration:
image

We are still able to run tests without the integration:
Tests

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -96,6 +96,7 @@
         "comment": "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
     },
     "cmake-tools.configuration.cmake.ctest.allowParallelJobs.description": "Allows ctests to be run in parallel, however the result output may be garbled as a result.",
+    "cmake-tools.configuration.cmake.ctest.testExplorerIntegrationEnabled.description": "Whether or not the integration with the test explorer is enabled. This is helpful to disable if you prefer using a different extension for test integration.",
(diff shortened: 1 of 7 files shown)


package.nls.json
```diff
@@ -96,6 +96,7 @@
         "comment": "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
     },
     "cmake-tools.configuration.cmake.ctest.allowParallelJobs.description": "Allows ctests to be run in parallel, however the result output may be garbled as a result.",
+    "cmake-tools.configuration.cmake.ctest.testExplorerIntegrationEnabled.description": "Whether or not the integration with the test explorer is enabled. This is helpful to disable if you prefer using a different extension for test integration.",
```
snehara99, line 99: I would say "Configures whether or not...", to make the description not a fragment, but up to you (nit)

snehara99:
does this need docs updates?

snehara99 (approved):

gcampbell-msft:
> does this need docs updates?

Good call, I'll check now if there are any docs that talk about the Test Explorer, and make sure I update them.

xisui-MSFT (approved):

Issue #3452, opened 2023-11-16
Add default-visible setting for advanced options in status bar


snehara99:
### Brief Issue Summary

Add setting to default advanced status bar options to general status bar visibility setting (if "icon", "visible", or "compact") when status bar is set to a non-hidden setting.

PR #3547, opened 2024-01-22
added inherit option and inheritDefault setting

snehara99:
## This change addresses item #3452 

This adds the `inherit` option to the `cmake.options.advanced.statusBarVisibility` setting, and also the `cmake.options.advanced.inheritDefault` setting. Here is the behaviour from the updated docs:

> If a `statusBarVisibility` option is set to `inherit`, it will inherit the more general `Cmake > Options : Status Bar Visibility` setting if it is not set to `hidden`. If the more general `Cmake > Options : Status Bar Visibility` setting is set to `hidden`, then the `inherit` option will default to what is set in the `inheritDefault` setting. The default option for `inheritDefault` is `visible`. 

These are now the default settings:
```json
"cmake.options.statusBarVisibility": "hidden"
"cmake.options.advanced": {
    "build": {
        "statusBarVisibility": "inherit"
    },
    "launch": {
        "statusBarVisibility": "inherit"
    },
    "debug": {
        "statusBarVisibility": "inherit"
    }
}


snehara99:
@sinemakinci1 It would be helpful if you could review the docs changes I made, thanks!

sinemakinci1 (approved):

sinemakinci1:
@snehara99  how does this change impact existing users who have configured their settings or did not configure their settings?

snehara99:
> @snehara99 how does this change impact existing users who have configured their settings or did not configure their settings?

It shouldn't at all

PR #3548, opened 2024-01-25
Presets V6 support (package and workflow)

andreeis:
This PR implements support for two new types of presets (package and workflow), following when possible the same design as build and test presets across the code base: similar classes, objects and APIs, same command palette actions or UI entry-points in project status view (except some status bar visibility default differences), same logic related to the environment where it would run or settings used/overwritten/expanded, event notifications, custom/default "add new" templates, ... etc ...

Below are various differences:
- by default, the package and workflow presets are hidden from the status bar, unless the visibility changes determine otherwise. Same visibility settings in the project status view as the other presets.
- the workflow preset is not tied to the configure preset selected for the whole project. The workflow can run build/test/or package steps from within the context of another configure preset, while setting everything back at the end of the execution
- while a workflow preset does not have any compatibility requirements between its own configure (step0) and the project configure, all the steps in the workflow must have the associated configure preset be exactly the one at step0.

Additionally, this PR fixes some long standing bugs (or not optimal functionalities) specific to all presets, like:
- when a configure preset is unset (because of various reasons, like failing validation of presets json file or exception or errors in expansion...) the extension used to erase from the workspace state the information about build/test/(and now package/workflow as well) presets that were associated with the configure preset that is about to be unset. Later on, we would go through some guess logic to spare the user from selecting all those presets again from UI when that configure preset is again set with success. With this PR, we don't delete the info from workspace state. If the configure preset is ever set again with success, we recover the info and all other presets are selected automatically. If the user wants a different selection of the other presets, we will store a different association in the workspace state that will not be deleted if the configure preset is unset again
- this PR also adds structure validation on top of json schema syntax validation. We detect and fail to load any preset if we have unresolved references (like "configurePreset" pointing  to a name that does not exist in the list of configure presets or when the workflow steps "name" cannot be found in the array of presets of type "type" (in the context of workflow step or when there is a configure preset mismatch).
- there was a bug that made presets show up in UI as "No preset is selected" even if the internal data would point to a valid preset (of any of the 5 kinds) while DisplayName was empty. Now it would be seen with the string used in the name field.
- skip over unnecessary string expansion (and get rid of lots of logging output) when the given string would not contain at all the pattern that normally triggers expansion, like ${var}

More comments in the code, more regression tests and detailed documentation in the docs folder will be following but we are merging this PR now as is (fully functional and thoroughly tested) because of the strings localization freeze deadline.


package.json
```diff
@@ -1914,6 +2064,15 @@
           ],
           "scope": "machine"
         },
+        "cmake.cpackArgs": {
```
gcampbell-msft, line 2067: Should there also be a setting for the `cpack` executable?

src/config.ts
```diff
@@ -316,6 +322,12 @@ export class ConfigurationReader implements vscode.Disposable {
     get rawCTestPath(): string {
         return this.configData.ctestPath;
     }
+    get rawCPackPath(): string {
+        return this.configData.cpackPath;
+    }
+    get cpackVars(): {[key: string]: boolean | number | string | string[] | util.CMakeValue} {
```
gcampbell-msft, line 328: Are these acutally defined? 

I don't see a setting defined for this in the package.json.

src/config.ts
```diff
@@ -340,6 +352,12 @@ export class ConfigurationReader implements vscode.Disposable {
     get ctestDefaultArgs(): string[] {
         return this.configData.ctestDefaultArgs;
     }
+    get packEnvironment() {
```
gcampbell-msft, line 355: What is the difference between `cpackVars` and `cpackArgs`? It feels like these should be the same settings. 

Also, I don't see these defined in the package.json.

src/drivers/cmakeDriver.ts
```diff
@@ -305,6 +305,21 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+    /**
+     * Get the environment variables that should be set at CPack and packaging time.
+     */
+    async getCPackCommandEnvironment(): Promise<Environment> {
+        if (this.useCMakePresets) {
+            return EnvironmentUtils.create(this._packagePreset?.environment);
+        } else {
+            let envs = this._kitEnvironmentVariables;
+            envs = EnvironmentUtils.merge([envs, await this.computeExpandedEnvironment(this.config.environment, envs)]);
+            envs = EnvironmentUtils.merge([envs, await this.computeExpandedEnvironment(this.config.testEnvironment, envs)]);
```
gcampbell-msft, line 317: This shouldn't be here, this is the `testEnvironment`, but this is the method for the cpack. 

Additionally, in the `useCMakePresets` branch, we now support using the settings to adjust the environment. This should be added.

src/paths.ts
```diff
@@ -225,6 +225,29 @@ class Paths {
         }
     }
 
+    async getCPackPath(wsc: DirectoryContext, overWriteCMakePathSetting?: string): Promise<string | null> {
+        const cpackPath = await this.expandStringPath(wsc.config.rawCPackPath, wsc);
+        if (!cpackPath || cpackPath === 'auto' || overWriteCMakePathSetting) {
```
gcampbell-msft, line 230: Is this because we do this in other locations in the codebase? What does the `auto` string actually do for us? I've not noticed this before, so I may just be ignorant to this feature.

src/preset.ts
```diff
@@ -324,6 +327,45 @@ export interface TestPreset extends Preset {
...
+    name: string;
+}
+
+export interface WorkflowPreset extends Preset {
+    name: string;
+    displayName?: string;
+    description?: string;
+    vendor?: VendorType;
+    steps: WorkflowStepsOptions[];
+    // Private fields
```
gcampbell-msft, line 364: What are these for? These aren't part of the docs for WorkflowPreset, so I'm wondering if there is another way that we can do this?

src/presetsController.ts
```diff
@@ -1129,12 +1540,25 @@ export class PresetsController {
...
             originalPresetsFile[presetType] = [];
         }
-        originalPresetsFile[presetType]!.push(newPreset);
+
+        switch (presetType) {
+            case "configurePresets":
+            case "buildPresets":
+            case "testPresets":
+            case "packagePresets":
+                originalPresetsFile[presetType]!.push(newPreset);
```
gcampbell-msft, line 1555: Why the difference in these two cases? But the other cases don't need differences?

gcampbell-msft (changes requested):

package.json
andreeis: Yes, I missed adding it in the package.json but it is in all other places.

src/config.ts
andreeis: In the beginning when things were less clear I think I had in mind something like configure variables (which now I see it doesn't exist as a setting either). I'll remove all traces of cpackVars. The "-D" defines can be sent via cpackArgs.

src/config.ts
andreeis: Exactly, I removed cpackVars in my recent update.

src/drivers/cmakeDriver.ts
andreeis: Thank you for reminding me of the more recent work done regarding settings and overrides. For the first IF path, I applied what I found is new in the CMake Tools code base. For the ELSE path, indeed I forgot code from the CTest case, which also has non-preset functionality but CPack does not function in that case.

src/preset.ts
andreeis: I later added functionality to look at every workflow step __binaryDir and __generator and didn't need these variables defined globally for a workflow (although, all the workflow steps need to have the same associated configure preset, meaning same __binaryDir and __generator for all, so we could still have one set of these variables defined globally). I removed them now.

src/presetsController.ts
andreeis: I had red squiggle because:

"Argument of type 'ConfigurePreset | BuildPreset | TestPreset | PackagePreset | WorkflowPreset' is not assignable to parameter of type 'WorkflowPreset'. Property 'steps' is missing in type 'ConfigurePreset' but required in type 'WorkflowPreset'."

If steps were not a required field I think the ".push" would work for all 5 preset types.

package.json
```diff
@@ -1158,10 +1252,26 @@
           "when": "cmake:enableFullFeatureSet",
           "command": "cmake.ctest"
         },
+        {
+          "when": "cmake:enableFullFeatureSet && useCMakePresets",
```
gcampbell-msft, line 1256: Just to keep the same style and pattern, please switch the order of these to have `command` first. It seems like there are some others that were already there like that, but we really should have `command` first in all of them.

package.json
```diff
@@ -1578,9 +1714,19 @@
           "group": "inline"
         },
         {
```
gcampbell-msft, line 1716: A similar command is missing for `viewPackageSettings`, which would show up in the Project Status View

src/projectStatus.ts
```diff
@@ -773,6 +929,52 @@ class TestPreset extends Node {
     }
 }
 
+class PackagePreset extends Node {
+
+    async initialize(): Promise<void> {
```
gcampbell-msft, line 934: Is there a reason we don't have code that lets us know that there are cpackArgs or cpackEnvironment settings that might override? Similar to the Configure/Build/Test presets?

package.nls.json
```diff
@@ -47,6 +51,10 @@
     "cmake-tools.command.cmake.cleanRebuildAll.title": "Clean Rebuild All Projects",
     "cmake-tools.command.cmake.ctest.title": "Run Tests",
     "cmake-tools.command.cmake.ctestAll.title": "Run Tests for All Projects",
+    "cmake-tools.command.cmake.cpack.title": "Run CPack",
```
gcampbell-msft, line 54: When playing with the extension, I wondered if we wanted to modify this name to include the fact that this is running cpack for a package preset? 

Currently I don't think there is support for just generically running cpack, similar to how there isn't support for just running cmake generically, so I would imagine it could be useful to have something similar to "Package Preset" or "Run Package Preset" or "Run CPack for Package Preset", something of the sort.

package.nls.json
```diff
@@ -147,6 +156,8 @@
     "cmake-tools.configuration.cmake.configureEnvironment.description": "Environment variables to pass to CMake during configure. When using CMake Presets, these are temporarily added to the environment provided by the active configure preset.",
     "cmake-tools.configuration.cmake.buildEnvironment.description": "Environment variables to pass to CMake during build. When using CMake Presets, these are temporarily added to the environment provided by the active build preset.",
     "cmake-tools.configuration.cmake.testEnvironment.description": "Environment variables to pass to CTest. When using CMake Presets, these are temporarily added to the environment provided by the active test preset.",
+    "cmake-tools.configuration.cmake.cpackArgs.description": "Additional arguments to pass to CPack. When using CMake Presets, these arguments are temporarily added to the arguments provided by the active package preset.",
+    "cmake-tools.configuration.cmake.packEnvironment.description": "Environment variables to set when packaging with CPack. When using CMake Presets, these are temporarily added to the environment used for CPack runs.",
```
gcampbell-msft, line 160: should this be `cpackEnvironment`? In order to get closer to `cpackArgs`?

src/cmakeProject.ts
```diff
@@ -135,6 +139,9 @@ export class CMakeProject {
     private disposables: vscode.Disposable[] = [];
     private readonly onUseCMakePresetsChangedEmitter = new vscode.EventEmitter<boolean>();
     public readonly cTestController: CTestDriver;
+    // do we need this?
```
gcampbell-msft, line 142: Is this a stale comment?

src/cmakeProject.ts
```diff
@@ -236,13 +245,21 @@ export class CMakeProject {
         if (this.configurePreset) {
             await this.workspaceContext.state.setBuildPresetName(this.folderName, this.configurePreset.name, null, this.isMultiProjectFolder);
             await this.workspaceContext.state.setTestPresetName(this.folderName, this.configurePreset.name, null, this.isMultiProjectFolder);
+            await this.workspaceContext.state.setPackagePresetName(this.folderName, this.configurePreset.name, null, this.isMultiProjectFolder);
```
gcampbell-msft, line 248: I might be misunderstanding this code, but why do we need to check for `this.configurePreset` in order to set a package preset name? It feels odd to set the other preset names based on the configure preset

src/projectStatus.ts
gcampbell-msft: Also, assuming you add this, make sure it reacts to real time settings changes.

src/cmakeProject.ts
```diff
@@ -513,6 +686,17 @@ export class CMakeProject {
     }
     private readonly _ctestEnabled = new Property<boolean>(false);
 
+    /**
```
gcampbell-msft, line 689: What are these for?

src/cmakeProject.ts
```diff
@@ -627,7 +811,7 @@ export class CMakeProject {
         if (drv) {
             await drv.asyncDispose();
         }
-        for (const disp of [this.statusMessage, this.targetName, this.activeVariant, this._ctestEnabled, this.isBusy, this.variantManager, this.cTestController]) {
+        for (const disp of [this.statusMessage, this.targetName, this.activeVariant, this._ctestEnabled, this.isBusy, this.variantManager, this.cTestController, this.cPackageController/*, this.cWorkflowController*/]) {
```
gcampbell-msft, line 814: left over comment?

src/cmakeProject.ts
```diff
@@ -1963,7 +2156,29 @@ export class CMakeProject {
 
     async ctest(): Promise<number> {
         const drv = await this.preTest();
-        return (await this.cTestController.runCTest(drv)) ? 0 : -1;
+        const retc = await this.cTestController.runCTest(drv);
```
gcampbell-msft, line 2159: Why the change to check for if `retc` is defined? The return type of `runCTest` is a `number`, so it should never return `undefined`.

gcampbell-msft: We should just be returning retc.

src/cmakeProject.ts
```diff
@@ -1963,7 +2156,29 @@ export class CMakeProject {
...
+        return (retc) ? 0 : -1;
+    }
+
+    async cpack(): Promise<number> {
+        this.isBusy.set(true);
+        const drv = await this.preTest();
+        const retc = await this.cPackageController.runCPack(drv);
```
gcampbell-msft, line 2166: Same comment here, why aren't we just returning `retc`? I think doing this would break the integrity of what cpack actually returned.

src/config.ts
```diff
@@ -158,6 +185,10 @@ export interface ExtensionConfigurationSettings {
     configureEnvironment: Environment;
     buildEnvironment: Environment;
     testEnvironment: Environment;
+    packEnvironment: Environment;
+    cpackPath: string;
+    cpackArgs: string[];
+    // no settings needed for workflow presets execution?
```
gcampbell-msft, line 191: Stale comment. I would agree that no settings are needed, since it actually calls into other presets execution stuff.

src/config.ts
```diff
@@ -386,6 +420,12 @@ export class ConfigurationReader implements vscode.Disposable {
     get ctestDefaultArgs(): string[] {
         return this.configData.ctestDefaultArgs;
     }
+    get packEnvironment() {
```
gcampbell-msft, line 423: See other comment, but should this be `cpackEnvironment`? I'm fine with it either way, just trying to think through what would be best.

src/cpack.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+}
+
+export class CPackDriver implements vscode.Disposable {
+    constructor(readonly ws: DirectoryContext) {}
+
+    // TODO: evaluate whether files like CPackSourceConfig.cmake or CPackConfig.cmake should have any impact on the package presets functionality
+    // same as CTestTestfile.cmake looks to have on test presets. Remove if not necessary and also review testingEnabled in ctest.ts
+    // (it is set/unset according to some logic but never queried).
+    private _packagingEnabled: boolean = false;
```
gcampbell-msft, line 31: What are these enabled settings? What determines them?

src/cpack.ts
```diff
@@ -0,0 +1,142 @@
...
+            log.info(localize('cpack.path.not.set', 'CPath path is not set'));
+            return -2;
+        }
+
+        let cpackArgs: string[];
+        if (driver.useCMakePresets && !driver.packagePreset) {
+            log.error(localize('package.preset.not.set', 'Package preset is not set'));
+            return -3;
+        } else {
+            const opts = driver.expansionOptions;
```
gcampbell-msft, line 114: Ah, this code makes it make more sense why `Run CPack` is the better command name, please ignore my other comment about possibly renaming that command.

package.nls.json
gcampbell-msft: Ignore this, I commented somewhere else why I realized that this actually makes the most sense. Resolving.

src/drivers/cmakeDriver.ts
```diff
@@ -333,6 +333,25 @@ export abstract class CMakeDriver implements vscode.Disposable {
         }
     }
 
+    /**
+     * Get the environment variables that should be set at CPack and packaging time.
+     */
+    async getCPackCommandEnvironment(): Promise<Environment> {
```
gcampbell-msft, line 339: Just as a note, we should test that the `NOTE` log statement only shows up in reasonable places. I think it likely does from reading all of the other code, but just making sure.

src/drivers/cmakeDriver.ts
```diff
@@ -600,6 +631,48 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+        } else {
+            log.info(localize('unsetting.package.preset', 'Unsetting package preset'));
+        }
+
+        await this.doSetPackagePreset(async () => {
+            await this._setPackagePreset(packagePreset);
+        });
+    }
+
+    private async _setPackagePreset(packagePreset: preset.PackagePreset | null): Promise<void> {
```
gcampbell-msft, line 650: Nit: (NOT TO BE ADDRESSED IN THIS PR). 

I've noticed many uses of null vs undefined in this repo, and at some point we might should refactor to only using `undefined`

src/extension.ts
```diff
@@ -826,6 +874,7 @@ export class ExtensionManager implements vscode.Disposable {
             this.activeTestPresetSub = cmakeProject.onActiveTestPresetChanged(FireNow, p => {
                 this.statusBar.setTestPresetName(p?.displayName || p?.name || '');
             });
+            // no package/workflow info in status bar
```
gcampbell-msft, line 877: We need to make a single decision regarding whether we show this in the status bar. I think need to be able to. 

Currently in this code, we have mixed signals. I see in the code (and tested to confirm this is the case) that we allow for settings.json settings to configure the visibility of package and workflow settings in the status bar. However, this information here says that opposite.

src/preset.ts
```diff
@@ -325,6 +328,42 @@ export interface TestPreset extends Preset {
...
+    generators?: string[];
+    variables?: { [key: string]: string | null | undefined };
+    configFile?: string;
+    output?: PackageOutputOptions;
+    packageName?: string;
+    packageVersion?: string;
+    packageDirectory?: string;
+    vendorName?: string;
+
+    // Private fields
```
gcampbell-msft, line 349: Didn't notice this earlier when I commented on the same thing in WorkflowPresets. I think, if possible, we should keep these interfaces as exact as possible compared to the Presets docs definition. 

Is there a way to reference these without having "private fields"?

src/preset.ts
```diff
@@ -338,11 +377,24 @@ interface Toolset {
 /**
  * Should NOT cache anything. Need to make a copy if any fields need to be changed.
  */
+
+// the strings don't match the spec???
```
gcampbell-msft, line 381: What strings don't match what spec?

src/preset.ts
```diff
@@ -1649,6 +2072,39 @@ export function testArgs(preset: TestPreset): string[] {
...
+    // Output
+    if (preset.output) {
+        preset.output.verbose && result.push('-V');
+        preset.output.debug && result.push('--debug');
+    }
+
+    return result;
+}
+
+// TODO: If needed, implement workflowArgs (or workflowStepsArgs) to return an array of args for each contained preset.
```
gcampbell-msft, line 2106: Stale comment?

src/workflow.ts
```diff
@@ -0,0 +1,160 @@
...
+import { ProjectController } from './projectController';
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+import * as logging from './logging';
+import { ConfigureTrigger } from './cmakeProject';
+const log = logging.createLogger('workflow');
+
+// class WorkflowOutputLogger implements OutputConsumer {
```
gcampbell-msft, line 17: stale comments

src/workflow.ts
```diff
@@ -0,0 +1,160 @@
...
+        for (const step of workflowPreset.steps) {
+            switch (step.type) {
+                case "configure":
+                    newConfigurePreset = getPresetByName(allConfigurePresets(driver.workspaceFolder), step.name);
+                    if (!newConfigurePreset) {
+                        log.error(localize('workflow.configure.preset.does.not.exist', 'The workflow step references a non existing configure preset: {0}', step.name));
+                        return -3;
+                    }
+
+                    if (step.name !== workflowPreset.steps[0].name) {
```
gcampbell-msft, line 75: Do we need this both here and above the switch statement? Technically it's double protection, but if we don't need it in both places, we should remove one (probably the one above the switch)

gcampbell-msft (changes requested):
There are still some changes that might need to be made, so I'll mark it as `Request Changes` only because there are various stale comments that should be removed and a couple of things I want to make sure we at least have a decision on before merging. 

However, overall, this looks very good and from my slight testing (I'll test more in the bug bash), it looks great!

src/status.ts
```diff
@@ -536,6 +536,7 @@ class BuildButton extends Button {
     }
 }
 
+// No status bar elements for Package and Workflow presets
```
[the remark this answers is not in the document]
sinemakinci1: I think we should allow the choice for both the status bar and side bar to be consistent.

src/cmakeTaskProvider.ts
```diff
@@ -551,6 +575,83 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
...
+            if (this.preset) {
+                packagePreset = await project?.expandPackagePresetbyName(this.preset);
+                if (!packagePreset) {
+                    log.debug(localize("package.preset.not.found", 'Package preset not found.'));
+                    this.writeEmitter.fire(localize("cpack.failed", "Package preset {0} not found. CPack failed.", this.preset) + endOfLine);
+                    this.closeEmitter.fire(-1);
+                    return;
+                }
+            }
+            const result: number | undefined = cmakeDriver ? await project?.runCPack(cmakeDriver, packagePreset, this) : undefined; //.runCPackCustomized(cmakeDriver, packagePreset, this) : undefined;
```
benmcmorran, line 601: What does the commented part of this line do?

src/cmakeTaskProvider.ts
```diff
@@ -551,6 +575,83 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
...
+                this.closeEmitter.fire(0);
+            }
+        }  else {
+            log.debug(localize("cmake.driver.not.found", 'CMake driver not found.'));
+            this.writeEmitter.fire(localize("cpack.failed", "CPack run failed.") + endOfLine);
+            this.closeEmitter.fire(-1);
+        }
+    }
+
+    // should we have one task per workflow or launch the specific tasks for each entry
```
benmcmorran, line 616: Not sure if I completely understand this, but wouldn't we need a task per workflow so that the workflows can programmatically invoked by other extensions?

src/config.ts
```diff
@@ -69,6 +69,19 @@ export interface AdvancedOptionConfig {
...
+        inheritDefault?: StatusBarInheritOptionVisibility;
+        statusBarLength?: number;
+    };
+    workflowPreset?: {
+        projectStatusVisibility?: ProjectStatusOptionVisibility;
+        statusBarVisibility?: StatusBarOptionVisibility;
+        inheritDefault?: StatusBarInheritOptionVisibility;
+        statusBarLength?: number;
+    };
+    // No room for package/workflow presets in status bar
```
benmcmorran, line 84: Is this just referring to the defaults? I see `statusBarVisibility` for package and workflow presets.

schemas/CMakePresets-v6-schema.json
```diff
@@ -0,0 +1,1579 @@
...
+                    "displayName": {
+                        "type": "string",
+                        "description": "An optional string with a human-friendly name of the preset."
+                    },
+                    "description": {
+                        "type": "string",
+                        "description": "An optional string with a human-friendly description of the preset."
+                    },
+                    "steps": {
+                        "type": "array",
```
gcampbell-msft, line 1287: This may be where you need to add `additionalProperties` as false. 

Also, if you want a schema to search through for comparison. https://cmake.org/cmake/help/latest/_downloads/3e2d73bff478d88a7de0de736ba5e361/schema.json

package.json
andreeis: Great find, yes, it was missing. Added now.

src/projectStatus.ts
andreeis: I did not realize that was missing after the last merge conflicts of bringing over latest functionality.

package.nls.json
andreeis: For testing, I saw ctestArgs and testEnvironment. That's why I went with cpackArgs and packEnvironment. But I agree cpackEnvironment fits better.

src/cmakeProject.ts
andreeis: Yes, thank you for catching it.

src/cmakeProject.ts
andreeis: I added the package and workflow case here because of the others. Also, note that we are not setting the names of the build/test/package/workflow presets based on the configure preset name. We are actually setting the names of the build/test/package/workflow presets to null.
I'm not sure I understand 100% either and also I wouldn't change that now, but it looks that if a configure set failed for any reason, we call this resetPresets API in order to not remember the association between the failed configure preset and whatever build/test/package/workflow presets it had associated before. I actually think we should leave that association in place because if in future setting that configure preset will succeed, it's convenient to have all others remembered. But since this has been the behavior for a long time (regarding build/test preset) it's safest to mimic the same for package/workflow.

src/cmakeProject.ts
andreeis: I don't know either. For package presets, I am only following the same as for test presets.
For test presets, it looks like these boolean variables are set depending on the presence on disk of some cmake files. CPack seems to consume similar input files as CTest and I kept the same coding. But even for test, I was not able to figure out what are they used for. Only what are they set to. I think we'll end up removing both.

src/cmakeProject.ts
andreeis: Yes great catch

src/cmakeProject.ts
andreeis: For cpack I did the same as for ctest. If thinking about both, I thought the number return codes greater than 0 mean success and those we would map to a 0 return code, whether any <=0 return code we'd want to be -1. I did not see it as a query for undefined but a query for positive value number. I can't seem to find documentation spec about ctest and cpack return codes.

src/cmakeProject.ts
andreeis: You are right. Do you think we should change for ctest as well?

src/cpack.ts
```diff
@@ -0,0 +1,142 @@
...
+        let cpackArgs: string[] = [];
+        if (!packagePreset && driver.packagePreset) {
+            packagePreset = driver.packagePreset;
+        }
+        if (!driver.useCMakePresets || !packagePreset) {
+            return undefined;
+        }
+
+        const opts = driver.expansionOptions;
+        const args = [];
```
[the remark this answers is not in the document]
andreeis: makes sense

src/extension.ts
andreeis: The spec says package/workflow should not be visible in status bar because limited space. After the latest FI I may have merged wrong some conflicts and I did not add correctly the settings part for package/workflow presets visibility.

src/preset.ts
andreeis: Stale comment. Spec says "Custom Workflow Preset" and we're generating "" for display name and description. This would be something specific to all presets "add new" wizard. I'll open a work item and no need to fix now before the string freeze.

src/preset.ts
andreeis: Yes but there is an idea behind worth discussing. All other presets can offer their command lines. Since the workflow is a collection of presets, we can offer a list of commands for it as well. I removed the comment and will open an issue on GitHub.

src/cmakeTaskProvider.ts
andreeis: Artefact of an older messier state of the feature implementation. Removed.

schemas/CMakePresets-v6-schema.json
andreeis: Thanks for spotting this and for the link. I wasn't aware I have such schema at my disposal for inspiration.

src/cmakeProject.ts
gcampbell-msft: Ah, I see now. That makes sense. Better to leave it as is. THanks!

src/cmakeProject.ts
gcampbell-msft: If you can confirm that it still works as expected with that change, I'm fine with it. However, noticing now that it was like this before your changes, so for safety reasons, I'm okay leaving it.

src/extension.ts
gcampbell-msft: See https://github.com/microsoft/vscode-cmake-tools/pull/3548#discussion_r1467829227, I think to provide consistency, we will allow the user to put it in the status bar, but by default, it won't be there.

src/preset.ts
gcampbell-msft: If there is a model for this in the build presets or other already existing presets, then I think it's fine. Thanks for pointing that out.

src/config.ts
gcampbell-msft: See https://github.com/microsoft/vscode-cmake-tools/pull/3548#discussion_r1467829227, I commented in another location as well, but we should allow status bar visibility, but it should only be in the project status view by default.

package.json
```diff
@@ -579,6 +626,44 @@
...
+        "category": "CMake"
+      },
+      {
+        "command": "cmake.projectStatus.workflow",
+        "title": "%cmake-tools.command.cmake.workflow.title%",
+        "when": "cmake:enableFullFeatureSet && useCMakePresets",
+        "icon": "$(run)",
+        "category": "CMake"
+      },
+      {
```
gcampbell-msft, line 655: Out of curiosity, is there any entry point from the projectStatus view?

src/presetsController.ts
```diff
@@ -734,6 +916,78 @@ export class PresetsController {
...
+
+        if (!currentWorkflowPreset) {
+            // No valid workflow preset matches the selected configure preset
+            await this.setWorkflowPreset(preset.defaultWorkflowPreset.name, false/*needToCheckConfigurePreset*/, false/*checkChangingPreset*/);
+        }
+    }
+
     private async checkConfigurePreset(): Promise<preset.ConfigurePreset | null> {
         const selectedConfigurePreset = this.project.configurePreset;
         if (!selectedConfigurePreset) {
```
gcampbell-msft, line 993: Not your change, but these options need to be "Select Configure Preset" and "Later", see the casing difference.

gcampbell-msft (approved):
Overall, I've tested and I think this looks good! There are some changes that we need to continue to monitor and test / fix on before releasing, but I think this is in a good state to be able to merge and get the strings in. 

Before merging, it would be best if you could update the description of the PR and log your design and under-the-hood changes so that if any bugs come up and we look back at this PR, it will be easier to identify what might have happened, since the diffs are so large.

gcampbell-msft (approved):

Issue #3557, opened 2024-01-31
$env{} not expanded in CMakePresets cacheVariables


doxxx:
### Brief Issue Summary

I have a `CMakePresets.json` that defines a configure preset with `cacheVariables` that reference environment variables using `$env{FOO}` in the cache variable value. For example:

```jsonc
{
    "version": 7,
    "configurePresets": [
        {
            "name": "base",
            "hidden": true,
            "binaryDir": "${sourceDir}/build",
            "generator": "Ninja Multi-Config",
            "cacheVariables": {
                "ARM_TOOLCHAIN_DIR": "$env{ARM_10_TOOLCHAIN_DIR}",
                "CMAKE_TOOLCHAIN_FILE": "${sourceDir}/arm-gcc-toolchain.cmake",
                "CMAKE_EXPORT_COMPILE_COMMANDS": "ON"
            }
        }
    ]
    // ... more presets removed for brevity
}
```

With CMake Tools v1.16.32 this works fine:

```
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.EXE" -DBOARD_TYPE=pca-00619 
-DSYSTEM_IP4_ADDR=10.10.10.23 
-DARM_TOOLCHAIN_DIR=C:/Tools/gcc-arm-none-eabi-10.3-2021.10 
-DCMAKE_TOOLCHAIN_FILE=C:/Users/gordon.tyler/dev/Synaptive/ModusV/embeddedfirmware/projects/arm-gcc-toolchain.cmake 
-DCMAKE_EXPORT_COMPILE_COMMANDS=ON 
-SC:/Users/gordon.tyler/dev/Synaptive/ModusV/embeddedfirmware/projects 
-BC:/Users/gordon.tyler/dev/Synaptive/ModusV/embeddedfirmware/projects/build 
-G "Ninja Multi-Config"
```

However, with the pre-release version of CMake Tools v1.17.8, this no longer works:

```
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.EXE" -DBOARD_TYPE=pca-00619 
-DSYSTEM_IP4_ADDR=10.10.10.23 
-DARM_TOOLCHAIN_DIR=$env{ARM_10_TOOLCHAIN_DIR} 
-DCMAKE_TOOLCHAIN_FILE=C:/Users/gordon.tyler/dev/Synaptive/ModusV/embeddedfirmware/projects/arm-gcc-toolchain.cmake 
[...]

PR #3563, opened 2024-02-01
Fix broken expansion

gcampbell-msft:
Fix for #3557. This was introduced in #3548. 

@andreeis I'm not sure why this was added, and it seems like it introduced bugs, so I'm removing it. Please let me know if there is a specific reason that it was added. 

No CHANGELOG entry because the issue is only in pre-release.

### DIFF
--- a/src/expand.ts
+++ b/src/expand.ts
@@ -5,7 +5,7 @@
 
 import * as vscode from 'vscode';
 import { createLogger } from './logging';
-import { replaceAll, fixPaths, errorToString, stringNeedsExpansion } from './util';
+import { replaceAll, fixPaths, errorToString } from './util';
 import * as nls from 'vscode-nls';
 import { EnvironmentWithNull, EnvironmentUtils } from './environmentVariables';
 import * as matchAll from 'string.prototype.matchall';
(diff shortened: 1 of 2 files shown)


moyo1997 (approved):

doxxx:
>> @andreeis I'm not sure why this was added, and it seems like it introduced bugs, so I'm removing it. Please let me know if there is a specific reason that it was added.

From the PR:
> skip over unnecessary string expansion (and get rid of lots of logging output)

I'd argue that it may be worth simply improving stringNeedsExpansion to recognize `$env{}` and `$penv{}`. The drawback is that it's a separate code site that now has to be aware of the different forms of variables, if others are added in the future.

moyo1997:
> > > @andreeis I'm not sure why this was added, and it seems like it introduced bugs, so I'm removing it. Please let me know if there is a specific reason that it was added.
> 
> From the PR:
> 
> > skip over unnecessary string expansion (and get rid of lots of logging output)
> 
> I'd argue that it may be worth simply improving stringNeedsExpansion to recognize `$env{}` and `$penv{}`. The drawback is that it's a separate code site that would have to be aware of the different forms of variables, if others are added in the future. Although a carefully crafted regex could match these and future variable forms, e.g. `\$[a-zA-Z]*\{.+?\}`.

It looks like same check (but more robust) is already done in expandstringhelper

Issue #2286, opened 2021-12-13
VS build environment is not set when using clang


bobbrow:
So what is the trick to get VSCode to call vcvars?
My preset:
```
{
    "name": "win-ninja-release",
    "displayName": "win-ninja-release",
    "description": "Ninja generator release",
    "generator": "Ninja",
    "binaryDir": "${sourceDir}/build/ninja/release",
    "inherits": ["ninja-general" , "win-vcpkg_settings" ],
    "cacheVariables": {
        "CMAKE_C_COMPILER" : "$penv{LLVMInstallDir}/bin/clang-cl.exe",
        "CMAKE_CXX_COMPILER" : "$penv{LLVMInstallDir}/bin/clang-cl.exe",
        "CMAKE_BUILD_TYPE": "Release",
    },
    "architecture": {
        "value": "x64",
        "strategy": "external"
     },
     "toolset": {
        "value": "host=x64",
        "strategy": "external"
     },
    "vendor": {
        "microsoft.com/VisualStudioSettings/CMake/1.0": {
          "hostOS": [ "Windows" ],
          "intelliSenseMode" : "windows-clang-x64",
          "enableMicrosoftCodeAnalysis" : true
        }
    }
},
```

_Originally posted by @Neumann-A in https://github.com/microsoft/vscode-cmake-tools/issues/1885#issuecomment-991155721_

bobbrow:
(Copied from #1885)

We currently only auto-configure the vcvars env for you if your clang-cl corresponds to a Visual Studio Kit we've previously scanned.

SupinePandora43:
My current workaround:

```bash
export CMAKE_C_COMPILER=clang
export CMAKE_CXX_COMPILER=clang++
cd build
cmake ..
cd ..
```

Diyou:
Without calling vcvars I'm getting
```powershell
[cmake] -- The C compiler identification is unknown
[cmake] -- The CXX compiler identification is unknown
```

And
```powershell
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.
```
When ninja is not in path eventhough it's installed with visual studio.

Seems to be unrelated to clang ... the environment just does not get set.
I need to launch vscode from a Visual Studio Developer Command Prompt (the x64 specifically because it defaults to x86 for some reason)

I'd like to keep the presets platform agnostic and free of absolute paths.

aiyolo:
[...]

Issue #3502, opened 2023-12-19
Clang-cl not working


Wolfleader101:
### Brief Issue Summary

Same issue as #2798 

![image](https://github.com/microsoft/vscode-cmake-tools/assets/25214835/58c35062-8fe9-4d1b-bb48-59f69bd73af1)

```
[cmake] CMake Error at C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.27/Modules/CMakeTestCCompiler.cmake:67 (message):
[cmake]   The C compiler
[cmake] 
[cmake]     "C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe"
[cmake] 
[cmake]   is not able to compile a simple test program.
[cmake] 
[cmake]   It fails with the following output:
```

This is only an issue with `clang-cl` , `clang` and `clang++` (GNU-Like clang) work with no issues, however i need clang-cl for windows specific libs.

My CMakePresets
```json
 "configurePresets": [
    {
      "name": "windows-clang-base",
      "hidden": true,
      "generator": "Ninja",
      "binaryDir": "${sourceDir}/out/build/${presetName}",
      "installDir": "${sourceDir}/out/install/${presetName}",
      "cacheVariables": {
        "CMAKE_C_COMPILER": "clang-cl.exe",
        "CMAKE_CXX_COMPILER": "clang-cl.exe"
      },
      "condition": {
        "type": "equals",
        "lhs": "${hostSystemName}",
        "rhs": "Windows"
      }
    },
     {
      "name": "clang-cl-x64-debug",
      "displayName": "Windows Clang x64 Debug",
      "inherits": "windows-clang-base",
      "architecture": {
        "value": "x64",
        "strategy": "external"
      },
      "cacheVariables": {
        "CMAKE_BUILD_TYPE": "Debug"
      }
    },
```

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.85.1",
  "cmtVersion": "1.16.32",
  "configurations": [
    {
      "folder": "c:\\code\\personal\\cpp\\FenrirEngine",
      "cmakeVersion": "3.27.2",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
[...]

Issue #3457, opened 2023-11-19
[Meta] Controlling the environment with presets


MathiasMagnus:
### Brief Issue Summary

This issue is borderline a discsussion and intends to collect feedback around finding a solution for a number of issues around the inability to properly control or infer the environment to drive toolchains. An honest effort to enumerate related issues:

- #2912
- #3436
- #2036
- #1897
- #1885
- #2286

Relevant upstream issues:

- [presets: sourcing environment scripts #21619](https://gitlab.kitware.com/cmake/cmake/-/issues/21619)

This linked issues, especially the upstream one has a fair amount of context. The takeaway in my read about all of it is:

- CMake (and Kitware) does not wish to concern itself with setting up an environment for configuring using makefile generators
- Users clearly want CMake to be concerned with that
  - It is a constant source of friction
  - IDE tooling can only mitigate the issue, but cannot solve it. CMake Tools can't help us in CI scenarios
- Presets already has a built-in mechanism to control the environment
  - The configuration env can propagate to build and tests in a well defined manner

The solution space is vast.

- Do what CMake Tools tries to do now and trigger extra logic for `"strategy": "external"` mixed with inspecting `CMAKE_[C|CXX]_COMPILER`
  - this won't help with shell/script/CI use-cases.
  - even in the IDE, this breaks with less popular toolchains.
    - `nvc++`, `amdclang++`, etc. I happen to be a GPGPU dev, so _this_ is my everyday use-case.
    - I constantly keep getting `Neither dumpbin, llvm-objdump nor objdump could be found. Can not take care of dll dependencies.` warnings on every target when using Vcpkg because CMake Tools cares not about the env in my preset.
- One of the issue authors advocates for supporting the `"cmakeGenerateCommand"` preset vendor extension
  - neither will this help with shell/script/CI use-cases.
- Have CMake persist env changes during configuration-time to build/test/package-time and enact env reqs in `CMAKE_TOOLCHAIN_FILE`s or `CMAKE_PROJECT_TOP_LEVEL_INCLUDES`.
  - This requires CMake changes
- Have tools respect environments setup through presets and have toolchains declare their env requirements in a preset-friendly way.
  - This is what I've been trying to advocate for in the upstream issue [here](https://gitlab.kitware.com/cmake/cmake/-/issues/21619#note_1356844).
- ??? (something else?)

[...]

Issue #2036, opened 2021-08-06
Presets: How to inherit the MSVC environment when toolchains are in use?


bobbrow:
When generating with Ninja, there is an expectation that the compilation environment is already set up.  This basically means that for MSVC projects, we need to figure out the architecture and toolset prior to invoking CMake.  When using Kits and Variants, this is not a problem because the user declares their intent.  For presets, we look at the value of `CMAKE_C|CXX_COMPILER` and if it looks like the MSVC compiler, we set up the environment (sneakily using the Kits) prior to invoking CMake.

The problem is that if MSVC is set as the compiler in a toolchain file instead of directly in the Presets file, we aren't aware of what needs to happen.  (this is actually the opposite the problem that VS has today)

Perhaps we can key off of the `architecture` and `toolset` values and require that one or both of them be set (with 'strategy == external') and if those are set, we take that as a sign that the user wants the MSVC environment configured.  According to CMake documentation [[1]](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_PLATFORM.html) [[2]](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html), these values can also drive the `Green Hills MULTI` and `XCode` generators, so we could just enable this for the `Ninja` generators and I think that would solve the issue.

cc: @esweet431

TheJCAB:
Keying it off of "toolset" == "external" definitely seems like a better idea than making people define "CMAKE_C|CXX_COMPILER".

But... the Visual Studio documentation states that it uses "CMAKE_C|CXX_COMPILER" to figure out the compiler to use. I just tried clang, and the VSCode extension doesn't seem to recognize "clang-cl.exe"

I still think "toolset" == "external" makes a lot of sense as the cue for this extension to go set up the build environment, with the architecture selection, MSVC, clang or otherwise based on the additional preset settings (with _perhaps_ adequate defaults, like targeting x64 if not specified, assuming cl.exe if not specified, etc...). If someone somehow doesn't specify "toolset", then you can wash your hands of any environment setting and just invoke CMake.

Thanx!

bobbrow:
I think the condition to configure the MSVC environment would be something like:
```ts
  // pseudocode
  // (perhaps put some warnings in place if the strategy is incorrect - Ninja will let you know, but if the developer
  // doesn't run in the terminal immediately they could be confused later)
[...]

PR #3606, opened 2024-02-22
Infer compiler from environment when using CMake presets

hokku-project:
## This change addresses item

My project is targeting 2 environments [ninja, linux, gcc, x64] and [ninja, windows, msvc, x64]
I have a following preset:

```
{
    "version": 6,
    "configurePresets": [
        {
            "name": "Example",
            "binaryDir": "...",
            "architecture": {
                "value": "x64",
                "strategy": "external"
            },
            "generator": "Ninja",
            "cacheVariables": {
                ...
            }
        }
    ]
}
```

### I expect vscode-cmake-tools:
  - use compiler kit selected using "CMake: Scan For Compilers"
  - if selected compiler is MSVC - configure environment as if vcvars*.bat was executed
  - if cmake was able to find cl.exe - it should pass a cmake compiler check

### Without this change:
  - If no compiler is provided in presets - cmake configure yields error "No CMAKE_CXX_COMPILER could be found."
  - If cl.exe is manually provided though the preset as "cl" or "cl.exe" - configuring and build steps succeed
  - If cl.exe is manually provided as an absolute path - compiler check fails, if circumvented - build fails, as it cannot find standard headers

### With this change
  - When no CMAKE_CXX_COMPILER  is provided in presets selected kit is used
  - when absolute path to cl.exe is provided though environment variables - build succeeds 
  - when cl.exe is provided - build succeeds

### Potential risks
  - I did not limit generator to Ninja and platform to win32, as in my opinion selected kit should be used as a fallback on every platform and for every generator, need your input on that

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -864,12 +864,11 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
             const compilerName: string | undefined = util.isSupportedCompiler(cxxCompiler) || util.isSupportedCompiler(cCompiler);
 
             // If CMake receives an absolute path to cl.exe it tries to perform a compiler check which is undesirable
-            if (compilerName)
-            {
-                if (cxxCompilerPath && expandedPreset.cacheVariables['CMAKE_C_COMPILER_WORKS'] === undefined) {
+            if (compilerName) {
+                if (cxxCompilerPath && path.isAbsolute(cxxCompilerPath) && expandedPreset.cacheVariables['CMAKE_C_COMPILER_WORKS'] === undefined) {
                     expandedPreset.cacheVariables['CMAKE_C_COMPILER_WORKS'] = true;
                 }
-                if (cCompilerPath && expandedPreset.cacheVariables['CMAKE_CXX_COMPILER_WORKS'] === undefined) {
+                if (cCompilerPath && path.isAbsolute(cCompilerPath) && expandedPreset.cacheVariables['CMAKE_CXX_COMPILER_WORKS'] === undefined) {
                     expandedPreset.cacheVariables['CMAKE_CXX_COMPILER_WORKS'] = true;
                 }
             }
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -809,6 +809,25 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
 
     if (preset.cacheVariables) {
         expandedPreset.cacheVariables = {};
+
+        // Lets infer the compiler from the environment, this might be overwritten later by a preset
+        if (preset.environment) {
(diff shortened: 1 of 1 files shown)


hokku-project:

oreyg:
Looking at the issue list, this change slightly improves situation with #2036, fixes #2286,  #3502
This change is also might be related to #1897

hokku-project:
@gcampbell-msft I recommend you to still take the part that can process the cl.exe if its passed as an absolute path (preset.ts:858-875)

hokku-project:
@gcampbell-msft I've updated the branch, merge it if you think that fits

src/preset.ts
```diff
@@ -836,12 +836,24 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
...
-            const cCompiler = getStringValueFromCacheVar(preset.cacheVariables['CMAKE_C_COMPILER'])?.toLowerCase();
+        if (preset.cacheVariables && expandedPreset.cacheVariables) {
+            const cxxCompilerPath = getStringValueFromCacheVar(expandedPreset.cacheVariables['CMAKE_CXX_COMPILER']);
+            const cCompilerPath = getStringValueFromCacheVar(expandedPreset.cacheVariables['CMAKE_C_COMPILER']);
+            const cxxCompiler = path.basename(cxxCompilerPath || "").toLocaleLowerCase();
+            const cCompiler = path.basename(cCompilerPath || "").toLocaleLowerCase();
             // The env variables for the supported compilers are the same.
             const compilerName: string | undefined = util.isSupportedCompiler(cxxCompiler) || util.isSupportedCompiler(cCompiler);
 
+            // If CMake receives an absolute path to cl.exe it tries to perform a compiler check which is undesirable
```
gcampbell-msft, line 847: Using the CMAKE_<LANG>_COMPILER_WORKS is risky. It is intentionally not documented by CMake. We don't want our implementation to rely on undocumented variables. 

What is the case that you're trying to solve? I've tested with a preset that uses an absolute path to the compiler and it seems to configure and build successfully.

oreyg: I've made an observation, that CMake check fails on otherwise valid compiler without a good reason. I have an assumption that it might be related to CMake version. Anyway, what I recommend to do is to @qarni to incorporate my changes, since there is only a husk of my original intentions left, and to avoid merging, and to close this issue.

gcampbell-msft: @oreyg Sorry for the delay, but just following up. We will make sure that when we get back to implementing @qarni's PR, we will keep this PR in mind and make sure that the changes this PR has that we need are incorporated. 
THanks!

tsondergaard:
See also #3457

Issue #3607, opened 2024-02-22
Wrong scope of `cmake.additionalKits`


OrkunTokdemir:
### Brief Issue Summary

The current scope of `cmake.additionalKits` is `resource` which means it is synchronized with other machines. Since the paths it contains are machine-specific paths. It doesn't make sense to share them with others. I think the correct scope should be `machine-overridable` instead of `resource`.

gcampbell-msft:
I've made comments in the PR as well, but, for now, I think that actually it is by design to be shared across users, because this setting supports macros like $(workspaceFolder) which allows for sharing across machines. This can help people share additional kits with a team. 

I'll close this issue, but if you have additional feedback please follow up. Thanks.

PR #3608, opened 2024-02-22
"Package.json: Change scope of cmake.additionalKits"

OrkunTokdemir:
## The purpose of this change
Since the paths it contains are machine-specific paths. It doesn't make sense to share them with others. The correct scope should be `machine-overridable` instead of `resource`.

Fixes: #3607

### DIFF
--- a/package.json
+++ b/package.json
@@ -3453,7 +3453,7 @@
           "items": {
             "type": "string"
           },
-          "scope": "resource"
+          "scope": "machine-overridable"
         },
         "cmake.revealLog": {
           "type": "string",


gcampbell-msft:
@OrkunTokdemir Thanks for the change. Just keeping you in the loop, we do plan to make this change but we are currently doing testing on a patch release that we will likely be doing next week. We will review and hopefully merge this after that. Thanks!

gcampbell-msft:
@OrkunTokdemir Actually, I'm not sure this change is fully necessary. While I agree with your change's idea, I believe we support expansion of this field, which would mean that things like `workspaceFolder`, which means that it can be easily shared between developers if these macros are used.

gcampbell-msft:
Since this is helpful if people use macros and can then share additional kits with teammates and other developers, I will close this for now. However, please feel free to continue a discussion about your thoughts on this here, and if we decide differently in the future, we can re-open.

PR #3612, opened 2024-02-23
Add option to enable launching cmake.exe from a VS dev env

qarni:


src/preset.ts
```diff
@@ -868,7 +869,7 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
                     shell: true
                 }).result;
 
-                if (!compilerLocation.stdout) {
+                if (vscode.workspace.getConfiguration('cmake').get('enableVSDevEnv')) {
```
qarni, line 872: @gcampbell-msft is it okay to call getconfig inline like this?

gcampbell-msft: Yes, this is the way to do this. 👍

src/preset.ts
```diff
@@ -868,7 +869,7 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
                     shell: true
                 }).result;
```
gcampbell-msft, line 871: The `compilerLocation` variable above can be removed.

gcampbell-msft: In addition, now that we aren't going to be using `where.exe` to determine whether to include the VS environment, we can remove the `whereOutput` variables and corresponding code.

gcampbell-msft (changes requested):
A couple of things can be removed, but overall this looks great. Marking as `request changes` for two reasons: some things need to be removed and pending our patch release.

gcampbell-msft:
Also, we need to make sure that we change the order that this happens. We discussed this offline, but now that we're by default including the VS developer prompt environment (based on the setting), we should start with this environment, and allow the preset to override it. 

Include VS Dev environment, overlay the presets expanded env. 

The best way to do this is likely to merge the vs dev environment on top of `process.env` above this line: `preset.environment = EnvironmentUtils.mergePreserveNull([process.env, preset.environment]);` which is near the top of the `expandConfigurePreset`. For code style, take it into it's own method.

gcampbell-msft:
We should decide whether this setting applies to CMakePresets, Kits, or both. Currently, the way that the setting is written, we are supporting it for both kits and presets, but the code is only for CMakePresets. I'm okay with either, but we should make sure to make it clear in the settings and make sure the implementation matches. 

@sinemakinci1 Do you have thoughts here?

sinemakinci1:
> We should decide whether this setting applies to CMakePresets, Kits, or both. Currently, the way that the setting is written, we are supporting it for both kits and presets, but the code is only for CMakePresets. I'm okay with either, but we should make sure to make it clear in the settings and make sure the implementation matches.
> 
> @sinemakinci1 Do you have thoughts here?

If we are having a setting that is widely applicable, I think it should implement for Kits as well so its a cohesive experience

gcampbell-msft:
> > We should decide whether this setting applies to CMakePresets, Kits, or both. Currently, the way that the setting is written, we are supporting it for both kits and presets, but the code is only for CMakePresets. I'm okay with either, but we should make sure to make it clear in the settings and make sure the implementation matches.
> > @sinemakinci1 Do you have thoughts here?
> 
> If we are having a setting that is widely applicable, I think it should implement for Kits as well so its a cohesive experience

@qarni We should keep this in mind. It may be that this already somewhat happens when the kit selected is a kit coming from VS, but we should make sure we cover those cases either way.

src/preset.ts
qarni: I think we still need whereExecutable here right https://github.com/microsoft/vscode-cmake-tools/blob/6ea698014683af1a4984bb419574448bea8eb93f/src/preset.ts#L946

gcampbell-msft: We could do that, but we don't really need to. If the point of this change is to rely on the setting, then we shouldn't have any "magic". We should apply the vs dev environment if the setting is on, and don't if not.

src/preset.ts
```diff
@@ -841,123 +841,84 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
...
-                            const cmakeGeneratorInstanceNormalized = path.normalize(cmakeGeneratorInstance);
-                            vsInstall = vsInstalls.find((vs) => vs.installationPath
-                                    && path.normalize(vs.installationPath) === cmakeGeneratorInstanceNormalized);
-
-                            if (!vsInstall) {
-                                log.warning(localize('specified.vs.not.found',
-                                    "Configure preset {0}: Visual Studio instance specified by {1} was not found, falling back on default instance lookup behavior.",
-                                    preset.name, `CMAKE_GENERATOR_INSTANCE="${cmakeGeneratorInstance}"`));
-                            }
+            if (compilerName && vscode.workspace.getConfiguration('cmake').get('enableVSDevEnv')) {
```
gcampbell-msft, line 844: At some point we may want to fully rely on only the `enableVSDevEnv` check, and not depend on what the compiler is, but I think this is good for now since we know that the VS dev env can't provide anything for any other compilers.

gcampbell-msft (approved):

oreyg:
One thing I don't like it's that this option is set globally for a workspace, what I propose instead is to utilize "vendor" field in preset (as mentioned in cmake documentation)
```json
"configurePresets": [
    {
      ...
      "vendor": {
        "microsoft.com/vscode/1.0": {
          "enableVSDevEnv": true
        }
    }
```

PR #3614, opened 2024-02-25
Chinese localization improvement

SHIINASAMA:
### This changes i18n.

- Modified the Chinese localization translation

## The purpose of this change

The term ‘cleanConfigureAll’ implies performing cleaning first and then proceeding with the reconfiguration.  This represents a sequential order rather than modification or dependency.  As a result, the sentence lacks clear meaning.  Using ‘并’ accurately conveys the practical intent of the operation.


gcampbell-msft:
@SHIINASAMA Since we have our own internal localization pipeline, we won't merge this fix directly. However, I'll open a bug internally to ensure that this feedback gets brought up with our internal localization, and if our loc team agrees with the change, it will get merged in a separate PR. I'll leave this PR open until I respond with an update.

gcampbell-msft:
Link for internal issue for team members: https://ceapex.visualstudio.com/CEINTL/_workitems/edit/950416.

gcampbell-msft:
@SHIINASAMA This change was merged here: https://github.com/microsoft/vscode-cmake-tools/pull/3634!

Just as an FYI, the reason we had to make the exact same change you made but through our internal channels, is simply because manual changes will be overwritten by our automation process if they are not originated through our localization pipeline. 

Thank you for the contribution!

PR #3618, opened 2024-02-28
fix loc string reference

gcampbell-msft:
Fixes an incorrect loc string reference in our package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2802,7 +2802,7 @@
                   ]
                 },
                 "statusBarLength": {
-                  "description": "%cmake-tools.configuration.cmake.options.advanced.statusBarLength.description%",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2802,7 +2802,7 @@
                   ]
                 },
                 "statusBarLength": {
-                  "description": "%cmake-tools.configuration.cmake.options.advanced.statusBarLength.description%",
+                  "description": "%cmake-tools.configuration.cmake.options.advanced.statusBarLength.markdownDescription%",
```
bobbrow, line 2805: I believe there are 12 locations in `package.json` that need to be updated because this string resource is resused.

bobbrow (approved):

PR #3620, opened 2024-03-01
1451 open build folder

hippo91:

- Adds an action in the command palette to append the current build directory in the explorer

## Other Notes/Information

For now there is no associated test because i do not know how to handle it. 
Where can i find some information on how to create a test?

### DIFF
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -2489,7 +2489,7 @@ export class CMakeProject {
     }
 
     /**
-     * Implementation of `cmake.appendBuildDirectoryToWorkspace` 
+     * Implementation of `cmake.appendBuildDirectoryToWorkspace`
      */
     async appendBuildDirectoryToWorkspace() {
         const binaryDir = await this.buildDirectory();
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1829,8 +1829,7 @@ export class ExtensionManager implements vscode.Disposable {
     /**
      * Appends the build directory of the active project to the current workspace
      */
-    async appendBuildDirectoryToWorkspace()
-    {
+    async appendBuildDirectoryToWorkspace() {
         await this.getActiveProject()?.appendBuildDirectoryToWorkspace();
     }
 


gcampbell-msft:
@hippo91 Is this purely a convenience method? It seems like it might be just as easy to navigate to the output folder in the sub-directories of the project as it is to run this command. 

I'm happy to take the change after the package.json is cleaned up (see other comment), but I want to make sure we understand the goal of the command. Thanks!

hippo91:
Thanks @gcampbell-msft .
Indeed when the build directory is inside the source directory itself, then this command is not meaningful.
However when the build is made out of the source directory this command allows to access it easily.
I am not sure i am clear...

package.json
```diff
@@ -824,6 +824,11 @@
         "title": "%cmake-tools.command.cmake.pinnedCommands.run.title%",
         "category": "CMake",
         "icon": "$(play-circle)"      
+      },
+      {
```
gcampbell-msft, line 828: Please add `"when": "cmake:enableFullFeatureSet"`

gcampbell-msft (changes requested):

package.json
hippo91: Done.
Just because i'am curious, what is the effect of "enableFullFeatureSet"?

package.json
```diff
@@ -1554,6 +1559,9 @@
         {
           "command": "cmake.projectStatus.setTestTarget",
           "when": "never"
+        },
```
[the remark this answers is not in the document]
hippo91: Yes, i forgot to remove it. Now it is done.

package.json
gcampbell-msft: Basically, we have a multi-step activation, which helps ensure that we only activate our extension when people want it activated. This `enabledFullFeatureSet` allows us to control when commands appear, so they only appear when we are fully activated.

hippo91: Ok.  Thanks for the explanation!

hippo91:
@gcampbell-msft i rebased on main. Hope everything is ok. By the way i did not add a test because i do not know how to start with this. Should i add one? If yes can you provide me some documentation/links that may help me?

Thanks.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Improvements:
 - Allow for users to add `--warn-unused-cli`. This will override our default `--no-warn-unused-cli`. [#1090](https://github.com/microsoft/vscode-cmake-tools/issues/1090)
 - Add option to disable kit scan by default when a kit isn't selected. [#1461](https://github.com/microsoft/vscode-cmake-tools/issues/1461)
 
+Features:
```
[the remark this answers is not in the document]
hippo91: @gcampbell-msft it's done. I also merged the two `Improvements` sections. Hope it is ok.

gcampbell-msft (approved):

PR #3639, opened 2024-03-07

csigs:


gcampbell-msft:
Relevant link: https://ceapex.visualstudio.com/CEINTL/_workitems/edit/968075

gcampbell-msft:
So much of this PR is already localized, I think we'll get unmanageable if we don't go ahead and merge this so that future loc PR's are smaller. Approving and merging.

gcampbell-msft (approved):

PR #3652, opened 2024-03-13
Properly set up environment for MSYS toolchains

Steelskin:
### This changes the environment for MSYS toolchains

- Set `MSYSTEM`, `MSYSTEM_PREFIX` environment variables for MSYS installations.
- Override `PATH` to prepend the MSYS paths.

## Other Notes/Information

This mimics the way MSYS sets up the environment with MSYS shells. In particular, it is necessary for the MSYS binaries to take precedence over the rest of the system or many packages may not be configured properly. For instance, if another MSYS installation is set up in `PATH`, some projects may erroneously use the wrong `sh` binary, resulting in incorrect paths during setup.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -304,26 +304,48 @@ function isMingw(bin: string): boolean {
 }
 
 async function asMingwKit(bin: string, kit: Kit): Promise<Kit> {
+    // Attempt to derive the MSYS environment name from the compiler prefix.
+    // See https://www.msys2.org/docs/environments/
+    const PREFIX_TO_ENVIRONMENT = new Map<string, string>([
+        ['/usr', 'MSYS'],
+        ['/ucrt64', 'UCRT64'],
+        ['/clang64', 'CLANG64'],
+        ['/clangarm64', 'CLANGARM64'],
+        ['/clang32', 'CLANG32'],
+        ['/mingw64', 'MINGW64'],
+        ['/mingw32', 'MINGW32'],
+    ]);
+
     const binParentPath = path.dirname(bin);
+    const prefixPath = path.dirname(binParentPath);
+    const msysPrefix = '/' + path.basename(prefixPath);
+    const msysEnvironment = PREFIX_TO_ENVIRONMENT.get(msysPrefix);
     const mingwMakePath = path.join(binParentPath, 'mingw32-make.exe');
     const mingwMakeExists = await fs.exists(mingwMakePath);
 
-    // During a scan, binParentPath must be a directory already in the PATH.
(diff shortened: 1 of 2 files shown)


src/kit.ts
```diff
@@ -304,26 +304,48 @@ function isMingw(bin: string): boolean {
...
-    kit.environmentVariables = { CMT_MINGW_PATH: ENV_PATH };
+    // binParentPath may not already be in the PATH (for instance, scanning
+    // was done from the "additionalCompilerSearchDirs" property). In addition,
+    // in order to prevent problems during cmake configuration, we should make
+    // sure that the MSYS paths are set before the rest of the path. In
+    // particular, this can be a problem for setups with multiple MSYS
+    // installations. In order to limit the scope of the environment variable,
+    // we only override the PATH if the MSYS environment is known.
+    if (msysEnvironment === undefined) {
+        kit.environmentVariables = { CMT_MINGW_PATH: `${binParentPath}` };
```
gcampbell-msft, line 334: What are the differences between `MT_MINGW_PATH` and `CMT_MINGW_PATH`? At a glance, it seems like it should be the same key.

src/kit.ts
```diff
@@ -304,26 +304,48 @@ function isMingw(bin: string): boolean {
...
+            MSYSTEM: `${msysEnvironment}`,
+            MSYSTEM_PREFIX: `${msysPrefix}`,
+            PATH: `${binParentPath}` + ';' + `${msysBinPath}` + ';${env:PATH}',
+        };
+    }
 
     if (mingwMakeExists) {
         // Check for working mingw32-make
-        const execMake = await proc.execute(mingwMakePath, ['-v'], null, { environment: { PATH: ENV_PATH }, timeout: 30000 }).result;
+        const execMake = await proc.execute(mingwMakePath, ['-v'], null, { environment: { PATH: `${binParentPath}` }, timeout: 30000 }).result;
```
gcampbell-msft, line 348: Should we use an environment from the kit? Maybe `kit.environmentVariables[CMT_MINGW_PATH]`? 

This may make future maintenance easier.

gcampbell-msft (changes requested):
Overall this looks like a reasonable change 👍, just a couple of comments a requested changes. Also, please make a CHANGELOG update and feel free to give yourself credit following the pattern already set in the CHANGELOG file.

src/kit.ts
Steelskin: Done.

Steelskin:
Thanks! I added myself to the changelog too.

gcampbell-msft (approved):

PR #3653, opened 2024-03-13
Show cmake output when version probe fails

geertj:
### This improves debuggability for users when the cmake version probe fails.

- When CMake Tools starts up, it tries to get information about the installed cmake version by executing the `cmake` binary. If the cmake invocation fails with a non-zero exit status, this change logs any standard output and standard error output.

### DIFF
--- a/src/proc.ts
+++ b/src/proc.ts
@@ -102,6 +102,7 @@ export interface ExecutionOptions {
     outputEncoding?: string;
     overrideLocale?: boolean;
     timeout?: number;
+    showOutputOnError?: boolean;
 }
 
 export function buildCmdStr(command: string, args?: string[]): string {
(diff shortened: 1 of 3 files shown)


geertj:
@microsoft-github-policy-service agree company="Amazon"

geertj:
(Apologies if it seemed like I was trying to merge this. I had imagined that the "Rebase" button would rebase the PR not merge this into main.)

src/proc.ts
```diff
@@ -207,6 +208,16 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
                 } else {
                     log.warning(localize({key: 'process.exit', comment: ['The space before and after all placeholders should be preserved.']}, 'The command: {0} exited with code: {1}', `${cmdstr}`, `${code}`));
                 }
+                if (options?.showOutputOnError) {
+                    if (stdout_acc) {
+                        const output = '> ' + stdout_acc.trimEnd().replace(/\n/g, '\n> ');
+                        log.warning(localize('process.exit.stdout', 'Command output on standard out:\n{0}', `${output}`));
+                    }
+                    if (stderr_acc) {
+                        const output = '> ' + stderr_acc.trimEnd().replace(/\n/g, '\n> ');
```
gcampbell-msft, line 217: I think that if we're going to inject newlines, it would be good to include tabs so that it looks good in the output window. Or, just let it print by itself, especially in the log.warning call. Our loggers add `[proc]` to the beginning of the message, or a tab to ensure it's clear. I would say line 217 is fine, but 218 we shouldn't put the newline in.

gcampbell-msft:
@geertj Just a couple small changes, please also add a CHANGELOG update. Thanks! Overall this looks like a good change!

gcampbell-msft (changes requested):
See other comments for small changes requested.

src/proc.ts
geertj: I was inserting that prefix because some of our error messages are multi-line. I'm not sure how common this is, so maybe we should just go with the simplest approach and let it print by itself per one of your suggestions:

```
[proc] The command: [...]/cmake-wrapper --version exited with code: 1
[proc] Command output on standard error:
> First line of multi-line error message.
> Second line of multi-line error message.
```

New output:
```
[proc] The command: [...]/cmake-wrapper --version exited with code: 1
[proc] Command output on standard error: First line of multi-line error message.
Second line of multi-line error message.
```

Happy to add a tab or other prefix to second and later lines you prefer.

gcampbell-msft: I think that if it's easy enough, it'd be best to add tabs to second or later lines. However, if you think this is risky, I'm okay without it. Approving now but feel free to make that change as well.

gcampbell-msft (approved):

src/proc.ts
geertj: That's an easy change, updated!

gcampbell-msft (approved):

Issue #3672, opened 2024-03-27
Errors setting the Windows SDK version in a Configure Preset


benschreiber:
### Brief Issue Summary

Beginning with CMake 3.27, you can specify a Windows SDK version in `CMAKE_GENERATOR_PLATFORM`. This variable can also be set with `cmake -A` or the `architecture` field in a configure preset. Consider the following presets.

```json
{
    "version": 6,
    "configurePresets": [
        {
            "name": "msvc",
            "generator": "Visual Studio 16 2019",
            "toolset": "host=x64",
            "architecture": "x64,version=10.0.19041.0",
            "cacheVariables": {
                "CMAKE_CXX_COMPILER": "cl.exe"
            }
        },
        {
            "name": "ninja",
            "generator": "Ninja Multi-Config",
            "toolset": {
                "value": "host=x64",
                "strategy": "external"
            },
            "architecture": {
                "value": "x64,version=10.0.19041.0",
                "strategy": "external"
            },
            "cacheVariables": {
                "CMAKE_CXX_COMPILER": "cl.exe"
            }
        }
    ]
}
```

From the CLI, the `msvc` preset works correctly.
```
cmake --preset msvc
...
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19045.
```

In VS Code, the `msvc` preset shows errors in the log when loading.
```
[ERROR:vcvarsall.bat] Invalid argument found : x64_x64=version
[ERROR:vcvarsall.bat] Error in script usage. The correct usage is:
```
but it does seem to work correctly?
```
[cmake] -- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19045.
```

Without specifying `version=10.0.19041.0`, I see
```
[cmake] -- Selecting Windows SDK version 10.0.22621.0 to target Windows 10.0.19045.
```

The `ninja` preset fails to find `cl.exe` when specifying `version=10.0.19041.0`, but it works without it.
```
[cmake] CMake Error at CMakeLists.txt:3 (project):
[cmake]   The CMAKE_CXX_COMPILER:
[cmake] 
[cmake]     cl.exe
[cmake] 
[cmake]   is not a full path and was not found in the PATH.
```

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.87.2",
  "cmtVersion": "1.17.17",
  "configurations": [
    {
[...]

PR #3675, opened 2024-03-29
Correctly parse target arch when version is included

gcampbell-msft:
This addresses #3672. 

The issue was that we weren't parsing the targetArch correctly when the `version` was included. See [here](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_PLATFORM.html#variable:CMAKE_GENERATOR_PLATFORM).

We pass a `includeVersion` boolean that defaults to false, because we also use this method to get the name of the script to call in arm scenarios, and we shouldn't include the version then. Only when we're getting the script args should we include the version.

### DIFF
--- a/src/installs/visualStudio.ts
+++ b/src/installs/visualStudio.ts
@@ -173,7 +173,7 @@ export function getHostTargetArchString(hostArch: string, targetArch?: string, a
     }
 
     const parsedTargetArch = targetArch.split(",");
-    const version = parsedTargetArch.filter(a => a.includes("=")).map(a => a.split("=")[1]).join(" ").trimEnd();
+    const version = parsedTargetArch.filter(a => a.includes("version=")).map(a => a.split("=")[1]).join(" ").trimEnd();
     // The platform can only be first in the list: https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_PLATFORM.html#variable:CMAKE_GENERATOR_PLATFORM
     // If the first entry in the list includes an "=", it is the version, so we need to use the host arch.
     targetArch = parsedTargetArch[0].includes("=") ? hostArch : parsedTargetArch[0];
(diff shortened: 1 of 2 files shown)


src/installs/visualStudio.ts
```diff
@@ -170,6 +172,12 @@ export function getHostTargetArchString(hostArch: string, targetArch?: string, a
         targetArch = hostArch;
     }
 
+    const parsedTargetArch = targetArch.split(",");
+    const version = parsedTargetArch.filter(a => a.includes("=")).map(a => a.split("=")[1]).join(" ").trimEnd();
```
moyo1997, line 176: my only concern here is because we don't check that they actually say "version=...", but just for = being present, this could lead to a state where their cmakepresets works in vscode, but fails on command line because of a typo for example (so inconsistency in behavior).

gcampbell-msft: Updated.

moyo1997 (approved):

src/installs/visualStudio.ts
```diff
@@ -170,6 +172,12 @@ export function getHostTargetArchString(hostArch: string, targetArch?: string, a
         targetArch = hostArch;
     }
 
+    const parsedTargetArch = targetArch.split(",");
+    const version = parsedTargetArch.filter(a => a.includes("version=")).map(a => a.split("=")[1]).join(" ").trimEnd();
```
bobbrow, line 176: There is code in `preset.ts` that parses toolsets (`parseToolset`).  Perhaps that should be made an `export` and refactored to someplace common.

gcampbell-msft: Good idea, I'll do that before changing. Thanks for that catch.

gcampbell-msft: Actually, update, I'll leave any possible refactor for later because this isn't for parsing the toolset, it's for parsing the `architecture` presets field, not toolset.

bobbrow: My bad.  It looks like that was introduced in 3.27 (after I stopped working on this).  In that case, you'd also want to update `getArchitecture` in `presets.ts` and make that the source of truth.

gcampbell-msft: Good idea, I'll make an issue to track that update.

PR #3695, opened 2024-04-08
Project Outline update

gcampbell-msft:
## The purpose of this change

This change modifies the Project Outline view to mirror what happens in Visual Studio, which shows a flat list of targets for each project. 

In this first edition of improving the Project Outline view, we only show non-generated targets and/or files. Additionally, we support the FOLDER property to group targets. 

We also slightly modified how items are viewed respective to what type of targets they are. 

For example, here is Bullet3 before and after. 
Before: 
image

After: 
image

Another example from the getting started guide from Azure RTOS:
Before:
image

After:
image


moyo1997 (approved):

spoyler:
Hello. Is there any a settings to return outline view to how it was before?

sinemakinci1:
@spoyler Hi, I opened an issue #3799 to track this request. For us to better understand your problem, why exactly would you like the prior view? Are there things you preferred about its organization of targets?

spoyler:
@sinemakinci1 Hi, Thanks for the answer.
When a project has many nested targets, then it is easier to navigate by its and the organization of the project looks similar to the arrangement of directories. With a flat list of goals, navigating the project becomes more difficult.

invy:
@sinemakinci1 CMake implies directory structure by "add_subdirectory". Also every library can have many "utility" targets, which are very seldom of interest. So basically, older view provides 'more natural filtering' based on the directory structure.
So, please add option to return the old outline view.

Issue #3588, opened 2024-02-13
Cmake extension does not prompt to select CmakeLists.txt file in 1.17


sinemakinci1:
### Brief Issue Summary

in 1.17, even when there is a CMakeLists.txt file, the Cmake Tools extension does not activate. An example repo I used can be found here: https://github.com/sinemakinci1/basic-cpp-copilot-example/tree/main/my-cmake-project
<img width="1201" alt="image" src="https://github.com/microsoft/vscode-cmake-tools/assets/99284450/d3228fc9-faf6-4f2b-ae1d-3508590127ee">

Since CMakeLists.txt file is embedded not at the top-level, it should prompt me to select a CMakeLists.txt file (this is part of the partial activation experience). Once I downgraded to 1.15, I saw the prompt, selected my CMakeList.txt file and kits, and now 1.16 and 1.17 versions of CMake Tools extension activate with the project.

sinemakinci1:
The hypothesis is that this came from a community PR to disable selecting CMakeList.txt when it is not in your root. We need to bring back that dropdown to some degree

YuriL180821:
@sinemakinci1 could you please provide any workaround to fix the problem locally on user's PC?

sinemakinci1:
@YuriL180821 Can you try the following?

1. Set this setting in your settings.json
`“cmake.sourceDirectory” : “${workspaceFolder}/<insert path to where the CMakeLists.txt lives in this project>”.` 

Then, reload the window and see if VSC will prompt for youto configure? Once prompted, you should press Yes, and see an option to select a CMakeLists.txt file up top (if embedded in a subdirectory).

In general, if you have the CMake Tools extension installed, you should be able to find the CMake: Configure command in the command palette which will fire that option to select a CMakeLists.txt file if detected in a sub-directory.

PR #3703, opened 2024-04-11
Update CMake Tools experience on open

gcampbell-msft:
Addresses #3588. 

The goal of this PR is to make it easier and more obvious for people to utilize the `cmake.ignoreCMakeListsMissing` setting. This setting allows people to turn off the searching for a CMakeLists.txt quick pick when they are opening a project that does contain a CMakeLists.txt somewhere, but they aren't intending on using the folder as a CMake project.

This changes the default for `cmake.configureOnOpen` to `true`. It also removes the "Would you like to configure" pop-ups. 

For the pop-up to select the CMakeLists.txt, we only pop this up on startup when `cmake.configureOnOpen` is true if we actually find a CMakeLists.txt within the folders that would've activated the extension. 

@vlavati FYI, this is making a slight modification from your PR here: #3276


snehara99 (approved):

vlavati:
@gcampbell-msft , thanks for the note!

I tried #3588 on current main branch and the dialog to select CMakelists is not shown only in case if "configureOnOpen" is persisted to "false". Then to configure the project it is possible to choose CMakelists.txt in vscode Explorer tree and run configure from context menu, or it is possible to run configure from the command palette (Ctrl+Shift+P).

In case of PR, if user opens folder and select "Not now" on "Would you like to configure project?", it is possible to see:
image
* on top - the dialog to select files with new option
* and on bottom right - the dialog to persist option "configureOnOpen"

In case of response "Not now", it looks confusing to see dialog to select CMakelists, but in case if option "configureOnOpen" is already persisted with "false", it is probably strange to see nothing if folder is opened.
May be it is possible to implement steps like this:
* on the folder is opened the extension tries to find projects (the configured projects in the settings or the project(CMakeLists.txt) in the root of folder)
* if nothing is found,
 * then dialog to select CMakeLists is shown if no option "cmake.ignoreCMakeListsMissing", the dialog should be with option to "ignoreCMakeListsMissing" like it is implemented in current PR, the continuation of dialog should not check option "configureOnOpen", because right now the usecase is transformed to the setup of the project.
 * else the option "configureOnOpen" is checked.

So the option "configureOnOpen" is checked only in case of existing and previously configured projects. And if user opens the folder for the first time, the dialog to select CMakeLists is shown.

I hope this workflow is possible to implement and it will not break any user experience.

src/config.ts
```diff
@@ -424,9 +424,6 @@ export class ConfigurationReader implements vscode.Disposable {
         return this.configData.cpackArgs;
     }
     get configureOnOpen() {
-        if (util.isCodespaces() && this.configData.configureOnOpen === null) {
```
bobbrow, line 427: You might want to keep a null check around for a while and flip it to true in case anyone accidentally put the default value in their settings.  You can remove the codespaces check though.

danniesim:
@gcampbell-msft At this point, we might want to reduce the scope of #3646 to:  Remove CTest project stubs of projects with ignoreCMakeLists set to true. That'll effectively make the folder scoped ignoreCMakeLists be the setting to enable/disable CMake Tools for a folder.

src/config.ts
gcampbell-msft: Good idea. Done.

src/cmakeProject.ts
```diff
@@ -876,11 +876,13 @@ export class CMakeProject {
                         fullPath: file
                     })) : [];
                     const browse: string = localize("browse.for.cmakelists", "[Browse for CMakeLists.txt]");
-                    items.push({ label: browse, fullPath: "", description: "Search for CMakeLists.txt on this computer" });
+                    const dontAskAgain: string = localize("do.not.ask.again", "[Do not ask again]");
```
elizabethmorrow, line 879: Nit: Is the common convention "Don't Ask Again" in other places in VSCode?

src/cmakeProject.ts
```diff
@@ -876,11 +876,13 @@ export class CMakeProject {
...
+                    items.push({ label: browse, fullPath: "", description: localize("search.for.cmakelists", "Search for CMakeLists.txt on this computer") });
+                    items.push({ label: dontAskAgain, fullPath: "", description: localize("do.not.ask.again.descriptoin", "Do not ask for CMakeLists.txt again in this folder. This will enable the cmake.ignoreCMakeListsMissing setting.") });
```
elizabethmorrow, line 881: Nit: spelling error "descriptoin"

src/cmakeProject.ts
sean-mcmanus: The usage in VS Code and the C/C++ extension is "Don't Show Again", see https://github.com/microsoft/vscode/blob/0991b0053f7664e5d88bb0b150822e6cc606a25a/src/vs/workbench/services/notification/common/notificationService.ts#L289 .

gcampbell-msft: Changing.

PR #3704, opened 2024-04-13
Sort pre-releases lower than releases

bobbrow:
### This changes behavior

**The following changes are proposed:**
When configuring with presets and using the Ninja generator on Windows with MSVC, deprioritize using any pre-releases installed when setting the environment.

## Other Notes/Information
I don't see a way currently to get CMake to use release compilers using just the Presets themselves.  This may be a development preference, so a setting may be warranted for this.  However, I didn't see an easy way to get settings evaluated during preset expansion.  I started work on that in case you would prefer that route, but am not finished and it could take a little while as there is a bit of plumbing involved.

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -1040,7 +1040,14 @@ async function tryApplyVsDevEnv(preset: ConfigurePreset) {
                         // that supports the specified toolset.
                         if (!vsInstall) {
                             // sort VS installs in order of descending version. This ensures we choose the latest supported install first.
-                            vsInstalls.sort((a, b) => -compareVersions(a.installationVersion, b.installationVersion));
+                            vsInstalls.sort((a, b) => {
+                                if (a.isPrerelease && !b.isPrerelease) {
+                                    return 1;
+                                } else if (!a.isPrerelease && b.isPrerelease) {
+                                    return -1;
+                                }
+                                return -compareVersions(a.installationVersion, b.installationVersion);
+                            });
 
                             for (const vs of vsInstalls) {
                                 // Check for existence of vcvars script to determine whether desired host/target architecture is supported.


gcampbell-msft:
@bobbrow To clarify, you're suggesting this because CMake doesn't work with release compilers using Presets? 

Before I approve since it seems reasonable, just wanted to get a full understanding of the context and reasoning behind the change. Thanks!

bobbrow:
> @bobbrow To clarify, you're suggesting this because CMake doesn't work with release compilers using Presets?
> 
> Before I approve since it seems reasonable, just wanted to get a full understanding of the context and reasoning behind the change. Does this stem from hitting issues with configure, build, etc?

What's happening is that Preview compilers are given priority over Release compilers because only a simple version check is happening and Preview, by definition, is a higher version number than Release.

gcampbell-msft:
@bobbrow So this is a general design choice which has CMake Tools prefer the release versions rather than prerelease? Got it.

gcampbell-msft (approved):

bobbrow:
> @bobbrow So this is a general design choice which has CMake Tools prefer the release versions rather than prerelease? Got it.

Yes, if you're ok with that.  It has the potential to change behavior for some folks unexpectedly.  If you get complaints, let me know and I might be able to wire up a setting to switch the behavior back.  But like I said above, it will require some plumbing.

gcampbell-msft:
> > @bobbrow So this is a general design choice which has CMake Tools prefer the release versions rather than prerelease? Got it.
> 
> Yes, if you're ok with that. It has the potential to change behavior for some folks unexpectedly. If you get complaints, let me know and I might be able to wire up a setting to switch the behavior back. But like I said above, it will require some plumbing.

I'm okay with that. There is a chance for different behavior, but I expect that would be mostly under the cover and most people wouldn't notice or care. We can always wire up a setting or revert if needed.

gcampbell-msft:
Approved. I'll merge when ready

PR #3717, opened 2024-04-23
CMake Quickstart

qarni:
### This improves the CMake QuickStart experience by allowing the user to dynamically create a CMakePresets.json file.

Quick Start wizard process:
1 proj name
2 proj lang
3 proj type
4 addl
5 selectTargets
6 select preset
7 select compiler

Error Checking:
Screenshot 2024-04-24 121759


gcampbell-msft:
Also, don't forget CHANGELOG.md updates

src/cmakeProject.ts
```diff
@@ -2824,92 +2844,133 @@ export class CMakeProject {
...
+        }
+
+        let init = [
             'cmake_minimum_required(VERSION 3.0.0)',
             `project(${projectName} VERSION 0.1.0 LANGUAGES ${langName})`,
-            '',
-            'include(CTest)',
-            'enable_testing()',
-            '',
+            '\n'
```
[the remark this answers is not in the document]
qarni: Yup!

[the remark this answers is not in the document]
qarni: Though it does leave an extra new line if ctest/cpack aren't included. Do we care about that?

src/presetsController.ts
```diff
@@ -392,18 +411,19 @@ export class PresetsController {
             }
 
             if (newPreset) {
-                const name = await this.showNameInputBox();
+
+                const before: preset.ConfigurePreset[] = await this.getAllConfigurePresets();
+                const name = await this.showNameInputBox() || newPreset.displayName || undefined;
```
qarni, line 416: @gcampbell-msft I thought about having a default name when the user selects a compiler for a new preset, so if they hit escape/don't write a name, it would make the preset with the default name. The current implementation uses the display name as the default name, but I think if there's certain characters cmake doesn't escape correctly(?), so would it be better to just name it "my preset" or something?

gcampbell-msft: What if we just use `newPreset.name`? We don't have to use `displayName`, and the `name` should be supported, right?

src/presetsController.ts
```diff
@@ -283,6 +284,24 @@ export class PresetsController {
...
+
+                        void vscode.window.showErrorMessage(
+                            localize('no.compilers.available', 'Cannot generate a CmakePresets.txt with Quick Start. No compilers available.'),
+                            {
+                                title: localize('learn.about.installing.compilers', 'Learn About Installing Compilers') ,
+                                isLearnMore: true
+                            })
+                            .then(item => {
+                                if (item && item.isLearnMore) {
+                                    open('https://code.visualstudio.com/docs/languages/cpp#_install-a-compiler');
```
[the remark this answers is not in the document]
qarni: Hmmm. I got this from [cmakeProjects.ts:2619](https://github.com/microsoft/vscode-cmake-tools/blob/b0dd816ec0ace305e0a4d98f7968c9ced689de00/src/cmakeProject.ts#L2619 )

src/presetsController.ts
qarni: offline discussion summary: the `name` is currently set as "placeholder" so for now we'll set the name as `displayName`

src/cmakeProject.ts
```diff
@@ -1746,7 +1748,9 @@ export class CMakeProject {
                 }
             }
         } else if (!this.configurePreset) {
-            throw new Error(localize('cannot.configure.no.config.preset', 'Cannot configure: No configure preset is active for this CMake project'));
+            await vscode.window.showErrorMessage(localize('cannot.configure.no.config.preset', 'Cannot configure: No configure preset is active for this CMake project'));
```
gcampbell-msft, line 1751: I thought we were going to remove the error message? Applies for above as well

src/cmakeProject.ts
```diff
@@ -2794,21 +2798,49 @@ export class CMakeProject {
...
 
+        // By now, we have a valid CMakeLists.txt so move onto CMakePresets.json
+        if (await fs.exists(mainPresetsFile)) {
+            log.info(localize('cmakepresets.already.configured', '[Quick Start] A CMakePresets.json is already configured.'));
+        } else {
+            if (!await this.presetsController.selectConfigurePreset(true)) {
+                await this.kitsController.setKitByName(SpecialKits.Unspecified);
+                log.info(localize('cmakepresets.not.created', '[Quick Start] CMakePresets.json not created.'));
+            } else {
+                log.info(localize('cmakepresets.created', '[Quick Start] CMakePresets.json created successfully.'));
```
gcampbell-msft, line 2820: I think we still need to set the `useCMakePresets` context value here as well, so that certain command pallete commands are visible when they should be.

gcampbell-msft (changes requested):
I think everything looks great! Just a couple more things to make sure we update.

src/cmakeProject.ts
qarni: I thought we were just changing it from the ugly throw new Error to a show error message? Should it be an info message instead? I think it should still be an error message.

gcampbell-msft: Ah, you're exactly right, I got confused from the other locations. Thanks.

src/cmakeProject.ts
qarni: Added issue #3734 to solve in a seperate PR

gcampbell-msft (approved):

PR #3730, opened 2024-04-30
Add support to configure default ActiveProject

sanore:
### The purpose of this change
When the user configured a default project in workspace configuration, during initialization it selects automatically the selected project.
* Add a workspace configuration option: "cmake.defaultActiveProject"
* Selects the configured project during initialization if configuration contains a valid project name

## Other Notes/Information
I've never used typescript before and therefore this is also my first pull request to vscode-cmake-tools. So please take a look and comment if I have forgot something. Thanks for your time! I'm open to suggestions for improvement

### Open Questions
- [x] Who is responsible for all translations?
- [x] Are there any recommendations how to setup unit tests for this feature?
- [x] Do I have forgot something?


sanore:

sanore:
Any missing information for this PR?

gcampbell-msft (changes requested):
Currently, while I like the idea of this PR, I don't believe it's ready to merge. Right now, it doesn't seem fully integrated with our current selection of the active Project support, for a couple of reasons. 

1. It is done outside of the `initActiveProject` method. While it doesn't HAVE to happen inside of it, based on the current code, `initActiveProject` is called in other places, and this method also sets the active project, so it feels like we should implement this behavior there.
2. This PR doesn't currently answer the following question: "How does this setting behave alongside the `cmake.autoSelectActiveFolder` setting? Does it override it? Does the `cmake.autoSelectActiveFolder` take precedence over the new `cmake.defaultActiveFolder`?

@sanore Marking this as "Request Changes" but please let me know if you have any questions, and also if you intend to make these changes and if we should keep this PR open for you to make those changes. For now, I'm marking this as a draft so that we can better understand the current state of open PR's. Thanks!

gcampbell-msft:
> Thank you @gcampbell-msft for you review! And yes, I would like to see this feature in the main branch
> 
> The main reason why it was implemented outside of the `initActiveProject` is that I want to keep the impact of this change as little as possibe. But I feel you prefere the ´initActiveProject´ function. So I have changed the implementation and included it in the prefered way. A small question, do you know a better way how to get the workspaceFolder based on a string? At the moment I compare the end of the path to get correct workspaceFolder, which works well, but to me it feels ugly.
> 
> To be backward compatible the `cmake.autoSelectActiveFolder` is always prefered over the new `cmake.defaultActiveFolder` configuration. Also for startup, if an active editor is open and `cmake.autoSelectActiveFolder` is enabled, it is also prefered over the `defaultActiveFolder`. Which makes sense to me, beacuase the new `defaultActiveFolder` property only should have an effect on initialization, afterwards changing the text editor should also change active folder if enabled.

I think this looks good! For your question about getting the workspace folder, could you use the `vscode.workspace.getWorkspaceFolder` method? It's already used right now in the `initActiveProject`, could it be used in your context as well? 
I'll wait for your update, but I think this PR looks pretty good now. 

One other nit, the CHANGELOG should be updated to reflect when it's actually going in based on our release numbers, could you create a new 1.20 section and Features section within it, and place this there? Thanks.

gcampbell-msft (approved):

PR #3732, opened 2024-05-01
Update CMake Tools Initial experience [exp]

gcampbell-msft:
This experimentally address #3588.

The goal of this PR is to make the initial experience of opening a folder that could potentially use the CMake Tools extension better. 

This includes making it easier and more obvious to use the `cmake.ignoreCMakeListsMissing` setting. This is done by adding an option in the quick pick to not see the quick pick again. Additionally, it we check to confirm that there is actually a CMakeLists.txt in the folder structure based on the same parameters as our activation events, before popping the quick pick. 

Lastly, we are experimentally removing the pop-ups that ask "Would you like to configure?", in favor of defaulting `cmake.configureOnOpen` to `true`.


moyo1997 (approved):

i-ky:
> Lastly, we are experimentally removing the pop-ups that ask "Would you like to configure?", in favor of defaulting `cmake.configureOnOpen` to `true`.

This does not provide the best experience if the project is supposed to be developed in container. There is no guarantee that configure will succeed while the folder is opened on host, because some of the required tools may only be installed in dev container and not on the host.

This can be worked around by setting `cmake.configureOnOpen` to `false` in `.vscode/settings.json` and to `true` in `.devcontainer/devcontainer.json` in `customizations.vscode.settings`, but this is a little bit awkward.

Issue #2577, opened 2022-05-29
Don't reconfigure when changing build types using multi-config generators


MathiasMagnus:
### Brief Issue Summary

When changing build types (Debug, Release, etc.) CMake Tools reconfigures the project, however when using multi-config generators, this may be omitted. Currently when changing configs, we can't issue a build right after, because "Another CMake operation/configuration is in progress" (or something similar) message appears. Depending on the project, a `nop` reconfiguration and subsequent generation can take 1-5 seconds.

Please remove the superfluous reconfigurations blocking builds when using multi-config generators.

_(I'm aware how this is still required to light up IntelliSense changes for extensions relying on `compile_commands.json`, but it should not block builds.)_

bobbrow:
Thanks for reporting this.  Yes, we agree that there doesn't need to be a reconfiguration for multi-config generators.

gcampbell-msft:
This was fixed in #3763, closing.

PR #3763, opened 2024-05-17
Skip full configure on variant change for multi-config generators

andreeis:
Fix for https://github.com/microsoft/vscode-cmake-tools/issues/2577. 
Don't fully configure after variant change for multi-config generators because such configures are generating all the build types and file-api jsons at once.
What is needed in this case is only to change the parameter for build and refresh IntelliSense via a configuration provider notification towards CppTools.

### DIFF
--- a/src/drivers/cmakeFileApiDriver.ts
+++ b/src/drivers/cmakeFileApiDriver.ts
@@ -271,7 +271,6 @@ export class CMakeFileApiDriver extends CMakeDriver {
             return 0;
         } else if (this.isMultiConfig && trigger === ConfigureTrigger.setVariant) {
             this._needsReconfigure = false;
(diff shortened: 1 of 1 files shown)


src/drivers/cmakeFileApiDriver.ts
```diff
@@ -269,6 +269,10 @@ export class CMakeFileApiDriver extends CMakeDriver {
             log.showChannel();
             log.info(proc.buildCmdStr(this.cmake.path, args));
             return 0;
+        } else if (this.isMultiConfig && trigger === ConfigureTrigger.setVariant) {
```
snehara99, line 272: Does the code in the else (line 275) still need to run if the only change is to prevent a reconfigure?

snehara99 (approved):

Issue #3765, opened 2024-05-17
Doubling of cpack arguments via package preset


andreeis:
### Brief Issue Summary

By mistake, arguments defined via the setting "cmake.cpackArgs" were appended twice to the command line in different entrypoints.

PR #3766, opened 2024-05-17
Fix doubling of cpack args passed by package preset

andreeis:
Fix for https://github.com/microsoft/vscode-cmake-tools/issues/3765.

### DIFF
--- a/src/cpack.ts
+++ b/src/cpack.ts
@@ -89,13 +89,6 @@ export class CPackDriver implements vscode.Disposable {
             }
         }
 
-        const opts = driver.expansionOptions;
-        const args = [];
-        for (const value of this.ws.config.cpackArgs) {
-            args.push(await expandString(value, opts));
-        }
-
-        cpackArgs = cpackArgs.concat(args);
         return cpackArgs;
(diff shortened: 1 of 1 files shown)


src/cpack.ts
```diff
@@ -89,13 +89,6 @@ export class CPackDriver implements vscode.Disposable {
             }
         }
 
-        const opts = driver.expansionOptions;
-        const args = [];
-        for (const value of this.ws.config.cpackArgs) {
```
snehara99, line 94: If this is doubling, why don't we keep this section and remove the corresponding if statements?

snehara99 (approved):

src/cpack.ts
andreeis: Good question. I figured either of the two places is a good candidate to stay or be removed. One place in here (getCPackArgs) and one in runCPack. The reason that made me chose this one would be that these args are not coming from the preset but from settings, like one additional thing defined by the users. The other switches constructed during getCPackArgs are coming from fields defined in the preset. Thank you for the review!

Issue #3767, opened 2024-05-17
Need to clean-configure when running workflow with workflow-configure-preset different than project-configure-preset


andreeis:
### Brief Issue Summary

Found issues on real world code when working on PureVirtual video. It is needed to run a clean configure when running a workflow that has a configure preset different than what the project is associated with. This is to avoid various configure errors that occur for non-clean configure.

PR #3768, opened 2024-05-17
Run clean-configure when workflow has different configure preset than project

andreeis:
Fix for https://github.com/microsoft/vscode-cmake-tools/issues/3767. 

Run clean-configure when workflow has different configure preset than project to avoid various configure errors.


benmcmorran (approved):
Changes look reasonable to me, although I'm curious why the clean step is necessary. I don't think there would be an automatic clean if you were to run a workflow preset from the CMake command line, for instance. Are we cleaning to handle cases where multiple presets re-use the same build directory?

Issue #3769, opened 2024-05-17
Fix UI issues in the CMake left panel related to package preset override of settings


andreeis:
### Brief Issue Summary

The star "*" that is displayed on top of the UI element of the package preset was not working properly on and off.

PR #3770, opened 2024-05-17
Fix UI for settings override of package presets

andreeis:
Fix for bug https://github.com/microsoft/vscode-cmake-tools/issues/3769. 

The "*" star used to represent settings overrides of various tools behind presets (like ctest.exe) was not working for cpack.exe. Some extra wiring needed to be done so that all cpack functionality was matching ctest and the UI to behave correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1738,7 +1738,7 @@
         },
         {
           "command": "cmake.projectStatus.selectPackagePreset",
-          "when": "view == cmake.projectStatus && cmake:enableFullFeatureSet && viewItem == 'packagePreset'",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1738,7 +1738,7 @@
         },
         {
           "command": "cmake.projectStatus.selectPackagePreset",
-          "when": "view == cmake.projectStatus && cmake:enableFullFeatureSet && viewItem == 'packagePreset'",
+          "when": "view == cmake.projectStatus && cmake:enableFullFeatureSet && viewItem =~ /packagePreset/",
```
snehara99, line 1741: just curious, what is this notation?

Issue #3309, opened 2023-08-21
Wrong compiler switch used to detect compiler version


jvanhamaki:
### Brief Issue Summary

Looks like wrong versionSwitch is obtained and used when configuring project which is using "chesscc" as C compiler.

Probably because the name involves "cc" which then gets "-v" as versionSwitch. This will result like this
`[proc] The command: chesscc -v exited with code: 1`

In this case correct versionSwitch would be "--version"

I could bypass the problem by renaming the compiler "chesscc" -> "chess"...

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.81.1",
  "cmtVersion": "1.15.31",
  "configurations": [
    {
      "folder": "<project secret>",
      "cmakeVersion": "3.27.1",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": false,
      "compilers": {
        "C": "/opt/asip/linux64/bin/LNa64bin/chesscc",
        "CXX": "chesscc"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "Release",
    "buildTypesSeen": [
      "Release"
    ],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 30,
    "executablesCount": 1,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
[cmake] -- Build files have been written to: <project secret>
[cmakefileapi-parser] Read reply folder: <project secret>/.cmake/api/v1/reply
[...]

PR #3777, opened 2024-05-20
Add chesscc to the allow-list of compilers.

andreeis:
Fix for https://github.com/microsoft/vscode-cmake-tools/issues/3309.
Add chesscc to the allow-list of compilers. Since we couldn't find an install to test nor the customer who requested the fix didn't reply, I am using the most common and most probably regular expression for interpreting the --version output. Same as the "cc" compiler, thinking that "chesscc" would resemble the original main "cc".

### DIFF
--- a/src/drivers/cmakeDriver.ts
+++ b/src/drivers/cmakeDriver.ts
@@ -1085,6 +1085,12 @@ export abstract class CMakeDriver implements vscode.Disposable {
             captureGroup: 1
         },
         {
+            name: "chesscc",
+            versionSwitch: "--version",
+            versionOutputRegexp: "version ([^\\s]+)",
+            captureGroup: 1
+        },
+        {
             name: "g++",
             versionSwitch: "-v",
             versionOutputRegexp: "version ([^\\s]+)",


src/drivers/cmakeDriver.ts
```diff
@@ -1084,6 +1084,12 @@ export abstract class CMakeDriver implements vscode.Disposable {
             versionOutputRegexp: "version ([^\\s]+)",
             captureGroup: 1
         },
+        {
+            name: "chesscc",
```
snehara99, line 1088: does this require any other changes for the compiler to be used properly? Has this been tested?

snehara99: nvm I see that you couldn't find the install to test

snehara99 (approved):

src/drivers/cmakeDriver.ts
andreeis: Couldn't test because I didn't find the compiler itself anywhere to install but 99%, a "chess**cc**" compiler would behave like **cc**, so I think the regexp (the same as for cc) is the right one.

Issue #3574, opened 2024-02-07
Update docs to use "scan for compilers" instead of "kits"


dc6c:
### Brief Issue Summary

The [documentation page](https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/kits.md#scan-for-kits) directs the user to run "Scan for Kits" from the command palette, but the actual command (in the current release of VSCode, v1.86.0, at least) is  "Scan for Compilers".

For better consistency, when the rest of the text talks about scanning for kits, it could say "scan for _compiler_ kits", so there is connection between the documentation and the command.

gcampbell-msft:
Likely we should merge this with #3624

gcampbell-msft:
@dc6c Could you confirm whether this is happening when you're using Presets or Kits? 

I want to confirm because while we likely should still update the docs, I think there are two different names for the commands depending on which workflow you are using. Thanks.

PR #3784, opened 2024-05-21
Update kits-variants doc with unsupported commands that are specific for presets

andreeis:
Fix for request https://github.com/microsoft/vscode-cmake-tools/issues/3574.

The kits-variants doc mentions "Scan for Kits" command which indeed is supported when kits/variants are ON (and not presets).
The user mentions that the command "Scan for Compilers" is available in the commands palette which probably happens when presets are ON (and not kits/variants).

Following the model of presets doc, which mentions all the kits-variants commands that are not available with presets on, this PR adds a paragraph enumerating all commands not available with kits-variants on. This should clear any confusion around "Scan for Kits" and "Scan for Compilers" commands.

### DIFF
--- a/docs/kits.md
+++ b/docs/kits.md
@@ -122,6 +122,21 @@ Keys:
 > **Note:**
 > To use Visual C++, both `visualStudio` and `visualStudioArchitecture` must be specified. Omitting either one won't work.
 
+## Unsupported commands
+
+The following commands are not supported when kits and variants are active:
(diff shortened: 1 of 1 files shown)


gcampbell-msft (approved):
Approved, but @andreeis  @qarni we should make sure to remember to update this if we modify this with quick start and work to default to presets.

gcampbell-msft:
Only docs, force merging.

Issue #2521, opened 2022-04-25
add option to reveal output panel only when an error occurs


scarf005:
### Brief Issue Summary

- when editing `CMakeLists.txt`, the extension automatically reconfigures cmake
- it also forcibly switches panel into output 
- it would be great to have an option to disable such behavior

### Additional Information

![image](https://user-images.githubusercontent.com/54838975/165031213-59846622-bb8b-4aa7-80e1-6fb2b7152d38.png)
before editing `CMakeLists.txt`, right panel on `terminal`
![image](https://user-images.githubusercontent.com/54838975/165031232-31310cb9-f53b-404e-afcb-6188ddef07cf.png)
after editing `CMakeLists.txt`, right panel on `output`

bobbrow:
Related: #1675

scarf005:
> "cmake.revealLog": "never"

Thank you, this option worked. However this option lacks a description, would be great to have one.
![image](https://user-images.githubusercontent.com/54838975/165399195-c813e06f-6a06-40b4-9fee-66d630ba703d.png)

ngagesmu:
It would be nice if the log was only revealed if exit code != 0

tuanpt-0634:
>It would be nice if the log was only revealed if exit code != 0
This is great if we have option "reveal on error". Or at least we can show build status (success/failed) on status bar or notification

wngtk:
How about use `"cmake.configureOnEdit": false`.

mhrstmnn:
> How about use `"cmake.configureOnEdit": false`.

Does not solve the problem, because CMake should still be configured. We just don't want to have to watch the configuration every time.

mhrstmnn:
> How about use `"cmake.configureOnEdit": false`.

Does not solve the problem, because CMake should still be configured. We just don't want to have to watch the configuration every time (unless an error occurs).

sr-tream:
Implemented this feature here: [cmake-build](https://marketplace.visualstudio.com/items?itemName=SR-team.cmake-tools-build-wrapper). This extension is a wrapper for CMake Tools, what can open output panel on errors.

NOTE: Required to call CMake commands via this extension!

PR #3785, opened 2024-05-21
implement  #2521 by adding output channel argument

stepeos:
Hi there, this is my first ts PR, so be adivesd. I really wanted this functionality so I'll throw the ball.

### The changes include:
an option to keep the output channel hidden unless there is a build or configuration error

1.  add optional boolean argument  to `showChannel` of the logging output channel
2. call showChannel after building/configuring so that the `"on error"` option can be triggered

## The purpose of this change
Building and configuring a project should not divert the programmers focus from the code when building. The user actually only needs to see the output window if an error occurs.

## Other Notes/Information
I see no other way than calling `showChannel` twice since option `"always"` expects the showChannel to pop before configuring starts while `"on error"` expects it after.


stepeos:
@gcampbell-msft any update on this? It's been months and it's not being merged?

gcampbell-msft:
@stepeos Thanks for the ping, we've had it on our backlog for awhile and just haven't gotten to it, thanks for the reminder.

gcampbell-msft:
@stepeos Overall it looks good, but could you add a CHANGELOG update?

package.json
```diff
@@ -3512,12 +3512,14 @@
           "enum": [
             "focus",
             "always",
-            "never"
+            "never",
+            "on error"
```
gcampbell-msft, line 3516: I think I'd rather have this enum value simply be `error`

src/logging.ts
```diff
@@ -23,7 +23,7 @@ enum LogLevel {
     Fatal,
 }
 
-type RevealLogKey = 'always' | 'never' | 'focus';
+type RevealLogKey = 'always' | 'never' | 'focus' | 'on error';
```
gcampbell-msft, line 26: Same comment as package.json

src/logging.ts
```diff
@@ -252,10 +252,17 @@ export class Logger {
         SingletonLogger.instance().clearOutputChannel();
     }
 
-    showChannel() {
+    showChannel(target_failed?: boolean) {
```
gcampbell-msft, line 255:
```suggestion
    showChannel(errorToShow?: boolean) {
```

gcampbell-msft: I think modifiying this makes it a little more generic about the error rather than "target_failed", somewhat of a nit change, but I think for future development it could help.

gcampbell-msft (changes requested):
Just a couple of small nit changes, otherwise I appreciate the contribution and it's a good idea! Sorry for the delay in reviewing. 

Also, could you please make a CHANGELOG update as well? Feel free to give yourself credit as other OSS contributors have in the changelog before.

src/logging.ts
stepeos: i used snake case instead for [fe04bb8](https://github.com/microsoft/vscode-cmake-tools/pull/3785/commits/fe04bb86d560ff7cd53e55066f7c2ea3c185beaa)

gcampbell-msft (dismissed):

gcampbell-msft (approved):

v-frankwang:
We verified the issue on CMake tools: v1.19.45 and it has been fixed.

ENV：
CMake Tools:v1.19.45
vscode: 1.92.2

Actual result:
image

PR #3816, opened 2024-06-05

csigs:


gcampbell-msft (approved):

gcampbell-msft:
Merging without build because it is only localized strings.

PR #3826, opened 2024-06-13
Fixes #3725 - user presets not read correctly

### LINKED ISSUE #3725: cmakeuserpresets.json not read by vscode-cmake-tools?
### Brief Issue Summary

My CMake project folder has a cmakeuserpresets.json file, but the preset(both configure preset and build preset) is not loaded by vscode-cmake-tools, so I can't use the preset defined in cmakeuserpresets.json.
Is there any option to open this feature? How can use the preset defined in cmakeuserpresets.json?

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

_No response_

qarni:
Fixes #3725 

Fixes issue where new presets couldn't inherit from configure presets in CmakeUserPresets.json, and couldn't expand from other build/test/package presets. 

Presets that have a parent in CmakeUserPresets.json are now added to CmakeUserPresets.json instead of CmakePresets.json.

Updated WorkflowPreset class to not extend Preset to adhere to https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#workflow-preset

Open to suggestions on a better interface name than "ConfigurePresetInheritPreset" :)

### DIFF
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -582,7 +582,6 @@ export class CMakeProject {
             workflowPreset,
             lightNormalizePath(this.folderPath || '.'),
             this.sourceDir,
-            this.getPreferredGeneratorName(),
             true,
             this.configurePreset?.name);
         if (!expandedWorkflowPreset) {
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -380,12 +380,8 @@ export interface WorkflowPreset {
     name: string;
     displayName?: string;
     description?: string;
-    hidden?: boolean;
-    environment?: EnvironmentWithNull;
     vendor?: VendorType;
-    condition?: Condition | boolean | null;
     isUserPreset?: boolean;
-
     steps: WorkflowStepsOptions[];
 
     __vsDevEnvApplied?: boolean; // Private field to indicate if we have already applied the VS Dev Env.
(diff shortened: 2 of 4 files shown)


src/preset.ts
```diff
@@ -564,6 +563,28 @@ function isInheritable(key: keyof ConfigurePreset | keyof BuildPreset | keyof Te
...
+
+    const originalUserPresetsFile: PresetsFile = getOriginalUserPresetsFile(folderPath) || { version: 8 };
+
+    const presetInherits = (presets: Preset[] | undefined, inherits: string | string[]) => presets?.find(p =>
+        Array.isArray(inherits)
+            ? inherits.some(inherit => inherit === p.name)
+            : inherits === p.name
+    );
+
+    if ((presetType === 'buildPresets' || presetType === 'testPresets' || presetType === 'packagePresets') &&
```
gcampbell-msft, line 577: We could avoid some duplication by pulling `(preset as ConfigurePresetInheritPreset).configurePreset` out into a variable before the if statement

src/preset.ts
```diff
@@ -564,6 +563,28 @@ function isInheritable(key: keyof ConfigurePreset | keyof BuildPreset | keyof Te
...
+        ((preset as ConfigurePresetInheritPreset).configurePreset &&
+            originalUserPresetsFile.configurePresets?.find(p => p.name === (preset as ConfigurePresetInheritPreset).configurePreset))) {
+        return true;
+    } else if (preset.inherits && presetInherits(originalUserPresetsFile[presetType], preset.inherits)) {
```
gcampbell-msft, line 581: From testing, I think we need a specific case for the "workflow preset". The workflow preset should not have the `inherit` property, it's actually an error that it extends Preset, because based on the docs here: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#workflow-preset, it shouldn't. 

Therefore, we need to check the first step of the workflow, and the configure preset that it uses, when we do this

gcampbell-msft (changes requested):
Besides a couple of suggestions and one change we need to make, this looks good. Just marking as "request changes" for those items.

src/preset.ts
qarni: Updated workflow preset to not extend Preset and removed the `inherit`, `environment`, `hidden`, `condition`, and `__inheritedPresetCondition` fields according to the cmake docs, so I had to take out some code from `expandWorkflowPreset` if you could double check that.

src/cmakeProject.ts
```diff
@@ -582,7 +582,6 @@ export class CMakeProject {
             workflowPreset,
             lightNormalizePath(this.folderPath || '.'),
             this.sourceDir,
-            this.getPreferredGeneratorName(),
```
gcampbell-msft, line 585: Why was this removed?

gcampbell-msft (approved):

PR #3835, opened 2024-06-16
The kit name for Clang/MSVC too long, put the arch at the front

lygstate:


gcampbell-msft:
@lygstate What scenarios does this break? It seems like the length of the name won't shorten any, it just changes the order. What scenario does this fix?

lygstate:
> @lygstate What scenarios does this break? It seems like the length of the name won't shorten any, it just changes the order. What scenario does this fix?

Because the arch is important, I want switch the target-arch, (and the redundant MSVC name is not important)

gcampbell-msft:
> > @lygstate What scenarios does this break? It seems like the length of the name won't shorten any, it just changes the order. What scenario does this fix?
> 
> Because the arch is important, I want switch the target-arch, (and the redundant MSVC name is not important)

Seems reasonable to me, I don't have any issues with reordering the name. I approved and kicked off the build.

gcampbell-msft (approved):

PR #3840, opened 2024-06-18
Fixes #3811 - checks for cmake.exe again if it wasnt present the previous time 

### LINKED ISSUE #3811: Installing cmake after launching VS Code results in claims of a cmake invalid location, forcing user to restart VS Code to proceed
### Brief Issue Summary

In environments where the `cmake` path points to a file that does not yet exist, installing `cmake` does not correct the issue requiring restaring VS Code to proceed.

An example is in projects where build dependencies are maintained through a local package manager such as `conda`.  When the project is loaded, the path to `cmake` for the project is invalid.  After installing the build dependencies, the `vscode-cmake-tools` extension does not attempt to use the newly installed `cmake`.

What is expected is that when attempting to use `cmake` an error due to the absence of the `cmake` executable should present a UI, and when the `cmake` executable is installed it should use this.  The current implementation performs this check once, presumably to prevent multiple UI notifications of the `cmake` path error.

It would be preferable to try to run the `cmake` program each time to account for the possibility of the program being installed, and upon failure only present the error notification UI on the first failure.

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

_No response_

qarni:
Fixes #3811.  CMakeTools now rechecks the same CMake Path to see if user installed CMake after launching VS Code, when it was previously an invalid path.

### DIFF
--- a/src/cmake/cmakeExecutable.ts
+++ b/src/cmake/cmakeExecutable.ts
@@ -29,10 +29,12 @@ export async function getCMakeExecutableInformation(path: string): Promise<CMake
         const normalizedPath = util.platformNormalizePath(path);
         if (cmakeInfo.has(normalizedPath)) {
             const cmakeExe: CMakeExecutable = cmakeInfo.get(normalizedPath)!;
-            await setCMakeDebuggerAvailableContext(
-                cmakeExe.isDebuggerSupported?.valueOf() ?? false
-            );
-            return cmakeExe;
+            if (cmakeExe.isPresent) {
+                await setCMakeDebuggerAvailableContext(
+                    cmakeExe.isDebuggerSupported?.valueOf() ?? false
+                );
+                return cmakeExe;
+            }
         }
 
         try {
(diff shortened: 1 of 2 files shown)


gcampbell-msft (approved):

src/cmake/cmakeExecutable.ts
```diff
@@ -29,10 +29,12 @@ export async function getCMakeExecutableInformation(path: string): Promise<CMake
         const normalizedPath = util.platformNormalizePath(path);
         if (cmakeInfo.has(normalizedPath)) {
             const cmakeExe: CMakeExecutable = cmakeInfo.get(normalizedPath)!;
-            await setCMakeDebuggerAvailableContext(
-                cmakeExe.isDebuggerSupported?.valueOf() ?? false
-            );
-            return cmakeExe;
+            if (cmakeExe.isPresent) {
```
elizabethmorrow, line 32: Should we log an error if cmake is not present here? 

Is this the case where the user has installed cmake between the time when they opened vscode and their first invocation of configure?

gcampbell-msft: @elizabethmorrow Correct, this is the case, but actually broader. It's any case where they open vscode, then install, then do any action where we invoke cmake.exe. 

I would agree it's a good idea to log here, @qarni we should do a quick PR for this as well

PR #3853, opened 2024-06-27
Set useCMakePresets context after quick start

### LINKED ISSUE #3734: usepresets context is not set after using quickstart
### Brief Issue Summary

When a CMakePreset is added through QuickStart, or if the user escapes out before adding a preset, the usepresets context is not set, and so no preset options pop up in the command pallete.

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

_No response_

qarni:
Ensure `useCMakePresets` context is set after making a CMakePreset.json with Quick Start, so the `Select Configure Preset` and `Add Configure Preset` options are available in the command pallete.

This works by updating the active project after the new project files are made, which in turn updates the use presets context.

In the future, it would be good to set up a file watcher to call this function to cover other use cases.

Fixes #3734.

### DIFF
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -2847,9 +2847,7 @@ export class CMakeProject {
         if (await fs.exists(mainPresetsFile)) {
             log.info(localize('cmakepresets.already.configured', '[Quick Start] A CMakePresets.json is already configured.'));
         } else {
-            await this.setUseCMakePresets(true);
             if (!await this.presetsController.selectConfigurePreset(true)) {
-            //if (!await extensionManager?.selectConfigurePreset(true)) {
                 await this.kitsController.setKitByName(SpecialKits.Unspecified);
                 log.info(localize('cmakepresets.not.created', '[Quick Start] CMakePresets.json not created.'));
             } else {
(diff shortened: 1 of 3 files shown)


src/cmakeProject.ts
```diff
@@ -2853,6 +2855,8 @@ export class CMakeProject {
             }
         }
 
+        await this.projectController?.updateActiveProject(this.workspaceFolder);
+
         // Regardless of the following configure return code,
         // we want full feature set view for the whole workspace.
         await enableFullFeatureSet(true);
```
gcampbell-msft, line 2862: One thing I noticed while testing is that line 2863 (in the new version), will technically force a second configure if the `automaticReconfigure` setting is enabled. I don't think this needs to block this PR, but it is something we should look into.

gcampbell-msft (dismissed):
Other than my comment that can be resolved in a separate PR, this looks good!

src/cmakeProject.ts
qarni: Will make a note to look into it in a follow up task

gcampbell-msft (approved):

PR #3858, opened 2024-06-30
Clean up remaining fallout in docs/ after relocation in #1324

kkm000:
## This change completes link clean-up in `docs/` after the repo relocation (#1324)

* Replace intra-repo absolute URLs with relative ones.
* Replace remaining pointers to `vector-of-bool` with `microsoft` in the URLs  
  that must remain absolute.

X-Ref-PR: #1324

### Other Notes/Information

This fixes all remaining links in `docs/` that should be relative but aren't, or are absolute by design but point to the old `vector-of-bool` repo home. The changeset is exhaustive, as found using VS Code's global search.


kkm000:

gcampbell-msft (approved):

gcampbell-msft:
No build needed, only docs.

PR #3859, opened 2024-07-01
clean configures when platform differs between old and new selected kit

snehara99:

-When the user changes the selected kit, I added a check to see if the platform changed, which triggers a clean configure.

### DIFF
--- a/src/kit.ts
+++ b/src/kit.ts
@@ -1370,7 +1370,8 @@ export function kitChangeNeedsClean(newKit: Kit, oldKit: Kit | null): boolean {
         vs: k.visualStudio,
         vsArch: k.visualStudioArchitecture,
         tc: k.toolchainFile,
-        preferredGenerator: k.preferredGenerator ? k.preferredGenerator.name : null
+        preferredGeneratorName: k.preferredGenerator ? k.preferredGenerator.name : null,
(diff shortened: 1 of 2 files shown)


src/kit.ts
```diff
@@ -1370,7 +1370,8 @@ export function kitChangeNeedsClean(newKit: Kit, oldKit: Kit | null): boolean {
         vs: k.visualStudio,
         vsArch: k.visualStudioArchitecture,
         tc: k.toolchainFile,
-        preferredGenerator: k.preferredGenerator ? k.preferredGenerator.name : null
+        preferredGeneratorName: k.preferredGenerator ? k.preferredGenerator.name : null,
+        preferredGeneratorPlatform: k.preferredGenerator && k.preferredGenerator.platform ? k.preferredGenerator.platform : null
```
jophippe, line 1374: Since we're already checking generator name and platform, should we check generator toolset as well? Or is that not important to check for cleaning?

gcampbell-msft: @jophippe @snehara99 I don't think it's crucial, but if it's simple to add I think that it could be worthwhile.

gcampbell-msft (dismissed):

src/kit.ts
snehara99: updated

gcampbell-msft (approved):

Issue #3727, opened 2024-04-29
YAMLException on configure


tysolov:
### Brief Issue Summary

When opening project, custom variants in my cmake-variants.yaml file are not loaded, seemingly due to a parsing error in a markdown file:

```---
title: README
...

# Large heading
Text
* Text with bullet  <!--- error here-->

## Another Header
More text
```

Due to this, my custom variants are not loaded upon opening vscode.  However, editing the cmake-variants.yaml file in any way (e.g., adding a newline to end) forces a reload, at which point the custom variants show up without any ostensible parsing errors.

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.85.1",
  "cmtVersion": "1.18.31",
  "configurations": [
    {
      "folder": "/path/to/project",
      "cmakeVersion": "3.26.4",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": false,
      "compilers": {
        "C": "/usr/bin/gcc",
        "CXX": "/usr/bin/g++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "Debug",
    "buildTypesSeen": [
      "Debug"
    ],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 4,
    "executablesCount": 2,
    "librariesCount": 2,
    "targets": [
      {
        "name": "my_project",
        "type": "STATIC_LIBRARY"
      },
      {
        "name": "my_project_lib",
        "type": "STATIC_LIBRARY"
      },
      {
        "name": "my_project_unit_tests",
        "type": "EXECUTABLE"
      }
    ]
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
...
[main] Copying compile_commands.json from /path/to/project/cmake-build-debug/compile_commands.json to /path/to/project/compile_commands.json
[expand] expanding /usr/bin/cmake
[expand] expanding -T
[expand] expanding test
[expand] expanding --output-on-failure
[variant] Error parsing /path/to/project/README.md: 
	YAMLException: name of an alias node must contain at least one character (7:2)
...
 7 | * Text with ...
------^
8 | 
 9 | ## Another Header ...
[...]

PR #3878, opened 2024-07-08
Add guard against reloading a file that's not a variants file

jophippe:
## This change addresses item #3727 

- Add an explicit guard when reloading the variants file to make sure that the file we're loading is actually a variants file, so we don't try to treat any other file as a variants file and throw a parsing exception.

## Other Notes/Information

Ideally, we would figure out why the reload variants functionality would be given a path other than a variants file, since our file watcher should only be watching for variants files, but I haven't figured that out yet


gcampbell-msft (approved):

gcampbell-msft (changes requested):
We need to update this to be in a 1.19 changelog section.

jophippe:
@gcampbell-msft Okay, I made a 1.19 section in the changelog and moved both my bugfixes into it. Does anything else need to be moved into the 1.19 section?

gcampbell-msft:
> @gcampbell-msft Okay, I made a 1.19 section in the changelog and moved both my bugfixes into it. Does anything else need to be moved into the 1.19 section?

I think that's it

gcampbell-msft (approved):

Issue #3821, opened 2024-06-09
Why CMAKE_BUILD_TYPE is always Debug?


zchrissirhcz:
### Brief Issue Summary

`CMake: Configure` always uses `-DCMAKE_BUILD_TYPE:STRING=Debug` even if I delete `CMAKE_BUILD_TYPE` or set `CMAKE_BUILD_TYPE=Release` in setting of CMake Tools extension, and I've already set environment variable: `export CMAKE_BUILD_TYPE=Release`

### CMake Tools Diagnostics

```shell
{
  "os": "darwin",
  "vscodeVersion": "1.90.0",
  "cmtVersion": "1.18.41",
  "configurations": [
    {
      "folder": "/Users/zz/work/cppsober/kcv",
      "cmakeVersion": "3.29.5",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": false,
      "compilers": {
        "C": "/Library/Developer/CommandLineTools/usr/bin/cc",
        "CXX": "/Library/Developer/CommandLineTools/usr/bin/c++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "Debug",
    "buildTypesSeen": [
      "Debug"
    ],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 7,
    "executablesCount": 1,
    "librariesCount": 3,
    "targets": [
      {
        "name": "birch-autotimer-update",
        "type": "UTILITY"
      },
      {
        "name": "imgui",
        "type": "STATIC_LIBRARY"
      },
      {
        "name": "konacv",
        "type": "STATIC_LIBRARY"
      },
      {
        "name": "test",
        "type": "EXECUTABLE"
      },
      {
        "name": "glfw",
        "type": "STATIC_LIBRARY"
      },
      {
        "name": "update_mappings",
        "type": "UTILITY"
      }
    ]
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "never",
      "configureOnOpen": false
    }
  ]
}
```

Luvideria:
I would like to add that I can't find the correct way to simply add more build types. The CMakeFiles I use have for instance ServRel or ToolRel as build types, which I can't access at all.

Evelyn-001:
[...]

PR #3879, opened 2024-07-09
Add "Unspecified" Build Type and Use by Default (Kits Only)

snehara99:
For #3821 

This makes it so that the default variant is "Unspecified" which basically lets CMake handle what build type to use.


src/rollbar.ts
```diff
@@ -84,7 +84,7 @@ class RollbarController {
      */
     error(what: string, additional: object = {}): void {
         log.error(what, JSON.stringify(additional));
-        debugger;
+        // debugger;
```
gcampbell-msft, line 87: Why this change?

snehara99: oh whoops sorry I was supposed to revert that. It was for testing. Will fix

snehara99: shoudl be fixed

snehara99:
@sinemakinci1 this had a docs change if you want to review

snehara99:
@gcampbell-msft What do you think of the wording updates in the doc file update? I made a distinction between the word "option" and "variant" that may be confusing. Do you have ideas for alternative wording?

gcampbell-msft (dismissed):

gcampbell-msft (approved):

PR #3888, opened 2024-07-10
Skip loading variants when using CMakePresets

jophippe:
### This changes when variants are loaded

- When initializing the cmakeProject, skip initializing the variantManager if CMakePresets is found the be used
- Initialize the variantManager when not using CMakePresets or if the project switches to not use CMakePresets
- Add a hasInitialized function to VariantManager so we know whether the variantManager has been initialized or not, so we know when to initialize variantManager and to prevent initializing multiple times
- Other minor cleanup

## Other Notes/Information

Still testing, will update when I'm done testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.19
 
+Improvements:
+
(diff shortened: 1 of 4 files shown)


gcampbell-msft (approved):

jophippe:
Testing done, ready to merge

Issue #3405, opened 2023-10-25
Unspecified doesn't let CMake guess what compilers and environment to use


gerard-ryan-immersaview:
### Brief Issue Summary

When the selected kit is `[Unspecified] Unspecified (Let CMake guess what compilers and environment to use)`, `-G Ninja` is added to the CMake command and `No CMAKE_C_COMPILER could be found.` 

### CMake Tools Diagnostics

```json
{
  "os": "win32",
  "vscodeVersion": "1.83.1",
  "cmtVersion": "1.15.31",
  "configurations": [
    {
      "folder": "c:\\Users\\gerard\\Desktop\\cmake-tools-mre",
      "cmakeVersion": "3.27.7",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": false,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": null
    }
  ]
}
```

### Debug Log

```shell
[main] Configuring project: cmake-tools-mre 
[main] Saving open files before configure/build
[driver] Start configure 
[driver] Running pre-configure checks and steps
[cmakefileapi-driver] Configuring using kit
[cmakefileapi-driver] Invoking CMake C:\Program Files\CMake\bin\cmake.EXE with arguments ["--no-warn-unused-cli","-DCMAKE_BUILD_TYPE:STRING=Debug","-DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE","-SC:/Users/gerard/Desktop/cmake-tools-mre","-Bc:/Users/gerard/Desktop/cmake-tools-mre/build","-G","Ninja"]
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.EXE" --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -SC:/Users/gerard/Desktop/cmake-tools-mre -Bc:/Users/gerard/Desktop/cmake-tools-mre/build -G Ninja
[...]

PR #3889, opened 2024-07-10
Allow CMake to select a default internal generator when the user selects "__unspec__" for the kit

snehara99:
## This change addresses item #3405 

CMake file api supports not passing in a generator in versions >= 3.15.0. I added a check for that version and using the file api in order to not pass in Ninja as a preferred generator when the user selects "__unspec__" as a kit and doesn't define any generator/preferredGenerators.

### DIFF
--- a/src/drivers/cmakeFileApiDriver.ts
+++ b/src/drivers/cmakeFileApiDriver.ts
@@ -38,7 +38,7 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 
 const log = logging.createLogger('cmakefileapi-driver');
 /**
- * The CMake driver with FileApi of CMake >= 3.15.0
+ * The CMake driver with FileApi of CMake >= 3.14.0
  */
 export class CMakeFileApiDriver extends CMakeDriver {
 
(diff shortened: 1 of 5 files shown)


src/drivers/cmakeFileApiDriver.ts
```diff
@@ -38,7 +38,7 @@ const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 
 const log = logging.createLogger('cmakefileapi-driver');
 /**
- * The CMake driver with FileApi of CMake >= 3.15.0
+ * The CMake driver with FileApi of CMake >= 3.14.0
```
gcampbell-msft, line 41: Why is this version being changed?

gcampbell-msft (commented):
Overall this looks good, assuming we confirm with testing that it works as expected and our tests pass, and I have one comment that needs addressing.

src/drivers/cmakeFileApiDriver.ts
snehara99: The version isn't changed. I think the version in the comment was incorrect for some reason. This is the actual version we check (unmodified) 
image

gcampbell-msft: Ahh, thanks for pointing that out, thanks! 👍

snehara99:
> Overall this looks good, assuming we confirm with testing that it works as expected and our tests pass, and I have one comment that needs addressing.

I've tested it a decent amount. I think it should be fine

gcampbell-msft (approved):

Issue #3870, opened 2024-07-03
macros for penv/env not expanded when cache variables for cmake(c/cxx)compilers are included in configure preset


qarni:
### Brief Issue Summary

$penv{PATH}/$env{PATH} are not expanded when cache variables for "CMAKE_C_COMPILER": "cl.exe", "CMAKE_CXX_COMPILER": "cl.exe" are included in configure preset. Likely they are overwritten by our vs dev environment magic.

Repro configure preset:

>  {
            "name": "preset",
            "displayName": "Visual Studio Enterprise 2022 IntPreview - amd64",
            "description": "Using compilers for Visual Studio 17 2022 (x64 architecture)",
            "generator": "Visual Studio 17 2022",
            "toolset": "host=x64",
            "architecture": "x64",
            "binaryDir": "${sourceDir}/out/build",
            "environment": {
                "PATH": "HELLOTHISISBAD;$penv{PATH}"
            },
            "cacheVariables": {
                "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}",
                "CMAKE_C_COMPILER": "cl.exe",
                "CMAKE_CXX_COMPILER": "cl.exe"
            }
        }

v-frankwang:
@qarni I used a simple project to reproduce the problem, using message in the CMakeList.txt file to print the value of the PATH, and it would not show the value of the macro. Does this reproduce your problem?

Actual result：
![image](https://github.com/microsoft/vscode-cmake-tools/assets/160998958/06a32367-8256-41b0-87e5-843a7a6de5c6)

Issue #3497, opened 2023-12-15
Support launching cmake.exe from a VS dev environment even when not using CL 


tristanlabelle:
### Brief Issue Summary

Switching the CMake compiler from cl to clang will make the CMake Tools extension stop launching `cmake.exe` from within a VS dev environment. This is problematic when relying on `ninja` from the VS installation or if using `midlrt.exe` from the Windows SDK, which fails if it cannot find `cl.exe` as a C preprocessor.

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.85.1",
  "cmtVersion": "1.16.32",
  "configurations": [
    {
      "folder": "<redacted>",
      "cmakeVersion": "3.28.20231108",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {
        "C": "cl",
        "CXX": "cl"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "RelWithDebInfo",
    "buildTypesSeen": [
      "RelWithDebInfo"
    ],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 184,
    "executablesCount": 0,
    "librariesCount": 184,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "always",
      "configureOnOpen": true
    }
  ]
}
```

### Additional Information

One solution could be to add a `"cmake.alwaysUseVSEnvironment"` setting.

This is related to #1836 as per discussion #3486.

gcampbell-msft:
We are not going to be able to solve this in 1.17, so I am moving this to 1.18. 

However, we have a proposal, please @tristanlabelle let us know what you think. 

We would propose that we, rather than doing the "magic" that we currently do, will expose a setting that, by default, disables the use of a VS dev environment. However, this would then allow users to enable this setting, which would introduce a dev environment so that your scenario is supported. 

What do you think?

gcampbell-msft:
Moved this to "On Deck" as we aren't likely to make it into the official 1.18 release. We do hope to merge this into pre-release (after releasing 1.18) as soon as we can. Thank you.

qarni:
Maybe related to #3260 and #1897

qarni:
[...]

Issue #3260, opened 2023-07-26
New machine experience: Ninja not found (in VS, even though it can find CMake in the VS installation just fine)


TheJCAB:
### Brief Issue Summary

This is a basic ease-of-use user experience issue.

A fresh machine, with VS 2022 installed on it and ready to go. Open a CMake/Ninja project folder VSCode, and try to build. This happens:

```
...
[cmake] CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
[proc] The command: "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" *redacted* -G Ninja exited with code: 1
```

So... it can find "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" just fine (it's not in the PATH). But it can't be bothered to put, say, "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja" in the PATH when calling CMake.

This sounds like a relatively simple fix. Ninja is included in VS along with CMake, and it should be usable right out of the box, no?

Thanx!

benmcmorran:
Thanks for the report! The suggestion sounds reasonable to me. As a workaround, are you able to launch VS Code from a Visual Studio Developer Command Prompt? That should have all the proper environment variables set for Ninja (and MSVC) to be easily discoverable.

@sinemakinci1 FYI this is related to our broader discussion about streamlining tool acquisition.

Thomas1664:
I'm experiencing the same issue. As a result, I'm unable to build anything using Ninja. The suggested workaround is not viable because opening VS Dev CMD is too complicated for the simple task of just opening a folder. It should "just work".

gcampbell-msft:
@Thomas1664 Could you provide exact repro steps for when you're seeing this? Are you using kits / presets? What type of project are you opening? Etc.

Yingzi1234:
@gcampbell-msft We can reproduce this issue on latest VSCode and you can get the details below.

**ENV:**
1.	VS: 17.11 Preview1.0
2.	VSCode: 1.88.1
3.	C/C++: v1.20.2
4.	CMake tools: v0.0.17
[...]

Issue #3143, opened 2023-04-26
CMakeTools breaks llvm's VS version detection


fischman-bcny:
### Brief Issue Summary

CMake Tools sets VCINSTALLDIR but not VCToolsInstallDir ([exclusion](https://github.com/microsoft/vscode-cmake-tools/blob/6825d8a9cf6f96f4ffb6a28fb5d7461f19cce26f/src/installs/visualStudio.ts#L263)). This combination fools clang into thinking that it's dealing with VS 2015 or older rather than 2017 or newer ([code](https://github.com/llvm/llvm-project/blob/main/llvm/lib/WindowsDriver/MSVCPaths.cpp#L505-L514)), whose heuristics derive from how (pre-)2015 and (post-)2017 `vcvarsall.bat` behave.

It would be nice for CMake Tools to pass either both variables or neither.

PR #3892, opened 2024-07-11
Add setting to control how we use VS Dev Env

gcampbell-msft:
Handles #3497, #1897, #3260, #3870, #3143

This adds a `cmake.useVsDeveloperEnvironment` setting that controls how we assess or apply the VS Developer Environment when presets are being used. This setting does not apply for Kits, only in the Presets mode. 

The available options are `always`, `never`, or `auto`. If `always`, we will unconditionally apply the VS Developer Environment (NOTE: the VS Developer Environment can only be applied in Windows), `never` means we'll never apply it, `auto` means that we'll do what we currently do, where we only apply it if certain compilers (cl, clang, clang-cl, clang-cpp) or generators (ninja), are being used. 

This also implemented a fix to ensure that we aren't incorrectly overriding environment variables when we apply the VS Dev Env.


src/preset.ts
```diff
@@ -709,6 +710,7 @@ async function getExpansionOptions(workspaceFolder: string, sourceDir: string, p
             presetName: preset.name
         },
         envOverride: preset.environment,
+        penvOverride: penvOverride,
```
qarni, line 713: is penv parent Env? maybe rename for clarity

src/preset.ts
```diff
@@ -988,12 +997,131 @@ async function expandConfigurePresetImpl(folder: string, name: string, workspace
...
+                    // If no toolset version specified then choose the latest VS instance for the given generator
+                    vsInstall = vs;
+                    break;
+                }
+            }
+        }
+    }
+
+    if (!vsInstall) {
+        // I don't think we want to do this anymore, or at least it's conditional on the mode being "auto". If it's "always", we just want to report that we couldn't find a vs isntallation.
```
qarni, line 1081: This makes sense to me

src/preset.ts
```diff
@@ -988,12 +997,131 @@ async function expandConfigurePresetImpl(folder: string, name: string, workspace
...
+}
+
+/**
+ * This method tries to apply, based on the useVsDeveloperEnvironment setting value and, in "auto" mode, whether certain preset compilers/generators are used and not found, the VS Dev Env.
+ * @param preset Preset to modify the parentEnvironment of. If the developer environment should be applied, the preset.environment is modified by reference.
+ * @param workspaceFolder The workspace folder of the CMake project.
+ * @param sourceDir The source dir of the CMake project.
+ * @returns Void. We don't return as we are modifying the preset by reference.
+ */
+async function tryApplyVsDevEnv(preset: ConfigurePreset, workspaceFolder: string, sourceDir: string): Promise<void> {
```
qarni, line 1115: we used to check if `preset.__vsDevEnvApplied` before doing this (to see if we've already applied the env??) do we still want to check `__parentEnvironment`?

gcampbell-msft: Yes, checking `__parentEnvironment` prohibits us from rechecking for the devenv (and thus re-invoking where.exe and more) when we have to expand the configure preset for all of the other presets (build, test, etc).

tristanlabelle:
Would we have a way to control the different flags to vsdevenv if our presets are not using cl as the compiler? For example, the target architecture, the windows sdk version and the vc toolset version?

src/preset.ts
```diff
@@ -988,12 +997,129 @@ async function expandConfigurePresetImpl(folder: string, name: string, workspace
...
+    shouldInterrogateForNinja: boolean;
+    compilerName?: string; // Only will have a value when `useVsDeveloperEnvironmentMode` is "auto"
+}
+
+/**
+ * @param opts Options to control the behavior of obtaining the VS developer environment.
+ * @returns Either the VS developer environment or undefined if it could not be obtained.
+ */
+async function getVsDevEnv(opts: VsDevEnvOptions): Promise<EnvironmentWithNull | undefined> {
+    const arch = getArchitecture(opts.preset);
```
tristanlabelle, line 1011: Our presets do not specify the architecture or toolset architecture by design, so that they can be used for native compilation on both x64 and arm64 machines. `getArchitecture` and `getToolset` return `x86` as a default value, which we don't support as a compilation target. Could they be changed to return a value based on the processor architecture? I believe this is consistent with the vcvarsall behavior when host and target arch are not specified

gcampbell-msft:
> Would we have a way to control the different flags to vsdevenv if our presets are not using cl as the compiler? For example, the target architecture, the windows sdk version and the vc toolset version?

To answer this question, yes, you can still control the target architecture by using the `preset.architecture` field. I believe the same is true for the `toolset` value. 
You can see some information here: https://github.com/microsoft/vscode-cmake-tools/issues/3672

src/preset.ts
```diff
@@ -988,12 +997,129 @@ async function expandConfigurePresetImpl(folder: string, name: string, workspace
...
+                "Configure preset {0}: Compiler {1} with toolset {2} and architecture {3} was not found, you may need to run the 'CMake: Scan for Compilers' command if this toolset exists on your computer.",
+                opts.preset.name, `"${opts.compilerName}.exe"`, toolset.version ? `"${toolset.version},${toolset.host}"` : `"${toolset.host}"`, `"${arch}"`));
+        } else {
+            log.error(localize('vs.not.found', "Configure preset {0}: No Visual Studio installation found that supports the specified toolset {1} and architecture {2}, you may need to run the 'CMake: Scan for Compilers' command if this toolset exists on your computer.",
+                opts.preset.name, toolset.version ? `"${toolset.version},${toolset.host}"` : `"${toolset.host}"`, `"${arch}"`));
+        }
+    } else {
+        log.info(localize('using.vs.instance', "Using developer environment from Visual Studio (instance {0}, version {1}, installed at {2})", vsInstall.instanceId, vsInstall.installationVersion, `"${vsInstall.installationPath}"`));
+        const vsEnv = await varsForVSInstallation(vsInstall, toolset.host!, arch, toolset.version);
```
tristanlabelle, line 1091: This omits `VCToolsInstallDir`, which is a problem because that's how LLVM detects a VS2017+ toolset: https://github.com/llvm/llvm-project/blob/a288d8dad3b3f459bce3f908cd564f2963060642/llvm/lib/WindowsDriver/MSVCPaths.cpp#L512

See `visualStudio.ts`:
```cpp
//'VCToolsInstallDir', - disabled temporarily as it breaks downlevel toolset selection
```

src/preset.ts
gcampbell-msft: Would something like using the `util.getHostArchitecture` method do what you're describing and get what you're expecting? We could default to that method rather than simply `x86`.

src/preset.ts
gcampbell-msft: Yes, I'm simultaneously investigating that for this bug, if I can confirm that it won't break any scenarios, we can likely include this in this PR.

src/preset.ts
gcampbell-msft: Updated. 👍

src/preset.ts
gcampbell-msft: @tristanlabelle I updated it, if you could please test this out in your scenarios that would be great.

gcampbell-msft:
Blocking comment for me to test changing different toolsets with devenv

src/preset.ts
tristanlabelle: It works!! 😍

tristanlabelle (approved):
Thanks! Super excited for this!

qarni (dismissed):

qarni:
Needs a changelog @gcampbell-msft  :)

test/unit-tests/presets.test.ts
```diff
@@ -1,13 +1,18 @@
...
 
 suite('Preset tests', () => {
     test('Parse architecture', () => {
         expect(getArchitecture({ name: 'test', architecture: 'x86' })).to.eq('x86');
         expect(getArchitecture({ name: 'test', architecture: 'amd64' })).to.eq('amd64');
         expect(getArchitecture({ name: 'test', architecture: { value: 'arm', strategy: 'set' } })).to.eq('arm');
         expect(getArchitecture({ name: 'test', architecture: { value: 'arm64', strategy: 'external' } })).to.eq('arm64');
-        expect(getArchitecture({ name: 'test' })).to.eq('x86');
+        if (os.platform() === "win32" || os.platform() === "linux") {
+            expect(getArchitecture({ name: "test" })).to.eq("x64");
```
moyo1997, line 12: is it verified that os.platform on windows arm64 is not also win32? 

asking because this is what i found in the docs:
process.platform[#](https://nodejs.org/dist/latest-v4.x/docs/api/process.html#process_process_platform)

What platform you're running on: 'darwin', 'freebsd', 'linux', 'sunos' or 'win32'

snehara99 (dismissed):
looks good to me!

src/preset.ts
```diff
@@ -1020,119 +1148,54 @@ async function tryApplyVsDevEnv(preset: ConfigurePreset) {
                 }
 
                 if (compilerName && whereExecutable) {
+                    // We need to construct and temporarily expand the environment in order to accurately determine if this preset has the compiler / ninja on PATH.
+                    // This puts the preset.environment on top of process.env, then expands with process.env as the penv and preset.environment as the envOverride
```
moyo1997, line 1152: should this be passing in nvironmentUtils.mergePreserveNull([process.env, preset.environment]] as the envorride instead or just preset.environement ? since the fallback behavior in var not in env field is to check parent env

src/preset.ts
```diff
@@ -1020,119 +1148,54 @@ async function tryApplyVsDevEnv(preset: ConfigurePreset) {
...
+                    if (!compilerLocation.stdout || shouldInterrogateForNinja) {
+                        developerEnvironment = await getVsDevEnv({
+                            preset,
+                            shouldInterrogateForNinja,
+                            compilerName
+                        });
                     }
                 }
             }
+        } else if (useVsDeveloperEnvironmentMode === "always") {
```
moyo1997, line 1190: what happens if user has this set to always but wants to use a different ninja  than the one is vs dev environment that they prepended to their path in environment field? does this scenario work, or we always use vs devenv ninja?

gcampbell-msft: If they already prepended to their path in the environment field, then we will find it in our `where.exe` invocation, and therefore won't add the ninja PATH.

src/preset.ts
gcampbell-msft: Good callout, fixing.

gcampbell-msft: Fixed.

gcampbell-msft:
@moyo1997 @qarni @snehara99 Committed some small fixes, please re-approve / re-review if needed.

test/unit-tests/presets.test.ts
gcampbell-msft: Fixed to be more explicit

qarni (approved):

Issue #3862, opened 2024-07-03
[Loc] "should NOT have additional properties" not localized


PhoebeTf:
### Brief Issue Summary

**Environment**:
VSCode 1.90.1+ CMake Tools v1.19.13(pre-release) 

**More info**:
Repro Language: All loc languages
Repro platform: Windows, Linux, Mac

**Repro steps**:
1. Open VScode and change the display language to CHS
2. Download and unzip [Version Validation](https://microsoft.sharepoint.com/teams/DD_VC/Shared%20Documents/Forms/AllItems.aspx?id=%2Fteams%2FDD%5FVC%2FShared%20Documents%2FCross%20Platform%2FCMake%2FTest%20CMake%20Projects%2FVersion%20Validation&p=true&ga=1)
3. Open above project with VS Code.
4. Open CMakePresets.json file
5. Launch output window and navigate to "CMake/Build" category
6. Check the output info

**Expected result**:
“should NOT have additional properties" localized

**Actual result**:
“should NOT have additional properties” not localized
![output_notloc](https://github.com/microsoft/vscode-cmake-tools/assets/37648162/87bfd609-3f86-46c0-9190-1a0f0c3044d8)

PhoebeTf:
Verified with CMake Tools v1.19.35 (pre-release), this issue has been fixed.
![image](https://github.com/user-attachments/assets/968f65b3-849b-4a70-b86d-500cedea3b56)

PR #3895, opened 2024-07-11
Localized the additional properties case in schema validation

snehara99:
## This change addresses item #3862 
localized it with same string so nothing changes for English users.


gcampbell-msft (approved):
Just a note that in the else branch, if we ever get bugs about that, we won't be able to fix because it's likely indeterminate what the string will be

PR #3899, opened 2024-07-14
Fix link in configure.md

Ashark:
The section is called "General options", not "Generic options".


gcampbell-msft (approved):

gcampbell-msft:
Doc fix, no tests needed.

PR #3902, opened 2024-07-15
Dispose of Launch terminal if environment has changed

jophippe:
### This changes when a Launch terminal gets re-used

- Create terminal Launch environment at the start of the function to compare it to existing terminals
- Compare the entire environment of the existing terminal with the current environment and dispose of existing terminal if any changes are found, so that the old environment is not used on later runs
- Remove the refreshLaunchEnvironment variable since the environment compare does the same thing, but more complete

## Other Notes/Information

Open to feedback for ways to compare environments that are better than a stringify equality


gcampbell-msft:
@jophippe Don't forget the CHANGELOG entry 👍

gcampbell-msft (dismissed):

gcampbell-msft (approved):

gcampbell-msft (approved):

Issue #3871, opened 2024-07-04
CMake configure preset installDir setting is evaluated relative to the binary directory instead of the source directory


Bekenn:
### Brief Issue Summary

From the [CMake documentation](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#configure-preset) for the `installDir` setting:
> If a relative path is specified, it is calculated relative to the source directory.

The CMake Tools extension appears to evaluate this setting relative to the binary directory.  If I supply a relative path and run the configure task using the CMake Tools extension, then the resulting cache variable for `CMAKE_INSTALL_PREFIX` begins with the binary directory.  If I instead invoke CMake directly with the same preset, then `CMAKE_INSTALL_PREFIX` begins with the source directory.

My best guess as to the cause is that the extension runs the configure step with the current working directory set to the binary directory.  Passing a relative path with `-DCMAKE_INSTALL_PREFIX=` will cause CMake to form a path relative to the current working directory rather than the source directory.  The CMake Tools extension should explicitly prepend the source directory in this case.

### CMake Tools Diagnostics

```shell
{
  "os": "darwin",
  "vscodeVersion": "1.91.0",
  "cmtVersion": "1.18.42",
  "configurations": [
    {
      "folder": "/Users/bekenn/Projects/altsup",
      "cmakeVersion": "3.29.3",
      "configured": true,
      "generator": "Ninja Multi-Config",
      "usesPresets": true,
      "compilers": {
        "CXX": "/opt/homebrew/opt/llvm/bin/clang++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": false
    }
  ]
}
```

### Debug Log

```shell
[expand] expanding cmake
[main] Configuring project: altsup 
[main] Saving open files before configure/build
[expand] expanding /Users/bekenn/Projects/altsup
[driver] Start configure 
[driver] Running pre-configure checks and steps
[expand] expanding /Users/bekenn/Projects/altsup
[cmakefileapi-driver] Configuring using preset
[...]

PR #3903, opened 2024-07-15
if installDir is a relative path, calculate it relative to the source directory

snehara99:
## This change addresses item #3871 

after installDir is expanded, if it is still a relative path, we calculate it according to the source directory ([docs](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#configure-preset)).

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -1137,6 +1137,9 @@ async function tryApplyVsDevEnv(preset: ConfigurePreset) {
 
 async function expandConfigurePresetHelper(folder: string, preset: ConfigurePreset, workspaceFolder: string, sourceDir: string, allowUserPreset: boolean = false) {
     if (preset.__expanded) {
+        if (preset.installDir) {
+            preset.installDir = util.resolvePath(preset.installDir, sourceDir);
+        }
         return preset;
     }
 
@@ -1204,6 +1207,9 @@ async function expandConfigurePresetHelper(folder: string, preset: ConfigurePres
     preset.environment = EnvironmentUtils.mergePreserveNull([inheritedEnv, preset.environment]);
 
     preset.__expanded = true;
+    if (preset.installDir) {
+        preset.installDir = util.resolvePath(preset.installDir, sourceDir);
+    }
     return preset;
 }
 
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -807,6 +807,9 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
     if (!preset) {
         return null;
     }
+    if (preset.installDir) {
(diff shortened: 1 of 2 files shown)


src/preset.ts
```diff
@@ -1137,6 +1137,9 @@ async function tryApplyVsDevEnv(preset: ConfigurePreset) {
 
 async function expandConfigurePresetHelper(folder: string, preset: ConfigurePreset, workspaceFolder: string, sourceDir: string, allowUserPreset: boolean = false) {
     if (preset.__expanded) {
```
gcampbell-msft, line 1139: A couple of questions,
1.  Why are we adding it in two locations that are basically the same? 
2. Is there not a place where we currently expand it relative to the binary directory as the bug says? If so, I would think that we need to fix that. 
3. Does the `util.resolvePath` handle if `installDir` is an absolute path? 

Instead of this, method, could we have the resolution in the overall `expandConfigurePreset` method?

gcampbell-msft (changes requested):

src/preset.ts
snehara99: For 1) I could tell this was recursive and didn't know if they were both exit situation. I basically wanted to guarantee it's done after the presets have been completely expanded. 
For 2) This should handle that case too. I think that was just one case that happened, but the fix is to always set an absolute path like the CMake docs say
For 3) And yes, if installDir is an absolute path, it returns it how it was

snehara99: Your last point makes sense yeah. I'll fix the location

snehara99: updated @gcampbell-msft

src/preset.ts
```diff
@@ -807,6 +807,9 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
     if (!preset) {
         return null;
     }
+    if (preset.installDir) {
```
gcampbell-msft, line 810: I think this needs to go in the if statement on line 852. We first need to expand any macros in the `installDir`, then we can normalize the path

gcampbell-msft (changes requested):

src/preset.ts
snehara99: ok this time it should be in the right place haha. `resolvePath` also normalizes the path: 
```
export function resolvePath(inpath: string, base: string) {
    const abspath = path.isAbsolute(inpath) ? inpath : path.join(base, inpath);
    // Even absolute paths need to be normalized since they could contain rogue .. and .
    return lightNormalizePath(abspath);
}
```

gcampbell-msft (dismissed):

gcampbell-msft (approved):

Issue #2781, opened 2022-10-08
[bug] Presets with `$env{}` are not checked for circular references.


hwhsu1231:
### Brief Issue Summary

This issue is related to the [topic](https://discourse.cmake.org/t/how-to-write-a-proper-configurepreset-using-mingw-compilers-with-mingw-makefiles-generator/6615/5?u=hwhsu1231) I posted in CMake Discourse.

In short, it should **FAIL** to configure the project when using `$env{}` like the following example:

```json
"environment": {
  "PATH": "C:/msys64/mingw64/bin;$env{PATH}"
},
```

And it failed as I expected when operating in Terminal.

```cmd
F:\Git-Repo\cmake-project-template-conan>cmake --preset win32-gcc-x64-mingw-debug
CMake Error: Could not read presets from F:/Git-Repo/cmake-project-template-conan: Invalid macro expansion
```

However, it somehow could work with CMake Tools in Visual Studio Code. According to what [Kyle Edwards](https://discourse.cmake.org/u/kyle.edwards) said:

> The `$env{...`} macro is supposed to reference other environment variables within the environment map, or the environment inherited from the parent process if the corresponding variable doesn’t exist. The problem is that if `PATH` contains `$env{PATH}`, then it’s referencing itself, creating a circular dependency. The purpose of `$penv{...}` is to break this cycle by only using the environment inherited from the parent process. If `"PATH": "$env{PATH}"` works in VSCode then it sounds like they didn’t implement their preset support properly.

### Environments and Versions

- Visual Studio Code: **1.71.2**
- CMake Tools: **1.13.11**

### Screenshots

![2af6809a9d09c181f7bf52a7e48f5f1d1744b57f](https://user-images.githubusercontent.com/71438617/194700266-55eb4e8f-ccc8-4640-a92a-84af9c3c9e6f.gif)

bobbrow:
It looks like we marked it as a warning instead of an error.

qarni:
Related to #2934

PR #3905, opened 2024-07-15
Preset file expansion on open/save and expansion validation

### LINKED ISSUE #2934: Expansion of ${generator} macro not working in preset conditions
### Brief Issue Summary

In order to reuse our company-internal CMake presets I wanted to create a repository with common base presets and leverage the `include` and `inherits` capabilities. Since the build presets differ for single-config and multi-config generators (mainly the `configuration` property), I tried to leverage the `condition` properties to check what the value of `${generator}` is. Unfortunately, the evaluation seems to fail and does not match CMake's command line behavior.

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.74.2",
  "cmtVersion": "1.12.27",
  "configurations": [
    {
      "folder": "c:\\Users\\WDAGUtilityAccount\\Desktop\\presets-test",
      "cmakeVersion": "3.25.1",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "useCMakePresets": "auto",
      "configureOnOpen": null
    }
  ]
}
```

### Debug Log

_No response_

### Additional Information

**Example Presets**

<details>
<summary>CMakePresets.json</summary>

```json
{
    "version": 4,
    "configurePresets": [
        {
            "name": "base-single",
            "hidden": true,
            "generator": "Ninja"
        },
        {
            "name": "base-multi",
            "hidden": true,
            "generator": "Ninja Multi-Config"
        },
        {
            "name": "dev1",
            "inherits": ["base-single"]
        },
        {
            "name": "dev2",
            "inherits": ["base-multi"]
        },
        {
            "name": "dev3",
            "generator": "Ninja"
        },
        {
            "name": "dev4",
            "generator": "Ninja Multi-Config"
        },
        {
            "name": "dev5",
            "inherits": ["base-single"],
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja"
            }
        },
        {
            "name": "dev6",
            "inherits": ["base-single"],
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        },
        {
            "name": "dev7",
            "generator": "Ninja",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja"
            }
        },
        {
            "name": "dev8",
            "generator": "Ninja",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        }
    ],
    "buildPresets": [
        {
            "name": "dev1-working",
            "configurePreset": "dev1",
            "environment": {
                "GENERATOR_MACRO": "${generator}"
            },
            "condition": {
                "type": "equals",
                "lhs": "${hostSystemName}",
                "rhs": "Windows"
            }
        },
        {
            "name": "dev1-single",
            "configurePreset": "dev1",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja"
            }
        },
        {
            "name": "dev1-multi-debug",
            "configurePreset": "dev1",
            "configuration": "Debug",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        },
        {
            "name": "dev1-multi-release",
            "configurePreset": "dev1",
            "configuration": "Release",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        },
        {
            "name": "dev2-single",
            "configurePreset": "dev2",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja"
            }
        },
        {
            "name": "dev2-multi-debug",
            "configurePreset": "dev2",
            "configuration": "Debug",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        },
        {
            "name": "dev2-multi-release",
            "configurePreset": "dev2",
            "configuration": "Release",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        },
        {
            "name": "dev3-single",
            "configurePreset": "dev3",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja"
            }
        },
        {
            "name": "dev3-multi-debug",
            "configurePreset": "dev3",
            "configuration": "Debug",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        },
        {
            "name": "dev3-multi-release",
            "configurePreset": "dev3",
            "configuration": "Release",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        },
        {
            "name": "dev4-single",
            "configurePreset": "dev4",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja"
            }
        },
        {
            "name": "dev4-multi-debug",
            "configurePreset": "dev4",
            "configuration": "Debug",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        },
        {
            "name": "dev4-multi-release",
            "configurePreset": "dev4",
            "configuration": "Release",
            "condition": {
                "type": "equals",
                "lhs": "${generator}",
                "rhs": "Ninja Multi-Config"
            }
        }
    ]
}
```

</details>

**Expected Behavior**

All presets whose conditions are evaluated to true are displayed.

**Actual Behavior**

_CMake > Select Configure Preset_ is missing the `dev5` configure preset:

image

_CMake > Select Build Preset_ does not show any of the associated build presets, whether the condition is met or not:

image

Also when a configure preset is selected for the first time, the first associated build preset is selected automatically even if its condition is not fulfilled and it is not visible in the dropdown:

image

For reference, these are the presets recognized by CMake:
```shell
$ cmake --list-presets=configure
Available configure presets:

  "dev1"
  "dev2"
  "dev3"
  "dev4"
  "dev5"
  "dev7"

$ cmake --list-presets=build
Available build presets:

  "dev1-single"       
  "dev2-multi-debug"  
  "dev2-multi-release"
  "dev3-single"       
  "dev4-multi-debug"  
  "dev4-multi-release"
```

**Other Observations**

With other properties such as `environment` the `${generator}` macro gets correctly expanded. Also, conditions with other macros such as `${hostSystemName}` work as expected. See `dev1-working`.

qarni:
Fixes #2934, #2781.

This change changes the way we do preset file expansion. Expansion now occurs upfront on file open or save, and **if there are no errors**, the expanded presets will be cached so repeated expansion is not necessary every time `getAll{type of preset}Presets()` is called.

The only time expansion will happen again is on set preset, which will then also apply the VS developer environment as needed. This is avoided on the initial expansion for performance.

If there **are** errors in expansion, the errors are shown in the problems panel per file. The presets file will also be set to undefined, which will invalidate that file and not allow any presets in that file to be recognized/selected as valid presets. This is in line with the CMake command line experience.

image

Some notes:
- The ${generator} macro should not work for packagePresets, but should now expand correctly for all other presets and evaluate conditions correctly
- Tertiary preset cache was added (really its just the original presets cache but used in a new way) to account for included json files that wouldnt be in the original presets files


src/preset.ts
```diff
@@ -776,43 +783,58 @@ async function expandCondition(condition: boolean | Condition | null | undefined
...
+            if (expandedPreset) {
+                const opts = await getExpansionOptions('${workspaceFolder}', sourceDir, expandedPreset);
+                preset.condition = await expandCondition(preset.condition, opts, errorHandler);
+            }
         }
     }
 }
 
-export async function expandConfigurePreset(folder: string, name: string, workspaceFolder: string, sourceDir: string, allowUserPreset: boolean = false): Promise<ConfigurePreset | null> {
+export async function expandConfigurePreset(folder: string, name: string, workspaceFolder: string, sourceDir: string, allowUserPreset: boolean = false, enableTryApplyDevEnv: boolean = true, errorHandler?: ExpansionErrorHandler): Promise<ConfigurePreset | null> {
```
gcampbell-msft, line 816: I wonder if we should pull the `enableTryApplyDevEnv` out of this method, and do it outside of this method so that it's fully decoupled. 

We could maybe do it upon selecting the configure preset.

src/preset.ts
```diff
@@ -1470,37 +1503,44 @@ export async function expandBuildPreset(folder: string, name: string, workspaceF
...
-            expandedPreset.nativeToolOptions[index] = await expandString(preset.nativeToolOptions[index], expansionOpts);
+            expandedPreset.nativeToolOptions[index] = await expandString(preset.nativeToolOptions[index], expansionOpts, errorHandler);
         }
     }
 
     // Other fields can be copied by reference for simplicity
     merge(expandedPreset, preset);
 
+    if (errorHandler) {
+        for (const error of errorHandler.tempErrorList || []) {
```
gcampbell-msft, line 1525: Why do we need to use a `tempErrorList`? Why not just use a single error list? It seems like we essentially already do this and then just assign it here?

src/presetsController.ts
```diff
@@ -420,7 +429,7 @@ export class PresetsController {
 
             if (newPreset) {
 
-                const before: preset.ConfigurePreset[] = await this.getAllConfigurePresets();
+                const before: preset.ConfigurePreset[] = preset.allConfigurePresets(this.folderPath);
```
gcampbell-msft, line 432: Shouldn't this include the user presets?

qarni: What do you mean? `allConfigurePresets()` also gets the user presets. 

Although, a related note is that `getAllConfigurePresets()` and `allConfigurePresets()` do exactly the same thing now that I took out expandVendor and expandConditions, except the former is in `PresetsController` and the latter is in `Preset`.

I wanted to get rid of `allConfigurePresets()` - but it has this comment "Don't use this function if you need to keep any changes in the presets" and I'm not sure what that is referring to or if it's just outdated.

qarni: Oh I see. One is async and the other is not

src/preset.ts
```diff
@@ -619,42 +619,48 @@ const referencedConfigurePresets: Map<string, Set<string>> = new Map();
  * They should NOT be used together.
  * They should Not call each other.
  */
-export function expandVendorForConfigurePresets(folder: string): void {
+export async function expandVendorForConfigurePresets(folder: string, sourceDir: string, workspaceFolder: string, errorHandler?: ExpansionErrorHandler): Promise<void> {
```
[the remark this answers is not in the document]
qarni: Yeah I agree. Also, is there a reason you added `expandConfigurePreset` in there? Can I cut it back to where it was before without `expandConfigurePreset`

src/presetsController.ts
gcampbell-msft: Ah, I see, yeah it would be great to only have one of those methods to avoid confusion. Are there any differences besides async? Are they using the same underlying code?

src/preset.ts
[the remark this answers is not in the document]
gcampbell-msft: The reason `expandConfigurePreset` was in there was because it needs to expand things like the generator and other things to ensure that the vendor can expand it's macros correctly, also accounting for inheritance. However, if we handle this in the overall method, I don't think we'll have to worry about it.

src/presetsController.ts
qarni: Nope they're exactly the same except for async. I'm thinking I'll just keep the async one..?

qarni: Will handle this in a follow up cleanup/refactor PR

src/preset.ts
```diff
@@ -1180,7 +1055,64 @@ async function tryApplyVsDevEnv(preset: ConfigurePreset, workspaceFolder: string
...
+
+    // TODO: We likely need to refactor to include these refs, for configure, build, test, etc Presets.
+    const refs = referencedConfigurePresets.get(folder);
+    if (!refs) {
+        referencedConfigurePresets.set(folder, new Set());
+    } else {
+        refs.clear();
+    }
+
+    // if enableTryApplyDevEnv is true, we have to check if the preset is valid in expandedPresetsFiles
```
gcampbell-msft: Remind me what case this is protecting against?

src/preset.ts
```diff
@@ -1180,7 +1055,64 @@ async function tryApplyVsDevEnv(preset: ConfigurePreset, workspaceFolder: string
...
+
+    errorHandlerHelper(preset.name, errorHandler);
+
+    // Other fields can be copied by reference for simplicity
+    merge(expandedPreset, preset);
+
+    return expandedPreset;
+}
+
+async function expandConfigurePresetImpl(folder: string, name: string, allowUserPreset: boolean = false, useOriginalPresets: boolean = false, errorHandler?: ExpansionErrorHandler): Promise<ConfigurePreset | null> {
```
gcampbell-msft, line 1088: We should add a comment explaining the `useOriginalPresets` variable here so that people understand the use case and why we have it. It's nice to have this functionality overall, but right now it's purely for an edge case, so it might even be worthwhile to have the edge case described in short in the comment for this function.

src/preset.ts
```diff
@@ -1180,7 +1055,64 @@ async function tryApplyVsDevEnv(preset: ConfigurePreset, workspaceFolder: string
     preset.__parentEnvironment = EnvironmentUtils.mergePreserveNull([process.env, developerEnvironment]);
 }
 
-async function expandConfigurePresetHelper(folder: string, preset: ConfigurePreset, allowUserPreset: boolean = false, errorHandler?: ExpansionErrorHandler) {
+export async function expandConfigurePreset(folder: string, name: string, workspaceFolder: string, sourceDir: string, allowUserPreset: boolean = false, enableTryApplyDevEnv: boolean = true, errorHandler?: ExpansionErrorHandler): Promise<ConfigurePreset | null> {
```
gcampbell-msft, line 1058: Similar comment here: I think we should add a comment for the parameters explaining the context and what they are used for, especially the `enableTryApplyDevEnv`. 

Also, correct me if I'm wrong, but I think the default for the `enableTryApplyDevEnv` could probably be `false` now right? 
Since it should only be true when we're actually setting the configure preset and re-expanding? It's just a default so as long as we set it correctly at invocation time, it's fine either way, just make sure it's documented, whichever direction you pick 👍

src/cmakeProject.ts
```diff
@@ -266,11 +266,24 @@ export class CMakeProject {
             return undefined;
         }
         log.debug(localize('resolving.config.preset', 'Resolving the selected configure preset'));
+
+        // We want to use the original unexpanded preset file to apply the dev env in expandConfigurePreset
+        // we have to first check if the preset is valid in expandedPresetsFiles since we won't be expanding the whole file here, only the path up for this preset
```
gcampbell-msft, line 271: I noticed a very similar comment and code to this inside of the `expandConfigurePreset` method. Do we need both? It feels like it should only be in one place, likely here?

src/cmakeProject.ts
```diff
@@ -266,11 +266,24 @@ export class CMakeProject {
...
+        if (!preset.getPresetByName(preset.configurePresets(this.folderPath), configurePreset)) {
+            return undefined;
+        }
+
+        // TODO: move applyDevEnv here to decouple from expandConfigurePreset
```
gcampbell-msft, line 276: The TODO is fine, besides our offline discussion about how in general we should try to avoid too many of these so we don't just continually grow TODO's. We discussed why this case is fine 👍, just figured I'd comment on it again.

src/expand.ts
```diff
@@ -122,22 +126,24 @@ export async function expandString<T>(input: string | T, opts: ExpansionOptions)
...
+            const expansion = await expandStringHelper(result, opts, errorHandler);
             result = expansion.result;
             didReplacement = expansion.didReplacement;
             circularReference = expansion.circularReference;
             i++;
         } while (i < maxRecursion && opts.recursive && didReplacement && !circularReference);
 
         if (circularReference) {
-            log.error(localize('circular.variable.reference', 'Circular variable reference found: {0}', circularReference));
+            log.error(localize('circular.variable.reference.full', 'Circular variable reference found: {0}', circularReference));
```
gcampbell-msft, line 137: Why the changes to the key value for the localize calls here? Is there a reason why we're adding `full` to the end of them all?

gcampbell-msft: Ah, I see why now, disregard.

src/preset.ts
```diff
@@ -449,11 +449,11 @@ export function setOriginalUserPresetsFile(folder: string, presets: PresetsFile
     originalUserPresetsFiles.set(folder, presets);
 }
 
-export function setPresetsFile(folder: string, presets: PresetsFile | undefined) {
-    presetsFiles.set(folder, presets);
+export function setExpandedPresetsFile(folder: string, presets: PresetsFile | undefined) {
```
gcampbell-msft, line 452: Just confirming, this is just a rename to make it more obvious about how we're saving the state of these files right?

src/expand.ts
qarni: needed to add a localized string for the errorHandler with almost the same wording but a bit different so i figured id do that to differentiate

src/preset.ts
qarni: yup!

src/preset.ts
```diff
@@ -1949,14 +1954,15 @@ async function expandWorkflowPresetImpl(folder: string, name: string, workspaceF
...
+        return expandWorkflowPresetHelper(folder, preset, workspaceFolder, sourceDir, true, enableTryApplyDevEnv, errorHandler);
     }
 
     log.error(localize('workflow.preset.not.found', 'Could not find workflow preset with name {0}', name));
+    errorHandler?.tempErrorList.push([localize('workflow.preset.not.found', 'Could not find workflow preset'), name]);
     return null;
 }
 
-async function expandWorkflowPresetHelper(folder: string, preset: WorkflowPreset, workspaceFolder: string, sourceDir: string, allowUserPreset: boolean = false) {
+async function expandWorkflowPresetHelper(folder: string, preset: WorkflowPreset, workspaceFolder: string, sourceDir: string, allowUserPreset: boolean = false, enableTryApplyDevEnv: boolean = true, errorHandler?: ExpansionErrorHandler) {
```
gcampbell-msft, line 1965: Not for this PR, but it might be wise to make a TODO for the fact that now that we have a collection for the Problems Pane for CMakePresets validation and expansion, there are lots of other entry points where we could use it, for example, confirming that step 1 of the workflow preset is a configure step, doesn't need a TODO comment, but maybe we could make a note or issue for ourselves somewhere

src/presetsController.ts
```diff
@@ -9,13 +9,15 @@ import { fs } from '@cmt/pr';
...
-import { ExpansionOptions, getParentEnvSubstitutions, substituteAll } from '@cmt/expand';
+import { ExpansionErrorHandler, ExpansionOptions, getParentEnvSubstitutions, substituteAll } from '@cmt/expand';
 import paths from '@cmt/paths';
 import { KitsController } from '@cmt/kitsController';
 import { descriptionForKit, Kit, SpecialKits } from '@cmt/kit';
 import { getHostTargetArchString } from '@cmt/installs/visualStudio';
 import { loadSchema } from '@cmt/schema';
 import json5 = require('json5');
+import { Diagnostic, DiagnosticSeverity, Position, Range } from 'vscode';
+import collections from './diagnostics/collections';
```
gcampbell-msft, line 20: **NIT**: Could we possibly follow the pattern of the other imports with `from '@cmt/collections`?

src/presetsController.ts
```diff
@@ -174,6 +175,11 @@ export class PresetsController {
         // Private fields must be set after validation, otherwise validation would fail.
         this.populatePrivatePresetsFields(presetsFile, file);
         await this.mergeIncludeFiles(presetsFile, presetsFile, file, referencedFiles);
+
+        // set the pre-expanded version so we can call expandPresetsFile on it
+        setPresetsFile(this.folderPath, presetsFile);
```
gcampbell-msft, line 180: What about the current code requires that we call `setPresetsFile` first? It seems to me that it's passed in as an argument to `expandPresetsFile`

gcampbell-msft (changes requested):
Left a bunch of comments, just wanting to make sure we don't forget to tidy up the code and make sure that everything is well understood and documented before merging. Overall, the changes look good, with just a couple of implementation questions.

gcampbell-msft:
Comment to remind to test these changes for this issue as well: #2934 , just wanted to remind since it's not a devenv issue, we don't want to forget it.

src/preset.ts
qarni: a work item maybe? it'll also need to be fleshed out a bit so we can add different kinds of errors without wiping the rest

src/preset.ts
qarni: added comment and yeah I set it to be false

src/preset.ts
qarni: making sure that the preset we are setting is valid in the expanded presets map, since expansion will be using the original presets map which is pre validation

src/presetsController.ts
qarni: Its passed to `expandPresetsFile` so we can update it by reference, but `setPresetsFile` is still required because each of the individual `expand___Preset` functions dont take a whole presets file as a parameter - we call `getPresetByName(configurePresets(folder, useOriginalPresets), name)` or something similar which reads the preset file that we already have saved

qarni: From my most recent change, we could do this same thing but read from original presets file instead of this semi processed version of expanded presets... though it would effectively be the same thing just a matter of it looking cleaner one way or the other

src/preset.ts
gcampbell-msft: Yeah I think that's a good idea

src/presetsController.ts
```diff
@@ -171,9 +172,20 @@ export class PresetsController {
         }
 
         presetsFile = await this.validatePresetsFile(presetsFile, file);
-        // Private fields must be set after validation, otherwise validation would fail.
-        this.populatePrivatePresetsFields(presetsFile, file);
-        await this.mergeIncludeFiles(presetsFile, presetsFile, file, referencedFiles);
+
+        if (presetsFile) {
```
gcampbell-msft, line 176: How does this solve the `include` test?

src/presetsController.ts
```diff
@@ -816,11 +829,25 @@ export class PresetsController {
...
 
+    async getAllBuildPresets(): Promise<preset.BuildPreset[]> {
+        return preset.buildPresets(this.folderPath).concat(preset.userBuildPresets(this.folderPath));
+    }
+
+    async getAllTestPresets(): Promise<preset.TestPreset[]> {
+        return preset.testPresets(this.folderPath).concat(preset.userTestPresets(this.folderPath));
+    }
+
+    async getAllPackagePresets(): Promise<preset.PackagePreset[]> {
```
gcampbell-msft, line 843: Are these used anywhere anymore?

gcampbell-msft (dismissed):
Still some comments to address and make sure we have understanding on, but it looks good to me.

src/presetsController.ts
qarni: Yes, I didnt merge the two similar functions bc making everything async was a big change so im going to follow up with it

src/presetsController.ts
qarni: we need to run mergeIncludeFiles to get the original presets file updated with all possible presets. Otherwise, when we try to select a preset and configure, and we try to grab the original preset to do a full expansion, it wouldn't exist. I made a copy by value so it doesn't get expanded after that. Any thoughts?

qarni:
TODO's for follow up:
- restore the last selected preset after an error
- reduce all the getallpresets into one function
- move `applyDevEnv` to decouple from `expandConfigurePreset`

gcampbell-msft (approved):

PR #3912, opened 2024-07-17
Update OneLocBuild process

gcampbell-msft:


gcampbell-msft:
Just a pipeline change, pushing through.

PR #3918, opened 2024-07-18
correct_inheritence_of_vendor_when_expanding

AdamBartosCodasip:

- Fix `vendor` inheritance in presets. Without this change, `vendor` fields are inherited only when the child vendor is undefined.


gcampbell-msft:
@AdamBartosCodasip We recently made some refactors and changes to the expanding of presets, could you re-assess this PR to make sure, 1) this is still necessary, and 2) whether this still does what you expect?

gcampbell-msft (changes requested):
See other comment.

kodsingalur:
This is not needed anymore. Thanks.

AdamBartosCodasip:
This is not needed anymore. Thanks.

PR #3933, opened 2024-07-24
Prevent second configure after `QuickStart` if the `automaticReconfigure` setting is enabled

### LINKED ISSUE #3910: quick start causes second configure if automaticReconfigure is enabled
One thing I noticed while testing is that line 2863 (in the new version), will technically force a second configure if the `automaticReconfigure` setting is enabled. I don't think this needs to block this PR, but it is something we should look into.

_Originally posted by @gcampbell-msft in https://github.com/microsoft/vscode-cmake-tools/pull/3853#discussion_r1662917321_

qarni:
Fixes #3910


gcampbell-msft (approved):

gcampbell-msft:
No need for re-test and re-build. Merging.

PR #3935, opened 2024-07-24
Set usepresets context after manually creating presets file

### LINKED ISSUE #3854: usepresets context is not set after manually creating a cmakepresets.json
### Brief Issue Summary

usepresets context is not set after manually creating a cmakepresets.json, so the Select Configure Preset and Add Configure Preset options are not available in the command palette.

Might need to add/update a file watcher to update the active project when changes are made to a cmakepreset.json file. Also maybe for cmakeuesrpresets.json as well.

Related to #3734.

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

_No response_

qarni:
Fixes #3854. Support for watching multiple paths (cmakepresets.json/cmakeuserpresets.json/include files) was broken due to a Chokidar bug. Created a wrapper class to handle this scenario.


src/presetsController.ts
```diff
@@ -1943,26 +1943,64 @@ export class PresetsController {
...
             this._sourceDirChangedSub.dispose();
         }
     }
 }
+
+/**
+ * FileWatcher is a wrapper around chokidar's FSWatcher that allows for watching multiple paths.
+ * Chokidar's support for watching multiple paths is currently broken, if it is fixed in the future, this class can be removed.
+ */
+class FileWatcher {
```
gcampbell-msft, line 1984: If we are implementing `dispose`, we should likely about extending vscode.Disposable.

qarni: I see that PresetsController also has dispose() but it doesnt extend vscode.Disposable - is there a difference here?

src/presetsController.ts
```diff
@@ -1943,26 +1943,64 @@ export class PresetsController {
...
+    private watchers: Map<string, chokidar.FSWatcher>;
+
+    public constructor(paths: string | string[], eventHandlers: Map<string, () => void>, options?: chokidar.WatchOptions) {
+        this.watchers = new Map<string, chokidar.FSWatcher>();
+
+        for (const path of Array.isArray(paths) ? paths : [paths]) {
+            try {
+                const watcher = chokidar.watch(path, { ...options });
+                Array.from(eventHandlers).forEach(([event, handler]) => watcher.on(event, handler));
```
gcampbell-msft, line 1993: One thing to consider here. When `.forEach` is called, the loop does not wait for all of the `forEach` methods to finish before moving to the next line. 

So for example, if it takes long enough to add all of the handlers, `this.watchers.set` may be called prior to all of the handlers being added. While I don't immediately see any issue arising from this, I worry whether the watcher would get set before the changes and then they don't get stored. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#description

The alternative would be to use a traditional for loop here. What do you think?

src/presetsController.ts
gcampbell-msft: FIrst, my mistake, I meant `implements`. And I would actually think that we should be doing it there too, because it's what stores the watcher

gcampbell-msft: https://vscode-api.js.org/classes/vscode.Disposable.html

The theory is that our wrapper contains a watcher/event listener (disposable), and then the controller holds that wrapper

gcampbell-msft: It's not a major difference in implementation, and it won't make a huge difference in performance, but it's technically correct and if we ever get high up enough that we can add it to the vscode list of disposables, then vscode could handle disposal when it terminates. So I would just add `implements vscode.Disposable` and leave it at that.

gcampbell-msft (approved):

Issue #3522, opened 2024-01-02
CMake presets parser not compliant with CMake


christophe-calmejane:
### Issue Summary

The presets parser used in VSCode (and VSStudio) is not compliant with the cmake specifications. It's for example possible to create a set of preset files that will correctly parse using VSCode, but not using CMake itself (VSCode parser is accepting things it shouldn't). I've created a ticket on the CMake issue tracker with an example of such files: https://gitlab.kitware.com/cmake/cmake/-/issues/25505

When carefully reading CMake presets specifications, it's quite clear that the issue is actually a bug in VSCode/VSStudio (by being not strict enough).

gcampbell-msft:
Thanks for letting us know! This is definitely something that we want to make sure is matching cmake.exe, I'll mark it as a bug and place it on our backlog!

christophe-calmejane:
Just so you know, I successfully tested the CMake dev branch that fixes the issue I reported (all links in the CMake ticket). It's using (required) the upcoming `version: 9` field, so I truly hope the VSCode plugin will support it in the near future.
Thanks a lot.

christophe-calmejane:
Hi, small update: `version: 9` will come in next cmake version, the code has been merged.
Hopefully VSCode plugin will be updated to support all the new `cmake presets` features.

gcampbell-msft:
@christophe-calmejane Following up on this, are there explicit items that you've noticed that we should fix? I agree that in general we need to fix it to match ALL of the spec, but I wondered if there are specific things that we should prioritize. Thanks!

christophe-calmejane:
Sure, I'll try to do a small setup and share it. It will require ‛version: 9‛ support though, since it's the only way (with official cmake parser) to share common preset files with multiple projects.

christophe-calmejane:
Hi,

please find attached a sample projects that works fine when using cmake presets parser (requires version 9 support that will land in 3.30).

The idea is this:
- A cmake presets file is shared accros multiple projects (can be in a submodule or whatever). This file can hold shared compile definition, shared targets, ...
- Each project define the minimum ‛project specific‛ definition and include the ‛shared file‛ to minimize maintenance

After discussing the 2 possibilities with other cmake members, the retained solution was to extend (thus the version 9 for presets definition) the ‛include‛ directive so it can search for ‛sourceDir‛ variable expansion.

So basically in order for my sample project to work with VSCode, the required modifications are:
[...]

PR #3941, opened 2024-07-25
evaluate inherit presets based on presets that should be visible from file

moyo1997:
addresses part of #3522


src/presetsController.ts
```diff
@@ -1629,28 +1629,28 @@ export class PresetsController {
...
-                rootPresetsFile.packagePresets = includeFile.packagePresets.concat(rootPresetsFile.packagePresets || []);
+                presetsFile.packagePresets = includeFile.packagePresets.concat(presetsFile.packagePresets || []);
             }
             if (includeFile.workflowPresets) {
-                rootPresetsFile.workflowPresets = includeFile.workflowPresets.concat(rootPresetsFile.workflowPresets || []);
+                presetsFile.workflowPresets = includeFile.workflowPresets.concat(presetsFile.workflowPresets || []);
             }
-
-            // Recursively merge included files
-            await this.mergeIncludeFiles(rootPresetsFile, includeFile, fullIncludePath, referencedFiles);
```
gcampbell-msft, line 1653: What happened to the recursive mergeIncludeFiles call?

src/preset.ts
```diff
@@ -1130,16 +1130,20 @@ export async function expandConfigurePreset(folder: string, name: string, worksp
...
-        return expandConfigurePresetHelper(folder, preset, false, usePresetsPlusIncluded, errorHandler);
-    }
-
-    if (allowUserPreset) {
-        preset = getPresetByName(userConfigurePresets(folder, usePresetsPlusIncluded), name);
+async function expandConfigurePresetImpl(folder: string, name: string, allowUserPreset: boolean = false, usePresetsPlusIncluded: boolean = false, errorHandler?: ExpansionErrorHandler, inheritedByPreset?: ConfigurePreset): Promise<ConfigurePreset | null> {
+    const presetList = inheritedByPreset ? inheritedByPreset.__file!.configurePresets : configurePresets(folder, usePresetsPlusIncluded);
+    const validInherit = presetList !== undefined && presetList.filter(p => p.name === name).length > 0;
+    if (validInherit) {
+        let preset = getPresetByName(configurePresets(folder, usePresetsPlusIncluded), name);
```
gcampbell-msft, line 1137: Do we need to use `presetList` here on line 1137 rather than configurePresets again? 

I guess it would be fine either way, because we've already validated? 

THis is true for all other methods as well

src/preset.ts
```diff
@@ -1530,29 +1534,33 @@ export async function expandBuildPreset(folder: string, name: string, workspaceF
...
-        };
-        return expandBuildPresetHelper(folder, preset, workspaceFolder, sourceDir, parallelJobs, preferredGeneratorName, true, enableTryApplyDevEnv, errorHandler);
+        if (allowUserPreset) {
+            preset = getPresetByName(userBuildPresets(folder), name);
+            if (preset) {
+                return expandBuildPresetHelper(folder, preset, workspaceFolder, sourceDir, parallelJobs, preferredGeneratorName, true, enableTryApplyDevEnv, errorHandler);
+            }
+        }
+
+        if (name === defaultBuildPreset.name) {
```
gcampbell-msft, line 1553: This is where we're breaking the tests right now, we should also check for this case along with `validInherit` because otherwise we don't get to this if statement.

src/preset.ts
```diff
@@ -1530,29 +1534,33 @@ export async function expandBuildPreset(folder: string, name: string, workspaceF
...
-    if (preset) {
-        return expandBuildPresetHelper(folder, preset, workspaceFolder, sourceDir, parallelJobs, preferredGeneratorName, false, enableTryApplyDevEnv, errorHandler);
-    }
-
-    if (allowUserPreset) {
-        preset = getPresetByName(userBuildPresets(folder), name);
+async function expandBuildPresetImpl(folder: string, name: string, workspaceFolder: string, sourceDir: string, parallelJobs?: number, preferredGeneratorName?: string, allowUserPreset: boolean = false, configurePreset?: string, enableTryApplyDevEnv: boolean = true, errorHandler?: ExpansionErrorHandler, inheritedByPreset?: BuildPreset): Promise<BuildPreset | null> {
+    const presetList = inheritedByPreset ? inheritedByPreset.__file!.buildPresets : buildPresets(folder);
+    const validInherit = presetList !== undefined && presetList.filter(p => p.name === name).length > 0;
+    if (validInherit) {
```
gcampbell-msft, line 1540: I think for all preset types except for configure, the `validInherit` check shouldn't block the default preset, if it's a default preset then we are also valid to inherit.

src/preset.ts
moyo1997: yes since the validation has already happened, we can call regular configurepresets. I'm calling the configurepresets here, to make sure we are getting the right/updated preset based on the usepresetsplusincluded

src/preset.ts
moyo1997: updated pr, thanks!

gcampbell-msft (dismissed):

gcampbell-msft (approved):

PR #3953, opened 2024-08-01
FOLDER grouping in default build target dropdown

itzandroidtab:

- Grouping default build targets dropdown based on CMake FOLDER property

## The purpose of this change
When a project has a lot of targets it becomes a hassle to find the right default build target. In the project outline the targets are grouped. This PR also groups the default build target dropdown. [Discussion #3949](https://github.com/microsoft/vscode-cmake-tools/discussions/3949). When clicking on the folder it will show the corresponding targets.

#### Previous
image

#### New
image
After selecting the folder:
image


itzandroidtab:

gcampbell-msft:
@itzandroidtab Our apologies for the delay, we've been in the midst of the 1.19 official version release, so we've not been able to review community PR's. We hope to get to this soon, thanks and we appreciate your patience.

gcampbell-msft (changes requested):
@itzandroidtab First, please include a CHANGELOG.md entry. 

Also, I like the idea of this PR and the screenshots you've provided look really good! I haven't taken a deeply close look at the code, but again it looks good. 

HOWEVER, I think that this is probably a feature change that each user should be able to control if they would like. I think it'd be a great idea to add a setting that would allow people to choose whether they want this to be a flat list, or whether they want it to be structured using the FOLDER property. I could imagine a scenario where they only have a couple of targets, but if they're using the FOLDER property, even though there are only a limited set of options, they have to do multiple clicks to select it, due to the new implementation you've added. 

I'd love to approve this feature, but I think it should have the added ability to control whether we use the flat list or the FOLDER grouping based on a setting. I currently don't have a strong opinion on whether the setting should default to the flat list or the FOLDER grouping. 

@moyo1997 @itodirel @sinemakinci1 FYI, thoughts?

itzandroidtab:
@gcampbell-msft What would you call this option? And should this option also apply to the side view?

gcampbell-msft:
> @gcampbell-msft What would you call this option? And should this option also apply to the side view?

I don't believe this option should apply to the side view, the FOLDER grouping should be used there always. The setting should only be used in the build target dropdown. 

For the name, it could be something like: 
`cmake.useFolderPropertyInBuildTargetDropdown`

itzandroidtab:
Thank you. I will work on this

gcampbell-msft:
@itzandroidtab Changing this to a draft PR until further changes are made. Please let us know if you have any questions but we will leave it in a draft state for a while, and then if no updates are made we will close. Thanks!

itzandroidtab:
@gcampbell-msft I have added the feature to the CHANGELOG.md and added the option to enable/disable this feature. By default it is turned on

image

src/cmakeProject.ts
```diff
@@ -123,6 +124,15 @@ export interface ConfigureCancelInformation {
     canceled: boolean;
 }
 
+/**
+ * A target with a name, but no output. Only used for targets that have a
+ * folder property.
+ */
+interface FolderTarget {
```
gcampbell-msft, line 131: Should this be defined in the same location as the RichTarget or NamedTarget? To keep things consistent?

src/cmakeProject.ts
```diff
@@ -2272,7 +2282,65 @@ export class CMakeProject {
...
+                        return {
+                            label: t.name,
+                            description: localize('target.to.build.description', 'Target to build')
+                        };
+                    }
+                    case 'rich': {
+                        return { label: t.name, description: t.targetType, detail: t.filepath };
+                    }
+                    case 'folder': {
+                        return { label: t.name, description: 'FOLDER' };
```
gcampbell-msft, line 2322: 'FOLDER' should be localized, I would think

gcampbell-msft (changes requested):
Some nit changes, and I'm conflicted about whether to default the setting to `true` or `false`. 

Personally, I don't think it's a very risky change, but I lean towards defaulting it to false, since it will affect the search capability and most projects are expecting a straight forward list

itzandroidtab:
I have changed the default to `false`

src/cmakeProject.ts
gcampbell-msft: Good point, I agree, thanks!

PR #3980, opened 2024-08-13
Update dependencies.

sean-mcmanus:
For https://github.com/microsoft/vscode-cmake-tools/security/dependabot/55 and some other issue in xml2js.


sean-mcmanus:
@gcampbell-msft The linter isn't working? I get the same issue in main.

gcampbell-msft:
@sean-mcmanus Looks like the errors are coming while parsing `api.ts`.

PR #3991, opened 2024-08-15
fix regression of kit environment variables

gcampbell-msft:
See title. 

Fixing issue where we incorrectly were merging the kit environment on top of `penv` before expanding. We don't support referencing the `environmentVariables` map.


moyo1997 (approved):
looks good! Only question is should we just completely remove penvoverride, since we should not be

gcampbell-msft:
@moyo1997 Well, technically it's going to have some level of `penv` supported either way, because `expandString` will call it's parent environment replacer regardless, so I think we might as well have that.

PR #4026, opened 2024-08-29
Don't check generator on PATH if manually specified

jophippe:
### This changes how the CMake generator gets selected

- Add a flag to the get best generator logic to tell whether we're selecting a default generator or a manually specified generator.
- If we're selecting a default generator, keep the old behavior and check whether the generator exists on PATH
- If we're selecting a manually specified generator (whether through the cmake.generator setting, the cmake.preferredGenerators setting, or from the kit), we can assume the generator exists and shouldn't check the PATH.

### DIFF
--- a/src/drivers/cmakeDriver.ts
+++ b/src/drivers/cmakeDriver.ts
@@ -727,21 +727,11 @@ export abstract class CMakeDriver implements vscode.Disposable {
         this._kitDetect = await getKitDetect(this._kit);
         log.debug(localize('cmakedriver.kit.set.to', 'CMakeDriver Kit set to {0}', kit.name));
         this._kitEnvironmentVariables = await effectiveKitEnvironment(kit, this.expansionOptions);
+        let usingDefaultGenerators = false;
 
         // Place a kit preferred generator at the front of the list
         if (kit.preferredGenerator) {
             preferredGenerators.unshift(kit.preferredGenerator);
-        }
-
-        // If no preferred generator is defined by the current kit or the user settings,
-        // it's time to consider the defaults.
-        if (preferredGenerators.length === 0
-            && !(this.usingFileApi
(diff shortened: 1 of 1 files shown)


src/drivers/cmakeDriver.ts
```diff
@@ -979,27 +981,39 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
-                }
-                if (gen_name === 'Unix Makefiles') {
-                    return this.testHaveCommand('make');
-                }
-                if (gen_name === 'MSYS Makefiles') {
-                    return platform === 'win32' && this.testHaveCommand('make');
+                // If we're not using the default generators, then the generator was manually specified
+                // in that case, assume it exists and don't check the PATH, only check if it's a supported generator.
+                // If we are using the default generators, then we should also check if it is on the PATH
+                if (!usingDefaultGenerators) {
```
gcampbell-msft, line 994: Can we use `usingFallbackGenerators`?

src/drivers/cmakeDriver.ts
```diff
@@ -979,27 +981,39 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
-                if (gen_name === 'Unix Makefiles') {
-                    return this.testHaveCommand('make');
-                }
-                if (gen_name === 'MSYS Makefiles') {
-                    return platform === 'win32' && this.testHaveCommand('make');
+                // If we're not using the default generators, then the generator was manually specified
+                // in that case, assume it exists and don't check the PATH, only check if it's a supported generator.
+                // If we are using the default generators, then we should also check if it is on the PATH
+                if (!usingFallbackGenerators) {
+                    return gen_name === 'Ninja' ||
```
gcampbell-msft, line 995: If we're going back to, if the user has `cmake.generator` set, we use it without conditions, I don't think we should have these conditions here, we should simply return `true`.

gcampbell-msft (changes requested):

PR #4044, opened 2024-09-05
determine shell for .cmd, .bat, .ps1

### LINKED ISSUE #4037: CMake executable cannot be a script
### Brief Issue Summary

The option "cmake.cmakePath" cannot refer to a script.
For example: (cmake.bat)
````
C:\path\to\cmake.exe $*
````
(In the real case, this is more complex as this triggers some remote execution)

Output:
````
[cmakeExecutable] CMake executable not found in cache. Checking again.
[proc] Executing command: C:/path/to/cmake.cmd --version
[proc] Executing command: C:/path/to/cmake.cmd -E capabilities
````

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

I already did some reverse engineering on the code.
proc.ts has a method called execute which accepts some options. If options.shell is set to true, it should enable the shell option as described in https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows
However cmakeExecutable.ts in the method getCMakeExecutableInformation doesn't pass this option. `const execOpt: proc.ExecutionOptions = { showOutputOnError: true };`

### LINKED ISSUE #4042: Error when ninja is not in PATH (but ninja can be executed by cmake)
### Brief Issue Summary

We have a slightly special setup, where we wrap `cmake` in a wrapper script (via `cmake.cmakePath`). The wrapper script helps us to isolate certain parts of the build system that should be configured in a specific way. All that is working well, however in recent versions of this extension, we started to get errors of the form:
```
[proc] The command: ninja --version failed with error: Error: spawn ninja ENOENT
[proc] The command: ninja-build --version failed with error: Error: spawn ninja-build ENOENT
[proc] The command: ninja --version failed with error: Error: spawn ninja ENOENT
[proc] The command: ninja-build --version failed with error: Error: spawn ninja-build ENOENT
```

Indeed we do not have `ninja` in the path. When our wrapper script is called with arguments for the build, it will put ninja on the path, so it can be found. But on a plain shell, `ninja` is not available.

Is that something that can be considered, or worked around?

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.92.2",
  "cmtVersion": "1.19.50",
  "configurations": [
    {
      "folder": "/home/memmenlauer/BDA/Src/BDAPipelineProcessing",
      "cmakeVersion": "3.30.2",
      "configured": false,
      "generator": "",
      "usesPresets": false,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "useCMakePresets": "auto",
      "configureOnOpen": false
    }
  ]
}
```

### Debug Log

_No response_

### Additional Information

_No response_

gcampbell-msft:
Test VSIX:
[cmake-tools.zip](https://github.com/user-attachments/files/16893465/cmake-tools.zip)

Fixes #4037. Fixes #4042 

We did not change anything regarding this between 1.18 and 1.19. This issue comes from a breaking change in VS Code, where in version 1.92, which came out in July, they update to node v20 which requires more strict options regarding proc.spawn: [Visual Studio Code July 2024](https://code.visualstudio.com/updates/v1_92#_electron-30-update).

### DIFF
--- a/src/proc.ts
+++ b/src/proc.ts
@@ -161,7 +161,7 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
 
     const spawn_opts: proc.SpawnOptions = {
         env: final_env,
-        shell: options.shell ?? determineShell(command)
+        shell: options.shell ?? process.platform === "win32" ? determineShell(command) : undefined
     };
     if (options?.cwd !== undefined) {
         util.createDirIfNotExistsSync(options.cwd);
--- a/src/proc.ts
+++ b/src/proc.ts
@@ -159,9 +159,13 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
         }
     }
 
+    if (options.shell === undefined && process.platform === "win32") {
+        options.shell = determineShell(command);
+    }
+
     const spawn_opts: proc.SpawnOptions = {
         env: final_env,
-        shell: options.shell ?? process.platform === "win32" ? determineShell(command) : false
+        shell: !!options.shell
     };
     if (options?.cwd !== undefined) {
         util.createDirIfNotExistsSync(options.cwd);
--- a/src/proc.ts
+++ b/src/proc.ts
@@ -161,7 +161,7 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
 
     const spawn_opts: proc.SpawnOptions = {
         env: final_env,
-        shell: options.shell ?? process.platform === "win32" ? determineShell(command) : undefined
+        shell: options.shell ?? process.platform === "win32" ? determineShell(command) : false
     };
     if (options?.cwd !== undefined) {
         util.createDirIfNotExistsSync(options.cwd);
--- a/src/proc.ts
+++ b/src/proc.ts
@@ -95,7 +95,7 @@ export interface ExecutionResult {
 
 export interface ExecutionOptions {
     environment?: Environment;
-    shell?: boolean;
+    shell?: boolean | string;
     silent?: boolean;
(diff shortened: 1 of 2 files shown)


gcampbell-msft:
Likely need more logic for only doing determine shell on Windows.

src/proc.ts
```diff
@@ -146,6 +158,11 @@ export function execute(command: string, args?: string[], outputConsumer?: Outpu
             log.debug(localize('execution.environment', '  with environment: {0}', JSON.stringify(final_env)));
         }
     }
+
+    if (options.shell === undefined && process.platform === "win32") {
```
snehara99, line 162: should `undefined` be treated differently than `false`?

gcampbell-msft: Yes. If it's false, then we are explicitly saying that we don't want the shell, if it's `undefined`, in the Windows case, we now need to check if it's a ps1 or bat script and ensure we handle it accordingly.

gcampbell-msft: However, I am pushing a change to change the order of the if statement, we want to check if it's windows first

snehara99: Ah gotcha

snehara99 (approved):

PR #4055, opened 2024-09-10
update node version, @vsce/vsce, and fix build errors from node updat…

gcampbell-msft:
Update our signing so that we support the new signing support partially documented here: https://gist.github.com/dtivel/c0b722fe1756c2b614d7096bde602c66. 

To complete this, it required updating our signing as well as our node version and vsce version, and the resulting changes required for building. 

NOTE: I've locally verified that the new signing process worked.


src/pr.ts
```diff
@@ -49,9 +49,9 @@ export namespace fs {
         return str;
     }
 
-    export function readFile(filePath: string, encoding: string = "utf8"): Promise<any> {
+    export function readFile(filePath: string, encoding: BufferEncoding = "utf8"): Promise<any> {
```
qarni, line 52: Why this change?

qarni (approved):

jobs/signing/SignVsix.proj
```diff
@@ -10,8 +10,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <FilesToSign Include="$(OutDir)\**\*.vsix">
-      <Authenticode>VsixSHA2</Authenticode>
+    <FilesToSign Include="$(OutDir)\vsix\extension.signature.p7s">
```
qarni, line 13: Hmm I also don't understand this change

src/pr.ts
gcampbell-msft: This is a change due to the update to node 20, the types changed and so we had to match the type change that node20 had.

jobs/signing/SignVsix.proj
gcampbell-msft: This is the update for signing. We can't sign the vsix with VsixSha2 anymore because it messes up the signing of the manifest.

gcampbell-msft: This is how the docs say we should sign

gcampbell-msft:
Blocking comment to test some with the vsix to ensure it behaves as expected.

gcampbell-msft:
NOTE: I performed a super fast smoke test to make sure things were behaving as expected, and they are.

PR #4063, opened 2024-09-13
ignore javascript files in codesign validation

gcampbell-msft:
See title.


jobs/release/prerelease.yml
```diff
@@ -29,6 +29,9 @@ extends:
         templateContext:
           type: releaseJob
           isProduction: true
+          sdl:
+            codeSignValidation:
+              additionalTargetsGlobPattern: -|*.js
```
jophippe, line 34: I think you should use the **\\*.js pattern instead of *.js, since that's the pattern that Code Sign Validation uses to include .js files by default

gcampbell-msft: Updated.

jophippe (approved):

PR #4078, opened 2024-09-18
Update to use CFS feed

qarni:
See title.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,3 @@
+registry=https://pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages/npm/registry/ 
+                        
+always-auth=true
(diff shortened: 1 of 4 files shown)


.npmrc
```diff
@@ -0,0 +1,4 @@
+# Delete this .npmrc file to use the npmjs registry instead of the Microsoft Azure Artifacts feed for local development
```
gcampbell-msft, line 1: Do we think we should put something referring only to OSS at the beginning of this comment? 
```suggestion
# [OSS Contributors] Delete this .npmrc file to use the npmjs registry instead of the Microsoft Azure Artifacts feed for local development
```

qarni: yeah this makes sense

gcampbell-msft (approved):

Issue #4059, opened 2024-09-12
CMakePresets.json can't inherit from preset in CMakeUserPresets.json


michaelbjerking:
### Brief Issue Summary

When trying to select a preset in UI the below error is printed i "output"

[preset] Preset unittest-gcc9 in CMakePresets.json can't inherit from preset unittest in CMakeUserPresets.json
[preset] Preset unittest-clang in CMakePresets.json can't inherit from preset unittest in CMakeUserPresets.json

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.93.0",
  "cmtVersion": "1.19.51",
  "configurations": [
    {
      "folder": "/home/mbjerking/workspace/test",
      "cmakeVersion": "3.25.20230103",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {
        "C": "/usr/bin/gcc",
        "CXX": "/usr/bin/g++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "always",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
[main] Configuring project: test 
[main] Saving open files before configure/build
[driver] Removing /home/mbjerking/workspace/test/build/unittest-gcc9/CMakeCache.txt
[driver] Removing /home/mbjerking/workspace/test/build/unittest-gcc9/CMakeFiles
[driver] Start configure 
[driver] Running pre-configure checks and steps
[driver] NOTE: You are configuring with preset unittest-gcc9, but there are some overrides being applied from your VS Code settings.
[cmakefileapi-driver] Configuring using preset
[...]

PR #4081, opened 2024-09-18
only set `isUserPreset` to `true` if it's defined in a user presets file

gcampbell-msft:
Fixes bug #4059. 

We were not showing presets as being available to select because we are determining that the preset is a user preset, when it really isn't. This is due to the fact that we were incorrectly setting `isUserPreset`, we should check which file the configure preset was defined in.

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -467,24 +467,39 @@ export function setPresetsPlusIncluded(folder: string, presets: PresetsFile | un
     presetsPlusIncluded.set(folder, presets);
 }
 
-export function setUserPresetsPlusIncluded(folder: string, presets: PresetsFile | undefined) {
+export function setUserPresetsHelper(presets: PresetsFile | undefined) {
     if (presets) {
+        // for each condition of `isUserPreset`, if we don't find file.path, then we default to true like before.
         if (presets.configurePresets) {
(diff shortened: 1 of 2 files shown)


michaelbjerking (approved):
Tested fix locally, and it fix the issue with CMakePresets.json "trying" to inherit from preset in CMakeUserPresets.json

gcampbell-msft:
@moyo1997 Ping for review, thanks!

src/preset.ts
```diff
@@ -467,24 +467,39 @@ export function setPresetsPlusIncluded(folder: string, presets: PresetsFile | un
     presetsPlusIncluded.set(folder, presets);
 }
 
-export function setUserPresetsPlusIncluded(folder: string, presets: PresetsFile | undefined) {
+export function setUserPresetsHelper(presets: PresetsFile | undefined) {
```
snehara99, line 470: Why did you make this a helper?

snehara99: oh nvm i guess you wanted a version without the `userPresetsPlusIncluded.set(folder, presets)` line

snehara99 (dismissed):

snehara99 (approved):

snehara99 (approved):

PR #4086, opened 2024-09-19
use npm directly

gcampbell-msft:
See title.


bobbrow:
Was there a problem with the tasks?  We use Npm for our extension too and wondering what this is for.

snehara99 (approved):

PR #4088, opened 2024-09-20
Ensure signing/packaging works as expected

gcampbell-msft:
See title.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3677,7 +3677,6 @@
     "@test": "out/test/"
   },
   "scripts": {
-    "vscode:prepublish": "yarn run compile-production",
     "compile": "yarn install && webpack --mode development",
     "compile-watch": "yarn install && webpack --mode development --watch --progress",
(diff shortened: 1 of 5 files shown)


kuchungmsft (approved):

package.json
```diff
@@ -3677,7 +3677,6 @@
     "@test": "out/test/"
   },
   "scripts": {
-    "vscode:prepublish": "yarn run compile-production",
```
benmcmorran, line 3680: Why is this removed? Will that break external contributors trying to package the VSIX themselves?

benmcmorran (approved):

package.json
gcampbell-msft: No, it just means that they'll need to do the same as our pipeline, `yarn run compile-production`, and then `yarn package` or the `vsce package` command

We have to remove this because in our pipeline, when we run `vsce package` it will do the `vscode:prepublish` command, which rebuilds and regenerates loc files, etc. This means that in our pipelines, the files that we're signed are getting corrupted when we try to package

Issue #4116, opened 2024-10-08
Test output isn't visible when failed


redstrate:
### Brief Issue Summary

When a test fails, the test output isn't visible... which is where the actual error is located. For example, in the code lens:

![image](https://github.com/user-attachments/assets/9cbf903a-b41e-4da7-a7fa-0e6b28c73e2b)

And when clicking the test in "Test Results":

![image](https://github.com/user-attachments/assets/6560ccec-a442-40aa-96f7-b7014feb35dc)

It's possible to work around this by pressing the "Show Result Output" button to view the actual output, but it isn't immediately obvious why I need to do that :sweat_smile: Is there something I'm missing or is this intentional?

I tracked down the [line to here](https://github.com/microsoft/vscode-cmake-tools/blob/ed37b19f512fb654ba64a6f32b5cf81b55351267/src/ctest.ts#L553) and I can even modify the extension to print the test output, but that seems wrong since it should have that log anyway. But if that would be better, I can make a pull request.

![image](https://github.com/user-attachments/assets/723aa8d4-49e9-4e1b-97a2-52e23c488cd6)

This doesn't have to do anything with my testing setup, even a Google.Test example I pulled from GH exhibits the same issue once I forced one of their tests to fail. Successful tests do print their log correctly.

Amy-Li03:
Hi @redstrate, thanks for your reporting!
@gcampbell-msft when a test fails, the test output does not show detailed error messages on the code line, but we can press the "Show Result Output" button to view the actual output. 
**Maybe this is by design?**

gcampbell-msft:
@redstrate This seems like a reasonable improvement to see the output without having to click a button, we'd be happy to accept a PR from you!

PR #4129, opened 2024-10-15
Improve test case failure/skip states and their output (#4116)

redstrate:
### This changes visible test output

- Change the test output message when encountering an error exit code.
- Check for skipped tests and mark them as skipped.

## The purpose of this change

Currently if a test fails, it's not immediately clear why that is because we give the exit code - and not the output from ctest. That is now fixed and the ctest output is shown which greatly improves the usefulness of the Test Results pane and other places where this output is shown like Code Lens.

Skipped test cases were also not actually marked as skipped, and that is now fixed as well.

## Other Notes/Information

Comparison screenshots to show the improvement:

### Failed Tests

| Before | After |
| --- | --- |
| Screenshot_20241015_125716 | Screenshot_20241015_125209 |

### Skipped Tests

| Before | After |
| --- | --- |
| Screenshot_20241015_125700 | Screenshot_20241015_125759 |

Unfortunately this google test example I used doesn't set test source location, but if it did then code lens would be affected as well like in this outdated screenshot from the issue:

image


redstrate:
@microsoft-github-policy-service agree company="KDAB"

gcampbell-msft (changes requested):
@redstrate This looks great! 

If you could please make a CHANGELOG entry, then we can approve and accept this PR! Thanks.

redstrate:
> @redstrate This looks great!
> 
> If you could please make a CHANGELOG entry, then we can approve and accept this PR! Thanks.

Done!

gcampbell-msft (approved):

src/ctest.ts
```diff
@@ -550,15 +550,19 @@ export class CTestDriver implements vscode.Disposable {
...
+                    new vscode.TestMessage(localize('test.failed.with.exit.code', '{0}\nTest {1} failed with exit code {2}.', output, testName, exitCode)),
                     failureDuration
                 );
             } else if (completionStatus !== undefined) {
-                this.ctestErrored(
-                    test,
-                    run,
-                    new vscode.TestMessage(localize('test.failed.with.completion.status', 'Test {0} failed with completion status "{1}".', testName, completionStatus))
-                );
+                if (completionStatus == "SKIP_REGULAR_EXPRESSION_MATCHED") {
```
gcampbell-msft, line 557: Needs to be `===` instead of `==`. Thanks.

gcampbell-msft (changes requested):
See other comments.

gcampbell-msft (approved):

PR #4139, opened 2024-10-23
ensure the icon sizes are correctly created

gcampbell-msft:
Some minor changes were needed to ensure that the sizes of the icons were correct when rendered into the IDE.


DeltaRazero:
If all SVG source icons have a tiny 'canvas' size so icons aren't scaled up, warping/distortion can occur due to rounding errors for the anchor point coordinates in the icon.
The font height property ensures all icons are big enough to prevent that. I think that is important to point out.

gcampbell-msft:
@DeltaRazero Interesting, okay, so you're saying that simply having the `normalize` set to `true` should fix the issue still?

DeltaRazero:
@gcampbell-msft Yes, I've tested this and verified the scaling is fixed by simply setting the `normalize` attribute.
`uE004_cmake-view-icon-1` and `uE005_cmake-view-icon-2` have bigger canvas sizes (36pt vs 16pt of the other icons), so the other icons aren't scaled and appear small. Setting `normalize` will scale them to 36pt.

As for the `fontHeight` option, I made a comparison: both of these pictures are taken with `normalize` set to `true` and VSCode interface with 6x zoom @ 1440p resolution.

If you look closely, you can already see some small rounding errors. For example, looking at the cross and arrows shapes in the custom icons, they appear rounded and don't have a uniform line thickness. With more complex shapes, this could result in off-looking or blurry icons. I believe this is a limitation of many font formats not being able to store tiny glyphs with enough precision, hence pre-scaling them fixes this.

### fontHeight: omitted
image

### fontHeight: 1000
image

moyo1997 (approved):

PR #4140, opened 2024-10-23
ensure we build before packaging in CG pipeline

gcampbell-msft:
See title.


gcampbell-msft:
Just a pipeline change, no need for tests/approval

PR #4149, opened 2024-10-29
Sanitize cmake.copyCompileCommands

### LINKED ISSUE #3874: copyCompileCommands to the original path still leaves a corrupted CDB on Windows
This bug was initially identified on all platforms in #3214, and was resolved for non-Windows platforms in #3500.  However, the problem still persists on Windows due to differences in path separators.  Example:

```json
"cmake.buildDirectory": "${workspaceFolder}/firmware/build"
```

on Windows gets sanitized to `your/path/to/firmware/build/compile_commands.json`.  But:

```json
"cmake.copyCompileCommands": "${workspaceFolder}/firmware/build/compile_commands.json"
```

does not get sanitized on Windows, and results in `your\path\to/firmware/build/compile_commands.json`.  I can confirm this is the problem by manually entering the absolute path with identical path separators as the sanitized input, and the feature will work as intended.  The result is there is still CDB corruption on Windows when using `${workspaceFolder}`.

_Originally posted by @recursivenomad in https://github.com/microsoft/vscode-cmake-tools/issues/3214#issuecomment-2198199560_

gcampbell-msft:
Fixes #3874.

Ensures that we're sanitizing the paths so that we accomplish two things:
1. Parsing the file will work as expected because the file path is sanitized on all platforms
2. Ensures that we sanitize all paths so that we can safely compare paths.


src/cmakeProject.ts
```diff
@@ -1541,20 +1541,21 @@ export class CMakeProject {
...
+                    const copyDest = util.lightNormalizePath(this.workspaceContext.config.copyCompileCommands);
+                    const expandedDest = util.lightNormalizePath(await expandString(copyDest, opts));
                     if (compdbPath !== expandedDest) {
                         const parentDir = path.dirname(expandedDest);
                         try {
                             log.debug(localize('copy.compile.commands', 'Copying {2} from {0} to {1}', compdbPath, expandedDest, 'compile_commands.json'));
                             await fs.mkdir_p(parentDir);
                             try {
                                 await fs.copyFile(compdbPath, expandedDest);
+                                compdbPaths.push(expandedDest);
```
moyo1997, line 1558: curious why we're also adding this path? It should be the same file as the previously added file in line 1546?

gcampbell-msft: No, it's not the same file. The one on line 1546 is where CMake / the build system outputs the compile_commands.json to start with. However, the `expandedDest` is wherever the user has defined in the setting `cmake.copyCompileCommands`, which could be somewhere different.

paulmaybee (approved):

moyo1997 (approved):

moyo1997 (approved):

PR #4202, opened 2024-12-06
Fix configure preset used in workflow preset defined in CMakeUserPresets.json

nikobockerman:
Previously, when workflow preset in CMakeUserPresets.json references configure preset that is also defined in CMakeUserPresets.json, this extension reports an error that the configure preset can't be found.
And that results in the CMakeUserPresets.json getting ignored along with all presets defined in it.

This commits fixes this problem by adding a lookup from userConfigurePresets in case the configure preset isn't found in configurePresets.


gcampbell-msft (dismissed):

gcampbell-msft:
@nikobockerman Approved, but can you also add a CHANGELOG.md update?

nikobockerman:

gcampbell-msft (approved):

Amy-Li03:
Verified this issue on cmake tools v1.20.50(pre-release) & v1.20.52(release), issue was fixed. We can find the workflow preset when workflow preset in CMakeUserPresets.json references configure preset that is also defined in CMakeUserPresets.json. Please see following recording:
Fixed

Issue #3598, opened 2024-02-19
Ability to pass custom target to "CMake: Build Target" command (aka `cmake.buildWithTarget`) as argument


i-ky:
### Brief Issue Summary

I would like to configure a task which would depend on building a specific target. It is possible to call a command from a task using `"inputs"`, `tasks.json` would look something like this:
```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Build Target",
      "command": "echo ${input:buildTarget}",
      "type": "shell",
      "problemMatcher": []
    }
  ],
  "inputs": [
    {
      "id": "buildTarget",
      "type": "command",
      "command": "cmake.buildWithTarget",
      "args": "targetName"
    }
  ]
}
```
Unfortunately, `cmake.buildWithTarget` does not support receiving target as argument, instead it unconditionally prompts user for target selection:
https://github.com/microsoft/vscode-cmake-tools/blob/2878725633841f0e63499bd51b1e26291c104884/src/cmakeProject.ts#L2066-L2067

This is not acceptable in my use case, because there is a _specific_ target I want to build and I don't want to bother workspace users with target selection (which would be non-user-friendly and error-prone).

Could you please make `cmake.buildWithTarget` accept target as argument?

gcampbell-msft:
@i-ky Thank you for the request / suggestion! I have placed it on the backlog, but if you want this feature quickly, we would certainly welcome an open-source contribution!

Thanks! 😊

hippo91:
@i-ky i made the #4206 PR. It would be nice if you could test it to confirm it does the job.

Yingzi1234:
This issue has been fixed on latest version 1.21.0(pre-release) , you can get the details below. Thanks!

![Image](https://github.com/user-attachments/assets/438e046b-5fde-47b3-9346-cac80b29e16c)

i-ky:
@hippo91, @gcampbell-msft, @Yingzi1234, thank you!

Issue #1601, opened 2020-12-17
CMake: Build Target from tasks.json to support compound debugging


timo-suoranta-varjo:
- I want to debug multiple individual executables together (server, client) that have no build time dependencies.
- I have `compounds` in `launch.json` that works for this.
- For each executable in my compound, I'd like to have `preLaunchTask` in `launch.json` to make sure those executables are rebuilt before debugging.
- In the preLaunchTasks I would like to be able to perform `CMake: Build Target`, so that I can build those targets.

andreeis:
Thank you for suggesting this feature idea. Currently, the command "CMake: Build Target" can't be hooked up to tasks.json, only "CMake: Build" (which builds the default target if defined or ALL otherwise).
You could use this one for now, even if this would cause the rebuild of targets that you don't need.

timo-suoranta-varjo:
Building all is not always an option. Building all of a large project can take a really long time. Some targets might not build at all, because the project has not yet been fully ported to a new environment.

I worked around this by using shell command `cmake --build .. --config .. --target .. -- -j ..` in `tasks.json`, essentially copying what `CMake: Build Target` outputs when it executes. But I need to copy paste an entry for every target.

MartyMcFlyInTheSky:
Yes, please make this work:

`{
    "version": "2.0.0",
    "tasks": [
        {
            "type": "cmake",
            "label": "CMake: build",
            "command": "build",
            "targets": [
                "${command:cmake.launchTarget}"
            ],
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "problemMatcher": [],
            "detail": "CMake template build task"
        }
    ],
}`

It should be as easy as just making the target name available (instead of the full path).

nickyvanurk:
I am also running in this issue, would love to use the build target from tasks

PR #4206, opened 2024-12-07
Fix_3598

hippo91:

- Add optional argument to `buildWithTarget` method of the `CMakeProject` class;
- Add optional argument to `buildWithTarget` method of the `ExtensionManager` class.

## Other Notes/Information

This MR enable the use of a custom task that executes `buildWithTarget` command given an argument.
The video below show this:

[BuildWithTarget.webm](https://github.com/user-attachments/assets/acf24660-8feb-4cce-8d67-3ff56f167df0)

Please note two things:
- I get a "Cannot substitute input variable ..." warning message
- This MR does not override the `CMake: Build Target` palette command whereas it was probably the original demand in #3598. Please not that this would probably require to solve #1601 ?

I did not add test for this. Please let me know if it is necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Features:
 
 Improvements:
 
+- Add ability to pass custom target to "CMake: Build Target" command as argument. [#3598](https://github.com/microsoft/vscode-cmake-tools/issues/3598)
 - Fix "Unable to resolve configuration with compilerPath" issue for Swift. [#4097](https://github.com/microsoft/vscode-cmake-tools/issues/4097)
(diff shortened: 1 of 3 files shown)


gcampbell-msft:
@hippo91 Overall, these changes look reasonable and look like they have very low risk for regressions. However, could you please add a CHANGELOG update and some screenshots or a video of how this works in the description? Thanks!

hippo91:
@gcampbell-msft i added an entry in the CHANGELOG and modified the MR description, adding a video.

gcampbell-msft (approved):

Issue #3232, opened 2023-07-07
Tasks: Configure Task lists only first folder in workspace


szekelyisz:
### Brief Issue Summary

I have several folders linked into my vscode workspace, mostly TypeScript projects and also a CMake one. When I run `Tasks: Configure Task`, the only choice to create a task with the CMake template is for the top and first folder. I had to drag my CMake project to the top to be given the right choice when creating a task, targeting my CMake project instead of a TypeScript one.

I would expect all templates be available to all folders in the workspace just like the way the NodeJS task generator works.

benmcmorran:
Thanks for the suggestion! Your suggested change makes sense to me. We may not be able to prioritize it immediately but would welcome a pull request if someone else gets to it first.

hippo91:
@szekelyisz i would like to tackle this issue but cannot reproduce clearly the problem. Can you give me a reproducer or at least at screencast illustrating it?

Yingzi1234:
@gcampbell-msft We ran into a problem when trying to reproduce the user's question. After running the command A, we did not see the selected folder name within the list within the command line. Is there any way we can see exactly which folder was run?

![Image](https://github.com/user-attachments/assets/b0bc6567-ed0d-4280-b2b7-55c62fe24d25)

**Customer:**
![Image](https://github.com/user-attachments/assets/bdc15e12-cd23-453d-9a53-3b400dbeca29)

szekelyisz:
> @gcampbell-msft We ran into a problem when trying to reproduce the user's question. After running the command A, we did not see the selected folder name within the list within the command line. Is there any way we can see exactly which folder was run?

Did you try with multiple folders in your workspace? I think the folder name is only shown then.

Yingzi1234:
@szekelyisz Yes I did create two different folders in the workspace, I'm not sure if the cause of your error is related to the configure project that was selected at the very beginning after run command CMake: configure, my reproduction steps are below, you can take a look.
1. Open a test file containing multiple folders
2. Click F1 to run command `CMake: Configure`
3. Choose one of the test folders
4. Click F1 to run command `Task: configure task`
![Image](https://github.com/user-attachments/assets/b974db44-a8e5-4497-9383-041429c5cb6b)

szekelyisz:
[...]

PR #4223, opened 2025-01-03
Fix 3232

hippo91:

- Two methods are added to the `ExtensionManager` class (`getAllCMakeProjects`, `getProjectForFolder`). They are just delegations to the `projectController` object.
- The `findBuildTask` method of the `CMakeTaskProvider` is expecting a workspace folder (under string form) as first argument. Its call site in the `cmakeDriver.ts`  file is updated accordingly
- The `provideTasks` method of the `CMakeTaskProvider` does not deal with the active project anymore but loop over all the CMake projects of the workspace to build the task list. 
- Accordingly, the `provideTask` method of the `CMakeTaskProvider` is expecting a workspace folder as 2nd argument so that the scope of each task is the given workspace folder. This folder is also used to specialize the task definition label in order for the definition to be unique for each task (even for the same task type of two different workspace folders). Last but not least, the workspace folder is also passed to the constructor of the `CustomBuildTaskTerminal` 
- The `CustomBuildTaskTerminal` does not rely only on the `activeProject` anymore. Instead, if a `workspaceFolder` member is defined, then the `getProject` method (formerly `getActiveProject` method) return the project associated to the workspace folder.

At first sight solving #3232 should have been as easy as looping over each `CMake` project of the workspace to build the list of available tasks (in the `provideTasks` method). But as lot of pieces of the `CMakeTaskProvider` and `CustomBuildTaskTerminal` were relying on working on the active project, i had no other choice that modifying them so that they could work on any project the workspace folder is given.

 
## Other Notes/Information
Please tell me if it is possible to add tests for task selection. I will be pleased to implement them.

### DIFF
--- a/src/cmakeTaskProvider.ts
+++ b/src/cmakeTaskProvider.ts
@@ -187,6 +187,10 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
 
     public static async provideTask(commandType: CommandType, workspaceFolder: vscode.WorkspaceFolder, useCMakePresets?: boolean, targets?: string[], presetName?: string): Promise<CMakeTask> {
         const taskName: string = localizeCommandType(commandType);
+        const project = await extensionManager?.projectController.getProjectForFolder(workspaceFolder.uri.path);
+        if (!project) {
+            log.error(localize("project.not.found", 'Project not found.'));
+        }
         let buildTargets: string[] | undefined;
         let preset: string | undefined;
         if (commandType === CommandType.build || commandType === CommandType.cleanRebuild) {
@@ -195,7 +199,7 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
         if (presetName) {
             preset = presetName;
         } else if (useCMakePresets) {
-            preset = await getDefaultPresetName(commandType);
+            preset = getDefaultPresetNameForProject(project, commandType);
         }
 
         const definition: CMakeTaskDefinition = {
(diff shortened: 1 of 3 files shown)


gcampbell-msft:
Reviewing now, @hippo91 could you please make a changelog update as well? 

You'll need to create a new section: 

```
## 1.21

Bug Fixes: 
- <>
```

gcampbell-msft (changes requested):
These changes look reasonable to me, and it looks like the various possible edge cases are covered. 

I approve, but marking as request changes until changelog is updated.

hippo91 (commented):
> These changes look reasonable to me, and it looks like the various possible edge cases are covered. 
> 
> I approve, but marking as request changes until changelog is updated.   

Thanks @gcampbell-msft. I updated the changelog.

gcampbell-msft (approved):

gcampbell-msft:
@hippo91 Have you found that this is working as expected? I'm finding that this breaks the ability to do local tests, because the `getProjectForFolder` always returns null, which, while in development, always breaks us when we call `throw`. 

It looks to me like the `getProjectForFolder` method should be passing the `folder.uri.fsPath` in the parameter, not the `folder.uri.path`.

hippo91:
@gcampbell-msft sorry i did not notice this bad behavior.

gcampbell-msft:
> @gcampbell-msft sorry i did not notice this bad behavior.

No worries! I have merged a change and fixed the issue, thanks for your contributions!

PR #4237, opened 2025-01-14
Fix localization pipeline

gcampbell-msft:
After adding the language service information to get localized, we needed to adjust the locid to not have invalid characters that were coming from some of the language service data. 

This fixes it such that it removes it both before and after the xlf generation so that we don't have to modify the code, but we get around the issue in the localization step.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -19,7 +19,6 @@ const tsProject = ts.createProject('./tsconfig.json', { typescript });
 const filter = require('gulp-filter');
 const vinyl = require('vinyl');
 const jsonc = require('jsonc-parser');
-const lodash = require("lodash");
 
 // Patterns to find schema files
 const jsonSchemaFilesPatterns = [
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -99,7 +99,7 @@ const traverseJson = (jsonTree, descriptionCallback, prefixPath) => {
 
 // remove invalid characters that aren't allowed in xlf.
 const removeInvalidXLFCharacters = (str) => {
-    return str.replace(/(<)|(>)/g, "");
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -97,6 +97,11 @@ const traverseJson = (jsonTree, descriptionCallback, prefixPath) => {
     }
 };
 
+// remove invalid characters that aren't allowed in xlf.
+const removeInvalidXLFCharacters = (str) => {
+    return str.replace(/[^a-zA-Z_]/g, "");
```
sean-mcmanus, line 102: Why are these invalid characters?

gcampbell-msft: They are invalid according to the tool that does the localization, the OneLocBuild task. I didn't find any details on why

benmcmorran (dismissed):

gulpfile.js
```diff
@@ -97,6 +97,11 @@ const traverseJson = (jsonTree, descriptionCallback, prefixPath) => {
     }
 };
 
+// remove invalid characters that aren't allowed in xlf.
+const removeInvalidXLFCharacters = (str) => {
+    return str.replace(/(<)|(>)/g, "");
```
sean-mcmanus, line 102: Can you replace `(<)|(>)` with `<|>`? I don't understand why you put them in a capture group that doesn't seem to be used (unless I'm misunderstanding something).

sean-mcmanus (changes requested):

gulpfile.js
gcampbell-msft: Yes, I originally thought that I needed it but just tested and your suggestion is correct. Pushed the changes.

sean-mcmanus (approved):

PR #4241, opened 2025-01-15
chore(CTestDriver): skip project building according to `cmake.buildBeforeRun` config

Dabsunter:
### This changes auto-build behavior when launching/debugging tests from the test explorer view.

The following changes are proposed in the CTest driver:

- Skip project building if `cmake.buildBeforeRun` is set to `false` (`true` by default, meaning the default installation will keep the current behavior)

## The purpose of this change

Sometime during debugging sessions, we might want to run/debug tests multiple time without any code change, and those unnecessary re-build steps can be very time consuming in big projects.

Having the auto-build feature on by default is fine, but we should be able to disable it.

## Other Notes/Information

I relied on the already existing `cmake.buildBeforeRun` config, but we could introduce a dedicated `cmake.ctest.buildBeforeTest` to give the user more granularity in their choice...


Dabsunter:
@microsoft-github-policy-service agree company="AMD"

gcampbell-msft (approved):
I agree with your assessment that it could also be useful to have a more granular test setting for this, but I think that for now, this should work, we can always make it more granular later. 

Approved.

PR #4247, opened 2025-01-21

gcampbell-msft:


gcampbell-msft:
Localization updates have been made, looks like everything is localized. Merging.

PR #4249, opened 2025-01-22
Some Modules Intellisense Fixes

### LINKED ISSUE #4245: [Language Services] Modules completions don't behave as expected in all cases
### Brief Issue Summary

If I type FindPython at the beginning of a line and use the completion, it helps me out and creates the `find_package` call with the right arguments. 

However, if I already have typed the `find_package` and am now putting `find_package(FindPython` and then accept the completion, it still creates the `find_package` command, resulting in 

`find_package(find_package(Python REQUIRED))`

This is erroneous.

gcampbell-msft:
Fix some quick info strings. 

Fix some module completions. 

Fixes #4245

### DIFF
--- a/assets/modules.json
+++ b/assets/modules.json
@@ -169,7 +169,7 @@
     },
     "CMakeFindDependencyMacro": {
         "name": "CMakeFindDependencyMacro",
-        "description": "TODO: Document me."
+        "description": "The ``find_dependency()`` macro wraps a :command:`find_package` call for a package dependency."
     },
     "CMakeFindFrameworks": {
         "name": "CMakeFindFrameworks",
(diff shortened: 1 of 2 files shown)


assets/modules.json
```diff
@@ -837,12 +879,18 @@
...
+        "description": "Find upx"
+    },
     "FindSQLite3": {
         "name": "FindSQLite3",
         "description": "Find the SQLite libraries, v3"
     },
-    "FindSquish": { "name": "FindSquish", "description": "-- Typical Use" },
+    "FindSquish": {
+        "name": "FindSquish",
+        "description": "-- Typical Use"
```
paulmaybee, line 892: What does this mean?

paulmaybee (approved):

assets/modules.json
gcampbell-msft: It's parsed from the cmake docs, it's likely that we could use some improvement there, but for now it's the first paragraph from the cmake docs.

Issue #3345, opened 2023-09-16
Allow test debugging without a launch configuration


quyykk:
### Brief Issue Summary

When I press on the Debug play button on a test in the Testing tab, it says in the output window:

```
[ctest] No launch configurations found.
```

Why do I need to provide a launch configuration when I want to debug a test? I can run the test, and debug the main target without any launch configuration. 

The same should apply when debugging a test.

gcampbell-msft:
@quyykk Thanks for the suggestion! Marking it a feature request and putting it on the backlog to try and assess when/if we can fit this work in! Thanks again.

quyykk:
@gcampbell-msft Seems like this feature (debugging a test) is actually broken.

I added a launch configuration to workaround and tried to debug a test, but instead of the test being launched with a debugger attached, it started my main application with a debugger. I'm assuming it uses my launch configuration? This should probably be reprioritized I think :smile:

benmcmorran:
@xisui-MSFT can you comment on how this scenario is supposed to work?

xisui-MSFT:
This is by design. VS Code requires a launch config in order to debug anything. CMake tools doesn't depend on extensions that provide debug adapters and so it's not appropriate to have a default one that uses "cppdbg" or something alike.

As for usage, @gcampbell-msft I remember you wrote a doc on creating a launch config for debugging tests?

quyykk:
> so it's not appropriate to have a default one that uses "cppdbg" or something alike.

This extension already uses cppdbg for Quick Debugging, so I don't really see the issue.

xisui-MSFT:
I'm pretty sure Quick Debugging is added after this, so the original design is also fine. However, since it's decided that this extension now have a higher level of dependency on cpptools, it's OK to add some default launch options. @benmcmorran

franneck94:
I still get the message:
[ctest] No launch configurations found.

But i have added this config from the offical cmake-toolsdoc:

```json
{
    "name": "(ctest) Launch",
    "type": "cppdbg",
    "request": "launch",
    "cwd": "${cmake.testWorkingDirectory}",
    "program": "${command:cmake.launchTargetPath}",
    "args": [ "${cmake.testArgs}"],
}

```

weyllor:
+1

IngmarPaetzold:
I am confused as well. I could set up the cmake build with gtest and let tests run with TestMate.
So why does running a test work out of the box, but debugging requires a "configuration"?
[...]

Issue #3153, opened 2023-05-01
Simplify debug configuration for ctest debug launch


PrimeMinisterOfGreed:
### Brief Issue Summary

it's not clear how to write a configuration in launch.json that can debug a single test, since the default configuration launch all tests.

IlVirtuoso:
i'll explain that confuse issue over there. When using TestMate c++ this extension provide a debug configuration that work by default without writing a new one in launch.json. The new cmake tools ask for a launch configuration when try to use a debugger, but after days of searching i did not encounter any tutorial that provide a way to , for example, pass the name of the selected test to the args in launch.json, this mean that also if i selected a test, all tests will be launched. If there is a way to pass such argument, this information should be provided by the extension itself or by a default launch.json configuration that indicate the presence of this argument.

xisui-MSFT:
@IlVirtuoso Can you provide the launch.json you are using?

gcampbell-msft:
@IlVirtuoso Following up on this, does the above launch.json work for you? Thanks

IlVirtuoso:
@gcampbell-msft  Sorry i saw the message just now, my previous json was that
```json
 {
            "name": "LaunchTest",
            "type": "cppdbg",
            "request": "launch",
            "program": "${workspaceFolder}/build/DistributedCrack_test",
            "args": [
                "--gtest_filter= ${relativeFile} "
            ],
            "stopAtEntry": false,
            "cwd": "${fileDirname}",
            "environment": [],
            "externalConsole": false,
            "MIMode": "gdb",
            "setupCommands": [
                {
                    "description": "Abilita la riformattazione per gdb",
                    "text": "-enable-pretty-printing",
                    "ignoreFailures": true
                },
                {
                    "description": "Imposta Versione Disassembly su Intel",
                    "text": "-gdb-set disassembly-flavor intel",
                    "ignoreFailures": true
                }
            ]
        },
```

i tried to catch the name of the entry and filter it in relativeFile with a custom action, but it failed. Trying to launch the .json you provide (in debug obliviously)  i obtain the error launch property 'cwd' is missing.

gcampbell-msft:
If you provide the same `cwd` from your example you sent, what happens then?

xisui-MSFT:
[...]

PR #4273, opened 2025-02-07
feat: add `cmake.ctest.testSuiteDelimiter.debugLaunchTarget`

Nerixyz:
### This changes debugging of tests

- The setting `cmake.ctest.testSuiteDelimiter.debugLaunchTarget` is added, which specifies a launch configuration to be used when debugging tests.

## The purpose of this change

When repeatedly debugging tests (something fails → fix → recompile → test → something else fails → ...), one usually uses the same launch configuration. Often, multiple launch configurations are present in a project (e.g. a specific one for tests that has `"args": ["${cmake.testArgs}"]`). The popup to select a test can get distracting.

### DIFF
--- a/docs/debug-launch.md
+++ b/docs/debug-launch.md
@@ -227,6 +227,8 @@ A couple of examples:
 
 Depending on your configuration or your settings, there may need to be additional configuration options set.
 
+To use a specific launch configuration when debugging tests, set `cmake.ctest.testSuiteDelimiter.debugLaunchTarget` to the desired name of the configuration (e.g. `(ctest) Launch`).
+
 ## Run without debugging
 
 You can run a target without debugging it, by running the *CMake: Run Without Debugging* from VS Code's command palette, by selecting the play button in the status bar or the play button to the left of the Launch node, or by pressing the keyboard shortcut (**Shift+Ctrl+F5**).
--- a/package.nls.json
+++ b/package.nls.json
@@ -112,6 +112,7 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     }, 
+	"cmake-tools.configuration.cmake.ctest.testSuiteDelimiter.debugLaunchTarget": "Target name from launch.json to start when debugging a test with CTest. By default and in case of a non-existing target, this will show a picker with all available targets.",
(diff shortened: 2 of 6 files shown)


docs/debug-launch.md
```diff
@@ -227,6 +227,8 @@ A couple of examples:
 
 Depending on your configuration or your settings, there may need to be additional configuration options set.
 
+To use a specific launch configuration when debugging tests, set `cmake.ctest.testSuiteDelimiter.debugLaunchTarget` to the desired name of the configuration (e.g. `(ctest) Launch`).
```
gcampbell-msft, line 230:
```suggestion
To use a specific launch configuration when debugging tests, set `cmake.ctest.debugLaunchTarget` to the desired name of the configuration (e.g. `(ctest) Launch`).
```

package.nls.json
```diff
@@ -112,6 +112,7 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     }, 
+	"cmake-tools.configuration.cmake.ctest.testSuiteDelimiter.debugLaunchTarget": "Target name from launch.json to start when debugging a test with CTest. By default and in case of a non-existing target, this will show a picker with all available targets.",
```
gcampbell-msft, line 115:
```suggestion
	"cmake-tools.configuration.cmake.ctest.debugLaunchTarget": "Target name from launch.json to start when debugging a test with CTest. By default and in case of a non-existing target, this will show a picker with all available targets.",
```

gcampbell-msft (changes requested):
We're happy to take these changes, the modifications are simple enough and it would provide value. 

However, a couple of comments for somethings need to be changed: 
1. See the other comments on this PR. 
2. Add a Changelog entry. You will need to add a section for 1.21
```
## 1.21

Features:

- <>
```
3. Please add a screenshot and/or gif to the description of the PR so that it's more obvious what this will look like in the settings and in practice.

gcampbell-msft (dismissed):

gcampbell-msft:
Just a changelog merge conflict. Merging.

zwassall-gentuity:
This seems to address #3153 and #3345 (duplicate of #3153)?

Amy-Li03:
Verified this on cmake tools v1.21.7(pre-release), we can see the cmake.ctest.debugLaunchTarget setting, and after use it, specifies a launch configuration to be used when debugging tests.
Fixed

PR #4278, opened 2025-02-11
update notices

gcampbell-msft:
see title.


gcampbell-msft:
No build needed, just updating cg.

PR #4299, opened 2025-02-25
add notification for uninstalling twxs.cmake

gcampbell-msft:
Now that we've removed our dependency on the twxs.cmake extension, we should suggest for users to uninstall it if they are no longer interested in using it, since removing the dependency doesn't automatically uninstall the extension. 

The notification looks like this: 

image

And if you click "uninstall twxs.cmake", it will direct you to the extensions pane where you can uninstall the extension yourself. 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@ Bug Fixes:
 
 - Fix regression where we weren't properly expanding copyCompileCommands path. [#4294](https://github.com/microsoft/vscode-cmake-tools/issues/4294)
 
-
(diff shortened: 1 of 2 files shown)


paulmaybee (approved):

gcampbell-msft:
This has already been approved, a commit was added to merge the changelog, this can go in

Amy-Li03:
Verified this on cmake tools v1.21.7(pre-release), we can see the notification for uninstalling twxs.cmake and to uninstall it successfully.
image

Issue #4296, opened 2025-02-24
It's been years I haven't had issues with PC freezing.


pps83:
### Environment

why is it building c++ doesn't use lowest cpu prio? I've been using Visual Studio, and building is a no issue. I've tried to use vscode with c++ and my pc is dying, mouse freezing etc.

### Bug Summary and Steps to Reproduce

Bug Summary:

Steps to reproduce:

Expected behavior:

### Configuration and Logs

```shell
no logs
```

sean-mcmanus:
@gcampbell-msft I'm not sure if CMake Tools can do anything about this or if CMake and/or Ninja are the cause and/or the OS.

gcampbell-msft:
@sean-mcmanus and @pps83 It would be great if you could provide the repro that you hit this issue with, so that we could not only attempt to repro, but also to understand if there are specifics about your project that might be handled differently in VS Code that we could investigate to see if there is a performance improvement we could make. 

As @sean-mcmanus pointed out, depending on the issue, it could be simply due to CMake or Ninja, but it could also be due to additional things that the CMake Tools extension attempts to do. 

Looking forward to getting more info. Thanks!

sean-mcmanus:
@gcampbell-msft I repro it when I build our cpptools/cpptools-srv binaries. It spawns a lot of  (too many?) cl.exe processes which can make the Windows UI and other app UI's freeze and be unusable. If those processes could be spawned with a lower "priority" that could fix the issue.

gcampbell-msft:
@sean-mcmanus Does it reproduce on the command-line?

sean-mcmanus:
@gcampbell-msft It appears that the issue is caused by CMake Tools using `--parellel 10` with a 4 core machine and `--parellel 14` with a 6 core machine. Using `--parellel <coreCount x 2>` seems to fix it.

i.e. logging shows `/usr/bin/cmake --build <buildPath> --parallel 14 --target install --`

But I'm not sure where the argument to parallel is coming from.

sean-mcmanus:
@gcampbell-msft It appears to be a regression from https://github.com/microsoft/vscode-cmake-tools/pull/2320 .

```ts
export function defaultNumJobs (): number {
    return os.cpus().length + 2;
}
```

sean-mcmanus:
@gcampbell-msft I have a PR with a fix at https://github.com/microsoft/vscode-cmake-tools/pull/4300 .

bobbrow:
[...]

PR #4300, opened 2025-02-25
Lower defaultNumJobs to prevent UI freezing and stop using --parallel when CMakePresets is used and parallelJobs is not set

sean-mcmanus:
Fixes (or at least improves) [4296](https://github.com/microsoft/vscode-cmake-tools/issues/4296)

Seems like a regression from https://github.com/microsoft/vscode-cmake-tools/pull/2817 -- it started using defaultNumJobs if parallelJobs is not set.

I see recommendations to use logical core count * 2 - 1, but I would avoid doing that unless users continue to hit UI freezing.

There may be an issue still with --parallel being used by ctest, but I would say leave that to fix later if anyone complains about it.

### DIFF
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -389,7 +389,7 @@ export class CMakeProject {
             buildPreset,
             lightNormalizePath(this.folderPath || '.'),
             this.sourceDir,
-            this.workspaceContext.config.parallelJobs,
+            this.workspaceContext.config.isDefaultValue("parallelJobs") ? undefined : this.workspaceContext.config.parallelJobs,
             this.getPreferredGeneratorName(),
             true,
             this.configurePreset?.name);
--- a/src/cmakeProject.ts
+++ b/src/cmakeProject.ts
@@ -389,7 +389,7 @@ export class CMakeProject {
             buildPreset,
             lightNormalizePath(this.folderPath || '.'),
             this.sourceDir,
-            this.workspaceContext.config.isDefaultValue("parallelJobs") ? undefined : this.workspaceContext.config.parallelJobs,
+            this.workspaceContext.config.numJobs,
             this.getPreferredGeneratorName(),
             true,
             this.configurePreset?.name);
--- a/src/config.ts
+++ b/src/config.ts
@@ -485,7 +485,7 @@ export class ConfigurationReader implements vscode.Disposable {
     }
 
     get numJobs(): number | undefined {
-        if (this.parallelJobs === undefined) {
+        if (this.isDefaultValue("parallelJobs")) {
             return undefined;
         } else if (this.parallelJobs === 0) {
             return defaultNumJobs();
--- a/src/config.ts
+++ b/src/config.ts
@@ -171,7 +171,7 @@ export interface ExtensionConfigurationSettings {
     configureArgs: string[];
     buildArgs: string[];
     buildToolArgs: string[];
-    parallelJobs: number | undefined;
+    parallelJobs: number;
     ctestPath: string;
     ctest: { parallelJobs: number; allowParallelJobs: boolean; testExplorerIntegrationEnabled: boolean; testSuiteDelimiter: string };
     parseBuildDiagnostics: boolean;
@@ -372,7 +372,7 @@ export class ConfigurationReader implements vscode.Disposable {
     get buildToolArgs(): string[] {
         return this.configData.buildToolArgs;
     }
-    get parallelJobs(): number | undefined {
+    get parallelJobs(): number {
         return this.configData.parallelJobs;
     }
     get ctestParallelJobs(): number | null {
--- a/src/config.ts
+++ b/src/config.ts
@@ -16,7 +16,7 @@ import { Environment } from '@cmt/environmentVariables';
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
 export function defaultNumJobs (): number {
-    return os.cpus().length + 2;
+    return os.cpus().length;
 }
 
 const log = logging.createLogger('config');
--- a/src/presets/preset.ts
+++ b/src/presets/preset.ts
@@ -1585,7 +1585,7 @@ async function getBuildPresetInheritsImpl(folder: string, name: string, workspac
             name: defaultBuildPreset.name,
             displayName: defaultBuildPreset.displayName,
             description: defaultBuildPreset.description,
-            jobs: parallelJobs || defaultNumJobs(),
+            jobs: parallelJobs,
             configurePreset
         };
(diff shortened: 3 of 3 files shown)


bobbrow:
> Fixes [4296](https://github.com/microsoft/vscode-cmake-tools/issues/4296)
> 
> Seemingly a regression from #2320 where that code appears to have been added.

Some code moved around in #2320, but this number has been around longer than that.

gcampbell-msft:
@bobbrow @sean-mcmanus How do we know that this fixes the issue? As @bobbrow pointed out, looking at the history, that number has been around for a long time.

gcampbell-msft:
@sean-mcmanus Just confirming, since we released a new version recently, the issue doesn't reproduce on only the latest version right? Meaning it's not a regression only in the latest version?

sean-mcmanus:
@gcampbell-msft I don't believe it's a regression from any recent release. It appears to fix the UI freezing when I tried it, but with no other CPU intensive work going on. If a user has other CPU intensive work such as another build going on, then it would not fix that issue.

sean-mcmanus:
@bobbrow I don't see where the previous code was using cpus() + 2.

bobbrow:
> @bobbrow I don't see where the previous code was using cpus() + 2.

In the PR you linked it was being used in `numJobs`

image

sean-mcmanus:
@bobbrow No, I tested it out on 2 machines. It's using coreCount + 2. Why would we not fix the default? I don't think users should have to modify this setting.

sean-mcmanus:
@bobbrow 

The package.json sets the cmake.parallelJobs to 0 by default, so it defaults to logicalCores*2 + 2.

image

gcampbell-msft:
@bobbrow What do you suggest the fix to be? Or are you suggesting that since it's a low amount of users that this is affecting, we should encourage them to simply use the parallel jobs settings in order to mitigate this?

sean-mcmanus:
@gcampbell-msft I believe there 2 separate issues.

1. I was a regression with --parallel being used.
2. Another is a pre-existing issue in which + 2 was being used when the "default" is specified, i.e. parallelJobs set to 0.

I believe both should be fixed.

I don't think we should use + 2 unless we have a really good known reason. It sounds like the current reason is that it was like that for a long time.

gcampbell-msft:
@pps83 Before approving and merging this PR, I'd want to confirm that this fixes your issue as well. 

Could you download this vsix (download the zip, rename the extension, then manually install) and try it out? 
[cmake-tools.zip](https://github.com/user-attachments/files/18989247/cmake-tools.zip)

src/cmakeProject.ts
```diff
@@ -389,7 +389,7 @@ export class CMakeProject {
             buildPreset,
             lightNormalizePath(this.folderPath || '.'),
             this.sourceDir,
-            this.workspaceContext.config.parallelJobs,
+            this.workspaceContext.config.isDefaultValue("parallelJobs") ? undefined : this.workspaceContext.config.parallelJobs,
```
gcampbell-msft, line 392: @sean-mcmanus Based on the `isDefaultValue` implementation, this seems to only check if the value is `undefined`. This makes me wonder if there's also a bug in the `isDefaultValue` implementation, shouldn't it be checking if the value equals the default value? 

As it stands, it seems like this would only ever return the `this.workspaceContext.config.parallelJobs`, unless the user explicitly set the value to `undefined`. 

Am I understanding this correctly?

gcampbell-msft: Another question, does this mean that in the presets case, we no longer use any default number of `parellelJobs`? This means that we won't parallelize at all in the default case? 

@bobbrow @iontodirel Any thoughts on this?

sean-mcmanus: @gcampbell-msft It is intentional that isDefaultValue is only checking for it being undefined and not equal to the actual default value. The reason is that undefined means the user has not set the value, while being 0 means they've actually set that.

In the preset case, no --parallel is set, but if the user is using Ninja, then it defaults to using "num logical cores" of parallelism, which is the same as if 0 were set for parallelJobs (the normal default as set by this PR).

gcampbell-msft: Gotcha, I think this makes sense then. I'll approve. 

It will go into pre-release in the next day and then the 1.21 official release when it comes.

gcampbell-msft (approved):

src/cmakeProject.ts
bobbrow: It feels to me like `ConfigurationReader.numJobs()` should be updated to check `isDefaultValue` so that we don't make this mistake again. And then we use `numJobs` everywhere instead of `parallelJobs`.  `parallelJobs` should also stop returning `| undefined` because it won't ever be `undefined`.

```ts
    get numJobs(): number | undefined {
        if (this.isDefaultValue("parallelJobs")) {
            return undefined;
        } else if (this.parallelJobs === 0) {
            return defaultNumJobs();
        } else {
            return this.parallelJobs;
        }
    }
```

gcampbell-msft: @bobbrow I think that's a good suggestion, @sean-mcmanus could you see if this works and try it out as a cleaner solution

gcampbell-msft (changes requested):
See other comments

src/cmakeProject.ts
bobbrow: > @bobbrow I think that's a good suggestion, @sean-mcmanus could you see if this works and try it out as a cleaner solution

Yeah, because I think the current iteration can pass 0 for this argument which will be unexpected later on.

sean-mcmanus: @bobbrow @gcampbell-msft I didn't bother changing numJobs() because it's not used by the preset code and only the ctest code. But it sounds like you want me to update that?

sean-mcmanus: @bobbrow I don't know what you mean by parallelJobs not being undefined. I can still be.

bobbrow: I can push an iteration with the changes I'm looking for if you want.

sean-mcmanus: Okay, yeah, I see it's 0 instead of undefined. I can change that.

gcampbell-msft: I think that it is from my perspective, I assume after the small refactors it still is working as expected for your project?

I think tracing through the code, it looks like everything is consistent with the original change, so I'm happy with it. 

I'll let @bobbrow review as well

sean-mcmanus: @gcampbell-msft Yes, it works as expected for me. My last push just fixed an issue/bug when parallelJobs was explicitly set to 0 -- I had removed the `parallelJobs || defaultNumJobs` at a spot, so not it'll be correctly set the defaultNumJobs.

bobbrow (approved):

src/cmakeProject.ts
bobbrow: Looks good to me.

gcampbell-msft (approved):

PR #3631, opened 2024-03-03
Google Test Fix for "Go to test" and "Debug Test"

danniesim:
## This change addresses items #3280 and #3362

The following changes are proposed:

- Get the location of the test definition from CTest's "DEF_SOURCE_LINE" property
- Fallback to previous behaviour if DEF_SOURCE_LINE does not exist
- Route debug test request to correct CTestDriver based on folder and test URI

## Other Notes/Information

Currently, the CTest executable does not output custom properties (defined using CMake's `define_property` function).

Hence, I have made a  pull request there to address that: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9305

Note that it falls back to the old behaviour without the CMake pull request merged (which will be CMake 3.29 and above if so).

fixes #3362

Edit:
During testing with my projects, I discovered that the root cause of #3280 is how multi-root workspaces are handled. Fixing now...

Edit 2:
I have fixed #3280 by routing a debug test request to the right CTestDriver using the URI of the test and project.

PR #4321, opened 2025-03-04
fix(ctest): don't require backtrace in presence of DEF_SOURCE_LINE

rjaegers:

- Don't check for presence of backtrace information when using DEF_SOURCE_LINE

## The purpose of this change

When using the DEF_SOURCE_LINE property on a test it is now required to have backtrace information present, while that backtrace information is not used. This change removes that dependency to enable support for `gtest_discover_tests` via the XML/JSON output. This will require an additional PR on the GoogleTest module in CMake.

## Other Notes/Information

Based on the great work done in #3631.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # What's New?
 
+## 1.21
+
+Bug fixes: 
+
+- Fix displaying "Go to Test" in the Test Explorer when the DEF_SOURCE_LINE property is set, but there is no backtrace information present. [4321](https://github.com/microsoft/vscode-cmake-tools/pull/4321) [@rjaegers](https://github.com/rjaegers)
+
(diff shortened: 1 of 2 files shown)


rjaegers:
@microsoft-github-policy-service agree company="Philips"

gcampbell-msft (changes requested):
Changes look great, could you also add a CHANGELOG entry? 
The 1.21 section hasn't been added yet, if you could do it. 

```
## 1.21

Bug fixes: 

- <>
```

rjaegers:
@gcampbell-msft added a CHANGELOG entry, hope I am in the right ballpark with the wording.

I have a follow-up PR ready that adds name-demangling for C++ symbols.

gcampbell-msft (approved):

PR #4325, opened 2025-03-05
Fix lowercase include paths on Mac

tringenbach:
## The purpose of this change

The C++ extension treats paths as case sensitive on MacOS. (In reality, sometimes they are, but it depends, usually not).

When CMakeTools converts them to lowercase and sends the include paths over, they don't work, and all the `#include`'s that reference files in those paths get red squiggles. 

I do not understand why converting to lower case would be desirable, even if it didn't break. The OS is at least case preserving, so the"normalized or canonical form really is the form the mixed case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # What's New?
 
+## 1.20.x
+
+Improvements:
+
+- No longer convert paths on lowercase on MacOS to enable cpp tools to resolve them. [#4325](https://github.com/microsoft/vscode-cmake-tools/pull/4325) [@tringenbach](https://github.com/tringenbach)
+
 ## 1.20.53
 
 Improvements:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # What's New?
 
-## 1.20.x
+## 1.21
 
 Improvements:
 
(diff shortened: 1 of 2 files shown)


gcampbell-msft:
@sean-mcmanus @Colengms Any thoughts on this PR? 

It sounds like it's possible that the C++ extension should be treating mac paths as case-insensitive on the C++ side

sean-mcmanus:
@gcampbell-msft We treat mac paths as case insensitive. We only normalize (change the case) of paths on Windows (we change it to uppercase internally). The PR changes seems good to me. I don't think lowercasing paths on mac is a good idea.

tringenbach:
FWIW, Disk Utility on MacOS does allow me to create a case sensitive APFS filesystem.

image

Perhaps more practical, when I saw lowercase paths in the logs in VSCode, that jumped out at me as "wrong". So even if you changed the C++ extension to accept it, it would "look wrong" and get blamed incorrectly for other problems. At least that's my logic.

I'm happy with whatever works though :)

gcampbell-msft:
I'm happy to take this then @tringenbach could you make a changelog update and pull it up to date with `main`? Then, I'll approve

tringenbach:

tringenbach:
> I'm happy to take this then @tringenbach could you make a changelog update and pull it up to date with `main`? Then, I'll approve

Sure! How's that?

gcampbell-msft:
@tringenbach Slightly updated, and approved, I will merge once tests and builds pass. Thanks!

gcampbell-msft (dismissed):

gcampbell-msft:
Already passed tests and build, just a merge conflict resolution for the changelog, merging.

PR #4330, opened 2025-03-06
Remove -T from ctest default arguments

### LINKED ISSUE #4327: CMake 4.0.0 rc2 - "There was an error running ctest to determine available test executables"
### Brief Issue Summary

Currently getting an error like this when trying to run tests via the extension tool:
```
[proc] The command: /home/jpr42/dk/cmake-4.0.0-rc2-linux-x86_64/bin/ctest --show-only=json-v1 -j14 -C Debug -T test --output-on-failure exited with code: 8
[ctest] There was an error running ctest to determine available test executables
```

When I run it manually I get the following:
```
ctest --show-only=json-v1 -j14 -C Debug -T test --output-on-failure --test-dir build
CMake Error in .:
  ctest_test Failed to change working directory to "": No such file or
  directory
```

However running `ctest -C Debug --output-on-failure` works fine.

I'm almost positive the cause is `-T test` when I add that to my command it fails in the command line for me. But I haven't deep dived into this.

I'm on Linux Ubuntu-22.04

### Additional Information

I was using CMake version 3.22.1 which worked fine.

jpr42:
The ctest -T <action> option is part of the Dashboard Client Mode, and is not necessary for simply running tests locally.

closes #4327

### DIFF
--- a/package.json
+++ b/package.json
@@ -2505,8 +2505,6 @@
           },
           "description": "%cmake-tools.configuration.cmake.ctestDefaultArgs.description%",
           "default": [
-            "-T",
-            "test",
             "--output-on-failure"
           ],
           "scope": "machine-overridable"


package.json
```diff
@@ -2505,8 +2505,6 @@
           },
           "description": "%cmake-tools.configuration.cmake.ctestDefaultArgs.description%",
           "default": [
-            "-T",
-            "test",
```
jpr42, line 2509: Furthermore this also fixes the issue vscode warning about not finding DartConfiguration.tcl

Since not all projects `include(CTest)` for testing.

jpr42:
Testing this locally it fails because vscode-cmake-tools relies on Test.xml from what I can tell

image

I believe the proper course of action is for vscode-cmake-tools to use `--output-junit` instead.

PR #4348, opened 2025-03-11
fix getProjectForFolder

gcampbell-msft:
Fixes the error in `getProjectForFolder`, it should be using `folder.uri.fsPath`. It was breaking the ability to test locally. 

Originated in #4223 

@hippo91 FYI, please confirm that this still fixes what you intended to fix with your PR.


snehara99 (approved):

hippo91:
@gcampbell-msft I confirm that this still fixes what i intended to fix with my PR.

PR #4351, opened 2025-03-12
Don't require LCOV to be found in test project

TSonono:
@gcampbell-msft This I believe would resolve the issue of running the single-root-ui tests on windows locally. Sorry for missing this. I don't think any other change I did in #4094 should affect Windows, since those things should only be triggered in Linux presets, and the tests should only be executed on Linux.

Can you let me know if this resolves your issue? Otherwise, could you send me the error you encounter?


gcampbell-msft:
@TSonono From a quick test, I believe this does fix the issue. Thanks!

gcampbell-msft (approved):

PR #4352, opened 2025-03-12
[Engineering / Tests] be more explicit in what we check

gcampbell-msft:

- 
- 
-

### DIFF
--- a/test/end-to-end-tests/single-root-UI/test/configure-and-build-presets.test.ts
+++ b/test/end-to-end-tests/single-root-UI/test/configure-and-build-presets.test.ts
@@ -57,6 +57,29 @@ suite('Build using Presets', () => {
 
         expect(await vscode.commands.executeCommand('cmake.configure')).to.be.eq(0);
         expect(testEnv.projectFolder.buildDirectory.isCMakeCachePresent).to.eql(true, 'no expected cache present');
+
+        // Read and parse the codemodel JSON file
+        const replyDir = path.join(testEnv.projectFolder.buildDirectory.location, '.cmake', 'api', 'v1', 'reply');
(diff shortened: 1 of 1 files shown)


test/end-to-end-tests/single-root-UI/test/configure-and-build-presets.test.ts
```diff
@@ -57,6 +57,29 @@ suite('Build using Presets', () => {
 
         expect(await vscode.commands.executeCommand('cmake.configure')).to.be.eq(0);
         expect(testEnv.projectFolder.buildDirectory.isCMakeCachePresent).to.eql(true, 'no expected cache present');
+
+        // Read and parse the codemodel JSON file
+        const replyDir = path.join(testEnv.projectFolder.buildDirectory.location, '.cmake', 'api', 'v1', 'reply');
+        const files = await fs.readdir(replyDir);
+        const codemodelFile = files.find(file => file.startsWith('codemodel'));
```
snehara99, line 64: Is this guaranteed? I'm not familiar with this file.

snehara99: ah wait I assume so since this is within our tests. nevermind!

snehara99 (approved):

test/end-to-end-tests/single-root-UI/test/configure-and-build-presets.test.ts
gcampbell-msft: To answer your question though, yes, the file api always outputs to a specific location. Then, we know it's 33 targets based on the project setup in our tests. 
Thanks

PR #4356, opened 2025-03-21
Path fix for lcov coverage implementation

TSonono:
I was able to reproduce the issue @makas005 encountered when using the coverage implementation on Windows. With this fix, I was able to get proper coverage population within the source files themselves. Linux still works with this change as well.

The path returned by the lcov parser could contain a carriage return in the case where it's a windows path. I'm guessing the parser was not written with the expectation of encountering windows paths. We now trim the path to ensure that any leading or trailing whitespaces or line terminator characters are removed from the path before calling the VS Code coverage API.

This change addresses the discussion at https://github.com/microsoft/vscode-cmake-tools/discussions/4302

@gcampbell-msft

### DIFF
--- a/src/coverage.ts
+++ b/src/coverage.ts
@@ -2,6 +2,7 @@ import * as vscode from 'vscode';
 import { lcovParser } from "@friedemannsommer/lcov-parser";
 import * as nls from 'vscode-nls';
 import * as logging from '@cmt/logging';
+import { platformNormalizePath } from './util';
 
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize: nls.LocalizeFunc = nls.loadMessageBundle();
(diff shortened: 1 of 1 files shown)


src/coverage.ts
```diff
@@ -32,7 +32,7 @@ export async function handleCoverageInfoFiles(run: vscode.TestRun, coverageInfoF
         }
         const sections = await lcovParser({ from: contents });
         for (const section of sections) {
-            const coverage = new vscode.FileCoverage(vscode.Uri.file(section.path),
+            const coverage = new vscode.FileCoverage(vscode.Uri.file(section.path.trim()),
```
gcampbell-msft, line 35: Do we also need to normalize the paths based on the os? 

For example, with our normalize methods in the utils file?

gcampbell-msft (approved):
I'm fine with this change, but I have one question/suggestion

src/coverage.ts
TSonono: I'm honestly not sure. I'll let you decide if that makes the most sense. I just tried changing the line to `const coverage = new vscode.FileCoverage(vscode.Uri.file(platformNormalizePath(section.path.trim())),` (after also importing that function) and that still works. The normalize function seemingly does not remove the "trailing" carriage return, so the `trim()` is still needed.

gcampbell-msft: Sorry for the delay in response, I think it's likely safer to go ahead and normalize, so if you could update with the `platformNormalizePath`, that would be great and we can approve and merge this PR. 

Thanks for your patience.

TSonono: Done, and no worries at all.

gcampbell-msft (approved):

PR #4357, opened 2025-03-21
feat: Limit number of splits by the delimiter

sandercox:
### This changes visible behavior/documentation

- specify the maximum number of times the ctest.testSuiteDelimiter regex can be applied to split a string
- adds an option ctest.testSuiteDelimiterMaxOccurrence

## The purpose of this change

Prevent empty/unwanted hierarchy levels in Test Explorer Hierarchy

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -112,6 +112,9 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     }, 
+    "cmake-tools.configuration.cmake.ctest.testSuiteDelimiterMaxOccurrence.markdownDescription": {
+        "message": "Maxmimum number of times the delimiter may be used to split the name of the test. `0` means no limit.",
+    },
 	"cmake-tools.configuration.cmake.ctest.debugLaunchTarget.description": "Target name from launch.json to start when debugging a test with CTest. By default and in case of a non-existing target, this will show a picker with all available targets.",
     "cmake-tools.configuration.cmake.parseBuildDiagnostics.description": "Parse compiler output for warnings and errors.",
     "cmake-tools.configuration.cmake.enabledOutputParsers.description": {
--- a/src/config.ts
+++ b/src/config.ts
@@ -173,7 +173,7 @@ export interface ExtensionConfigurationSettings {
     buildToolArgs: string[];
     parallelJobs: number;
     ctestPath: string;
-    ctest: { parallelJobs: number; allowParallelJobs: boolean; testExplorerIntegrationEnabled: boolean; testSuiteDelimiter: string; debugLaunchTarget: string | null };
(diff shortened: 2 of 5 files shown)


src/config.ts
```diff
@@ -605,7 +608,7 @@ export class ConfigurationReader implements vscode.Disposable {
         parallelJobs: new vscode.EventEmitter<number>(),
         ctestPath: new vscode.EventEmitter<string>(),
         cpackPath: new vscode.EventEmitter<string>(),
-        ctest: new vscode.EventEmitter<{ parallelJobs: number; allowParallelJobs: boolean; testExplorerIntegrationEnabled: boolean; testSuiteDelimiter: string; debugLaunchTarget: string | null }>(),
+        ctest: new vscode.EventEmitter<{ parallelJobs: number; allowParallelJobs: boolean; testExplorerIntegrationEnabled: boolean; testSuiteDelimiter: string; testSuiteDelimiterMaxOccurrence: number, debugLaunchTarget: string | null }>(),
```
olegkhar, line 611: Should be a semicolon after `number` instead of comma.

src/ctest.ts
```diff
@@ -694,10 +694,30 @@ export class CTestDriver implements vscode.Disposable {
         let parentSuiteItem = testExplorerRoot;
         let testLabel = testName;
 
+        function limited_split(regexString: string, subject: string, maxCount: number): string[]
+        {
+            if(maxCount === 0)
```
olegkhar, line 699: Can you please add a space between `if` and `(`?  And also for the `while`, etc.  Thanks!

src/config.ts
sandercox: thanks for pointing it out

src/ctest.ts
sandercox: ah so used my tools format for me I guess it didn't on this file.

sandercox:
@olegkhar thanks for taking the time to review and point out these small corrections. Applied them as suggested.

package.nls.json
```diff
@@ -112,6 +112,9 @@
             "Markdown text between `` should not be translated or localized (they represent literal text) and the capitalization, spacing, and punctuation (including the ``) should not be altered."
         ]
     }, 
+    "cmake-tools.configuration.cmake.ctest.testSuiteDelimiterMaxOccurrence.markdownDescription": {
+        "message": "Maxmimum number of times the delimiter may be used to split the name of the test. `0` means no limit.",
```
olegkhar, line 116: Just noticed the misspelling of "Maximum"

olegkhar: And the trailing comma is probably causing the build failure.

sandercox:
addressed the earlier linting errors. I think it's ready in it's current state.

olegkhar (approved):

olegkhar:
> addressed the earlier linting errors. I think it's ready in it's current state.

Thanks for your PR!

Issue #4404, opened 2025-04-15
[Bug] `UTILITY` targets with source files can interfere with cpptools `includePath` generation when Build Target is `all`


malsyned:
### Brief Issue Summary

Some `UTILITY` targets, such as the ones created by `Doxygen.cmake`, have a list of sources attached. If one of these targets happens to be placed at index `[0]` in the list of configurations passed to `CppConfigurationProvider.updateFileGroup()` by `updateConfigurationData()`, then `provideConfigurations()` for those sources will provide incomplete configurations for those sources.

This is because such a `UTILITY` target breaks the assumption articulated in this comment in `getConfiguration()`:

```typescript
return configurations?.values().next().value; // Any value is fine if the target doesn't match
```

### CMake Tools Diagnostics

```shell
// With "all" selected as Build Target:
{
  "os": "linux",
  "vscodeVersion": "1.99.2",
  "cmtVersion": "1.13.0",
  "configurations": [
    {
      "folder": "/home/malsyned/work/preferred/fsg/preferred-fsg",
      "cmakeVersion": "4.0.0",
      "configured": true,
      "generator": "Unix Makefiles",
      "usesPresets": true,
      "compilers": {
        "C": "/home/malsyned/packages/LLVM-ET-Arm-19.1.5-Linux-x86_64/bin/clang",
        "CXX": "/home/malsyned/packages/LLVM-ET-Arm-19.1.5-Linux-x86_64/bin/clang++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "Debug",
    "buildTypesSeen": [
      "Debug"
    ],
    "requests": [
      "file:///home/malsyned/work/preferred/fsg/preferred-fsg/src/main.c"
    ],
    "responses": [
      {
        "uri": "file:///home/malsyned/work/preferred/fsg/preferred-fsg/src/main.c",
        "configuration": {
          "includePath": [],
          "defines": [],
          "compilerPath": "/home/malsyned/packages/LLVM-ET-Arm-19.1.5-Linux-x86_64/bin/clang++",
          "compilerArgs": [],
          "compilerFragments": [
            "--target=arm-none-eabi"
          ]
        }
      }
    ],
    "partialMatches": [],
    "targetCount": 44,
    "executablesCount": 2,
    "librariesCount": 12,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
[...]

PR #4405, opened 2025-04-15
Stop UTILITY targets from affecting IntelliSense

malsyned:

Avoid passing `UTILITY` targets to `updateConfigurationData()` so that they don't interfere with the assumptions made by `getConfiguration()`.

## The purpose of this change

Prevent `UTILITY` targets with sources, such as the ones created by `Doxygen.cmake`, from causing incorrect IntelliSense generation.


malsyned:

malsyned:

malsyned:
The strategy of this PR relies on the assumption that `UTILITY` targets (i.e. targets created by `add_custom_target()`)  aren't going to contribute anything of value to C/C++ compilation information. I think that's a pretty safe assumption, and certainly safer than the assumption made in the current source.

If filtering them all out indiscriminately seems like too extreme or risky, I can imagine a few other strategies for addressing #4404, though,

In increasing order of complexity:
1. Filter out configurations that don't contribute any non-`undefined`, non-empty properties before arbitrarily returning the first remaining one.
2. Make a new composite configuration by iterating over all target configurations and using each non-`undefined`, non-empty value encountered for each property
3. Like 2, but combine array properties like `includePath` instead of picking the first one

I have implementations for (1) and (3) already written up, and (2) would be pretty simple to make. If you'd prefer a PR for any of those strategies, let me know and I'll push it shortly after.

malsyned:
@gcampbell-msft I'm having a really hard time seeing why this change would affect the macOS and Windows integration tests but not the Linux ones. It's hard to see through all of the spurious "Error" lines about deprecation warnings and "Hello" output, but I think it's saying that the failing test is "Good Run test with coverage", is that right?

Can you give me some guidance on how to even start troubleshooting this? I don't have access to a Windows or macOS PC, and this change in the cpptools integration shouldn't be going anywhere near anything test explorer related, I don't think.

gcampbell-msft:
@malsyned Unfortunately these are known flaky issues with our tests. The Mac has a flakiness in one test and the Windows occasionally hits a timeout.

malsyned:
Still no love from the Mac CI test, huh?

@gcampbell-msft does this PR look good otherwise to you?

gcampbell-msft:
@malsyned What about an option where we simply sort the configurations we provide, and essentially de-prioritize the Utility targets? It makes me nervous to fully remove a target from intellisense inclusion, but if we simply de-prioritize it such that it's more of a last resort, this feels safer. 

This is very similar to the Option 1 you mentioned, and I do think I prefer something like that, so that it can still be used if there are no other configurations available.

malsyned:
> @malsyned What about an option where we simply sort the configurations we provide, and essentially de-prioritize the Utility targets? It makes me nervous to fully remove a target from intellisense inclusion, but if we simply de-prioritize it such that it's more of a last resort, this feels safer.
> 
> This is very similar to the Option 1 you mentioned, and I do think I prefer something like that, so that it can still be used if there are no other configurations available.

OK yeah, no problem. I can think of a couple ways to implement this, do you have a preference?

## 1. Explicitly look at UTILITY targets and move them to the back of the line

This has to happen at `updateConfigurationData()` time because target type isn't easily available at `provideConfigurations()` time.

```diff
diff --git a/src/cpptools.ts b/src/cpptools.ts
index 50a3a50d..34fefc19 100644
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -595,7 +595,11 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
             // Update only the active build type variant.
             if (config.name === opts.activeBuildTypeVariant || (!opts.activeBuildTypeVariant && config.name === "")) {
                 for (const project of config.projects) {
-                    for (const target of project.targets) {
+                    // Move utility targets to the end so executable and library
+                    // targets which contribute more to IntelliSense
+                    // configuration take precedence.
+                    const targets = project.targets.sort((a, b) => Number(a.type === 'UTILITY') - Number(b.type === 'UTILITY'));
+                    for (const target of targets) {
                         // Now some shenanigans since header files don't have config data:
                         // 1. Accumulate some "defaults" based on the set of all options for each file group
                         // 2. Pass these "defaults" down when rebuilding the config data
```

## 2. Only use non-contributing configurations as a last resort

This can happen at `provideConfigurations()` time. It's more like an implicit check for UTILITY targets. It might be marginally more future-proof or self documenting?

```diff
diff --git a/src/cpptools.ts b/src/cpptools.ts
index 50a3a50d..a6cfe68b 100644
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -315,6 +315,27 @@ export function getIntelliSenseMode(cptVersion: cpptools.Version, compilerPath:
     }
 }
 
+/**
+ * Try to find a target configuration with some populated properties.
+ *
+ * All targets get defaults for `compilerPath`, `compilerArgs`, and
+ * `compilerFragments`, even `UTILITY` targets defined with
+ * `add_custom_command()` that provide no other useful configuration, so if
+ * possible, return one with more than just those populated.
+ */
+function fallbackConfiguration(configurations: Map<string, cpptools.SourceFileConfigurationItem> | undefined) {
+    if (!configurations) {
+        return undefined;
+    }
+    for (const item of configurations.values()) {
+        const { configuration: { includePath, defines, intelliSenseMode, standard} } = item;
+        if (includePath.length || defines.length || intelliSenseMode || standard) {
+            return item;
+        }
+    }
+    return configurations.values().next().value;
+}
+
 /**
  * The actual class that provides information to the cpptools extension. See
  * the `CustomConfigurationProvider` interface for information on how this class
@@ -350,7 +371,7 @@ export class CppConfigurationProvider implements cpptools.CustomConfigurationPro
         if (this.activeTarget && configurations?.has(this.activeTarget)) {
             return configurations!.get(this.activeTarget);
         } else {
-            return configurations?.values().next().value; // Any value is fine if the target doesn't match
+            return fallbackConfiguration(configurations);
         }
     }
 
```

gcampbell-msft:
@malsyned I think I like option 2 as it's more generic, and as you said, it might be more future-proof, thanks!

malsyned:
@gcampbell-msft just rebased this against the latest post-1.21.36 `main`. It uses the implementation you said you preferred, and from my end it's ready for merging. Let me know if there's anything else you want me to do on it.

malsyned:
Sure. Want me to also change `## 1.21` to `## 1.21.36`?

malsyned:
I just noticed that https://github.com/microsoft/vscode-cmake-tools/pull/4491 got merged with its changelog update still in the 1.21 section. Want me to make a little PR to move that, once the 1.22 section exists?

malsyned:
@gcampbell-msft I just pushed an update that does what you asked, adding a 1.22 section and moving this change into it. LMK if you want to make these other two changes I mentioned above:

* change `## 1.21` to `## 1.21.36`
* make a little PR to move the line about #4491 into the 1.22 section

gcampbell-msft (approved):

malsyned:
@gcampbell-msft I can't tell if the macOS CI failure is real or spurious, can you?

PR #4412, opened 2025-04-17
fix the error in the `Cache Editor` the `key` translation in simplified Chinese

rise-worlds:
#4417

fix the error in the `key` translation in simplified Chinese: `密钥` change to `缓存键`.

### This changes 

- `key` translation: `密钥` change to `缓存键`

## The purpose of this change

fix translation error in simplified Chinese.


rise-worlds:

gcampbell-msft:
@Amy-Li03 Could you please create a celoc bug?

Issue #4418, opened 2025-04-18
[Feature] CTest: Extract `TestMessage` preoprties from test failure output using regular expressions


malsyned:
## Background

CTest can be used to include tests from C++ unit test frameworks like [CppUTest](https://cpputest.github.io/) and [GoogleTest](https://github.com/google/googletest) into a project's suite of automated tests. CMake Tools already has support for associating tests with source locations using the `DEF_SOURCE_LINE` property. This property is set by GoogleTest, and [support for it is approved for merging into CppUTest](https://github.com/cpputest/cpputest/pull/1848) as well.

Beyond that, though, CMake Tools doesn't attempt to examine failure output to determine the source location for test failures. It just always reports the first line of the test as the location of any failures. It also doesn't do anything to take advantage of Test Explorer's`actualOutput`/`expectedOutput` diffing capabilities.

## Proposed Feature

I'm proposing adding the capability to examine test failure output for source file, source line, message, actual output, and expected output. The format of the output differs between test frameworks, and a single project's test suite could conceivably have heterogeneous output formats for different kinds of tests (smoke, integration, unit, etc.)  so it wouldn't make sense to hard-code any particular output parsing. Instead, Regular Expressions with capture groups could be used in a way similar to how Tasks use them in Problem Matcher patterns.

### Additional Information

I have a PR that implements this feature, I'll be posting it soon. I've been trying it out with CppUTest on my projects and it's working great, adding detail and specificity to the results.

The configuration is based on the one used for Problem Matcher patterns, centered around regular expressions and match group indexes. It's not identical to Problem Matchers, though, because of the different needs of this scenario; chiefly, the need to capture expected and actual output.

In my current implementation, the setting `cmake.ctest.failurePatterns` is a list of patterns with the fields `regexp`, `file`, `line`, `message`, `actual`, and `expected`. My settings for CppUTest look like this, for example:
```json
"cmake.ctest.failurePatterns": [
    {
        "regexp": "(.*?):(\\d+):.*\\n((?:\\texpected *<(.*)>\\n\\tbut was *<(.*)>\\n)?(?:\\t.*\\n)*)",
        "expected": 4,
        "actual": 5
      }
]
```

[...]

PR #4420, opened 2025-04-18
Search CTest output for failure locations (#4418)

malsyned:

Adds configuration setting `cmake.ctest.failurePatterns`, a list of regular expressions and capture group indices for matching file, line, message, expected output, and actual output from failing tests.

## The purpose of this change

Populate `TestMessage` objects with more specific information about test failures than is available through the `DEF_SOURCE_LINE` property.

## Other Notes/Information

I based the design of this feature around the design of Task Problem Matchers. It lacks some of the advanced features of Problem Matchers, like named patterns, pattern inheritance, multi-regexp patterns, and loops. I'm happy to add those to the implementation if needed, but I thought it better to start with the simplest thing that could work and grow it in response to feedback.

## Screenshots

image

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -2255,6 +2255,51 @@
           "markdownDescription": "%cmake-tools.configuration.cmake.ctest.testSuiteDelimiter.markdownDescription%",
           "scope": "machine-overridable"
         },
+        "cmake.ctest.failurePatterns": {
+          "type": ["string", "array"],
+          "items": {
+            "type": ["string", "object"],
+            "properties": {
+              "regexp": {
+                "type": "string",
+                "description": "%cmake-tools.configuration.cmake.ctest.failurePatterns.regexp%"
+              },
+              "file": {
+                "type": "integer",
+                "description": "%cmake-tools.configuration.cmake.ctest.failurePatterns.file%",
+                "default": 1
+              },
+              "line": {
+                "type": "integer",
+                "description": "%cmake-tools.configuration.cmake.ctest.failurePatterns.line%",
+                "default": 2
+              },
+              "message": {
+                "type": "integer",
+                "description": "%cmake-tools.configuration.cmake.ctest.failurePatterns.message%",
+                "default": 3
+              },
+              "actual": {
+                "type": ["integer" ],
+                "description": "%cmake-tools.configuration.cmake.ctest.failurePatterns.actual%"
+              },
+              "expected": {
+                "type": "integer",
+                "description": "%cmake-tools.configuration.cmake.ctest.failurePatterns.expected%"
+              }
+            }
+          },
+          "default": [
+            {
+              "regexp": "(.*?):(\\d+): *(?:error: *)(.*)"
+            },
+            {
+                "regexp": "(.*?)\\((\\d+)\\): *(?:error: *)(.*)"
+            }
+          ],
+          "markdownDescription": "%cmake-tools.configuration.cmake.ctest.failurePatterns.markdownDescription%",
+          "scope": "machine-overridable"
+        },
         "cmake.ctest.debugLaunchTarget": {
           "type": "string",
           "default": null,
--- a/src/ctest.ts
+++ b/src/ctest.ts
@@ -132,6 +132,68 @@ interface ProjectCoverageConfig {
     coverageInfoFiles: string[];
 }
 
+interface FailurePattern {
+    regexp: string;
+    file?: number;
+    line?: number;
+    message?: number;
+    actual?: number;
+    expected?: number;
+}
+
+type FailurePatternsConfig = string | string[] | FailurePattern[];
+
+export function searchOutputForFailures(patterns: FailurePatternsConfig, output: string): vscode.TestMessage[] {
+    output = stripCarriageReturns(output);
+    const messages = [];
+    patterns = Array.isArray(patterns) ? patterns : [patterns];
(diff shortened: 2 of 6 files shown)


gcampbell-msft (commented):
Could we use a more strict type than `Object` when defining the settings/emitters?

malsyned:
I have been doing some work to make sure that this feature is powerful enough to match GoogleTest and GoogleMatch failure patterns. I think it is. I went in thinking it might be challenging to match the fairly complicated set of expected vs. actual patterns, but actually, GoogleTest and GoogleMock provide so much additional info in their messages that would be hidden if `TestMessage.actual` and `.expected` were filled in, that I don't think I want to match them anyway.

The regexp for GoogleTest/GoogleMock is `"(.+?):((?:\\d|#)+): *(?:[Ff]ailure|[Ee]rror)\n+((?:[^\n].*\n)*)\n"`, and it is able to match multiple failures in the same test if they are reported:
image

So, that's increased my confidence in this with its current feature set.

src/ctest.ts
```diff
@@ -132,6 +133,61 @@ interface ProjectCoverageConfig {
...
+function matchToTestMessage(pat: FailurePattern, match: RegExpMatchArray): vscode.TestMessage {
+    const file = match[pat.file as number];
+    const line = pat.line ? parseLineMatch(match[pat.line]) : 0;
+    const message = pat.message && match[pat.message]?.trim() || 'Test Failed';
+    const actual = pat.actual ? match[pat.actual] : undefined;
+    const expected = pat.expected ? match[pat.expected] : undefined;
+
+    const testMessage = new vscode.TestMessage(normalizeCRLF(message));
+    testMessage.location = new vscode.Location(
+        vscode.Uri.file(file), new vscode.Position(line, 0)
```
malsyned, line 168: @gcampbell-msft I have a small hesitation about this line, I would love for you to weigh in on it. It's assuming that the output file name matched by the regex is an absolute path suitable for passing to `Uri.file()`. It always has been in my testing, and I think due to the fact that CMake tends to generate build systems that operate exclusively on absolute paths, I'm guessing it's likely that it always will.

However, if you think otherwise, and you think there's a reasonable directory to resolve relative paths against, I would be happy to add something to this PR.

malsyned:
@paulmaybee @gcampbell-msft Is there anything else you all would like on this PR before you'd consider merging it?

src/ctest.ts
gcampbell-msft: I don't have any concerns, as the behavior when there are no patterns defined should be the same, correct? 

With it being a new feature, I'm okay with the assmption that Uri.file() will work, again, as long as it still works the same if this feature isn't being used.

src/ctest.ts
```diff
@@ -132,6 +133,61 @@ interface ProjectCoverageConfig {
...
+    return testMessage;
+}
+
+function normalizeLF(s: string) {
+    return s.replace(/\r\n?/g, '\n');
+}
+
+function normalizeCRLF(s: string) {
+    return s = s.replace(/\r?\n/g, '\r\n');
+}
```
gcampbell-msft: These feel like they should be in the `util.ts` class

gcampbell-msft (approved):
I think this is a cool feature! It seems to present very little risk for regressions of other features and is purely additive, so I think this looks good, and it seems you have done good testing. 

I left just a comment or two, but overall I think this looks great!

package.json
```diff
@@ -2271,6 +2271,68 @@
...
+                    "type": "string"
+                  }
+                ]
+              }
+            },
+            {
+              "type": "string"
+            }
+          ],
+          "default": [
```
malsyned, line 2419: @gcampbell-msft in a few comments on this PR you've said some variation of
> There is also minimal risk for regression, as by default this will not perform any additional computation because the default is without this

But this PR does add a couple of default regular expressions that will be tried against test output, so it would potentially change behavior in the absence of intentional user configuration. Should I empty out the default so that this feature is disabled by default?

src/ctest.ts
malsyned: Correct, with no patterns defined, we'll never get to this line of code.

My concern is that the string being passed to `Uri.file()` is extracted from ctest output that probably originates in some unit test framework code that uses `__FILE__`. That's how CppUTest and GoogleTest do it, for example. GCC-like compilers generally don't absolute-ize relative paths passed on the command line, so in general it's possible for the output of these test executables to contain relative paths, which aren't handled well by `Uri.file()`.

However, this isn't the general case, this is in the context of CMake, which generally ensures compilers are passed absolute paths, and uses absolute paths in its output messages. That's why I *think* we can get away with passing the file path directly to `Uri.file()` without worrying about trying to make it an absolute path first, but also why I wanted a second opinion.

malsyned:
> I think this is a cool feature! It seems to present very little risk for regressions of other features and is purely additive, so I think this looks good, and it seems you have done good testing.
> 
> I left just a comment or two, but overall I think this looks great!

I'm glad you like it! Thanks for the review. I'll work on your comments today.

malsyned:
@gcampbell-msft just rebased this against the latest post-1.21.36 `main`. The only outstanding question I think is whether you want me to make `cmake.ctest.failurePatterns` an empty array instead of being pre-populated with some generic regexps. Other than that I think it's ready for merging if you do.

gcampbell-msft (approved):

Issue #1976, opened 2021-07-06
Feature Req: Access to CMakePresets' cacheVariables and and environment variables from tasks.json and launch.json


hwmaier:
For those working with multiple presets it would be beneficial if the `environment` and `cacheVariables` specified in the current configuration preset and build preset (in CMakePresets.json) can be queried by command substitutions from `tasks.json` and `launch.json`.

Suggested syntax:
```
${command:cmake.configureSettings.DEVICE}   // to access DEVICE cacheVariable
${command:cmake.configureEnvironment.JLINK_ID}   // to access JLINK_ID preset environment
${command:cmake.buildEnvironment.BUILD_NO}   // to access BUILD_NO build environment variable
```
This would also align with the principle "CMakePresets.json should be the source of truth for all settings related to configure, build, and test." mentioned in https://github.com/microsoft/vscode-cmake-tools/blob/develop/docs/cmake-presets.md.

andreeis:
@hwmaier, thank you for this feature request. We were just thinking of doing some improvements in this area last week and we were about to open a similar report ourselves.

gruenedd:
That is actually a very important missing feature.
Especially since this **works for standard visual studio** (aka `.vs/launch.vs.json`) by using `${cmake.YOUR_CACHE_VARIABLE}`.

For now i ended up with the following workaround in `CMakeLists.txt` which might be useful for others:

```cmake
set (VSCODE_SETTINGS_PATH "${CMAKE_CURRENT_SOURCE_DIR}/.vscode/settings.json")
if (EXISTS "${VSCODE_SETTINGS_PATH}")
    file (READ "${VSCODE_SETTINGS_PATH}" VSCODE_SETTINGS_JSON)
else()
    set (VSCODE_SETTINGS_JSON "{}")
endif()
string (JSON VSCODE_SETTINGS_JSON_NEW SET ${VSCODE_SETTINGS_JSON} "your_var" "\"${YOUR_CMAKE_VARIABLE}\"")
string (JSON VSCODE_SETTINGS_JSON_NEW SET ${VSCODE_SETTINGS_JSON_NEW} "your_var2" "whatever_value"")
# and so on ...
if (NOT "${VSCODE_SETTINGS_JSON_NEW}" STREQUAL "${VSCODE_SETTINGS_JSON}")
  file (WRITE "${VSCODE_SETTINGS_PATH}" "${VSCODE_SETTINGS_JSON_NEW}")
endif()
```

[...]

PR #4422, opened 2025-04-21
Add `cmake.cacheVariable` command

malsyned:
### This changes visible behavior & documentation

Add a new command, `cmake.cacheVariable`, which can be used in `launch.json` and `tasks.json` as a command input, allowing substitution of cache variables defined in `CMakeLists.txt` or in the `configurePresets.cacheVariables` property of `CMakePresets.json`.

The `inputs` section can be used to pass a CMake cache variable `name` to `cmake.cacheVariable`, and that input can in turn be used for variable substitution in property values in the `configuration` section.

Example:

```jsonc
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Debug with J-Link",
            "request": "launch",
            "type": "cortex-debug",
            "executable": "${command:cmake.launchTargetPath}",
            "cwd": "${workspaceFolder}",
            "servertype": "jlink",
            "interface": "jtag",
            "device": "${input:DEVICE}"
       }
    ],
    "inputs": [
        {
            "id": "DEVICE",
            "type": "command",
            "command": "cmake.cacheVariable",
            "args": {
                "name": "DEVICE",
                "default": "rm44l520"
            }
        }
    ]
}
```

## The purpose of this change

To allow `CMakeLists.txt` to serve as a single point of truth for configuration information specific to the user's project. In my case, for example, I can keep the part number of my project's microcontroller in a CMake cache variable, and use it to populate `cortex-debug` properties like `"device"` and `"svdFile"`. The description in #1976 looks to me like it has the same use case in mind, but I bet there are plenty of other situations where this would come in handy.

## Other Notes/Information

The referenced issue tickets mention providing access not just to cache variables, but also to environment variables. This PR doesn't address the environment variable part of those feature requests.


malsyned:
I would love to get a maintainer's take on bcbc15e to confirm that 1) I should be doing that and 2) I'm doing it the right way.

src/cmakeProject.ts
```diff
@@ -2874,6 +2874,22 @@ export class CMakeProject {
         return env;
     }
 
+    async cacheVariable(name?: string, defaultValue?: string): Promise<string | null> {
+        if (await this.needsReconfigure()) {
```
gcampbell-msft, line 2878: I don't think we should do this. While it's a good idea, we should leave it up to the user to control in their tasks.json or launch.json whether they want to enforce a configure first. I would recommend taking this out.

gcampbell-msft (changes requested):
Overall, this seems like a good idea and allows people to get variables set in their CMake project, whether from CMakePresets, CMakeLists.txt, or anywhere else. With a couple of small changes, this looks like a good addition with minimal risk. Thanks!

src/cmakeProject.ts
malsyned: I can certainly take this out, it's as easy as reverting bcbc15ef3e3813f3ec97cfff299f34641c24c09f. Before I do, I want to give you some insight into why I added it, in case you find it persuasive.

Backing it out out out creates the potential for surprising, inconsistent behavior when *a)* `cmake.configureOnEdit` is `false`, and *b)* `cmake.cacheVariable` is used along side any of the `cmake.launch*` or `cmake.getLaunch*` commands. Since all of those commands start with a similar block of code, either by calling [`CMakeProject.getLaunchTargetPath()`](https://github.com/microsoft/vscode-cmake-tools/blob/eb5a8060394270784055ba87ab939802361bdee9/src/cmakeProject.ts#L2657) or [`CMakeProject.prepareLaunchTargetExecutable()`](https://github.com/microsoft/vscode-cmake-tools/blob/eb5a8060394270784055ba87ab939802361bdee9/src/cmakeProject.ts#L2788), it is possible to get either a stale cache value or a fresh one depending on the order in which the commands appear in a `launch.json` or `tasks.json` configuration.

Here's a practical example. I set up a `cmake.cacheVariable` input like so:
```jsonc
    "inputs": [
        {
            "id": "cmakeCache.DEVICE",
            "type": "command",
            "command": "cmake.cacheVariable",
            "args": { "name": "DEVICE" }
        }
```
and a debug configuration with these two lines in this order (in this case, a `cortex-debug` configuration, but the example should generalize to other use cases):
```jsonc
            "device": "${input:cmakeCache.DEVICE}",
            "executable": "${command:cmake.launchTargetPath}",
```

I then go and add the following line to `CMakeLists.txt`
```cmake
set(DEVICE RM44L520 CACHE INTERNAL "")
```

The first time I try to start this debugging session, it will fail because the cache won't yet have that change incorporated and the debug server will receive wrong arguments. However, since `cmake.launchTargetPath` gets evaluated before trying to start up, a reconfigure will happen **anyway**, just too late to help. So if I try immediately to start the debug session again, it will succeed.

If, on the other hand, those two `launch.json` lines had been written in the other order:
```jsonc
            "executable": "${command:cmake.launchTargetPath}",
            "device": "${input:cmakeCache.DEVICE}",
```
the debug session would have started successfully on the first attempt.

I guess this leads to a deeper question: how do you decide which commands will try to do a reconfigure if needed, and which won't? It's not really documented in the [command substitution documentation](https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-settings.md#command-substitution). I figured that since this command read from the cache, it made sense to put it in the category of commands that try to reconfigure if needed.

If this hasn't been persuasive, let me know, I'll back out this block and update the PR right away.

gcampbell-msft: @malsyned I personally don't love the design of having the forcing of the reconfigure, as it forces a reconfigure even if a user doesn't want it, however, you've convinced me, and I think it's better to stay consistent with the other similar commands. I'm good with the way you have it. 

Thanks for the in-depth explanation!

gcampbell-msft (dismissed):

malsyned:
OK awesome.

Looks like the CI is failing on localization. I admit to some wishful mimicry when it comes to that part of the PR. I'll see if I can figure out what the problem is and fix it. Any pointers on what I might be doing wrong would be appreciated.

src/cmakeProject.ts
```diff
@@ -2874,6 +2874,22 @@ export class CMakeProject {
...
+            await this.configureInternal(ConfigureTrigger.api, [], ConfigureType.Normal);
+        }
+        if (!name) {
+            void vscode.window.showErrorMessage(localize('cachevariable.missing.required.argument.name', 'cacheVariable: missing required argument "name".'));
+            return null;
+        }
+        const value = (await CMakeCache.fromPath(await this.cachePath))?.get(name)?.value ?? defaultValue;
+        if (value === null || value === undefined) {
+            void vscode.window.showErrorMessage(localize('cachevariable.not.found', `Variable "${name}" not found in CMake cache, and no default provided.`));
```
gcampbell-msft, line 2887: So for the localize method, instead of using string interpolation, you should use something like this

```suggestion
            void vscode.window.showErrorMessage(localize('cachevariable.not.found', `Variable "{0}" not found in CMake cache, and no default provided.`, name));
```

malsyned: Gotcha. I found a bunch of other examples in `cmakeProjects.ts` of localize being called like that, but with the quotes in the third argument like `` `"${name}"` ``. I followed that pattern and pushed the fix before seeing your comment. Let me know if you want me to swap it out for your exact suggested change.

malsyned: I agree with you on that nit, I was just playing "when in Rome" doing it the other way. Happy to revise.

src/cmakeProject.ts
```diff
@@ -2874,6 +2874,22 @@ export class CMakeProject {
...
+        if (await this.needsReconfigure()) {
+            await this.configureInternal(ConfigureTrigger.api, [], ConfigureType.Normal);
+        }
+        if (!name) {
+            void vscode.window.showErrorMessage(localize('cachevariable.missing.required.argument.name', 'cacheVariable: missing required argument "name".'));
+            return null;
+        }
+        const value = (await CMakeCache.fromPath(await this.cachePath))?.get(name)?.value ?? defaultValue;
+        if (value === null || value === undefined) {
+            void vscode.window.showErrorMessage(localize('cachevariable.not.found', 'Variable "{0}" not found in CMake cache, and no default provided.', `${name}`));
```
gcampbell-msft, line 2887: Even further, you can just use `name` directly as the parameter, it's already a string

malsyned: d'oh. Of course. Sorry for the noise, I was moving too fast.

gcampbell-msft: No worries! Thanks

gcampbell-msft (approved):

malsyned:
I pulled in your mac CI flakiness fix, in the hopes of getting a green light.

PR #4427, opened 2025-04-23
Update devcontainer.json

jody89:
[

- 
- 
-


gcampbell-msft (approved):

i-ky:
@gcampbell-msft, I think this PR needs to be reverted. It turns `devcontainer.json` into invalid JSON.

PR #4440, opened 2025-04-29
Add `cmake.exclude` setting to allow for multi-root scenarios to ignore certain folders

### LINKED ISSUE #1069: Ignore some root folders in a multi-root workspace

### LINKED ISSUE #4112: Improve multi-root CMake detection
### Brief Issue Summary

Related open issues: #2338, #1069.
Related closed issues: #3278, #3170.

## Scenario
In a multi-root workspace, people are hitting issues where, when some of the roots are CMake, and some are not, the CMake extension is attempting to activate and do work for all of the roots, despite some not being CMake. 

This includes attempting to add the root to the Testing support, attempting to configure intellisense for it, etc. 

## Possible Solutions
We believe the best way to solve this is to fix the design around how we detect and integrate with multi-root support. We should better detect at a root level whether each folder is itself a CMake project and then handle that accordingly for the various things that we provide with the extension (i.e. Test integration, configuring, etc.)

We want to be able to correctly detect this so that we can ensure that the CMake Tools extension is inactive for folders that aren't CMake projects. 

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

_No response_

gcampbell-msft:
Add `cmake.exclude` setting to allow for users to, in both the `User` and `Workspace` vscode settings, specifically ignore folders in a multi-root scenario that they don't want the CMake Tools extension to acknowledge. 

Fixes #1069.
Fixes #4112. 

We believe that this solves the issues that users had mentioned in #4112 and associated bugs, because it allows users to specifically ignore folders that they don't want acknowledged. There are two main scenarios that this could happen. 

1. "True" multi-root. This is a scenario where a user opens a workspace file that has multiple folders opened. In this case, users can utilize the `cmake.exclude` setting to ignore certain folders. 
2. Multiple sub-folders multi-root. This is a scenario where a user opens a single workspace folder, but they have nested subfolders (either CMake or non-CMake). The first time this is opened, CMake Tools will ask if you want to configure the project, and if so, you will be prompted to select a folder. This folder will then be set as the `cmake.sourceDirectory`. In this case, users can be selective in their `cmake.sourceDirectory` setting, and by omitting any folders they don't want handled by CMake Tools, they can effectively be ignored. 

This may still hit, in a "True" multi-root scenario, a pop-up asking you if you want to configure with CMake Tools in a non-CMake folder, but you can use this feature to ignore it after this.


gcampbell-msft:
@danniesim I'd appreciate your review and feedback, as you've worked and helped post issues in this space. You can either test with this locally, or you could use this .vsix. 
[cmake-tools.zip](https://github.com/user-attachments/files/19961336/cmake-tools.zip)
Reminder: You'll need to download the file and modify the file extension to be .vsix.

gcampbell-msft:
TODO: Investigate and confirm that default active folder is handled correctly and the active folder selection, related settings: `cmake.autoSelectActiveFolder`, `cmake.defaultActiveFolder`

paulmaybee (dismissed):

danniesim:
> @danniesim I'd appreciate your review and feedback, as you've worked and helped post issues in this space. You can either test with this locally, or you could use this .vsix. [cmake-tools.zip](https://github.com/user-attachments/files/19961336/cmake-tools.zip) Reminder: You'll need to download the file and modify the file extension to be .vsix.

Will do. Thanks!

paulmaybee (dismissed):

gcampbell-msft:
@danniesim I've made a couple of updates, so it'd be better to test locally rather than the previously uploaded vsix, just FYI. Thanks!

gcampbell-msft:
@danniesim I've gotten approval from some team members, so please keep me in the loop with any testing you do. I'd love to merge this PR this week or next. I understand you likely have many other things taking your attention, so no worries, I just wanted to check with you since you've helped contribute to this area. Thanks!

snehara99 (approved):

danniesim:
@gcampbell-msft It works, nice one, and thanks to all contributors. However, absolute paths are not great. Paths relative to the workspace file would be better as it would be portable to anywhere/path/machine the repo is cloned to. See screenshot for how it works on my workspace:
Screenshot 2025-05-06 at 10 18 24

gcampbell-msft:
@danniesim Thanks for the feedback! I did consider this, but for the initial MVP I wanted to make sure we didn't introduce confusion, and relative paths introduce difficulties if people set this at the user level. However, we are more than willing to consider this, please create a feature request and we'll do our best to prioritize it accordingly. Thanks!

danniesim:
@gcampbell-msft Perhaps something like this "${workspaceFolder}/project_folder" (The Python extension supports this)

Erlkoenig90:
> @gcampbell-msft It works, nice one, and thanks to all contributors. However, absolute paths are not great. Paths relative to the workspace file would be better as it would be portable to anywhere/path/machine the repo is cloned to.

I strongly agree, I would be very helpful to be able to put something like this into the `.code-workspace` file:

```json
{
	"folders": [
		{
			"name": "MainWorkspace",
			"path": "."
		},
		{
			"name": "Subproject1",
			"path": "Subproject1"
		},
		{
			"name": "Subproject2_Lib",
			"path": "Subproject2/Lib"
		},
		{
			"name": "Subproject2_App",
			"path": "Subproject2/App"
		}
	],
	"settings": {
		"cmake.exclude": [
			"${workspaceFolder:MainWorkspace}",
			"${workspaceFolder:Subproject2_Lib}"
		]
	}
}
```

ie. allow substitution with `${workspaceFolder}`, also allow providing a folder name. This would allow us to have library projects in the workspace that can't be built stand-alone, but are included by other subprojects in CMake. Excluding the library from the CMake extension would avoid trying to configure it or cluttering the folder list.

Additionally it would be very nice being able to add something like `"cmake.configureSettings": { "CMAKE_MODULE_PATH": "${workspaceFolder:Subproject2_Lib}/cmake" }` to the `.code-workspace` file but that's a different story.

Issue #4443, opened 2025-04-30
[Bug] Repeat Execution in Test Presets Causes CTest error


ogagliardo:
### Brief Issue Summary

When I specify the execution.repeat option in testPresets, the vscode cmake tool is appending "--repeat until-pass:3" to the list of ctest arguments. 
```json
"testPresets": [
    {
      "name": "native-x86-gcc",
      "configurePreset": "native-x86-gcc",
      "execution": {
        "repeat": {
          "mode": "until-pass",
          "count": 3
        }
      }
    }
]
```
Because the argument is quoted it is treated as a single argument, and ctest (v3.30.1) throws an error:
```bash
$ /opt/tools/vcpkg/downloads/tools/cmake-3.30.1-linux/cmake-3.30.1-linux-x86_64/bin/ctest "--repeat until-pass:3"
CMake Error: Unknown argument: --repeat until-pass 3
CMake Error: Run 'ctest --help' for all supported options
```
If I run the command manually, without quotes, it works:
```bash
$ /opt/tools/vcpkg/downloads/tools/cmake-3.30.1-linux/cmake-3.30.1-linux-x86_64/bin/ctest --repeat until-pass:3
Test project /home/user/project/Bld/native-x86-gcc/Tst
      Start  1: Path.To.Test.Case
 1/44 Test  #1: Path.To.Test.Case ..........
```

On line 105795 of main.js:
```js
        preset.execution.repeat && result.push(`--repeat ${preset.execution.repeat.mode}:${preset.execution.repeat.count}`);
```
The entire repeat option is inside a single set of backticks, causing the space to be included in the argument, which adds the quotes.
If each part of the option is a separate parameter to push(), the arguments are passed to ctest properly:
```js
        preset.execution.repeat && result.push('--repeat', `${preset.execution.repeat.mode}:${preset.execution.repeat.count}`);
```

NOTE: when ctest v3.28.3 is used instead, the invalid argument "--repeat until-pass:3" is discarded silently, it is only with v3.30.1 that this actually causes an error. So the malformed argument was silently failing in previous versions of ctest as well.

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.99.3",
  "cmtVersion": "1.20.53",
  "configurations": [
    {
      "folder": "/home/user/project",
      "cmakeVersion": "3.30.1",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {
[...]

PR #4444, opened 2025-04-30
Broke repeat execution flag and value into separate arguments to ctest (#4443)

ogagliardo:

- split 'repeat' flag and value into separate arguments to ctest


gcampbell-msft (changes requested):
I approve of this change, but could you add a Changelog to track the change and please feel free to credit yourself in the changelog following the pattern. Thanks!

ogagliardo:

Issue #4283, opened 2025-02-16
CompilationDatabase - performance bottleneck for mid-large C++ projects (shlex)


borjamunozf:
### Brief Issue Summary

Hello everyone.
In our company we have been really struggling with the decision to use VSCode for C++ codebase. Most of the time it has been really frustrating: crashes, massive slowness & buggy.

Lately some details has been fixed or improved, but it's not enough to make a satisfying experience to work with mid-large C++ codebases. So I decided to see if we can do anything to improve this extension first, later the c++ ext itself, identify pain points and propose ideas:

Debugging the extension found that for the Configure Step with CMake Extension, roughly a total of **~35seconds** is spent with the creation of the Compilation Database object. This could make sense, our **compile_commands.json** ranges between 40Mb to 120Mb.

Inside the code, a basic timer shows the following:
- Read file ~1 to 5 second
- Parse JSON ~ 2 seconds
- Pushing json compile commands to database array: ~1.2s

And the ~95% time consumer, **shlex.split**!:

```typescript
arguments: cur.arguments ? cur.arguments : [...shlex.split(cur.command)]
```

Considering how we can speed up this, some of the ideas:
1) Is it possible to defer to a lazy initialzation? And when the compile command is requested, do by demand the cmd.argumets shlex.split.
2) If it's not possible and we need to have the full arguments for all commands when the object is created, then:
2.1 Optimize the shlex. I'm non a Typescript dev, but we'll see how to fine tune the algorithm if possible.
2.2 WASM Rust - not so complex to use the shlex from Rust wasm_bindgen:
- Tested and the time reduced from **~30s** to **~4s**
3. Would us benefit to handle the CompilationDatabase as a WASM Rust component perhaps?

With default shlex - 30s
![Image](https://github.com/user-attachments/assets/12ca9657-126b-4883-a7d6-d6dc0dc88ab4)

With Rust WASM shlex: ~3.7s
![Image](https://github.com/user-attachments/assets/7e33f5c8-6643-4628-8465-450102f51631)

I see that the shlex.split uses a Generator & concatenation. Could it possible to use an array instead and also avoid concatenation? This could be the first approach.

Thanks!
Hopefully this will be the base to continue improvement the extension performance.

### CMake Tools Diagnostics

```shell

```

### Debug Log

```shell

```

gcampbell-msft:
[...]

Issue #4279, opened 2025-02-13
CMake stops working: "cmake .viewLog" or others says command not found


borjamunozf:
### Brief Issue Summary

Hello. 

- C++ real world project, mid-large size company one.
- Ubuntu 20.04 WSL
- It has been happening for multiple releases. Latest one also: 1.19.52

Extension seems to stop working randomly. Sometimes it takes a few hours, other it crashes.

_NODE_OPTIONS has been added in ~/.bashrc already._

Error messages:
- Any CMake command stops working, so I can't get or execute the log diagnostics or any other.

```
cmake.logDiagnostics: command not found:
cmake.viewLog command not found`
```

The wokaround is to restart vscode and opening the project again.

Any guidance or help how I can provide you more info to help work this out?

### CMake Tools Diagnostics

```shell

```

### Debug Log

```shell

```

borjamunozf:
I can't reproduce it other machine, it's coming from a specific developer. 

1) Nothing specials, he remarks just using VSCode with C++ extension & CMake. Sometimes it crashes in a few minutes, sometimes takes longer: Go to Definition, navigate through code... He does not identify any specific crash pattern
2) Bisection procedure identified the CMake Extension as the faulty one.

I'm thinking I'll try to debug the extension live perhaps and see where it breaks when it fails. 
I'll come back to you with more results.

Amy-Li03:
Hi @borjamunozf , thanks for your reply!
We are very happy that you are debugging the extension to investigate this issue, happy coding!

borjamunozf:
More info with **strace** confirms the heap issue.

**FATAL ERROR: Reached heap limit "**

 protect(0x3137cf7c0000, 262144, PROT_READ|PROT_WRITE) = 0
futex(0x421f76d0, FUTEX_WAKE_PRIVATE, 1) = 1
futex(0x421f7680, FUTEX_WAKE_PRIVATE, 1) = 1
futex(0x421f76d0, FUTEX_WAKE_PRIVATE, 1) = 1
futex(0x421f7680, FUTEX_WAKE_PRIVATE, 1) = 1
futex(0x421f76d0, FUTEX_WAKE_PRIVATE, 1) = 1
futex(0x421f7680, FUTEX_WAKE_PRIVATE, 1) = 1
futex(0x421f76d4, FUTEX_WAKE_PRIVATE, 1) = 1
futex(0x421f7680, FUTEX_WAKE_PRIVATE, 1) = 1
[...]

PR #4458, opened 2025-05-11
improv: shlex.split() function heap allocs & perf for large compile commands

borjamunozf:
## This change addresses item #4283 

- Coming from a discussion regarding #4283 and also related to #4279 I appreciated a huge heap usage when working with our large compile_commands.json generated by the build system.

The shlex (and CompileDatabase infoByFilePath) is only responsible for an average usage in the heap of ~4GB across our different projects. It also takes an avg of 35-40sg to parse.

- After digging a bit in the code and benchmarking it, it looks like an almost direct improvement to the **split** function is to avoid the naive string concatenation and extra heap allocations with a token array and joining at the end.

## The purpose of this change

- Reduce heap usage at least 50-60% in mid to large CompileDatabase
- Improve response times after the configuring step and processing of the Compile Database.
- Function signature and functionality respected avoiding possible side effects.

## Other Notes/Information

Probably there could be room to improve or optimize further, but I think this is a almost direct and easy change to apply.

Tests seems to pass ok.

Some dummy file to bench:
```

import { CompilationDatabase, CompileCommand } from './compilationDatabase';
import { fs } from './pr';
import * as v8 from 'v8';
import * as shlex from './shlex';
import { time } from 'console';
import * as util from './util';

const databasePaths = [
    '/home/borjamf/workspace/vscode-cmake-tools/src/compile_commands.json',
];

function logHeapUsage() {
    const heapStats = v8.getHeapStatistics();
    console.log('Heap Statistics:');
    console.log(`Total heap size: ${heapStats.total_heap_size / 1024 / 1024} MB`);
    console.log(`Used heap size: ${heapStats.used_heap_size / 1024 / 1024} MB`);
    console.log(`Heap size limit: ${heapStats.heap_size_limit / 1024 / 1024} MB`);
}

async function buildCompileCommands(databasePaths: string[]): Promise<CompileCommand[]> {
    const database: CompileCommand[] = [];
    for (const path of databasePaths) {
        if (!await fs.exists(path)) {
            continue;
        }

        const fileContent = await fs.readFile(path);
        try {
            const content = JSON.parse(fileContent.toString()) as CompileCommand[];
            database.push(...content);
        } catch (e) {
            return database
        }
    }
    return database;
}

async function main() {
    const cmd_commands: CompileCommand[] = await buildCompileCommands(databasePaths);
    logHeapUsage();

    console.log("Before building the map")
    console.time('Total time spent building infoByFilePath: current-shlex');
    const infoByFilePath = cmd_commands.reduce(
                (acc, cur) => acc.set(cur.file, {
                    directory: cur.directory,
                    file: cur.file,
                    output: cur.output,
                    command: cur.command,
                    arguments: cur.arguments ? cur.arguments : [...shlex.split(cur.command)]
                }),
                new Map<string, CompileCommand>()
    );

    console.timeEnd('Total time spent building infoByFilePath: shlexSplit');

    console.log("After building the map")
    logHeapUsage();
}

main();
```

Heap stats with current split: **4GB** + time 48.452s
imagen

Heap stats after optimized split: **600Mb** + time ~16s
imagen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Features:
 Improvements:
 
 - No longer convert paths on lowercase on MacOS to enable cpp tools to resolve them. [#4325](https://github.com/microsoft/vscode-cmake-tools/pull/4325) [@tringenbach](https://github.com/tringenbach)
+- Speedup & reduce heap allocations in shlex split module function. Significant gains for mid-large compile_commands.json - CompilationDatabase construction. [#4458](https://github.com/microsoft/vscode-cmake-tools/pull/4458) [@borjamunozf](https://github.com/borjamunozf)
 
 Bug Fixes:
 
--- a/src/shlex.ts
+++ b/src/shlex.ts
@@ -76,7 +76,6 @@ export function* split(str: string, opt?: ShlexOptions): Iterable<string> {
     }
 }
 
-
 /**
  * Quotes a string for safe use in a shell command.
  * If the string contains special characters, it will be wrapped in double quotes and any existing double quotes will be escaped.
--- a/src/shlex.ts
+++ b/src/shlex.ts
@@ -13,21 +13,23 @@ export function* split(str: string, opt?: ShlexOptions): Iterable<string> {
     opt = opt || {
         mode: process.platform === 'win32' ? 'windows' : 'posix'
     };
+
     const quoteChars = opt.mode === 'posix' ? '\'"' : '"';
     const escapeChars = '\\';
     let escapeChar: string | undefined;
-    let token: string | undefined;
+    let token: string[] = [];
     let isSubQuote: boolean = false;
 
     for (let i = 0; i < str.length; ++i) {
         const char = str.charAt(i);
+
         if (escapeChar) {
             if (char === '\n') {
                 // Do nothing
             } else if (escapeChars.includes(char)) {
-                token = (token || '') + char;
+                token.push(char);
             } else {
-                token = (token || '') + escapeChar + char;
+                token.push(escapeChar, char);  // Append escape sequence
             }
             // We parsed an escape seq. Reset to no escape
             escapeChar = undefined;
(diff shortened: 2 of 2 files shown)


gcampbell-msft (changes requested):
Good news! I hunted through some historical testing I did of this, and this is identical to some changes I had tested out, but you were able to perform some better benchmarking, so I'm excited to take this! 

Before approving and merging though, could you please add a CHANGELOG entry? Please credit yourself following the pattern people have used before.

gcampbell-msft:
@borjamunozf Oh, and there are some formatting errors that the linter is complaining about, that's why the builds are failing. Could you fix this as well? 

It's simple enough that I could fix it, but then we would need a different approver, so it'd be best if you could make those changes. Thanks!

borjamunozf:
> @borjamunozf Oh, and there are some formatting errors that the linter is complaining about, that's why the builds are failing. Could you fix this as well?
> 
> It's simple enough that I could fix it, but then we would need a different approver, so it'd be best if you could make those changes. Thanks!

There it goes! Thanks :)

gcampbell-msft (approved):

borjamunozf:

gcampbell-msft:
> @microsoft-github-policy-service agree

Could you try again? It seems like the license/cla still hasn't reported.

borjamunozf:

borjamunozf:
Don't know, it seems that it's not picking it up. Not sure if I'm missing something. 

Is the only step required as far as I recall, right?

borjamunozf:
It does redirect to my home, but perhaps you can trigger this?

https://cla-assistant.io/check/microsoft/vscode-cmake-tools?pullRequest=4458

Mentioned in the [COMMON_ISSUES cla](https://github.com/cla-assistant/cla-assistant/blob/main/COMMON_ISSUES.md)

gcampbell-msft:
> It does redirect to my home, but perhaps you can trigger this?
> 
> https://cla-assistant.io/check/microsoft/vscode-cmake-tools?pullRequest=4458
> 
> Mentioned in the [COMMON_ISSUES cla](https://github.com/cla-assistant/cla-assistant/blob/main/COMMON_ISSUES.md)

Did you also try that link? I attempted it and nothing seemed to happen

borjamunozf:
Finally. I'm dumb, the commits were pushed with my company email account.
Fixed and seems that CLA is ready.

Sorry for the bother!

borjamunozf:
The macOS pipeline is failing always in the same tests. I have read that you mentioned about a flaky test, could be this?
Should we just wait then?

gcampbell-msft:
@borjamunozf I will force merge this, other builds are failing due to this, and we recently tried to fix this but it doesn't seem to have worked. Thanks for this contribution, merging.

PR #4460, opened 2025-05-14
Add support for presets v10

gcampbell-msft:
Adds `graphviz` and `$comment`


gcampbell-msft:
@TSonono Sorry for pinging again, but it looks like the tests for the test coverage are failing again, this time it seems to be crashing. Any ideas? We really want to integrate these tests to ensure we don't regress, but we may need to remove the tests if they aren't consistent. Thanks again for your effort here

TSonono:
@gcampbell-msft I tried to look in to it a little today. My theory is that when the single root project is opened in a test context, the test explorer integration is not appearing, which means that the invocation of `testing.coverage.uri` is not resulting in anything occurring. I'm not sure why this is the case, the extension code by default disables the text explorer integration during testing, but I added code for testing this feature that would not disable it in the tests for this particular feature.

I wonder if there's anything that could have changed either in the extension or vscode that makes it not appear when running tests, I'm not sure. I'm also not sure what our path forward should be here. It's a feature that is not really straight forward to write a test for since the VS Code APIs don't really provide good interfaces for doing just that.

laurenprinn (approved):

PR #4479, opened 2025-06-10
Potential fix for hanging build 

PILIX123:
### This fixes a bug where build would hang when "cmake.buildTask" is set to true

- Moving the worskSpaceFolderObj before the task fetching. And passing it to the filter to get the build tasks


gcampbell-msft (changes requested):
This looks like a reasonable and good fix to me, can you add a changelog entry as well?

PILIX123:
> @microsoft-github-policy-service agree

Issue #4464, opened 2025-05-20
[Feature] Add support for std args c++26/gnu++26/c++2c/gnu++2c/c++23preview


sean-mcmanus:
### Request Overview

The change would be similar to the PR that added c++23 at https://github.com/microsoft/vscode-cmake-tools/pull/2476 . I could do it myself depending on the ETA your team expects to be able to do it.

The corresponding vscode-cpptools change is at https://github.com/microsoft/vscode-cpptools/pull/13608/files and  https://github.com/microsoft/vscode-cpptools-api/pull/60 for C/C++ extension version 1.26.0. The 6.3 version of vscode-cpptools has not yet been published to npm but should be available.

Without this change, CMake Tools incorrectly reports a project using c++23 as using c++latest with MSVC mode and clang/gcc modes reports c++23 instead of c++26.

There's a Makefile Tool issue at https://github.com/microsoft/vscode-makefile-tools/issues/735 .

sean-mcmanus:
@Amy-Li03 Why is this marked as Blocked? Can I submit a PR to fix it or did the CMake Tools team have an ETA on when they would have time to get to it?

Amy-Li03:
@sean-mcmanus It seems the Blocked status is default, and we are very happy for you to provide a PR to fix it!

sean-mcmanus:
This is actually largely a no-op so I'm closing it (I still have a PR, but it's only minor changes). The shipping product already correctly handles c++26/etc. -- the issue I was hitting previously was due to cmake itself using c++latest instead of c++23preview.

PR #4482, opened 2025-06-14
Cleanup standard args handling

sean-mcmanus:
## This change addresses item https://github.com/microsoft/vscode-cmake-tools/issues/4464

Cleanup how the standard args are handled to avoid confusing developers.

Add input handling for converting c++23/26 to c++20 (the latest version supported with v5).

Update vscode-cpptools to 7.1.1, which has some minor improvements (less files).

## Other Notes/Information

Follow up to the previous PR for C++23 at https://github.com/microsoft/vscode-cmake-tools/pull/2476 .


sean-mcmanus:
I can't test this until https://github.com/microsoft/vscode-cpptools-api/pull/64 is checked in.

sean-mcmanus:
Oh, it looks like the current version of cmake tools after v6 doesn't check the standard flags?

sean-mcmanus:
I need to test this still...

UPDATE: It doesn't work. c++23 projects are still using c++latest for some reason -- looks like it's building with c++latest instead of c++23preview. CMake itself may need to be updated. Investigating...

It appears CMake itself isn't supporting c++23preview.

src/cpptools.ts
```diff
@@ -64,15 +64,11 @@ interface TargetDefaults {
...
+function parseCppStandard(std: string, canUseGnu: boolean): StandardVersion {
     const isGnu = canUseGnu && std.startsWith('gnu');
-    if (std.endsWith('++23') || std.endsWith('++2b') || std.endsWith('++latest')) {
-        if (canUseCxx23) {
-            return isGnu ? 'gnu++23' : 'c++23';
-        } else {
-            return isGnu ? 'gnu++20' : 'c++20';
-        }
-    } else if (std.endsWith('++20') || std.endsWith('++2a')) {
+    if (std === 'c++latest' || std.endsWith('++26') || std.endsWith('++2c') ||
```
bobbrow, line 69: This doesn't seem right to me.  For MSVC, I thought that C++23preview == C++26 == C++latest so we should send C++26.  For the other compilers, I'm not aware of any restriction and C++23 should send C++23 and C++26 should send C++26.

bobbrow: So perhaps the only thing I would change from the original is that ++latest and ++23preview send C++26 now.

sean-mcmanus: No, c++23preview is c++23. This code only executes for v4 of vscode-cpptools API, which only supports c++20, so c++23/26 are converted to that.

bobbrow: What I meant was that C++23preview isn't sent by cmake. It sends C++latest instead.

But yeah, if this code doesn't run for v7 of the API, then you don't need that.

sean-mcmanus: @bobbrow I removed the check for c++23preview -- unless you know of a way for cmake to still send that to cpptools <=V4. The only difference in behavior is whether it falls back to using c++20 or an older default, which might be c++17.

PR #4501, opened 2025-06-29
Fix 4345

hippo91:

* Create a new test collection that is the exclusion of request.exclude tests from the request.include ones

* Flatten the request.include and request.exclude test collections so that all leafs (i.e tests and not suites) are collected


hanniavalera:
@hippo91 Hi, would you be open to adding a `CHANGELOG.md` update? (You can place it under the latest version, 1.21).

hanniavalera (approved):
Thanks for fixing this!

PR #4518, opened 2025-07-17
Update API changes that add methods with command result information

gcampbell-msft:
This implements the API changes suggested/made in this PR: https://github.com/microsoft/vscode-cmake-tools-api/pull/10. 

### Backwards-compatibility

This is designed to be backwards compatible with older versions of the API by simply adding methods. While the new parameters are optional and the return type can be ignored which would make this back-compat, the pre-existing methods threw an error if they hit a non-zero return code. If this is changed, users may be broken and be forced to update to this version.

@benmcmorran @bobbrow @laurenprinn @moyo1997 @paulmaybee Let me know what you think about this, it may be acceptable to simply add a non-void return type, add optional parameters to the current methods. 

### Storage of tests

This adds the ability to "cache" our tests for retrieval even when the CMake version is an older. You'll notice in ctest.cs `refreshTests` that, before this PR, we only stored in `this.tests` our current tests in the branch logic where the CMake version was 3.14 or beyond. This would make our support for our `listTests` API depend on whether you had a more recent version of CMake. To solve this, we also store the "legacy" tests if it's an older version of CMake. 

### Refreshing of tests

This modifies our logic to allow for refreshing of tests even if the Test Explorer integration is turned off. This allows for our support of the `listTests` API. 

### Wider support of cancellation of commands

This adds the ability to cancel commands through our API. To support this, the ability to combine multiple cancellation tokens into one was added so that we can handle cancellation through the API or through the UI at the same time. 

## The purpose of this change

Implements the changes made in the above linked PR. This adds the ability, through the extension API to be able to build, configure, clean, install, and reconfigure and get the result information (return code, stdout, and stderr) returned to you. Additionally, it lets you list the current build targets and current tests available.


package.json
```diff
@@ -3836,7 +3836,7 @@
     "tsconfig-paths": "^3.11.0",
     "tslint": "^6.1.3",
     "typescript": "^4.1.5",
-    "vscode-cmake-tools": "^1.2.0",
+    "vscode-cmake-tools": "file:../vscode-cmake-tools-api",
```
gcampbell-msft: Builds/tests will fail until the https://github.com/microsoft/vscode-cmake-tools-api/pull/10 PR merges and this is updated to the new version

src/api.ts
```diff
@@ -83,27 +84,59 @@ class CMakeProjectWrapper implements api.Project {
         return withErrorCheck('configure', async () => (await this.project.configure()).result);
     }
 
+    async configureWithResult(cancellationToken?: vscode.CancellationToken): Promise<api.CommandResult> {
+        return this.project.configure(undefined, cancellationToken);
+    }
+
     build(targets?: string[]): Promise<void> {
-        return withErrorCheck('build', () => this.project.build(targets));
```
bobbrow, line 87: Is this changing because you want to maintain the same behavior of returning the result code even though the API itself was supposed to return nothing (void)?

gcampbell-msft: Ah, no, good point, this is an artifact of me prototyping with changing the existing method. I'll fix.

gcampbell-msft: Fixed.

bobbrow: Ah, I see. I didn't look at the definition of `withErrorCheck`.

package.json
```diff
@@ -22,7 +22,7 @@
     "syntaxes"
   ],
   "engines": {
-    "vscode": "^1.88.0"
+    "vscode": "^1.95.0"
```
itodirel: is this change needed by the PR, or just a consequence of vscode updating?

package.json
itodirel: as I understand it, this is needed until vscode-cmake-tools-api is published, but we won't merge it until that?

src/cmakeTaskProvider.ts
```diff
@@ -356,14 +357,17 @@ export class CustomBuildTaskTerminal implements vscode.Pseudoterminal, proc.Outp
     }
 
     constructor(private command: string, private targets: string[], private workspaceFolder?: vscode.WorkspaceFolder, private preset?: string, private options?: { cwd?: string; environment?: Environment }) {
+        super();
     }
 
-    output(line: string): void {
+    override output(line: string): void {
         this.writeEmitter.fire(line + endOfLine);
+        super.output(line);
```
itodirel, line 368: can we add a comment for this change?

src/ctest.ts
```diff
@@ -872,9 +857,85 @@ export class CTestDriver implements vscode.Disposable {
             };
         }
 
+    }
+
+    /**
+     * @brief Refresh the list of CTest tests
+     * @returns 0 when successful
+     */
+    async refreshTests(driver: CMakeDriver): Promise<number> {
```
itodirel, line 890: what does refresh means?  can we add more description to the comment?

src/ctest.ts
```diff
@@ -872,9 +857,85 @@ export class CTestDriver implements vscode.Disposable {
...
+            const tests = result.stdout?.split('\n')
+                .map(l => l.trim())
+                .filter(l => /^Test\s*#(\d+):\s(.*)/.test(l))
+                .map(l => /^Test\s*#(\d+):\s(.*)/.exec(l)!)
+                .map(([, id, tname]) => ({ id: parseInt(id!), name: tname! })) ?? [];
+
+            this.legacyTests = tests;
+
+            if (refreshTestExplorer && this._legacyTests) {
+                this.refereshTestsInTestExplorer(driver, ctestArgs, "LegacyCTest");
```
itodirel: can we add a comment describing the difference between these two branches?

src/ctest.ts
```diff
@@ -872,9 +857,85 @@ export class CTestDriver implements vscode.Disposable {
...
+    }
+
+    /**
+     * @brief Refresh the list of CTest tests
+     * @returns 0 when successful
+     */
+    async refreshTests(driver: CMakeDriver): Promise<number> {
+        // NOTE: If the cmake.ctest.testExplorerIntegrationEnabled is disabled, we should return early and not initialize
+        // the testExplorer.
+        const refreshTestExplorer = driver.config.testExplorerIntegrationEnabled && !(util.isTestMode() && !util.overrideTestModeForTestExplorer());
```
itodirel, line 891: this function does a little bit too much, I recommend splitting the driver.execute command branches into two other functions

src/util.ts
```diff
@@ -1378,3 +1378,25 @@ function globWrapper(globPattern: string, cwd: string): Promise<boolean> {
         });
     });
 }
+
+export function createCombinedCancellationToken(...tokens: (vscode.CancellationToken | undefined)[]): vscode.CancellationToken {
```
itodirel, line 1397: recommend add a comment describing the lifetime of the combined token, this does more than create a combined cancellation token

src/ctest.ts
```diff
@@ -1291,7 +1358,7 @@ export class CTestDriver implements vscode.Disposable {
                             extensionManager.cleanOutputChannel();
                         }
                         const buildResult = await project.build(undefined, false, false);
-                        if (buildResult !== 0) {
+                        if (buildResult.result !== 0) {
```
itodirel: was this wrong before?

itodirel (approved):

src/extension.ts
```diff
@@ -2450,7 +2450,7 @@ async function setup(context: vscode.ExtensionContext, progress?: ProgressHandle
 
     return { getApi: (version: api.Version) => {
         // Since our API is backwards compatible, we can make our version number match that which was requested.
-        if (version === api.Version.v1 || version === api.Version.v2) {
+        if (version === api.Version.v1 || version === api.Version.v2 || version === api.Version.v3) {
             ext.api.version = version;
```
moyo1997, line 2454: not sure what this piece of code does, but curious why there isn't v4 here even though api.ts has 
    version: api.Version = api.Version.v4;

package.json
gcampbell-msft: Correct. This PR won't be merged until we merge the api, and update the npm package.

src/ctest.ts
gcampbell-msft: No, this is because of us adding more information to the return value of the build method.

src/extension.ts
gcampbell-msft: Good catch.

gcampbell-msft: Fixed.

src/util.ts
gcampbell-msft: Updated and improved the lifetime handling and disposing.

src/ctest.ts
gcampbell-msft: Updated.

package.json
gcampbell-msft: Not needed. Removed.

src/ctest.ts
gcampbell-msft: Updated to have an extraction method that calls driver.executeCommand with parameters defining the behavior and a callback to update the tests.

src/ctest.ts
gcampbell-msft: Added a comment.

src/cmakeTaskProvider.ts
gcampbell-msft: Added a comment

moyo1997 (approved):

PR #4523, opened 2025-07-22
Build tests prior to executing them

epistax:
### This changes the build-before-test behavior

- When build before test is enabled, the extension will review the list of selected tests, which projects they are in, and request a minimum build from CMake.
- This is done by explicitly asking CMake to build the test executables. 

## The purpose of this change

- When building before test, the current behavior is to pass no build target to CMake, resulting in the ALL target.
- This will require that everything built by ALL works in order to run _any_ test. The only alternative is for the user to disable build before testing entirely, which is also not ideal.
- It's also possible that the particular target needed to built the test is excluded from _all_ in which case the current build isn't sufficient to prepare the selected test(s) to run.

## Other Notes/Information

I have never contributed to this project before and I ran into difficulties running the tests. I was not able to find any testss related to the change and I'd be willing to expand the test base with a little bit of instruction. I was not able to find a proper venue for getting started with the testing. There appears to be some required machine setup prior to running the tests. Thank you for any assistance!
This change opens up the possibility for very long command lines when invoking CMake. I am not aware if munging is taken care of automatically or if I need to introduce something myself.


epistax:
@microsoft-github-policy-service agree company="Markem-Imaje"

hanniavalera:
Superseded by #4881, which rebases this concept onto the current main branch. Thank you @epistax for the original contribution!

PR #4527, opened 2025-07-29
Update TextMate grammar based on vim-syntax

pboettch:
I upgraded my grammar-generation script I use to generate the vim-syntax-highlighting-rules to put out a TextMate-grammer.

The script can be found here: https://github.com/pboettch/vim-cmake-syntax/tree/textmate-grammer-generation

- Improved syntax highlighting grammar
- Fine grained keyword/command
- Builtin-variables vs user-variables
- Properties only in property-commands

## Other Notes/Information

This is a proposal for discussion. I learnt TextMate-grammar just recently and I'm relatively new to vsCode. There is 
a lot of space for improvements. Modules are not at all handled, as an example.

Here is a screenshot of a CMake-Files using the Monokai Dimmed-theme

image
image


hanniavalera:
Hi @pboettch, thank you for this PR, and for the time you put into researching the vim syntax as a reference! The ideas here are genuinely useful and several of them are being brought in.

Our grammar has evolved toward a regex-generalized architecture rather than exhaustive keyword enumeration, meaning we prefer patterns like `CMAKE_\w+_COMPILER_VERSION` over listing every language variant individually. 

This keeps the file maintainable as CMake grows and avoids false-positives. The tradeoff is that it doesn't match the structure of a vim syntax file very directly, so the translation isn't one-to-one.

The improvements from this PR are being incorporated in a rebased form that layers your ideas on top of the current grammar's design, preserving the scoped-command architecture, the existing scope names and regex families, and the fixes already present in main.

You'll be credited in the changelog for the inspiration. Thanks again for the contribution!

snehara99 (approved):

pboettch:
Don't worry, I didn't do this PR for the glory. I just know from the past that finding something generic regarding CMake syntax is hard to impossible and also CMake does not provide anything in their source tree which would make a syntax-tree design easier - except the documentation.

Issue #3799, opened 2024-05-30
Add setting to toggle Project Outline View


sinemakinci1:
### Brief Issue Summary

We have received customer feedback that some users prefer to prior Project Outline View.

This item is to track a setting users can toggle if they want to revert to this prior experience.

PR #4538, opened 2025-08-08
Implement outlineViewType setting that changes the Project Outline View (#3799)

ar1m4n:
This change addresses item #[3799]

This changes visible behavior, documentation and configuration]

- New config option "**cmake.outlineViewType**" that accepts one of 2 values - **_list_** or **_tree_**
- If "**cmake.outlineViewType**" is set to "**_list_**", the Project Outline View will show all cmake targets in a "flat" view
- If "**cmake.outlineViewType**" is set to "**_tree_**", the Project Outline View will show all cmake targets in a hierarchical view

The purpose of this change is to enable users to choose witch type of view to use. In my case, I find the "**_tree_**" view much more useful 

Changes:

- A ConfigurationReader argument is added to ProjectNode`s constructor
- ProjectNodes::update uses the  "**cmake.outlineViewType**" to produce a different output
- An on change event handler is added to refresh the Outline View each time the value of the new config option is changed

### DIFF
--- a/docs/cmake-settings.md
+++ b/docs/cmake-settings.md
@@ -66,6 +66,7 @@ Options that support substitution, in the table below, allow variable references
 | `cmake.options.advanced` | Advanced options for CMake Tools. | See package.json | no |
 | `cmake.options.statusBarVisibility` | Controls visibility of the status bar. | `hidden` | no |
 | `cmake.outputLogEncoding` | Encoding to use for tool output. | `auto` | no |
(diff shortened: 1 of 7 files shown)


ar1m4n:

hanniavalera:
Here is what the changes look like (tested locally):
image
image

hanniavalera (approved):

snehara99 (approved):

PR #4539, opened 2025-08-08
bookmarks: Add bookmarks and filtering for the outline view

bradphelan:
### This changes visible behaviour

This pull request introduces bookmarks and filtering capabilities to the outline view in VS Code CMake Tools. Users can now bookmark specific entries for quick access and apply filters to streamline navigation within complex outlines. These features aim to enhance productivity and improve the user experience for developers working with large CMake projects.

### Add filtering to the outline view

image

### Add bookmarking to the outline view.

image

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6950,9 +6950,9 @@ tapable@^2.1.1, tapable@^2.2.0:
   integrity sha512-GNzQvQTOIP6RyTfE2Qxb8ZVlNmw0n88vp1szwWRimP02mnTsx3Wtn5qRdqY9w2XduFNUgvOwhNnQsjwCp+kqaQ==
 
 tar-fs@^2.0.0:
-  version "2.1.3"
-  resolved "https://pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages/npm/registry/tar-fs/-/tar-fs-2.1.3.tgz#fb3b8843a26b6f13a08e606f7922875eb1fbbf92"
-  integrity sha1-+zuIQ6JrbxOgjmBveSKHXrH7v5I=
+  version "2.1.4"
+  resolved "https://pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages/npm/registry/tar-fs/-/tar-fs-2.1.4.tgz#800824dbf4ef06ded9afea4acafe71c67c76b930"
+  integrity sha1-gAgk2/TvBt7Zr+pKyv5xxnx2uTA=
   dependencies:
     chownr "^1.1.1"
     mkdirp-classic "^0.5.2"


bradphelan:
@microsoft-github-policy-service agree company="Moduleworks"

src/ui/projectOutline/projectOutline.ts
```diff
@@ -662,6 +712,67 @@ export class ProjectOutline implements vscode.TreeDataProvider<BaseNode> {
...
+    public async setSearchTerm(term: string) {
+        this._searchTerm = term;
+        this._filterNode.updateSearchTerm(term);
+        await vscode.commands.executeCommand(
+            'setContext',
+            'cmake:outlineFiltered',
+            !!this._searchTerm
+        );
+        // Update description to show current filter
+        const description = this._searchTerm ? `Filtered: "${this._searchTerm}"` : '';
```
hanniavalera: @bradphelan Hi, it looks like there is a build error here with `description`. Could you remove it or use it?

src/ui/projectOutline/projectOutline.ts
```diff
@@ -19,7 +19,7 @@ interface NamedItem {
 /**
  * Base class of nodes in all tree nodes
  */
-abstract class BaseNode {
+export abstract class BaseNode {
```
gcampbell-msft, line 22: Why do we need to export the BaseNode?

src/ui/projectOutline/projectOutline.ts
```diff
@@ -35,6 +35,56 @@ abstract class BaseNode {
...
+        item.id = this.id;
+        item.tooltip = this.searchTerm ?
+            'Click to edit filter or clear with X button' :
+            'Click to set a filter for the project outline';
+        item.contextValue = 'filter';
+
+        // Add appropriate icon and command
+        if (this.searchTerm) {
+            item.iconPath = new vscode.ThemeIcon('filter-filled');
+            item.description = `(${this.searchTerm.length} chars)`;
```
gcampbell-msft: Does this description appear anywhere in the UI? If so, we should likely localize

src/ui/projectOutline/projectOutline.ts
```diff
@@ -35,6 +35,56 @@ abstract class BaseNode {
...
+        } else {
+            item.iconPath = new vscode.ThemeIcon('filter');
+            item.description = 'Click to filter';
```
gcampbell-msft: We need to localize this, so it'll look something like this `localize('click.to.filter', "Click to filter")`

src/ui/projectOutline/projectOutline.ts
```diff
@@ -35,6 +35,56 @@ abstract class BaseNode {
...
+        }
+
+        item.command = {
+            command: 'cmake.outline.search',
+            title: 'Edit Filter',
```
gcampbell-msft: Similar comment, this title should be localized.

src/ui/bookmarks.ts
```diff
@@ -0,0 +1,227 @@
...
+        );
+
+        // Set collapsible state based on whether there are children
+        if (this.getChildren().length > 0) {
+            item.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
+        } else {
+            item.collapsibleState = vscode.TreeItemCollapsibleState.None;
+        }
+
+        item.tooltip = `${this.bookmark.name}\nProject: ${this.bookmark.projectName || "N/A"}\nType: ${this.bookmark.type}\nFolder: ${this.bookmark.folderName}`;
```
gcampbell-msft: This is on the edge of needing to be localized, the only string that would get localized would be `Project`, so while it's only one string, we should likely still localize

src/ui/bookmarks.ts
```diff
@@ -0,0 +1,227 @@
...
+            name: string;
+            projectName: string;
+            folderName: string;
+            type: string;
+        }[]>("cmake.bookmarks", []);
+        this.bookmarks.clear();
+        for (const bookmark of saved) {
+            this.bookmarks.set(bookmark.id, { ...bookmark });
+        }
+        log.info(`Loaded ${this.bookmarks.size} bookmarks`);
```
gcampbell-msft: Localize

src/extension.ts
```diff
@@ -2414,6 +2438,85 @@ async function setup(context: vscode.ExtensionContext, progress?: ProgressHandle
         vscode.commands.registerCommand('cmake.outline.configureAll', () => runCommand('configureAll')),
         // add parameters that give a more specific configureTrigger
         vscode.commands.registerCommand('cmake.outline.configureAllWithDebugger', () => runCommand('configureAllWithDebugger', ConfigureTrigger.projectOutlineConfigureAllWithDebugger)),
+        // add parameters that give a more specific configureTrigger
+        vscode.commands.registerCommand(
+            "cmake.outline.search",
+            async () => {
+                const result = await vscode.window.showInputBox({
+                    prompt: "Enter a search term to filter the Project Outline",
```
gcampbell-msft: localize

src/extension.ts
```diff
@@ -2414,6 +2438,85 @@ async function setup(context: vscode.ExtensionContext, progress?: ProgressHandle
...
+                } else if (node instanceof SourceFileNode) {
+                    nodeName = node.name;
+                } else if (node instanceof DirectoryNode) {
+                    nodeName = node.pathPart;
+                } else {
+                    nodeName = node.id;
+                }
+
+                const wasAdded = await ext.getBookmarksProvider().toggleBookmark(node);
+                const action = wasAdded ? 'added to' : 'removed from';
```
gcampbell-msft: localize

src/extension.ts
```diff
@@ -2414,6 +2438,85 @@ async function setup(context: vscode.ExtensionContext, progress?: ProgressHandle
...
+                const nodeName = node instanceof TargetNode ? node.name : node instanceof SourceFileNode ? node.name : node instanceof DirectoryNode ? node.pathPart : node.id;
+                void vscode.window.showInformationMessage(`Bookmark for "${nodeName}" removed`);
+                ext.getProjectOutline().refresh(); // Refresh the outline to remove the bookmark from the list
+            }
+        ),
+        vscode.commands.registerCommand(
+            "cmake.bookmarks.clearAll",
+            async () => {
+                const result = await vscode.window.showWarningMessage(
+                    'Are you sure you want to clear all bookmarks?',
```
gcampbell-msft: localize

src/extension.ts
```diff
@@ -2414,6 +2438,85 @@ async function setup(context: vscode.ExtensionContext, progress?: ProgressHandle
...
+                    'Yes', 'No'
```
gcampbell-msft: What are your thoughts @hanniavalera on this confirmation message and whether it is necessary? @bradphelan Curious on your thoughts as well. 

We often get feedback that VS Code can overload users with notifications, so sometimes we try to mitigate adding notifications. However, I can understand the usage here.

package.json
```diff
@@ -3024,6 +3118,7 @@
                     "hidden",
                     "inherit"
                   ],
+                  "default": "hidden",
```
gcampbell-msft, line 3121: What effects does making this hidden by default cause?

gcampbell-msft: Looks like this is specific to the workflow preset, so this is a slightly different intent than the rest of this PR. As long as we're okay with this change, I'm okay with it.

src/ui/projectOutline/projectOutline.ts
hanniavalera: I had compilation errors if I didn't export as I need to bridge `BaseNode` across multiple files (i.e. bookmarks integrating with the project outline).

src/extension.ts
hanniavalera: That's fair, I just find it helpful to prompt a notification/warning given that this is a more significant action than the single-unit toggling of the bookmarks. Let me know if you have any other thoughts!

bradphelan:
Thankyou everyone for picking this up and running with this. Sorry I didn't have bandwidth to contribute after my first commits.

bradphelan:
Though. I'm not sure bookmarks are working in the pre-release? Where do the bookmarks end up in the display? A screen cast of me trying it out.


https://github.com/user-attachments/assets/cdda4e15-daf4-48af-8768-c567dbe0484d

bradphelan:
Ok. I closed VSCode and then restarted and the bookmarks appear.

image

I'm not sure if this is a problem or not.

PR #4552, opened 2025-08-19
Make autofocus on search field in cache view work

simhof-basyskom:
### This changes the behaviour of the CMake cache view

- replace the `autofocus` attribute on the `<input>` tag for the search/filter field (which seems not to work) with an `onload` handler

## The purpose of this change

The focus should be on the search/filter field when opening the cache view to make it easier to find specific values to change. If I understand correctly this was already the intention but did not work as desired.


simhof-basyskom:
@microsoft-github-policy-service agree company="basysKom"

PR #4556, opened 2025-08-25
make microbuild updates for signing

gcampbell-msft:
See title.

### DIFF
--- a/jobs/release.yml
+++ b/jobs/release.yml
@@ -6,7 +6,7 @@ parameters:
 - name: signType
   displayName: Signing Type
   type: string
-  default: test
+  default: real
   values:
     - test
     - real


jobs/release.yml
```diff
@@ -3,12 +3,11 @@
 # ==================================================================================
 
 parameters:
-- name: SignTypeOverride
-  displayName: Signing Type Override
+- name: signType
+  displayName: Signing Type
   type: string
-  default: default
+  default: test
```
sean-mcmanus: Normally, I would think you would want to default to `real` signing, with `test` signing being a special case. We only test sign for internal-only builds, but I don't know if that matters. Up to you.

sean-mcmanus (dismissed):

sean-mcmanus (approved):

Issue #4567, opened 2025-09-05
[Bug] Compiler Warnings/Errors not shown in Problems Window


laussPoettinger:
### Brief Issue Summary

When compiling with CMake extension (not task.json), the compilation output from the Output window is not transferred to the problems window.

**Compiler: TI ARM 5.6.9**

## Example
The output window produces the following output
`[build] "main.cpp", line 20: warning #190-D: enumerated type mixed with another type`

The existing problem matcher for **Wind River Diab** would match the above output
`^\"(.*)\",\s+(?:line\s+(\d+):\s+)?(info|warning|(?:|fatal |catastrophic )error)\s+\((.*)\):\s+(.*)$`

However the cmake extension prepends `[build]` in front of the compiler output, causing the regex match to fail.

Simple solution would be to remove the `^` in front of the regex 

### CMake Tools Diagnostics

```shell
Not provided
```

### Debug Log

```shell
Not provided
```

Cassie-02:
Hi @laussPoettinger We couldn't repro this issue on vs code version 1.103.2 + cmake Tools version 1.22.8, could you please help double check if I missed something? Or could you please provide a demo project?
Couldn't find TI ARM 5.6.9, used ti-cgt-armllvm_2.1.0.LTS to compiler. We download ti-cgt-armllvm_2.1.0.LTS from https://www.ti.com/tool/ARM-CGT
1. Download and extract below project, open it
[TI_ARM_CMake_Project.zip](https://github.com/user-attachments/files/22203910/TI_ARM_CMake_Project.zip)
2. Ctrl+Shift+P, click CMake: Configure
3. Click "Select a Kit", select "TI ARM Clang Complier" 
4. Ctrl+Shift+P, click CMake: Build

<img width="790" height="188" alt="Image" src="https://github.com/user-attachments/assets/d2bd5802-47f3-4b96-8cab-caffaa3641d0" />

gif:
![Image](https://github.com/user-attachments/assets/354651d1-60d0-45bb-8743-e2e40d31712f)

laussPoettinger:
Hi, 
thank you for investigating my issue and for preparing the example.
The problem with the example project is, that is doesn't produce the expected warning (or any other warning). 
So the issue cannot be reproduced. 

Let me clarify:
As soon as I press F7, the compilation starts and write the compilation output into VS Code output window (like in your example), including all errors and warnings. 
[...]

PR #4568, opened 2025-09-05
Fix Issue #4567 - change diagnostic regex to match compiler errors/wa…

laussPoettinger:
Change diagnostic regex to match compiler errors/warnings from output window

## This change addresses item #[[4567]]

- Remove "^" from problme matches diagnostic regex for diab, ghs and iar compilers, to show compiler errors/warnings in the problems window.


laussPoettinger:

PR #4573, opened 2025-09-18
Fix CMake version conflicts in macOS CI pipeline

hanniavalera:
This PR fixes the macOS CI workflow that was failing due to using the system CMake 4.x.x instead of our required CMake 3.18.3.

## Investigation
- Found that the workflow was installing CMake 3.18.3 via lukka/get-cmake, but tests were calling `/usr/local/bin/cmake` directly
- PATH environment updates were ineffective because the hard-coded path bypassed PATH lookup

## Solution
Created a wrapper script at `/usr/local/bin/cmake` that redirects to the installed CMake 3.18.3, preserving the original binary.

### DIFF
--- a/.github/workflows/ci-main-mac.yml
+++ b/.github/workflows/ci-main-mac.yml
@@ -39,35 +39,25 @@ jobs:
             cmakeVersion: 3.18.3
             ninjaVersion: 1.10.1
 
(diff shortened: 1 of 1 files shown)


hanniavalera:
@microsoft-github-policy-service agree company="Microsoft"

PR #4578, opened 2025-09-24
Fix support of cmake 4.1.0 and new cmake-file-api error files

JDESTM:
### This changes improve support of CMake 4.1.0

- Do not throw exception of no index files are found, return null and let callee deal with that. Which is currently already well done.


JDESTM:
@microsoft-github-policy-service agree company="STMicroelectronics International N.V."

gcampbell-msft:
@JDESTM This change looks like a good workaround fix to me. Please add a CHANGELOG entry, but other than that, approved.

JDESTM:
@gcampbell-msft, thanks for your review, I added the missing CHANGELOG entry.

PR #4583, opened 2025-10-06
feature: Don't expand presets when running configure or build

In-line:
# Summary of Changes
This pull request adds an enhancement to the CMake Tools extension by introducing a new configuration option, cmake.expandCMakePresets, which allows users to choose whether to expand CMake presets or use the preset name directly in the command line arguments. The default value for this new configuration is set to true.

Main objective is to gather feedback and ideas.

Although, this is against wishes of CMake documentation. This gives users clear choice and fixes many outstanding issues with the support of CMake Presets.

If this PR isn't accepted I will create a fork of CMake Tools with argument expansion entirely removed as it's the only reason I don't use this excellent extension.

# Key Changes

Updated package.json to include new configuration cmake.expandCMakePresets.

# Source Code Updates:

Updated CMakeProject and CMakeDriver classes to handle the new configuration option.

Modified methods within cmakeProject.ts, cmakeDriver.ts, cmakeFileApiDriver.ts, cmakeLegacyDriver.ts, and cmakeServerDriver.ts to conditionally expand or use preset names based on expandCMakePresets.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3578,6 +3578,12 @@
           "description": "%cmake-tools.configuration.cmake.useCMakePresets.description%",
           "scope": "resource"
         },
+        "cmake.expandCMakePresets": {
+          "type": "boolean",
+          "default": true,
+          "description": "%cmake-tools.configuration.cmake.expandCMakePresets.description%",
+          "scope": "resource"
+        },
         "cmake.useVsDeveloperEnvironment": {
             "type": "string",
             "enum": [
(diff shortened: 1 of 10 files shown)


In-line:

gcampbell-msft:
@In-line To help us understand, could you point us to the current outstanding issues with our CMake Presets support? We're definitely interested in solving any open issues with our CMake Presets support.

PR #4596, opened 2025-10-18
added REQUIRED to the list of cmake operators  and added a regex to capture integers

DARKMOONlite:
## The purpose of this change

To improve readability of CMake code by distinguishing numeric values from other. and adding REQUIRED to the list of parameters that should get some syntactic icing sugar.

I recently was running into issues with a large cmake project and had troubles reading it. After realising that most of my worries came from the fact that I had a deprecated cmake vscode extension that was ruining all syntax highlighting, I decided to try and improve it by submitting what i believe would be a good change to the repo.

Note: I was able to get this working on the main branch as well but since  this commit d217cacd507ca49bdc4e06620071949017842cba, effectively removed the advanced syntax highlighting that i was fond of in the latest update I decided to submit the pull request here where they could be combined together to make it very nice to look at. IDK if that was correct, if you want i can submit the pull request from this [change](9b0e68ff9709228a7cefe7b595775c894f9011f9) 

### Example:
This was just a sample `CMakeLists.txt` file, using the `Andromeda Italic Bordered` theme with the following pull request https://github.com/EliverLara/Andromeda/pull/40 which improves its syntax highlighting for cmake.
this pull request can be effectively replicated by adding the following code to the `settings.json` file
```json

"editor.tokenColorCustomizations": {
            "[Andromeda Italic Bordered]": {
                "textMateRules": [
                    {
                        "scope": "string.unquoted",
                        "settings": {
                            "foreground": "#D5CED9"
                        }
                    },
                    {
                    "scope": "support.type",
                        "settings": {
                            "foreground": "#c74ded"
                        }
                    }
                ]
            }
        }
```

image

As you can see.
Very Mindful. Very Demure.

### DIFF
--- a/syntaxes/CMake.tmLanguage.json
+++ b/syntaxes/CMake.tmLanguage.json
@@ -32,6 +32,7 @@
     { "include": "#quoted" },
     { "include": "#string" },
     { "include": "#generatorExpression" },
+    { "include": "#integerString"},
     { "include": "#unquoted" }
   ],
   "repository": {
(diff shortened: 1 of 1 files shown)


hanniavalera:
Hi @DARKMOONlite :) Thanks for the contribution and for caring about CMake readability! 🙂

I have a couple of concerns with the current approach:

REQUIRED as an operator: In CMake's grammar, REQUIRED is a command-specific keyword (e.g., for `find_package()`), not
a conditional operator like `AND`/`OR`/`EQUAL` that the operators pattern is designed for. See the CMake `if()` docs (https://cmake.org/cmake/help/latest/command/if.html) for the list of actual condition operators vs. the `find_package()` docs (https://cmake.org/cmake/help/latest/command/find_package.html) where REQUIRED lives. Mixing them would give REQUIRED the wrong semantic scope (`keyword.operator`), which could confuse themes and other tooling.

Integer/version highlighting: The regex \bv?\d+(?:\.\d+)*\b is quite broad as CMake treats nearly everything as
strings, so this could false-positive on target names, component names, or other identifiers that happen to contain
digits. See CMake language docs on arguments (https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#unquoted-argument) for context on why this is tricky.

I'd suggest narrowing the scope. For example, only highlighting numbers in specific contexts where they're
unambiguously numeric (like version arguments). Happy to discuss further if you'd like to iterate on this! Thanks again :)

hanniavalera:
Closing pull request due to inactivity, feel free to ping me if you have any questions or thoughts!

DARKMOONlite:
Hey @hanniavalera, okay I understand your points, I'll try focussing on "unambiguously numeric" variables, It may take me a bit as I haven't thought about this in a while, but ill get back to you on it. 
Thanks
S

Issue #4421, opened 2025-04-20
[Bug] cmaketools tirelessly cleans before building


YexuanXiao:
### Brief Issue Summary

I use SSH to remotely build on the server, but even though I set `"cmake.clearOutputBeforeBuild": false` in the local user settings, workspace settings, and remote server settings, cmaketools still cleans before building, causing the build cache to be completely ineffective.

### CMake Tools Diagnostics

```shell
{
  "os": "linux",
  "vscodeVersion": "1.99.3",
  "cmtVersion": "1.20.53",
  "configurations": [
    {
      "folder": "/home/bizwen/deque",
      "cmakeVersion": "4.0.1",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": false,
      "compilers": {
        "C": "/usr/bin/clang",
        "CXX": "/usr/bin/clang++"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
none
```

Amy-Li03:
Hi @YexuanXiao, thanks for reporting here!

Verified this issue on CMake Tools v1.20.53 and v1.21.17(pre-release), when I use SSH to remotely build, and set "cmake.clearOutputBeforeBuild": false in **workspace settings**, cmake tools will not clean before building.
Details please see following video:

![Image](https://github.com/user-attachments/assets/3e76534d-4cf4-4104-836f-6bba07d85425)

gcampbell-msft:
@YexuanXiao The `cmake.clearOutputBeforeBuild` setting does not change any behavior regarding a build clean, it is specific to whether the output window will be cleared.

ElemenTP:
Did you enable "cmake.buildTask"? I met the same problem you mentioned, when "cmake.buildTask" is enabled.

YexuanXiao:
> Did you enable "cmake.buildTask"? I met the same problem you mentioned, when "cmake.buildTask" is enabled.

[...]

PR #4605, opened 2025-11-05
Fix Issue #4421 - always clean rebuild

RedSkittleFox:
### This fixes a bug which would always run clean rebuild instead of build

- Update the `findBuildTask` function to properly account for the two existing templates.
- Properly select the `build` task instead of `clean rebuild` task. The cleaning in the `CMake: Clean Rebuild` command is handled separately and currently is being done twice. 

## Other Notes/Information

The debugging process with code snippets and screenshots has been documented in the respective issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Bug Fixes:
 - Fix kits from `cmake.additionalKits` not being shown when `cmake.showSystemKits` is `false`. [#4651](https://github.com/microsoft/vscode-cmake-tools/issues/4651)
 - Fix how `jobs` is handled in build presets. Also update how `cmake.parallelJobs` is handled as a fallback when a build preset does not define `jobs`. [#4176](https://github.com/microsoft/vscode-cmake-tools/issues/4176)
 - Fix diagnostics to handle when there isn't a command in the error output. [PR #4765](https://github.com/microsoft/vscode-cmake-tools/pull/4765)
+- Fix bug in which running "CMake: Build" would always run "CMake: Clean Rebuild" when `cmake.buildTask` is enabled [#4421](https://github.com/microsoft/vscode-cmake-tools/issues/4421) [@RedSkittleFox](https://github.com/RedSkittleFox)
 
 ## 1.22.28
 
--- a/src/cmakeTaskProvider.ts
+++ b/src/cmakeTaskProvider.ts
@@ -330,11 +330,15 @@ export class CMakeTaskProvider implements vscode.TaskProvider {
                 if (defaultTask.length >= 1) {
                     return defaultTask[0];
                 } else {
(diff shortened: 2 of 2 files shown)


hanniavalera:
Hi @RedSkittleFox, thank you for working on fixing this bug! 

Just wanted to let you know that the CI pipeline is failing due to linting issues in your changes. It may have to do with the spacing in front of the if and also the equality with `==` as opposed to `===`.

Also, I noticed there might be some duplication in the logic. It looks like you have overlapping conditions for handling 2 tasks. The current change shows: First condition handles `length === 1 || length === 2`. Then immediately after, another condition handles `length === 1 || length === 3`. Could you clarify and elaborate on the intended logic here?

Once you fix the linting issues, the tests should pass. You can run `yarn lint --fix` locally to automatically fix some of these issues before pushing!

Thanks again for your contribution! Let me know if you need any help with these changes.

RedSkittleFox:
> Hi @RedSkittleFox, thank you for working on fixing this bug!
> 
> Just wanted to let you know that the CI pipeline is failing due to linting issues in your changes. It may have to do with the spacing in front of the if and also the equality with `==` as opposed to `===`.
> 
> Also, I noticed there might be some duplication in the logic. It looks like you have overlapping conditions for handling 2 tasks. The current change shows: First condition handles `length === 1 || length === 2`. Then immediately after, another condition handles `length === 1 || length === 3`. Could you clarify and elaborate on the intended logic here?
> 
> Once you fix the linting issues, the tests should pass. You can run `yarn lint --fix` locally to automatically fix some of these issues before pushing!
> 
> Thanks again for your contribution! Let me know if you need any help with these changes.

Hey, I've fixed the the formatting. The first condition only handles the case 2. The diff is a bit confusing.

RedSkittleFox:
Okay, this is embarrassing. Give me a moment, I wrongly assumed that running the vs-code's format command is going to resolve the issue. I need to look into that.

RedSkittleFox:
Okay, now I've fixed it and confirmed it with `yarn run lint`. I've looked into the error and it turns out that the vs-code's formatter fixed my issue but caused another one that wasn't visible in the local diffs. Sorry for the confusion.

YexuanXiao:
Can we add some diagnostic information and terminate the task to avoid the current confusing behavior?

RedSkittleFox:
> Can we add some diagnostic information and terminate the task to avoid the current confusing behavior?

This issue has been resolved in my branch, I think no diagnostics required?

RedSkittleFox:
I've rebased and fixed conflicts. Could someone please review this MR, it has been pending for almost 4 months.

hanniavalera:
Hi @RedSkittleFox, I will prioritize looking at this, thank you for the ping!

Issue #1379, opened 2020-07-22
Cmake-tools does not send Windows SDK version to cpptools


michelleangela:
Issue found from internal usage of extensions.

1. Install the following Windows SDK from VS 16.7 preview.
![image](https://user-images.githubusercontent.com/38734287/88120840-bae05700-cb78-11ea-92d7-76d62e982789.png)

2. In Cmake configuration, select "Windows SDK version 10.0.18362.0 to target Windows 10.0.19041”

In user's provided example of configurations recieved from cmake-tools to cpptools, Windows SDK version is missing.
```
   Custom configurations received:
  uri: file:///x%3A/jcab-ms/CMakeTest1/test1.cpp
  config: {
  "defines": [
    "WIN32",
    "_WINDOWS",
    "NDEBUG"
  ],
  "standard": "c++17",
  "includePath": [
    "~/cmaketest1/mylib/inc",
    "~/cmaketest1/mydll/inc"
  ],
  "intelliSenseMode": "msvc-x64",
  "compilerPath": "c:/program files (x86)/microsoft visual studio/2019/preview/vc/tools/msvc/14.27.29009/bin/hostx64/x64/cl.exe",
  "compilerArgs": [
    "/DWIN32",
    "/D_WINDOWS",
    "/GR",
    "/EHsc",
    "/Zi",
    "/O2",
    "/Ob1",
    "/DNDEBUG",
    "-MD"
  ]
}

```

3. Cpptools extension then seem to get the SDK version based on what's installed, rather than from a specified version
```
sending compilation args for X:\JCAB-MS\CMAKETEST1\TEST1.CPP
  include: ~\CMAKETEST1\MYLIB\INC
  include: ~\CMAKETEST1\MYDLL\INC
  include: C:\PROGRAM FILES (X86)\MICROSOFT VISUAL STUDIO\2019\PREVIEW\VC\TOOLS\MSVC\14.27.29009\INCLUDE
  include: C:\PROGRAM FILES (X86)\MICROSOFT VISUAL STUDIO\2019\PREVIEW\VC\TOOLS\MSVC\14.27.29009\ATLMFC\INCLUDE
  include: C:\PROGRAM FILES (X86)\WINDOWS KITS\10\INCLUDE\10.0.19041.0\UM
  include: C:\PROGRAM FILES (X86)\WINDOWS KITS\10\INCLUDE\10.0.19041.0\SHARED
  define: WIN32
  define: _WINDOWS
[...]

Issue #125, opened 2017-03-08
Select SDK under Windows


amigo421:
please remind - is there (in extension settings) an ability to select from installed WinSDKs which one should be used for the build? 
is it possible to add this setting button to status bar in a line of the toolchain, build and debug type? 

P.S. also, I've [raised the request](https://github.com/Microsoft/vscode-cpptools/issues/534) in cpp-tools repository regarding include paths for IntelliSense and navigation, based on the info of the current build. would be nice if the functionality from your extension, about  -which toolchain and SDK are currently selected, can be exposed for the navigation, but this is just a question , you see - I 'm third party here :) , not a developer ...

vector-of-bool:
That would go into the build environment detection code. Currently it just uses whatever environment batch scripts it can find for a few Visual Studio installs, but I haven't looked much into a corresponding environment selection for Windows SDKs

amigo421:
I'll investigate this , and let you know about the results
please keep the request opened for 3-4 days

vector-of-bool:
There's been talk of reworking the "build environment" stuff into something more fleshed out and akin to QtCreator's _Kits_. I could see this going in that direction.

Djoulihen:
As mentioned in #1332, this could probably be implemented by adding a parameter for visual studio kits. The kits currently have "visualStudioArchitecture" which is passed to vcvarsall.bat (after some naming conversion) when retrieving the environment variables necessary for the msvc compiler to work.

Adding a "visualStudioWinSDKVersion" parameter could allow the user instruct cmake-tools to also pass that parameter to vcvarsall.bat to correctly select the Windows SDK instead of taking the default (latest ?) one. 

I've also looked at using environment variables to control the selection of the SDK but did not come up with a satisfactory solution.

lygstate:
T

> Adding a "visualStudioWinSDKVersion" parameter could allow the user instruct cmake-tools to also pass that parameter to vcvarsall.bat to correctly select the Windows SDK instead of taking the default (latest ?) one.

How to using visualStudioWinSDKVersion parameter?

chrisjaquet:
This would be a very handy feature. For various reasons (e.g supporting older software) I need to be able to specify an older Win10 SDK than is selected by default. 

[...]

PR #4619, opened 2025-11-24
Add visualStudioArguments for MSVC kit and handle it properly

lygstate:
### This changes Select SDK under Windows
This is partially for https://github.com/microsoft/vscode-cmake-tools/issues/125
And this is also can be used to implement

Closes: https://github.com/microsoft/vscode-cmake-tools/issues/1905
Closes: https://github.com/microsoft/vscode-cmake-tools/issues/1379

A new visualStudioArguments is provided.

## The purpose of this change
As the newest windows sdk is not compatible with MSVC 2015, so we need stick to specific version of windows sdk
by using `visualStudioArguments`


lygstate:
cc @gcampbell-msft  ping for take a look

mcodilla:
Hey @lygstate, thanks for the contribution! the change generally looks reasonable, there are just a couple of things we’d like to address before merging

The current structure looks like a string array where meaning is implied by order. That makes it a bit hard to reason about. Would you be willing to revise this to a more self-describing structure (for example, a JSON object so each value's purpose is explicit)?

If you’re up for making those adjustments, I think we’d be in a good place to proceed

lygstate:
> Hey @lygstate, thanks for the contribution! the change generally looks reasonable, there are just a couple of things we’d like to address before merging
> 
> The current structure looks like a string array where meaning is implied by order. That makes it a bit hard to reason about. Would you be willing to revise this to a more self-describing structure (for example, a JSON object so each value's purpose is explicit)?
> 
> If you’re up for making those adjustments, I think we’d be in a good place to proceed

I would keep this, if MSVC have more arguments in future, then we have no need update the code.

mcodilla:
That makes sense. Let's proceed with this as is. All that is needed at this point is a changelog update. Would you be willing to add that in?

PR #4620, opened 2025-11-25
Make tests easier to run

cwalther:
### This changes the developer experience (but not the user experience)

The following changes are proposed:

- ~Lift cmake_minimum_required of test cases to 3.5.0~
- Skip server mode tests on CMake >= 3.20
- Fix sysroot access test
- Clarify that fakebin must be built before running tests

(More details in the individual commit messages.)

## The purpose of this change

Make it easier for a new developer of CMake Tools to run tests.

## Other Notes/Information

I was working on a new feature for CMake Tools and, being new to the project, was having some trouble running the tests. After some investigation, it all came down to two hurdles:
1. It is not obvious that the tests are intended to be run with a specific old version of CMake (3.18).
2. It is not obvious that before running tests one needs to build the executables in test/fakebin/ using the test/fakeOutputGenerator/ project.

To fix 1., instead of just writing instructions to install CMake 3.18, I decided to make the tests work with the current version 4.2, because the [feature](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11419) I am working on requires CMake 4.3 (not released yet) and I will want to test with that too. The changes required for that turned out to be relatively minor, and are contained in the first ~three~ two of the attached commits.

To fix 2., I made the tests that without fakebin executables would fail in obscure ways, instead fail with a clear message telling the user how to build the missing executables. This is the last commit.

Together, these changes allow me to successfully run all tests (`yarn smokeTests`, `unitTests`, `integrationTests`, `endToEndTestsSuccessfulBuild`, `endToEndTestsSingleRoot`, `endToEndTestsMultiRoot`, `backendTests`) with both CMake 4.2.0 and 3.18.6 on my Ubuntu 20.04 system.

Would these changes be welcome in the project? Or would you prefer other ways to mitigate these hurdles?

Since these are not user-visible changes, I did not add an entry to CHANGELOG.md. Should I?

PR #4627, opened 2025-12-03
Fix IntelliSense in projects with compiler arguments in CMAKE_<LANG>_COMPILER

cwalther:

- Pass mandatory compiler arguments from CMake 4.3 to cpptools

## The purpose of this change

Fix IntelliSense in projects with compiler arguments in `CMAKE_<LANG>_COMPILER` by passing those arguments, newly exposed in the file API by CMake 4.3, to cpptools so it can properly determine system include paths and built-in preprocessor macro definitions.

### Problem

In CMake, the [`CC`](https://cmake.org/cmake/help/latest/envvar/CC.html) or similar environment variables as well as the [`CMAKE_<LANG>_COMPILER`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html) variables set by toolchain files can, in addition to the compiler path, contain mandatory command line arguments that must be passed to every invocation of the compiler. Previously, these arguments were not exposed in the File API, therefore CMake Tools as an IntelliSense configuration provider had no way to pass them on to `cpptools`, therefore `cpptools` would call the compiler without them when trying to determine system include paths and built-in preprocessor macro definitions, potentially resulting in non-working code navigation.

For example, for building for embedded systems with different CPU architectures we use toolchain files with lines like

```cmake
set(CMAKE_CXX_COMPILER "${INDEL_TOOLCHAIN_LOCATION}/bin/clang++${CMAKE_HOST_EXECUTABLE_SUFFIX};--config;aarch64-indel-elf.cfg;-mcpu=cortex-a72")
```

where the referenced config file contains arguments like `--target`, `--sysroot`, `-stdlib++-isystem` that influence include paths and definitions. Currently, Visual Studio Code `cpptools` calls the compiler without these arguments when trying to determine the include paths and definitions, and gets back wrong values related to the host platform instead of the embedded system.

### Steps to Reproduce

Making a generic minimal reproducing example is a bit tricky because CMake expects working compilers (including sysroots). Here is something that works for me on an x86_64 Linux system with Clang 18 installed (Ubuntu 20.04, `clang-18` package from official archive.ubuntu.com/security.ubuntu.com):

CMakeLists.txt:
```cmake
cmake_minimum_required(VERSION 3.24)
project("compilerargs" C)
add_executable("test" main.c)
```

toolchain.cmake:
```cmake
set(CMAKE_C_COMPILER "/usr/bin/clang-18;-target;i386-linux-gnu")
```

CMakePresets.json:
```json
{
	"version": 6,
	"configurePresets": [
		{
			"name": "Test",
			"generator": "Ninja",
			"binaryDir": "${sourceDir}/output",
			"toolchainFile": "toolchain.cmake"
		}
	]
}
```

main.c:
```c
#include <something.h>

int main() {
	return hello;
}
```

/usr/include/i386-linux-gnu/something.h:
```c
int hello = 42;
```

#### Steps to reproduce

1. Put the first four files above in a folder anywhere and the last in the named location.
2. Open the folder in Visual Studio Code.
3. Click _Build_ in the status bar and select preset _Test_ when prompted.
4. The build completes successfully and you can run `./output/test; echo $?` to get the expected result `42`. `file output/test` confirms that it was compiled for i386.
5. Open main.c.

#### Actual result

- The first line `#include <something.h>` has red squiggly underlines. Ctrl-clicking on it does nothing.
- Hovering on `hello` shows no tooltip.
- For reasons I don’t understand that may be related to remnants from previous experiments done on my system, ctrl-clicking on `hello` does work and takes me to its definition in …/i386-linux-gnu/something.h. I would expect it to fail as well.
- _CMake: Log Diagnostics_ and _C/C++: Log Diagnostics_ give no indication that the required `-target i386-linux-gnu` arguments have made it from CMake to cpptools, in particular the C/C++ diagnostics show the wrong `system include: /usr/include/x86_64-linux-gnu` for main.c.

#### Expected result

- No red squiggly underlines.
- Hovering on `hello` shows a tooltip “int hello”.
- Ctrl-clicking on either `<something.h>` or `hello` opens …/i386-linux-gnu/something.h.
- _CMake: Log Diagnostics_ shows `"compilerFragments": ["-target i386-linux-gnu"]` for main.c.
- _C/C++: Log Diagnostics_ shows the correct `system include: /usr/include/i386-linux-gnu` for main.c.

### Proposed Solution

CMake has just merged [an addition to the file API to expose these arguments](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11419), to be released in CMake 4.3. I therefore propose the attached commits to retrieve them from the file API and pass them on to cpptools.

~Only the last two commits are part of this change, the rest are #4620, which is required to actually be able to test with CMake 4.3 (so don’t use the combined diff for review, but the individual commits). But if there are objections to that, I will of course rework it. This change does not strictly depend on #4620, some of the added tests will just be skipped without it.~

An update to vscode-cmake-tools-api is required, submitted in microsoft/vscode-cmake-tools-api#15, so there is a dependency version bump in package.json. CONTRIBUTING.md tells me to ping “our team” in that case, but I don’t know who that is (maybe @gcampbell-msft?).

### Open Questions

- I am not sure what to do about the use of `shlex.split()`, see comments in [cpptools.ts:471](https://github.com/indel-ag/vscode-cmake-tools/blob/390eb3b3da5bbd313527d3b4e94a3b39c96d57d1/src/cpptools.ts#L471-L476). Possible options:
  - Leave it as it is and declare the (rare, as far as I can tell) case where it causes problems unsupported.
  - Add a second shlex-split function that does what I would expect and what is needed here.
  - Change the behavior of `shlex.split` to be what I would expect and what is needed here. However I fear that would break other things, because presumably the current behavior, which is explicitly tested for in test/unit-tests/shlex.test.td, was introduced for a reason in #1663.

  Any preferences? Any comment from @elahehrashedi who wrote that behavior?


cwalther:
Rebased on the updated #4620 to fix merge conflicts. No changes required.

cwalther:
Rebased on main, with minor manual conflict resolution. Verified that it still works, using the _Steps to Reproduce_ above and CMake 4.3.1. Tests also still pass (although #4891 was required for me).

I am guessing that the proper location for the changelog entry is now _1.24_, because as far as I understand, version 1.23.51 has now been released as the final 1.23 release, and there is a new pre-release 1.24.0.

cwalther:

cwalther:
Rebased to fix merge conflict in changelog.

cwalther:
I will suspend further changelog conflict fixing whack-a-mole until asked or other changes become necessary.

hanniavalera:
Hi @cwalther, I really appreciate this contribution! I would like to properly review it, would you mind placing a changelog entry for it under 1.24 under Improvements? I understand it is a little bit of a whack-a-mole, but would appreciate your time so we can take this in!

cwalther:
Rebased and moved changelog entry from Features to Improvements.

I see you made some changes to `shlex.split` in #4898. I will check whether these resolve my “open question” next.

cwalther:
#4898 does not completely resolve the open question: `shlex.split`still does a different thing than a POSIX shell would in that it still does not consume quotes that have done their duty (`splitUnix('"fo o" bar')` is `['"fo o"', 'bar']` instead of the expected `['fo o', 'bar']`). However its handling of backslashes and mismatched quotes is now improved, which further reduces the (already small, in my opinion) set of cases where the way it is used here causes incorrect results.

cwalther:
CI fails because https://github.com/microsoft/vscode-cmake-tools-api/pull/15 is not released yet. Until it is, you can test with a local checkout using this change:

```diff
diff --git a/package.json b/package.json
index 7ca2066c..f2cabc34 100644
--- a/package.json
+++ b/package.json
@@ -4497,7 +4497,7 @@
     "tsconfig-paths": "^3.11.0",
     "tslint": "^6.1.3",
     "typescript": "^4.1.5",
-    "vscode-cmake-tools": "^1.7.0",
+    "vscode-cmake-tools": "file:../vscode-cmake-tools-api",
     "vscode-nls-dev": "^3.3.2",
     "webpack": "^5.104.1",
     "webpack-cli": "^5.1.4"
```

hanniavalera:
@cwalther got in your PR for the API + published it, all should be set now :) would you be alright resolving the conflicts? please let me or @mcodilla know if you have any questions!

cwalther:
Thanks, looks good!

Rebased on main with manual conflict resolution in CHANGELOG.md, no other changes required; added yarn.lock changes for API package update.

cwalther:
I’m not sure why the tests fail on Windows. I see no reason why CMake should choose _cl.exe_ when I explicitly tell it to use _cross-compile-gcc.exe_. Looks like that somehow doesn’t get through. I guess I’ll have to set up a Windows development system and debug it.

cwalther:
Before I do that, a shot in the dark: This reminds me of a [similar problem I had while working on the CMake part](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11419#note_1729549): Could it be that in this Windows case CMake defaults to a Visual Studio generator that does not support other compilers than MSVC? Let’s try with an explicit generator.

@hanniavalera or @mcodilla, could you start the workflow again?

cwalther:
Still the same failure. Too bad. Let me try to reproduce and debug it.

cwalther:
Got it. I need to pass the generator using `-D`, not `-G`, otherwise it appends another `-G` that overrides it. There also was a forward/backward-slash path mismatch. This passes for me on Windows now. Hope it doesn’t break the others.

hanniavalera:
@cwalther, thank you so much for the ongoing iterations and contributions, i was about to reply to inform you that is what i was theorizing, but we'll see if we're missing something! (i doubt it, though!) thanks a lot!

hanniavalera (approved):

hanniavalera:
Approved and merged! Thanks a lot once again! I really appreciate the diligence :)

cwalther:
Thanks for the review!

PR #4634, opened 2025-12-10
Updating dependencies

hanniavalera:
### The following changes are proposed:

This pull request updates the `js-yaml` dependency to version 4.1.1. This change ensures the project uses the latest patch of `js-yaml`, which may include important bug fixes or security updates.

### The purpose of this change
Updated the `js-yaml` package version from 4.1.0 to 4.1.1 in both `package.json` and `ThirdPartyNotices.txt` to ensure consistency and use of the latest release.

### DIFF
--- a/ThirdPartyNotices.txt
+++ b/ThirdPartyNotices.txt
@@ -4474,7 +4474,7 @@ SOFTWARE.
 
 ---------------------------------------------------------
 
-js-yaml 4.1.0 - MIT
+js-yaml 4.1.1 - MIT
 https://github.com/nodeca/js-yaml#readme
 
 Copyright (c) 2011-2015 by Vitaly Puzrin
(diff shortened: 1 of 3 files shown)


ThirdPartyNotices.txt
```diff
@@ -4474,7 +4474,7 @@ SOFTWARE.
 
 ---------------------------------------------------------
 
-js-yaml 4.1.0 - MIT
+js-yaml 4.1.1 - MIT
```
gcampbell-msft, line 4477: Did you manually modify this? 
If so, ping me and we can talk about how to grab this from our CG pipelines and update it.

hanniavalera: now updated this with the CG pipeline, no change besides what I introduced!

gcampbell-msft (approved):

PR #4675, opened 2026-01-29
Add a regexp for problem matching on the msvc linker output

### LINKED ISSUE #4077: Allow for Additional Problem Matchers to be defined in the settings
### Brief Issue Summary

When using CMake, it is a common use case that you would integrate a linter like clang-tidy or PCLint Plus, or some processing scripts directly into your cmake build as prebuild commands with add_custom_command or add_custom_target.

These tools generate warnings that are not in the format that CMake Tools expects, and thus dont show up in the "Problems" output.

Currently, the only solution to this is to create a VSCode task that invokes CMake, and then apply your problem matchers there.

This is inconvenient, because you can no longer run the build through the CMake Tools menus, and need to use your custom task.

It would be wonderful if CMake Tools exposed a setting that allowed us to add additional problem matchers to its internal tasks, such that we can aggregate all of the errors from all of our tools without needing to run the build from outside of CMake Tools.

### CMake Tools Diagnostics

```shell
N/A
```

### Debug Log

```shell
N/A
```

### Additional Information

N/A

bradphelan:
## This change addresses item #4077

It doesn't specifically fix #4077 but it adds the missing problem matcher for MSVC linker errors.

## Problem

When running CMake: build with msvc all the build errors are found and added to the problems pane. However linker errors are not found.

This is of particular concern with using github copilot in an agent loop. The agent expects to be able to see the problems via get_errors and if the linker errors are lost then the llm gets very confused because it will see that the build has completed and then it will proceed to run my tests which may not have been built and then it goes quickly downhill.

I've made a separate REGEX for the linker to avoid messing with the already complex regexp used for the compiler. This means two regexp per line which I hope is not considered bad form or a performance issue.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Features:
 - triple: Add riscv32be riscv64be support. [#4648](https://github.com/microsoft/vscode-cmake-tools/pull/4648) [@lygstate](https://github.com/lygstate)
 - Add command to clear build diagnostics from the Problems pane. [#4691](https://github.com/microsoft/vscode-cmake-tools/pull/4691)
 
+Improvements:
+- Add MSVC linker error problem matching to the Problems pane. [#4675](https://github.com/microsoft/vscode-cmake-tools/pull/4675) [@bradphelan](https://github.com/bradphelan)
+
 Bug Fixes:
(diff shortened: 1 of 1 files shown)


bradphelan:
@hanniavalera I added more regexps and more tests

bradphelan:
@hanniavalera The big hack now is that get_errors from copilot doesn't support diagnostics without a source file. I've hacked the diagnostics to produce a dummy source file for linker errors. Now copilot can consume these errors correctly

hanniavalera:
Hi @bradphelan, thanks for this contribution, looking over this now! :)

PR #4738, opened 2026-02-18
Add extensionPack

gcampbell-msft:

- 
- 
-


hanniavalera (approved):

VGRSTM:
May I ask the rationale and/or vision behind this contribution?
Thank you.

sinemakinci1:
@VGRSTM 

The C/C++ Dev Tools extension includes experiences that build on existing C/C++ extensions to improve the end-to-end C++ workflow in VS Code. GitHub Copilot is installed by default in VS Code, and our goal is to ensure C++ developers get consistent, high quality C++ experiences out of the box. To do so, the C/C++ DevTools extension adds workflow enhancements that rely on services implemented by our other C++ extensions.
Declaring this dependency ensures that everything required for those features is automatically present, works together, and provides the optimal default C++ experience across VS Code.

VGRSTM:
> @VGRSTM
> 
> The C/C++ Dev Tools extension includes experiences that build on existing C/C++ extensions to improve the end-to-end C++ workflow in VS Code. GitHub Copilot is installed by default in VS Code, and our goal is to ensure C++ developers get consistent, high quality C++ experiences out of the box. To do so, the C/C++ DevTools extension adds workflow enhancements that rely on services implemented by our other C++ extensions. Declaring this dependency ensures that everything required for those features is automatically present, works together, and provides the optimal default C++ experience across VS Code.

@sinemakinci1 Thanks for answer.
At a high level, I like the idea.
In detail, having an extension forced on the end user (pulled when asking for the CMake Tools extension), while this extension has no publicly shared code base, is an issue for me. CMake Tools source code is exposed, but this one does not. (If I am wrong, please correct me; a source code hyperlink would be very welcome.)

VGRSTM:
@sinemakinci1 & @gcampbell-msft 

From a technical and policy standpoint, I consider this contribution quite unusual. Here, we have an extension acting as an extension pack.
In my opinion, the purpose of an extension pack is to bundle separate extensions together so they can be easily installed at one time, according to your own documentation [extension-marketplace](https://code.visualstudio.com/docs/configure/extensions/extension-marketplace).

I would welcome a new extension pack or an enhancement to an existing one like [ms-vscode.cpptools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools), which bundles [ms-vscode.cmake-tools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools) and the new [ms-vscode.cpp-devtools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpp-devtools).

sinemakinci1:
@VGRSTM Our goal was to provide the best default experience for our users, since this new extension is a companion to the CMake tools extension, which is why we chose this dependency pattern. If you decide this setup doesn't work for you, you can uninstall the C++ devtools extension at any point and your existing CMake and C/C++ tooling will continue to work as before.

Thanks for raising this, feedback like this is helpful as we iterate on the technical approach and the way we communicate these changes.

halflifefan:
@sinemakinci1 & @gcampbell-msft

I agree with you @VGRSTM, thanks for linking my issue here.

This extension is already part of the C/C++ Extension Pack, so users who want the full C++ experience already get it. Adding it as an extension pack entry in CMake Tools means it gets silently installed on users who deliberately chose CMake Tools alone rather than the full pack.

The marketplace desc saying "Enhanced development tools for C++" also undersells that it's a Copilot integration. You only find that out in the Overview. Users who don't use Copilot end up with an extra closed-source extension that does nothing for them.

Would it be possible to keep this dependency only in the C/C++ Extension Pack where it already lives, or at the very least have it respect `chat.disableAIFeatures` so it doesn't activate for users who've opted out of AI tooling?

halflifefan:
I've opened a PR to revert this change: #4800

JHartzer:
@gcampbell-msft  I, like others, intentionally install ms-vscode.cmake-tools directly because I specifically don't want or need ms-vscode.cpp-devtools as it breaks other workflows. This feature is already covered by ms-vscode.cpptools-extension-pack and this extension doesn't directly depend on cpp-devtools, therefore this change should be reverted with #4800

Issue #3015, opened 2023-02-10
API: wrong target build artifact paths


bernhardbreuss:
### Brief Issue Summary

The path to the target´s build artifacts returned by the API do not resolve to the actual location of the artifact on the disk when the target is defined in another directory than `cmake.sourceDirectory`.

cmake.getLaunchTargetPath returns:
`C:\work\src\incubator\buildpath\build\src\buildpath.exe`
API `project.codeModel.configurations[0].projects[0].targets[0].artifacts` gives
`C:\work\src\incubator\buildpath\build\src\src\buildpath.exe`

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.75.0",
  "cmtVersion": "1.13.45",
  "configurations": [
    {
      "folder": "c:\\work\\src\\incubator\\buildpath",
      "cmakeVersion": "3.25.1",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": false,
      "compilers": {
        "C": "C:/msys64/mingw64/bin/cc.exe",
        "CXX": "C:/msys64/mingw64/bin/c++.exe"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": true,
    "hasCodeModel": true,
    "activeBuildType": "Debug",
    "buildTypesSeen": [
      "Debug"
    ],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 1,
    "executablesCount": 1,
    "librariesCount": 0,
    "targets": [
      {
        "name": "buildpath",
        "type": "EXECUTABLE"
      }
    ]
  },
  "settings": [
    {
      "communicationMode": "automatic",
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Additional Information

Demo Project: [buildpath.zip](https://github.com/microsoft/vscode-cmake-tools/files/10705684/buildpath.zip)

Converting the build artifact path to an absolute path is done in different ways.
- API: https://github.com/microsoft/vscode-cmake-tools/blob/7b8f2e0a0e7eeac6b5241f2438de77f163622514/src/drivers/cmakeFileApi.ts#L481

- for `cmake.getLaunchTargetPath`: https://github.com/microsoft/vscode-cmake-tools/blob/7b8f2e0a0e7eeac6b5241f2438de77f163622514/src/drivers/cmakeFileApi.ts#L376

[...]

PR #4752, opened 2026-02-24
Fix wrong path created for artifacts when parsing code model

bkoperski:
### Problem

When parsing code model fetched using CMake File API, the path to the artifact was created using `path.join` with `build` directory specified inside `target-buildpath-*.json`. The result was then converted to absolute path using `convertToAbsolutePath` function, which in turn also prepended the path with top-level build directory. This resulted in an incorrect path with duplicated segments in the middle, e.g. given example from the issue [#3015](https://github.com/microsoft/vscode-cmake-tools/issues/3015):

`target-buildpath-*.json`:

```
{
	"artifacts" : 
	[
		{
			"path" : "src/buildpath"
		}
	],
    /* ... */
	"paths" : 
	{
		"build" : "src",
		"source" : "src"
	},
    /* ... */
}
```

`codemodel-v2-*.json`:

```
{
    /* ... */
	"paths" : 
	{
		"build" : "<workspace>/build",
		"source" : "<workspace>"
	},
    /* ... */
}
```

Path to artifact was created as `<workspace>/build/src/src/buildpath`, while the correct path should be `<workspace>/build/src/buildpath`.

### Correction

Changed the logic to append top-level build directory from `codemodel-v2-*.json` when creating the artifact path if the path from `target-buildpath-*.json` is relative, and to use it as-is if it's already absolute. This approach is in line with [CMake File API specification](https://cmake.org/cmake/help/v3.14/manual/cmake-file-api.7.html#codemodel-version-2-target-object) for `artifacts.path`, which states that:

> If the file is inside the top-level build directory then the path is specified relative to that directory. Otherwise the path is absolute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Improvements:
 - Allow preset modification commands to target CMakeUserPresets.json. The target file is determined by the focused editor, or by prompting the user when both files exist. [#4564](https://github.com/microsoft/vscode-cmake-tools/issues/4564)
 
 Bug Fixes:
+- Fix wrong path created for artifact when parsing code model using CMake File API. [#3015](https://github.com/microsoft/vscode-cmake-tools/issues/3015)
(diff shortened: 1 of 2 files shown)


bkoperski:

hanniavalera (approved):

PR #4800, opened 2026-03-08
Remove C/C++ DevTools from extensionPack

halflifefan:

- Remove `ms-vscode.cpp-devtools` from `extensionPack`

## The purpose of this change

`ms-vscode.cpp-devtools` is a Copilot integration extension that is already included in the C/C++ Extension Pack. Adding it as an extension pack dependency of CMake Tools causes it to be silently installed on users who chose CMake Tools standalone. Users who don't use Copilot end up with a closed-source extension that has no effect on their workflow.


halflifefan:

JHartzer:
I would like to throw my support behind this change. I intentionally only install `ms-vscode.cmake-tools` because  `ms-vscode.cpp-devtools` breaks other workflows. The devtools extension is not a dependency of this extension and should therefore not be installed by default without user consent. The previous change is also redundant given the existence of `ms-vscode.cpptools-extension-pack`.

troberti:
Yes, get rid of this. It's absurd that this very useful extensions sneakily imports another extension that I don't want.

PR #4813, opened 2026-03-11
Add Copilot Instructions for CMake Tools

hanniavalera:
Adds `copilot-instructions.md` so that GitHub Copilot is grounded in the repo's architecture, conventions, and common pitfalls when assisting contributors.

What's included

- Domain knowledge - two operating modes (presets vs. kits/variants), generator types (single-config vs. multi-config), code model, variable expansion, cross-platform considerations
- Project conventions - @cmt/* path alias, rollbar error boundaries, telemetry helpers
- Architecture table - maps every major layer (drivers, project, presets, kits, UI, diagnostics, debugger, config, tests) to its primary source files
- Mandatory rules - canonical data paths, vscode-nls localization boilerplate, module-scoped logging, path handling, settings update checklist, CHANGELOG requirement
- Testing checklist - unit tests, both operating modes, both generator types, cross-platform
- "Where to start" guide - quick lookup from symptom → relevant source files

Rationale:

Copilot without repo-specific context tends to guess at file locations, miss the presets-vs-kits duality, assume single-config generators, use console.log instead of the module logger, or skip localization. This prompt steers it toward the repo's actual patterns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ Bug Fixes:
 - Fix how `jobs` is handled in build presets. Also update how `cmake.parallelJobs` is handled as a fallback when a build preset does not define `jobs`. [#4176](https://github.com/microsoft/vscode-cmake-tools/issues/4176)
 - Fix diagnostics to handle when there isn't a command in the error output. [PR #4765](https://github.com/microsoft/vscode-cmake-tools/pull/4765)
 - Fix bug in which running "CMake: Build" would always run "CMake: Clean Rebuild" when `cmake.buildTask` is enabled [#4421](https://github.com/microsoft/vscode-cmake-tools/issues/4421) [@RedSkittleFox](https://github.com/RedSkittleFox)
+- Fix issue with hover provider not checking for undefined. [#4812](https://github.com/microsoft/vscode-cmake-tools/issues/4812)
 
 ## 1.22.28
 
--- a/src/languageServices/languageServiceData.ts
+++ b/src/languageServices/languageServiceData.ts
@@ -180,14 +180,12 @@ export class LanguageServiceData implements vscode.HoverProvider, vscode.Complet
         }
 
         const hoverSuggestions = this.commands[value] || this.variables[value] || this.modules[value] || this.modules[`Find${value}`];
-
-        const markdown: vscode.MarkdownString = new vscode.MarkdownString();
-        markdown.appendMarkdown(hoverSuggestions.description);
-        hoverSuggestions.syntax_examples?.forEach((example) => {
-            markdown.appendCodeblock(`\t${example}`, "cmake");
-        });
-
         if (hoverSuggestions) {
+            const markdown: vscode.MarkdownString = new vscode.MarkdownString();
+            markdown.appendMarkdown(hoverSuggestions.description);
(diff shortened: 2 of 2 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,130 @@
...
+| Targets / target types | `CMakeProject.targets`, `.executableTargets`, or `codeModelContent` |
+| Active preset | `CMakeProject.configurePreset` / `.buildPreset` / `.testPreset` / `.packagePreset` / `.workflowPreset` |
+| Active kit | `CMakeProject.activeKit` |
+| Merged preset list | `PresetsController` |
+| Cache entries | `CMakeDriver.cmakeCacheEntries` |
+| Extension settings | `ConfigurationReader` (`src/config.ts`) — never `vscode.workspace.getConfiguration()` directly |
+
+### Always handle both operating modes
+
+Check `CMakeProject.useCMakePresets`. If a code path only works in one mode, it is a bug waiting to happen.
```
gcampbell-msft: This comment feels slightly misleading, I think we'd want to be clear that if we don't check this, and it only works in one mode without considering the other, it's a bug waiting to happen. As it reads right now, it feels like it might avoid features that are only for one specific mode

.github/copilot-instructions.md
```diff
@@ -0,0 +1,130 @@
...
+### Use the module-scoped logger — never `console.log`
+
+```typescript
+import * as logging from '@cmt/logging';
+const log = logging.createLogger('my-module');
+```
+
+### `async`/`await` — never swallow errors
+
+Avoid raw `.then()` chains and empty `catch` blocks. Wrap top-level event handlers in `rollbar.invokeAsync()`.
```
gcampbell-msft: This is mostly true, probably true enough to leave in. 

However, there are cases, like for pop-up notifications, where using `.then` is useful

.github/copilot-instructions.md
```diff
@@ -0,0 +1,130 @@
...
+
+`package.json` (`contributes.configuration`), `src/config.ts` (`ConfigurationReader`), and `docs/cmake-settings.md`.
+
+### Every PR needs a CHANGELOG entry
+
+One entry under the current version in `CHANGELOG.md`, in the appropriate section (`Features:`, `Improvements:`, or `Bug Fixes:`), describing user-visible behavior in the repo's existing tense and category style.
+
+## Testing checklist
+
+- [ ] `yarn test:unit` passes
```
gcampbell-msft: Does our code actually have `test:unit`?  I don't believe we have that anywhere defined.

.github/copilot-instructions.md
hanniavalera: i adjusted it, thank you for the feedback!

.github/copilot-instructions.md
hanniavalera: i see, i overlooked that!

.github/copilot-instructions.md
hanniavalera: sorry, i thought i edited this to `yarn unitTests`

hanniavalera: fixed it now

gcampbell-msft (approved):

PR #4816, opened 2026-03-12
Remove vendor schema

halflifefan:

- Remove vendor properties from the JSON schema of CMakePresets.json

## Other Notes/Information

Since we throw a hard error if CMakePresets.json doesn't comply with the schema, we shouldn't have any vendor-specific properties in the schema unless they belong to VS Code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Improvements:
 - Honor `debugger.workingDirectory` from the CMake File API when debugging a target, so that the `DEBUGGER_WORKING_DIRECTORY` target property is used as the debugger working directory. [#4595](https://github.com/microsoft/vscode-cmake-tools/issues/4595)
 
 Bug Fixes:
+- Fix duplicate launch configurations in the ctest debug picker when both `.vscode/launch.json` and a `.code-workspace` file exist. [#4586](https://github.com/microsoft/vscode-cmake-tools/issues/4586)
 - Users can now debug with or without launch configurations from the test explorer. `cmake.ctest.neverDebugTestsWithLaunchConfiguration` can be set to `true` to bypass the quick pick menu where the launch configuration is selected. [#4790](https://github.com/microsoft/vscode-cmake-tools/issues/4790)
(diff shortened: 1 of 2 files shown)


hanniavalera:
Hi @halflifefan, thank you for filing this issue and putting together a fix so quickly! The reasoning in the PR description is spot on :)

The one concern I have is that removing the vendor schema entirely also removes the editor IntelliSense/autocomplete users get when editing CMakePresets.json (things like hostOS, intelliSenseMode, codeAnalysisRuleset, etc.). CMakeTools also reads hostOS at runtime to filter presets by OS, so having the schema validate that field is still useful.

What we could do instead is:
- Relax the intelliSenseMode enum (either make it a plain "type": "string" or add the missing shorthand modes like clang-x64)
- Add "additionalProperties": true to the microsoft.com/VisualStudioSettings/CMake/1.0 object so unknown vendor sub-keys won't cause issues anymore

That way we can keep the nice autocomplete experience but stop rejecting valid presets. What do you think?

halflifefan:
@hanniavalera

Thanks for reviewing my PR. Good call, I thought this was exclusive to VS 2026.

hanniavalera:
@halflifefan Thanks for taking my feedback! Also, if you could add a CHANGELOG.md entry noting of the issue you are fixing and credit yourself if you'd like to, that would be great!

halflifefan:
@hanniavalera Totally slipped my mind, all done :)

halflifefan:
Oh wow, turns out I even NEEDED this setting in VS Code, that fixed all of my IntelliSense issues on ClangCL. Amazing!

I wish I knew how to fix clangd similarly as well..

hanniavalera:
> Oh wow, turns out I even NEEDED this setting in VS Code, that fixed all of my IntelliSense issues on ClangCL. Amazing!
> 
> I wish I knew how to fix clangd similarly as well..

Great! Happy to hear that :) I am approving this! 

Please feel free to file your issues with clangd so we can take a look at it, thanks!

PR #4830, opened 2026-03-19
Construct minimal regex when running unit tests in parallel

theuke:
### This changes behavior of how CTest generates regex for running parallel unit tests

- Minimize the regex used to indicate which unit tests are to be run when many unit tests are selected.
- Current behavior is to have a single "strict match" regex for every single unit test case which results in ctest crashing out due to the size of the aggregate regex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ Bug Fixes:
 - Fix diagnostics to handle when there isn't a command in the error output. [PR #4765](https://github.com/microsoft/vscode-cmake-tools/pull/4765)
 - Fix bug in which running "CMake: Build" would always run "CMake: Clean Rebuild" when `cmake.buildTask` is enabled [#4421](https://github.com/microsoft/vscode-cmake-tools/issues/4421) [@RedSkittleFox](https://github.com/RedSkittleFox)
 - Fix issue with hover provider not checking for undefined. [#4812](https://github.com/microsoft/vscode-cmake-tools/issues/4812)
+- Fix bug in which CTest is unable to run large amount of tests in parallel due to regex exceeding command line length limits [#4829](https://github.com/microsoft/vscode-cmake-tools/issues/4829) [@theuke](https://github.com/theuke)
 
 ## 1.22.28
 
--- a/src/ctest.ts
+++ b/src/ctest.ts
@@ -166,6 +166,96 @@ export function searchOutputForFailures(patterns: FailurePatternsConfig, output:
     return messages;
 }
 
+/**
+ * Finds the minimal regex fragments to match targets from the superset.
+ * If a target is a prefix of a forbidden string, it returns an exact match.
+ * Otherwise, it returns a minimal prefix.
+ *
+ * @param superset : the list of all individual test names (individual and or group names)
+ * @param targets : the list of test names (individual and or group names) we want to match.
+ * @returns : the list of regex fragments to match the targets without matching any other test from the superset
+ */
+export function getMinimalRegexFragments(superset: string[], targets: string[]): string[] {
+    interface TrieNode {
+        children: Map<string, TrieNode>;
+        forbiddenCount: number;
+    }
+
+    const targetSet = new Set(targets);
+    const forbidden = superset.filter(s => !targetSet.has(s));
+
+    if (targets.length === 0) {
+        return [];
+    }
+    if (forbidden.length === 0) {
+        return ["^."];
+    }
+
+    const root: TrieNode = { children: new Map(), forbiddenCount: 0 };
+
+    // 1. Build Trie with forbidden strings
+    for (const s of forbidden) {
+        let current = root;
+        for (const char of s) {
+            if (!current.children.has(char)) {
+                current.children.set(char, { children: new Map(), forbiddenCount: 0 });
+            }
+            current = current.children.get(char)!;
+            current.forbiddenCount++;
+        }
+    }
+
+    const fragmentSet = new Set<string>();
+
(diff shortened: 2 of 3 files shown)


theuke:

hanniavalera:
Hi @theuke, thank you for re-flagging the issue and submitting a PR for it! 

Could you please update the branch? I will take a look at this today!

theuke:
@hanniavalera branch has been re-based. Thanks for getting to this so quickly!

Also, one thing that is not explicitly covered in the unit tests is the fact that this approach continues to work correctly even when "discover tests" is used (i.e. as in something like gtest discover_tests). Namely

1. If I hit "run" on `TestSuiteA`, the vscode testing panel selects all of the tests under `TestSuiteA` at the time "run" was pressed.
2. CMake tool extension then "rebuilds" the project as neccessary (e.g. because new tests were added / removed / changed).
3. The VSCode test panel then passes in the initial "target" tests list and the UPDATED "all tests" into the cmake extension.
4. This results in the minimal regex being constructed such that ONLY the unit tests that were selected get run. I.e. it doesn't sneakily end up running the "new" unit tests under `TestSuiteA` that were discovered in step 2.
    - I call this out because this is the existing behavior (prior to my change) and my change does NOT change this behavior.

hanniavalera:
> @hanniavalera branch has been re-based. Thanks for getting to this so quickly!
> 
> Also, one thing that is not explicitly covered in the unit tests is the fact that this approach continues to work correctly even when "discover tests" is used (i.e. as in something like gtest discover_tests). Namely
> 
> 1. If I hit "run" on `TestSuiteA`, the vscode testing panel selects all of the tests under `TestSuiteA` at the time "run" was pressed.
> 2. CMake tool extension then "rebuilds" the project as neccessary (e.g. because new tests were added / removed / changed).
> 3. The VSCode test panel then passes in the initial "target" tests list and the UPDATED "all tests" into the cmake extension.
> 4. This results in the minimal regex being constructed such that ONLY the unit tests that were selected get run. I.e. it doesn't sneakily end up running the "new" unit tests under `TestSuiteA` that were discovered in step 2.
>    
>    * I call this out because this is the existing behavior (prior to my change) and my change does NOT change this behavior.

@theuke Good callout :) the behavior is preserved correctly by design since `getMinimalRegexFragments` takes the updated superset alongside the original targets, so any newly discovered tests end up in the forbidden set and get excluded automatically.

That said, it'd be worth adding a small unit test to lock this in explicitly, such as:

```
test('newly discovered tests (e.g. via gtest_discover_tests) are not accidentally included', () => {
    const superset = ['Suite.Test1', 'Suite.Test2', 'Suite.TestNew']; // TestNew added after build
    const targets  = ['Suite.Test1', 'Suite.Test2'];                  // original user selection
    const result   = getMinimalRegexFragments(superset, targets);
    expect(result).to.deep.eq(['^Suite\\.Test1', '^Suite\\.Test2']);
});
```

Keeps the guarantee self-documenting and regression-proof going forward. Either way, if you do not have the bandwidth, I can throw this in right after. Let me know!

theuke:
@hanniavalera So the scenario you listed it is already covered by two of the unit test cases. So the regex behavior is covered. What is technically not covered by unit tests is ensuring that `getMinimalRegexFragments()` is always passed in the up to date "all tests" list. That would have to be integ tested somehow and I don't think is worth it since that would just be us testing the expected behavior of the VSCode test panel.

Case 1:
```ts
        test('handles regex special characters properly', () => {
            const superset = ['Test[A]', 'Test[B]'];
            const targets = ['Test[A]'];
            // Forbidden is ['Test[B]']
            // 'Test[' is shared. 'Test[A' is unique.
            const result = getMinimalRegexFragments(superset, targets);
            expect(result).to.deep.eq(['^Test\\[A']);
        });
```

Case 2:
```ts
        test('complex edge cases: nested suites, swallowed prefixes, special characters', () => {
            const superset = [
                'Suite',            // Target: gets swallowed by prefix sharing with forbidden
                'Suite.Test1',      // Target: logically nested
                'Suite.Test2',      // Target: logically nested
                'Suite.Test3',      // Forbidden
                'OtherSuite.Test1', // Target
                'OtherSuite.Test2', // Target
                'O[ther]',          // Forbidden: share 'O' prefix
                'A+B.Test',         // Target: gets swallowed + special chars
                'A+B.Test2'         // Forbidden
            ];

            const targets = [
                'Suite',
                'Suite.Test1',
                'Suite.Test2',
                'OtherSuite.Test1',
                'OtherSuite.Test2',
                'A+B.Test'
            ];

            const result = getMinimalRegexFragments(superset, targets);

            expect(result).to.have.members([
                '^Suite$',
                '^Suite\\.Test1',
                '^Suite\\.Test2',
                '^Ot',
                '^A\\+B\\.Test$'
            ]);
            expect(result.length).to.eq(5);
        });
    });
```

hanniavalera:
@theuke Fair point, you're right :) the exclusion behavior is already demonstrated in both of those cases. And agreed on the integration side, that's testing VS Code's contract rather than ours. Thanks for clarifying. LGTM from my end, I'll get this merged! Thank you so much for your contributions to CMake Tools!

theuke:
Nice! Thanks for the quick turn around time.

Also, I think this PR can / should be linked to issue #4830 and therefore can close out that issue.

theuke:
@hanniavalera Also, just for my knowledge, which version of the extension will have this change? I take it that should be any version >= 1.22.29? Or will it be in 1.23.X since that seems to be the current pre-release version? Lastly, when is this change expected to make its way into the pre-release version?

PR #4846, opened 2026-03-27
Add Support for CMake Presets Version 11

hanniavalera:
### Summary

Adds support for [CMake Presets version 11](https://cmake.org/cmake/help/v4.3/manual/cmake-presets.7.html), introduced in CMake 4.3. The only schema change from v10 is that `execution.jobs` in test presets can now be an empty string, equivalent to passing `--parallel` with no value (letting CTest auto-detect the job count).

### Changes

**Schema & validation**
- New CMakePresets-v11-schema.json - copy of the v10 schema with `testPresets.execution.jobs` widened to accept either an integer or an empty string (`anyOf [integer, string with maxLength: 0]`).
- presetsParser.ts - bumped `maxSupportedVersion` from 10 → 11; added explicit v10 routing so the v11+ fallthrough uses the new schema.
- package.json - updated `jsonValidation` entries for `CMakePresets.json` and `CMakeUserPresets.json` to reference the v11 schema.

**Runtime behavior**
- preset.ts - `testArgs()`: when `execution.jobs` is an empty string, pushes only `--parallel` (no value). When it's a number, behavior is unchanged (`--parallel <N>`).

**Tests**
- validation.test.ts - added two new test cases:
  - v10: validates that `graphviz` (added in v10) is accepted on v10 and rejected on v9.
  - v11: validates that an empty-string `execution.jobs` is accepted on v11 and rejected on v10.

**Docs**
- cmake-presets.md - updated supported version range to "version 2 through 11".
- CHANGELOG.md - added entry under Features.

### Testing

- `yarn run compile` - clean build, no errors.
- `npx tsc -p test.tsconfig.json --noEmit` - test project type-checks cleanly.
- New validation tests cover both the accept and reject paths for v10 and v11 schema changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4476,7 +4476,7 @@
     "tsconfig-paths": "^3.11.0",
     "tslint": "^6.1.3",
     "typescript": "^4.1.5",
-    "vscode-cmake-tools": "^1.5.0",
+    "vscode-cmake-tools": "^1.6.0",
     "vscode-nls-dev": "^3.3.2",
     "webpack": "^5.104.1",
     "webpack-cli": "^5.1.4"
--- a/yarn.lock
+++ b/yarn.lock
@@ -7843,10 +7843,10 @@ vinyl@^3.0.0, vinyl@^3.0.1:
     replace-ext "^2.0.0"
     teex "^1.0.1"
 
-vscode-cmake-tools@^1.5.0:
-  version "1.5.0"
-  resolved "https://pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages/npm/registry/vscode-cmake-tools/-/vscode-cmake-tools-1.5.0.tgz#208d15ab4364861f7bad17282bb4f1d47539248a"
-  integrity sha1-II0Vq0Nkhh97rRcoK7Tx1HU5JIo=
+vscode-cmake-tools@^1.6.0:
+  version "1.6.0"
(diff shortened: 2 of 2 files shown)


src/presets/preset.ts
```diff
@@ -256,7 +256,15 @@ export interface TestPresetPrivate {
     __generator?: string; // Getting this from the config preset
 }
 
-export interface TestPreset extends api.TestPreset, TestPresetPrivate, PresetPrivate {}
+// v11+: execution.jobs can be an empty string (--parallel with no value).
+// Widen the type from the API's number to number | string.
+export interface TestExecutionOptions extends Omit<api.ExecutionOptions, 'jobs'> {
```
gcampbell-msft: We shouldn't have to do this, though it was a good idea while testing, we should instead plan to approve and merge the new version of the API and then ingest it here before merging this PR

src/api.ts
```diff
@@ -98,7 +98,9 @@ class CMakeProjectWrapper implements api.Project {
 
     get testPreset() {
         logApiTelemetry('getTestPreset');
-        return this.project.testPreset ?? undefined;
+        // Cast: internal TestPreset widens execution.jobs to number | string (v11+),
+        // but the external API still exposes the narrower number type.
+        return this.project.testPreset as api.TestPreset | null ?? undefined;
```
gcampbell-msft: Another place we can update once we ship the new version of the API

src/presets/preset.ts
hanniavalera: Sounds good!

PR #4873, opened 2026-04-09
Add cmake.languageServerOnlyMode setting for language services without project integration

### LINKED ISSUE #4516: [Feature] Language Server Only Mode
### Request Overview

The  CMake Tools extension provides a language server for cmake files. However, it forces this functionality to be coupled with integration of the cmake executable.  This is undesirable for containerized development where the cmake executable is not present in the same environment as vscode.
While the language server is functional without cmake installed, the CMake Tools extension has undesirable behavior.  When saving any file in a project containing any cmake files, the extension will do one of the following:
* Pop-up an executable selection dialog, which will take focus from the current file.
* Pop-up an error toast that cannot be muted because it has "error" severity
Given how frequently saving a file occurs, these behaviors are very intrusive.

This feature request is to provide an option for disabling cmake build/test integration and provide only the language server functionality for syntax highlighting on environments with no cmake executable.

### Additional Information

_No response_

Omotola:
## Summary
Adds a new `cmake.languageServerOnlyMode` setting that enables CMake language services (hover, completion) while disabling all project integration (kit selection, configure, build, test). This is useful for users who want CMake language support but don't have CMake installed locally.

## Related Issue
Closes #4516

## Changes
- **Setting**: Added `cmake.languageServerOnlyMode` (boolean, defaults to `false`, machine scope)
- **Config**: Updated `src/config.ts` with new setting, getter, and EventEmitter
- **Extension lifecycle**: Early return in `postWorkspaceOpen()` to skip kit scanning and configure-on-open
- **Project integration**: Suppressed save-triggered reconfigure, bad-executable toasts, and driver initialization when enabled
- **UI**: Gates `cmake:enableFullFeatureSet` context key to hide build/test/kit commands and views
- **Documentation**: Added setting to `cmake-settings.md` and CHANGELOG

## Testing
Tested Locally, confirmed that with the new setting on and a fake cmake path, there's no auto-configure when saving cmakefiles, and during manual configure output shows a message indicating language server only mode is enabled.
image

## Notes
- The language services implementation was already fully decoupled from the CMake executable, so no changes to language service code were needed
- The setting requires manual user configuration; it does not auto-detect missing CMake


src/cmakeProject.ts
```diff
@@ -843,15 +843,27 @@ export class CMakeProject {
...
         const cmakeInfo = await this.getCMakeExecutable();
         if (!cmakeInfo.isPresent) {
-            // Do not show a popup here to avoid duplicate "Bad CMake executable" messages.
-            // The canonical user-facing error is shown when a command actually needs a driver
-            // and getCMakeDriverInstance() validates the executable.
-            telemetry.logEvent('CMakeExecutableNotFound');
+            if (!this.workspaceContext.config.languageServerOnlyMode) {
+                void vscode.window.showErrorMessage(localize('bad.executable', 'Bad CMake executable: {0}. Check to make sure it is installed or the value of the {1} setting contains the correct path', `"${cmakeInfo.path}"`, '"cmake.cmakePath"'));
+                telemetry.logEvent('CMakeExecutableNotFound');
+            } else {
```
hanniavalera, line 848: We intentionally removed the error message from this handler because the same error is already shown to the user inside getCMakeDriverInstance() when a command actually needs the driver. This PR re-introduces that error message for users who aren't in language-server-only mode, which means they'd see the same "Bad CMake executable" popup twice. I think the fix is to preserve the base branch's behavior here and only add the language-server-only suppression of the telemetry event, but I could be wrong, let me know what you find out!

```
if (!cmakeInfo.isPresent) {
    if (!this.workspaceContext.config.languageServerOnlyMode) {
        telemetry.logEvent('CMakeExecutableNotFound');
    } else {
        log.debug(localize('bad.executable.suppressed.language.server.only', '...'));
    }
}
```

src/cmakeProject.ts
```diff
@@ -1491,15 +1504,23 @@ export class CMakeProject {
...
             if (!this.useCMakePresets && !this.activeKit) {
                 log.debug(localize('not.starting.no.kits', 'Not starting CMake driver: no kit selected'));
                 return null;
             }
 
             const cmake = await this.getCMakeExecutable();
             if (!cmake.isPresent) {
-                void vscode.window.showErrorMessage(localize('bad.executable', 'Bad CMake executable: {0}. Check to make sure it is installed or the value of the {1} setting contains the correct path', `"${cmake.path}"`, '"cmake.cmakePath"'));
-                telemetry.logEvent('CMakeExecutableNotFound');
+                if (!this.workspaceContext.config.languageServerOnlyMode) {
```
hanniavalera: The method already returns null at the top when `languageServerOnlyMode` is enabled, so the second `languageServerOnlyMode` check further down inside the `!cmake.isPresent` branch can never actually execute. I'd remove it to keep things clear.

src/extension.ts
```diff
@@ -711,6 +714,10 @@ export class ExtensionManager implements vscode.Disposable {
         if (!project) {
             return;
         }
+        if (project.workspaceContext.config.languageServerOnlyMode) {
+            log.debug('Skipping CMake project integration during workspace open because language-server-only mode is enabled.');
```
hanniavalera, line 718: The `log.debug(...)` call passes a plain English string instead of using localize(...). The rest of the file consistently uses `localize()` for log messages, which is our convention.

hanniavalera (changes requested):
The config fixture update is good, but it would be worth adding at least a basic test that confirms getCMakeDriverInstance() returns null when the mode is on, and that doCMakeFileChangeReconfigure skips reconfiguration if possible :)

src/extension.ts
Omotola: I actually got a build error on this which is why localize is omitted: 

Image

The build was failing because localize() requires corresponding entries in all the i18n translation bundles.

hanniavalera: makes sense, all good then!

hanniavalera (approved):

PR #4891, opened 2026-04-16
Update test-electron to fix offline testing

cwalther:
### This changes the developer experience (but not the user experience)

- Update test-electron to fix offline testing

## The purpose of this change

Allow testing of CMake Tools when Microsoft servers are being uncooperative.

## Other Notes/Information

@vscode/test-electron 2.3.8 has a bug that causes tests to error out when https://update.code.visualstudio.com/api/releases/stable returns "429.608 TooManyRequests" (https://github.com/microsoft/vscode-test/issues/285, fixed in version 2.5.1). Update it to the latest version (2.5.2) to fix that.

Output before:

```
yarn run v1.22.22
warning cmake-tools@1.13.0: The engine "vscode" appears to be invalid.
$ yarn run pretest && node ./out/test/unit-tests/runTest.js
warning cmake-tools@1.13.0: The engine "vscode" appears to be invalid.
$ tsc -p test.tsconfig.json
Failed to parse response from https://update.code.visualstudio.com/api/releases/stable as JSON
Failed to get JSON
Failed to run tests
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Output after:

```
yarn run v1.22.22
warning cmake-tools@1.13.0: The engine "vscode" appears to be invalid.
$ yarn run pretest && node ./out/test/unit-tests/runTest.js
warning cmake-tools@1.13.0: The engine "vscode" appears to be invalid.
$ tsc -p test.tsconfig.json
⠼ Resolving version...Failed to parse response from https://update.code.visualstudio.com/api/releases/stable?released=true as JSON
Error retrieving VS Code versions, using already-installed version 1.116.0 Failed to get JSON
✔ Validated version: 1.116.0
✔ Found existing install in /home/cwalther/projects/cmake-tools/vscode-cmake-tools/.vscode-test/vscode-linux-x64-1.116.0
…
(test proceeds successfully)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # What's New?
 
+## 1.24
+
+Bug Fixes:
+- Update testing framework to fix bugs when running tests of CMake Tools without a reliable internet connection. [#4891](https://github.com/microsoft/vscode-cmake-tools/pull/4891) [@cwalther](https://github.com/cwalther)
+
 ## 1.23
 
 Features:
--- a/package.json
+++ b/package.json
@@ -4468,7 +4468,7 @@
     "@typescript-eslint/eslint-plugin": "^5.59.11",
     "@typescript-eslint/eslint-plugin-tslint": "^5.59.11",
     "@typescript-eslint/parser": "^5.59.11",
-    "@vscode/test-electron": "^2.3.8",
+    "@vscode/test-electron": "^2.5.2",
     "@vscode/vsce": "^3.1.0",
     "chai": "^4.3.0",
     "chai-as-promised": "^7.1.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -552,11 +552,6 @@
   resolved "https://pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages/npm/registry/@sinonjs/text-encoding/-/text-encoding-0.7.1.tgz"
   integrity sha512-+iTbntw2IZPb/anVDbypzfQa+ay64MW0Zo8aJ8gZPWMMK6/OubMVb6lUPMagqjOPnmtauXnFCACVl3O7ogjeqQ==
 
-"@tootallnate/once@1":
-  version "1.1.2"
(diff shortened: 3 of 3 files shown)


snehara99:
Hi @cwalther, thanks for your contribution! Could you agree to the Contributor's Agreement so I can review this? Thanks!

cwalther:
I already did that in my previous PR #4620 and I’m not sure how I would do it again. In fact, the “license/cla” CI check seems to agree. Can you clarify?

snehara99:
> I already did that in my previous PR #4620 and I’m not sure how I would do it again. In fact, the “license/cla” CI check seems to agree. Can you clarify?

Although the license/cla check is passed, we still check that it was agreed to manually. We need this for every PR. You can agree by commenting something like this:

```
company="<insert company name>"
```

cwalther:

cwalther:
Hmm, if that version somehow has problems on macOS and Windows, then maybe scrap this. I only tested on Linux. I guess I could try 2.5.1 instead. I‘ll take a closer look next week.

snehara99:
> Hmm, if that version somehow has problems on macOS and Windows, then maybe scrap this. I only tested on Linux. I guess I could try 2.5.1 instead. I‘ll take a closer look next week.

seems like it was just a timeout issue. a rerun allowed it to pass. I'll still validate locally before merging

snehara99:
@cwalther could you add a changelog entry? Even if it isn't user-facing, we like to include bug fixes

cwalther:
OK, now with a changelog entry. I hadn’t added one because it didn’t seem like in the past pure devDependencies updates got any.

I am guessing that the proper section for the changelog entry is 1.24, because as far as I understand, version 1.23.51 has now been released as the final 1.23 release, and there is a new pre-release 1.24.0 – but I may still be misunderstanding the versioning process.

snehara99:
@cwalther you added the changelog entry correctly, thank you! This will be merged in shortly

snehara99 (approved):
validated locally

cwalther:
Thank you for the quick turnaround!

snehara99:
no problem, thank you for contributing!

PR #4892, opened 2026-04-16
Enhance generator mismatch detection and clean-up logic

### LINKED ISSUE #4890: [REGRESSION] Ninja generator no longer being used after update
### Brief Issue Summary

Cmake Tools no longer detects Ninja generator after update.

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.116.0",
  "cmtVersion": "1.23.51",
  "configurations": [
    {
      "folder": "NO",
      "cmakeVersion": "4.2.3",
      "configured": true,
      "generator": "Visual Studio 18 2026",
      "usesPresets": false,
      "compilers": {}
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
[cmake] -- Selecting Windows SDK version 10.0.26100.0 to target Windows 10.0.26200.
[cmake] CMake Error: Error: generator : Visual Studio 18 2026
[cmake] Does not match the generator used previously: Ninja
[cmake] Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
[cmake] 
[cmake] -- Configuring incomplete, errors occurred!
```

### Additional Information

My build tools should not change their own configuration after an update.

hanniavalera:
This pull request addresses a regression affecting Visual Studio kits with Ninja-based build caches, enhances generator selection and fallback logic for Visual Studio kits, and introduces improved detection and handling of generator mismatches to prevent build failures. It also updates documentation to clarify generator selection behavior and adds targeted unit tests for generator mismatch detection.

**Bug fixes and generator selection improvements:**

* Fixed a regression where Visual Studio kits with an existing Ninja-based build cache would fail after updating CMake Tools due to a generator mismatch. Ninja is now again preferred when available, and for stale VS kits missing `preferredGenerator`, the correct VS generator is derived at runtime as a fallback. The extension now auto-cleans the build directory when a generator mismatch with `CMakeCache.txt` is detected, including during driver reloads triggered by generator-related setting changes. [[1]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR12-R16) [[2]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fR637-R656) [[3]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fL802-R835) [[4]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fR1646-R1659)

* Improved generator selection logic for Visual Studio kits:
  - Kits scanned for known VS versions now get a `preferredGenerator` at scan time.
  - If a kit was scanned before support for a VS version, the correct generator is derived at runtime and only used after Ninja/Unix Makefiles fallbacks.
  - Documentation and code comments clarify the generator precedence order and fallback behavior. [[1]](diffhunk://#diff-4675bf8026fca3f7fe2f6ce3a17ad4502c73a0783404fa3db29c0813f7d437c2L9-R14) [[2]](diffhunk://#diff-4675bf8026fca3f7fe2f6ce3a17ad4502c73a0783404fa3db29c0813f7d437c2L55-R56) [[3]](diffhunk://#diff-4675bf8026fca3f7fe2f6ce3a17ad4502c73a0783404fa3db29c0813f7d437c2L150-R159) [[4]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fL814-R849) [[5]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fR863-R872) [[6]](diffhunk://#diff-e57cc44808ddb61cb47246af909ead60f99221e5b15674442f1cf6970401f6ffL836-R837) [[7]](diffhunk://#diff-e57cc44808ddb61cb47246af909ead60f99221e5b15674442f1cf6970401f6ffL879-R882)

**Testing and reliability:**

* Added a pure helper function `generatorMismatch` to detect generator mismatches between the desired generator and the one cached in `CMakeCache.txt`, and exported it for direct unit testing. [[1]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fR151-R162) [[2]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fL35-R35)
* Added comprehensive unit tests for generator mismatch detection, covering a variety of scenarios including cache parsing and different generator combinations.

**Documentation updates:**

* Updated `docs/kits.md` to clarify how Visual Studio kits handle generator selection, when and how `preferredGenerator` is set or derived, and the order in which generators are considered. Also documented the new safety net for cleaning the build directory on generator mismatches. [[1]](diffhunk://#diff-4675bf8026fca3f7fe2f6ce3a17ad4502c73a0783404fa3db29c0813f7d437c2L9-R14) [[2]](diffhunk://#diff-4675bf8026fca3f7fe2f6ce3a17ad4502c73a0783404fa3db29c0813f7d437c2L55-R56) [[3]](diffhunk://#diff-4675bf8026fca3f7fe2f6ce3a17ad4502c73a0783404fa3db29c0813f7d437c2L150-R159)


src/drivers/cmakeDriver.ts
```diff
@@ -802,6 +822,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
             if (needsCleanIfKitChange && (newBinaryDir === oldBinaryDir)) {
                 await this._cleanPriorConfiguration();
             }
+            await this._cleanIfGeneratorChanged(this._generator?.name);
```
gcampbell-msft: Could this result in two back to back cleans? Could we combine with the check above?

src/drivers/cmakeDriver.ts
```diff
@@ -812,14 +833,8 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
         // Place a kit preferred generator at the front of the list
-        // For VS kits that don't have preferredGenerator (e.g., scanned before a VS version was added),
-        // try to derive it from the VS installation.
-        let kitPreferredGenerator = kit.preferredGenerator;
-        if (!kitPreferredGenerator && kit.visualStudio) {
-            kitPreferredGenerator = await getVsKitPreferredGenerator(kit);
-        }
-        if (kitPreferredGenerator) {
-            preferredGenerators.unshift(kitPreferredGenerator);
+        if (kit.preferredGenerator) {
```
gcampbell-msft, line 836: Does this simply upstream the code that was removed into the setup of the kits themselves?

gcampbell-msft: We need to make sure we don't lose that behavior

gcampbell-msft: Based on this, it seems like we're removing the behavior that we're _adding_ to the documentation explaining that we'll set the preferred generator to the VS generator. Just want to make sure this still is present. Is it happening somewhere else?

src/drivers/cmakeDriver.ts
hanniavalera: You're right, will do!

src/drivers/cmakeDriver.ts
hanniavalera: The runtime derivation is still present, just moved. Here's what changed:

**Before (what you reviewed):** The derivation was removed entirely, and `_setKit` only checked `kit.preferredGenerator` directly. That did lose the behavior.

**After (force-pushed and I iterated on the fix):** `getVsKitPreferredGenerator` is back in kit.ts (with platform/toolset bugs fixed to match scan-time logic). The key difference from the original code is where the derived generator lands in the priority list:

```typescript
// 1. Kit's preferredGenerator → front of list (scan-time VS kits win here)
if (kit.preferredGenerator) {
    preferredGenerators.unshift(kit.preferredGenerator);
}

// 2. Defaults if nothing set
preferredGenerators.push({ name: "Ninja" });
preferredGenerators.push({ name: "Unix Makefiles" });

// 3. Derived VS generator → END of list (last-resort fallback)
if (!kit.preferredGenerator && kit.visualStudio) {
    const derived = await getVsKitPreferredGenerator(kit);
    if (derived) {
        preferredGenerators.push(derived);  // push, not unshift
    }
}
```

The original code used `unshift` (front), which made the derived VS generator beat Ninja, that's what caused #4890. Now it's `push` (end), so Ninja is still tried first (restoring documented behavior), but the VS generator is available as a fallback when Ninja isn't on PATH. I also combined the two clean checks into one as you suggested.

test/unit-tests/kit-scan.test.ts
```diff
@@ -395,6 +395,133 @@ suite('Kits scan test', () => {
...
+            };
+            expect(kit.kitChangeNeedsClean(newKit, null)).to.be.false;
+        });
+
+        test('Returns true when preferredGenerator changes from undefined to VS generator', () => {
+            const oldKit: kit.Kit = {
+                name: 'Visual Studio Community 2026 Release - amd64',
+                visualStudio: 'VisualStudio.18.0',
+                visualStudioArchitecture: 'x64',
+                isTrusted: true
```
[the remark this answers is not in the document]
hanniavalera: Not required

gcampbell-msft (approved):

PR #4900, opened 2026-04-22
Update paths.test.ts

Pgns-bsb:

- 
- 
-


hanniavalera:
Hi @Pgns-bsb, thanks for the contribution! Could you provide some context on this change? What error or issue were you encountering that prompted adding the /// <reference types="node" /> directive? Also, was there a reason you targeted the copilot/allow-path-alterations-pre-cmake-search branch instead of main?

PR #4916, opened 2026-04-29
Add cmake.launchConfig setting for customizable run configurations

### LINKED ISSUE #4904: [Feature] Make "Run without debuggnig" configurable (for embedded dev)
### Request Overview

The "Run without debugging" action just tries to execute the build artifact on host. 

This makes it useless for embedded MCU development where the compilation artifact is a firmware file intended to be run on a microcontroller. To be run it needs to be written to MCU in a vendor-specific way.

Please make this action configurable, so that it can invoke a programmer to flash the firmware onto MCU. This is already working for debugging, where there is a "cmake.debugConfig" option that uses whatever debugging configuration the developer has configured. 
Don't know what the ideal way to implement this would be, right now I use a custom task (defined in tasks.json) that invokes the programmer.

### Additional Information

_No response_

hanniavalera:
This pull request introduces a major new feature to VS Code CMake Tools: the ability to customize the "Run Without Debugging" command via a new `cmake.launchConfig` setting. This allows users to delegate launching to a VS Code task or to run a custom program (such as a firmware flashing tool), with support for argument, environment, and working directory customization. The implementation includes schema, documentation, and UI/logic changes, as well as telemetry and user guidance for the new behavior. Suggested by @positron96 and addresses #4904 

Key changes include:

**New Feature: Customizable Run Without Debugging**

* Added the `cmake.launchConfig` setting, allowing users to configure "Run Without Debugging" to either delegate to a VS Code task (`task`) or run a custom program (`program`). This supports variable substitution and custom arguments, environment, and working directory. [[1]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR8) [[2]](diffhunk://#diff-f0cc20103c5b45ee9ed9da1f1983848bf1417993b2fb7f90c2d22225d2955491R48) [[3]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R2946-R2999) [[4]](diffhunk://#diff-9b29d8b47a74097a04da353b72a8632d485d658282457b97b4caf26575a42dd0R253-R269)

**Core Implementation and Lifecycle Management**

* Implemented logic in `CMakeProject` to handle the new launch modes, including running a VS Code task or a custom program instead of the default executable, with proper terminal/task lifecycle and cleanup. [[1]](diffhunk://#diff-3f6cd275fa3ae395afd1373e06642826f95dc0f3af2f23f3d16dfd412297125fR45) [[2]](diffhunk://#diff-3f6cd275fa3ae395afd1373e06642826f95dc0f3af2f23f3d16dfd412297125fR912-R915) [[3]](diffhunk://#diff-3f6cd275fa3ae395afd1373e06642826f95dc0f3af2f23f3d16dfd412297125fR3409) [[4]](diffhunk://#diff-3f6cd275fa3ae395afd1373e06642826f95dc0f3af2f23f3d16dfd412297125fR3419-R3427) [[5]](diffhunk://#diff-3f6cd275fa3ae395afd1373e06642826f95dc0f3af2f23f3d16dfd412297125fR3478-R3493)
* Added telemetry event logging to track usage of the new launch modes.

**Configuration and API Updates**

* Extended the configuration schema and settings interface to support the new `launchConfig` property and expose it through the configuration reader. [[1]](diffhunk://#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R14) [[2]](diffhunk://#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R212) [[3]](diffhunk://#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R474-R476)

**User Guidance and Documentation**

* Updated documentation and settings descriptions, including warnings when both `cmake.launchConfig` and `cmake.debugConfig` fields are set (as the latter are ignored for the launch path), and improved error/warning messages for ambiguous or missing tasks. [[1]](diffhunk://#diff-9b29d8b47a74097a04da353b72a8632d485d658282457b97b4caf26575a42dd0R253-R269) [[2]](diffhunk://#diff-3f6cd275fa3ae395afd1373e06642826f95dc0f3af2f23f3d16dfd412297125fR3478-R3493)

These changes provide advanced users and embedded developers with much more flexibility for customizing their build/run workflows directly from CMake Tools.


gcampbell-msft:
With any new feature, let's be verbose in the description of the PR and include screenshots of the feature. While I plan to checkout and try it out, it makes reviewing a little bit simpler when new UI is added.

src/ui/projectStatus.ts
```diff
@@ -786,6 +789,10 @@ class LaunchNode extends Node {
         return [this.launchTarget!];
     }
 
+    dispose(): void {
+        this.launchTarget?.dispose();
```
gcampbell-msft, line 793: Curious on what about this needs to be disposed?

src/ui/projectStatus.ts
```diff
@@ -1092,13 +1101,39 @@ class LaunchTarget extends Node {
...
         if (!treeDataProvider.cmakeProject) {
             return;
         }
         this.label = treeDataProvider.cmakeProject.launchTargetName || await treeDataProvider.cmakeProject.allTargetName;
+        this.updateLaunchConfigDescription();
+    }
+
+    private updateLaunchConfigDescription(): void {
+        const launchCfg = treeDataProvider.cmakeProject?.workspaceContext.config.launchConfig;
+        if (launchCfg?.task) {
```
gcampbell-msft, line 1124: Are task and program mutually exclusive? If so, we should make sure we document which we prefer

src/cmakeProject.ts
```diff
@@ -908,7 +909,10 @@ export class CMakeProject {
         log.debug(localize({key: 'disposing.extension', comment: ["'CMake Tools' shouldn't be localized"]}, 'Disposing CMake Tools extension'));
         this.disposeEmitter.fire();
         this.termCloseSub.dispose();
+        this.launchTaskEndSub.dispose();
         this.launchTerminals.forEach(term => term.dispose());
```
gcampbell-msft, line 913: Should we also clear the launchTerminals?

package.json
```diff
@@ -2943,6 +2943,60 @@
...
+                    "type": { "type": "string" }
+                  }
+                }
+              ]
+            },
+            "program": {
+              "type": "string",
+              "description": "%cmake-tools.configuration.cmake.launchConfig.program.description%"
+            },
+            "args": {
```
gcampbell-msft, line 2975: Do the args and following properties apply to both scenarios of task and program? I imagine that the task itself has these already, how is this handled?

gcampbell-msft (commented):
Partial review, more review to come 👍

hanniavalera:
> With any new feature, let's be verbose in the description of the PR and include screenshots of the feature. While I plan to checkout and try it out, it makes reviewing a little bit simpler when new UI is added.

@gcampbell-msft sounds fair, forgot to put the screenshots up, will do!

src/ui/projectStatus.ts
hanniavalera: From what I understood, it releases the `config.onChange('launchConfig', ...)` subscription registered in `initialize()` so the node stops reacting to setting changes after teardown, let me know if it makes sense!

src/ui/projectStatus.ts
hanniavalera: Yes they are! I added on to task in addition to the program description to make note of this!

src/cmakeProject.ts
hanniavalera: Thanks for calling this out! I now added `this.launchTerminals.clear()` after the dispose loop, matching the pattern used right next to it for `launchTaskExecutions`, does that make sense to you?

package.json
hanniavalera: They only apply to program mode as VS Code tasks already own those fields. I've tightened the descriptions to say "(ignored in task mode)" and added a defensive `log.warning` in `runLaunchAsTask` if a user sets them anyway, so the override is visible without being disruptive.

PR #4926, opened 2026-05-05
Fix GNU LD regex matching CMake status lines as diagnostics (#4910)

### LINKED ISSUE #4910: [Bug] Incorrect gnuld message displayed in Problems panel as diagnostics
### Brief Issue Summary

CMake Tools is creating diagnostics for informational build output lines like:
`-- Zephyr version: 4.3.0 (/path/to/zephyr), build: v4.3.0`

These appear in VS Code's Problems panel with:
- Owner: `cmake-build-diags`
- Severity: Information (8)
- Source: `cmake GNULD`

```
[{
	"resource": "/path/to/project/build/-- Zephyr version: 4.3.0 (/path/to/zephyr), build",
	"owner": "cmake-build-diags",
	"severity": 8,
	"message": "v4.3.0",
	"source": "cmake GNULD",
	"startLineNumber": 1,
	"startColumn": 1,
	"endLineNumber": 1,
	"endColumn": 1000,
	"origin": "extHost7"
}]
```

Users see "problems" in the Problems panel when there are no actual issues, making it harder to identify real build errors and warnings.

Root Cause:

Found the incorrect regex in `microsoft/vscode-cmake-tools/src/diagnostics/gnu-ld.ts`:

```
{
    // path/to/ld[.exe]: message (without trailing colon)
    regexPattern: /^(.*ld(?:\.exe)?):\s+(.+)(?<!:)\s*$/,
    matchTypes: [MatchType.Full, MatchType.File, MatchType.Message]
}
```

Expected behavior:

Informational build status messages should not appear in the Problems panel. Only actual warnings and errors should be shown as diagnostics.

Actual behavior:

Normal build output containing "ld:" or "build:" is incorrectly parsed as GNU linker diagnostics.

### CMake Tools Diagnostics

```shell
"os": "darwin",
  "vscodeVersion": "1.114.0",
  "cmtVersion": "1.24.4"
```

### Debug Log

```shell
-
```

### Additional Information

_No response_

Omotola:
## Fix GNU LD diagnostic regex matching CMake status lines (#4910)

  Problem

  The GNU linker diagnostic parser incorrectly matches informational CMake build output as linker diagnostics. For
  example, Zephyr build status lines like:

   -- Zephyr version: 4.3.0 (/path/to/zephyr), build: v4.3.0

  appear in the VS Code Problems panel as errors with source: cmake GNULD, because the regex .*ld in the pattern
  matches the ld suffix in build.

  Root Cause

  Five regex patterns in src/diagnostics/gnu-ld.ts use .*ld to match the linker binary path. This is too broad — it
  matches any word ending in ld (e.g., build, child, ould), not just actual linker paths like /usr/bin/ld or
  C:\mingw\bin\ld.exe.

  Fix

  Changed .*ld → (?:.*[/\\])?ld in all five patterns. This requires ld to either:

   - appear at the start of the line, or
   - be preceded by a path separator (/ or \)

  This ensures only real linker binary paths are matched while rejecting arbitrary words that happen to end in ld.

  Testing

  Added two new unit tests that reproduce the false positive from #4910:

   - Zephyr version line: -- Zephyr version:
    4.3.0 (/path/to/zephyr), build: v4.3.0
   - Generic CMake status line: -- Some tool version:
    1.0 (/path/to/tool), build: v1.0

  Both tests fail before the fix and pass after. All existing GNU LD diagnostic tests continue to pass.

  Changes

   - src/diagnostics/gnu-ld.ts — Tightened 5 regex patterns
   - test/unit-tests/diagnostics.test.ts — Added 2 negative test cases
   - CHANGELOG.md — Added bug fix entry

  Fixes #4910


hanniavalera:
hi! nice fix for the Zephyr case. one concern before merge: the new prefix `(?:.*[/\\])?ld` requires `ld` to be preceded by `/`, `\`, or start-of-line, which drops cross-compiler linkers invoked by bare name (common in Zephyr / ARM bare-metal / Yocto where the toolchain is on `PATH`):

```
arm-none-eabi-ld: cannot find -lfoo
aarch64-linux-gnu-ld: warning: cannot find entry symbol _start
```

These previously matched via `.*ld` and now silently disappear from the Problems panel. Affects [pattern 4](https://github.com/microsoft/vscode-cmake-tools/blob/d5c64cddf0ab4d7e44c840f86318d207728690ec/src/diagnostics/gnu-ld.ts#L24) and [pattern 5](https://github.com/microsoft/vscode-cmake-tools/blob/d5c64cddf0ab4d7e44c840f86318d207728690ec/src/diagnostics/gnu-ld.ts#L28), and also [1](https://github.com/microsoft/vscode-cmake-tools/blob/d5c64cddf0ab4d7e44c840f86318d207728690ec/src/diagnostics/gnu-ld.ts#L12)/[2](https://github.com/microsoft/vscode-cmake-tools/blob/d5c64cddf0ab4d7e44c840f86318d207728690ec/src/diagnostics/gnu-ld.ts#L16)/[3](https://github.com/microsoft/vscode-cmake-tools/blob/d5c64cddf0ab4d7e44c840f86318d207728690ec/src/diagnostics/gnu-ld.ts#L20).

suggested tweak: allow `-` alongside path separators, and constrain the prefix to a filename-like token so sentences can't match:

```js
(?:[^\s:]*[/\\\-])?ld(?:\.exe)?
```

this still rejects the Zephyr `build:` line while catching `arm-none-eabi-ld:` and friends.

also worth adding a positive test next to the new ones in [`diagnostics.test.ts`](https://github.com/microsoft/vscode-cmake-tools/blob/d5c64cddf0ab4d7e44c840f86318d207728690ec/test/unit-tests/diagnostics.test.ts#L636-L647) for `arm-none-eabi-ld:` so this can't regress again. the existing gnuld tests only cover `/path/to/ld[.exe]` style inputs, which is how this gap slipped through.

snehara99 (approved):

hanniavalera (approved):

PR #4930, opened 2026-05-05
Add onConfigureResult event to notify on configure success or failure

### LINKED ISSUE #4021: [Feature Request]: Get notified when cmake configure failed
### Brief Issue Summary

currently, If cmake-tools configure succeed. other extension get notified through function `onCodeModelChanged` in cmakeTools API. but if configure failed, other extension has nothing to do with it, the extension didn't know the configuration is failed. If cmake tools can notify other extension that configuration is failed in its api, other extension which depends on cmakeTools can do something.

### CMake Tools Diagnostics

_No response_

### Debug Log

_No response_

### Additional Information

_No response_

Omotola:
## Summary

Adds a new \onConfigureResult\ event to the CMake Tools API that fires after every configure attempt, regardless of success or failure. This allows dependent extensions to detect and react to configure failures.

## Problem

Currently, the only configure-related event exposed is \onCodeModelChanged\, which only fires on **successful** configuration. Extensions that depend on CMake Tools have no way to know when a configure has **failed** — they simply never get notified.

## Solution

Added a new \onConfigureResult\ event on \CMakeProject\ that fires with a \ConfigureResult\ payload (containing \exitCode\ and \
resultType\) after every configure attempt. The event is exposed through the public API wrapper (\CMakeProjectWrapper\).

The event fires from all configure exit paths:
- Cached configuration (success or failure)
- \NoCache\ early return
- Normal configure flow (success and failure)

## Testing

Added smoke tests for both scenarios:
- **badProject**: Asserts \onConfigureResult\ fires exactly once with a non-zero exit code on failure
- **goodProject**: Asserts \onConfigureResult\ fires exactly once with exit code 0 on success

## Note

The \vscode-cmake-tools-api\ package will need a corresponding type update to formally expose \onConfigureResult\ to external consumers via the typed interface.

Fixes #4021


snehara99:
does this need a changelog update?

snehara99 (approved):

PR #4932, opened 2026-05-06
Fix pretest-buildfakebin on Windows

cwalther:
### This changes the developer experience (but not the user experience)

- Add `--config` arguments to make `yarn pretest-buildfakebin` work with multi-config generators.
- Use it in CI.

## The purpose of this change

While working on https://github.com/microsoft/vscode-cmake-tools/pull/4627#issuecomment-4380175931, I noticed that the `yarn pretest-buildfakebin` command that I had recently introduced in #4620 did not work on Windows, because it wasn’t prepared for multi-config generators such as the _Visual Studio_ used by default on Windows (I had only tested on Linux, where I think single-config _Unix Makefiles_ is used).

The attached commit fixes that, and to make sure it stays fixed, changes the CI jobs to use that command instead of a dedicated CMake build action. That also reduces some duplication (because those jobs were where I copied the procedure from to begin with).

## Other Notes/Information

By the way, just out of curiosity, I tested this with different CMake versions at https://github.com/indel-ag/vscode-cmake-tools/actions?query=branch%3Abuildfakebin-test – the original 3.18.3/3.31.6 (of which the former skips my new tests from #4627), 3.31.12, and the latest 4.3.2. Just so you know that works in case you ever want to update these versions.

### DIFF
--- a/.github/workflows/ci-main-linux.yml
+++ b/.github/workflows/ci-main-linux.yml
@@ -40,11 +40,7 @@ jobs:
             ninjaVersion: 1.10.1
 
       - name: Build fake compilers for tests
-        uses: urkle/action-cmake-build@v1.0.0
-        with:
-          source-dir: ${{ github.workspace }}/test/fakeOutputGenerator
-          configure-options: -DCMAKE_INSTALL_PREFIX:STRING=${{ github.workspace }}/test/fakebin
-          install-build: true
+        run: yarn pretest-buildfakebin
 
       - name: Run backend tests
         run: yarn backendTests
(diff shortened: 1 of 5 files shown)


cwalther:

hanniavalera (approved):

cwalther:
Thanks for the quick turnaround!

I think I’m done for now, my itch is scratched, so I will stop spamming you with PRs now! 🙂

hanniavalera:
> Thanks for the quick turnaround!
> 
> I think I’m done for now, my itch is scratched, so I will stop spamming you with PRs now! 🙂

We are always happy to take in contributions, thank you so much for contributing to CMake Tools and your detailed work!

Issue #2449, opened 2022-03-22
Run program before configuring


jasonbeach:
### Discussed in https://github.com/microsoft/vscode-cmake-tools/discussions/2403

<div type='discussions-op-text'>

<sup>Originally posted by **jasonbeach** February 28, 2022</sup>
Is it possible to have the extension run an external program prior to configuring cmake?  The immediate use case is pulling in conan dependencies.  I currently have a build script that runs `conan install` before configuring cmake.  I can create a build task that runs the script, but that doesn't work with the cmake widgets down at the bottom of the vscode window. 

I assume I could add an custom target to cmake to run conan, but that seems tedious to do for multiple repos and I've tried hard to keep my cmake agnostic to conan. </div>

Feature request:
  It would be nice to be able to run a task (either a shell executable or I guess build task) before cmake configures. See above discussion.  It kind of works if you check the CMake:Build Task option and then set a default task, again as described in the discussion.  The only thing that's a little clumsy is that it runs if cmake hasn't run yet then this fails because the task (in my case `conan install`) really needs to be run before cmake configures not before the build. You can attempt the build (it will fail), then configure cmake, and then rerun the build (it succeeds this time).  I can add another task that runs cmake and make the default depend on that, and that works, but....it's very....circumlocutious.  It would be much better to just be able to configure a task to run to run before cmake configures.

lanyizi:
It would be really nice if CMake tools could let me launch Visual Studio's Developer Command prompt before configuring CMake.
Currently, in order to let CMake locate the required toolset, I have to either launch CMake executable inside the developer command prompt, or start entire VS Code from it, which is not very convenient.

starball5:
@lanyizi https://stackoverflow.com/a/73366471/11107541

PR #4960, opened 2026-06-03
Add setting "preConfigureTask"

erdemiru:
## This changes introduce a new feature

- Add a new setting "preConfigureTask"
  - If set, the task bearing the name set is called before CMake configure run.

## The purpose of this change

The purpose of this change is to enable users to set up tasks before CMake configuration, allowing for greater automation. This is particularly useful when external tools are required to generate source code and CMake files before the cmake configure stage. 

## Other Notes/Information

While pre-build and post-configure tasks are already supported, they are insufficient when additional tools are needed to generate CMake files. In my use case, toolchain flags, such as included paths, libraries and also list of generated source files are all based on input JSON or XML configuration files.  Using execute_process() or custom commands have many other drawbacks in such cases:
- file(GLOB ) has to used in configure time since the list of generated files are dynamic based on the xml input.
- when included cmake files are updated by the external tool, re-configure has to be triggered.
- ...


erdemiru:
@microsoft-github-policy-service agree company="QualCode"

package.nls.json
```diff
@@ -272,6 +272,7 @@
     "cmake-tools.configuration.cmake.emscriptenSearchDirs.description": "Directories where Emscripten may be installed.",
     "cmake-tools.configuration.cmake.mergedCompileCommands.description": "Recursively collect and merge all compile_commands.json found in the cmake.buildDirectory.",
     "cmake-tools.configuration.cmake.copyCompileCommands.description": "Copy compile_commands.json to this location after a successful configure.",
+    "cmake-tools.configuration.cmake.preConfigureTask.description": "If set, this named task will be executed before a successful CMake configure.",
```
jonahgraham: I don't think the `preConfigureTask` is conditional on the configure being successful.

```suggestion
    "cmake-tools.configuration.cmake.preConfigureTask.description": "If set, this named task will be executed before CMake configure.",
```

jonahgraham (approved):
FWIW I approve of this as I think it resolves #2449 too.

IIUC It leaves one small side-effect. If people have done the workaround suggested in https://github.com/microsoft/vscode-cmake-tools/issues/2449#issuecomment-1077833316 then if those users apply this change and leave pre run task on, and run the configure as a task, the pre configure task will run twice. If my understanding is correct, then the release notes may want to point this out.

package.nls.json
erdemiru: Thanks @jonahgraham,  I've updated the wording with your suggestions.

jonahgraham (approved):

hanniavalera:
Hi @erdemiru, thanks for the contribution, nicely scoped feature! :) And good call flagging the double-run caveat in the setting description (plus @jonahgraham's related #2449 note).

One thing I think needs fixing: `executePreConfigureTask()` awaits `vscode.tasks.executeTask(task)`, which only waits for the task to start, not finish. So configure can race ahead while the pre-configure task is still running, which breaks the core use case (such as, a task that generates files CMake needs before it configures). It should wait for the task to complete via `onDidEndTaskProcess` (like cmakeBuildRunner.ts does), and ideally abort configure on a non-zero exit.

Also, this currently runs for `ConfigureType.ShowCommandOnly`, which shouldn't trigger side effects, that should be guarded so it only runs for real configures.

erdemiru:
Hi @hanniavalera, thanks for highlighting the points! Now awaiting task completion via onDidEndTaskProcess instead of just executeTask(). Configure will abort on non-zero exit. Also, added a guard for ConfigureType.ShowCommandOnly. Let me know if it looks good!

hanniavalera:
Hi @erdemiru - thank you so much for taking in my feedback and once again, for your contribution! It all looks good to me, we deeply appreciate your time and efforts! Will be approving :)

hanniavalera (approved):

PR #4966, opened 2026-06-10
fix: prompt kit/preset selection for command substitutions in launch.json

OrkunTokdemir:

- Make `queryCMakeProject` call `ensureActiveConfigurePresetOrKit()` before executing the query, matching the existing guard in `runCMakeCommandForProject`.
- When `${command:cmake.launchTargetPath}` (or related substitution commands like `cmake.launchTargetDirectory`, `cmake.launchTargetFilename`, `cmake.launchTargetName`, `cmake.getLaunchTargetPath`, etc.) is used in `launch.json` and no kit or preset is active, the extension now shows the quick pick selection dialog instead of failing with an error.

## The purpose of this change

When building or running via the status bar buttons, the extension prompts users to select a kit or preset if none is active. However, when launching through `launch.json` using `${command:cmake.launchTargetPath}`, the same scenario produced an error because `queryCMakeProject` skipped the `ensureActiveConfigurePresetOrKit()` check that `runCMakeCommandForProject` already performs.

This change adds the same guard so that all command substitutions behave consistently with the build/configure commands.


hanniavalera (approved):

OrkunTokdemir:
@hanniavalera Thanks.

PR #4972, opened 2026-06-14
Fix broken README badges

Maxxxxior:
Replaced broken `shields.io` marketplace badges with working `vsmarketplacebadges.dev` endpoints.
Wrapped all badges in anchor tags to fix navigation issues and ensure they open the correct links.

---

### This changes visible behavior in `README.md`.

- Replaced broken `shields.io` marketplace badge URL's with reliable `vsmarketplacebadges.dev` endpoints.
- Wrapped all badges in anchor tags (`<a>`) to ensure they navigate to the correct URLs when clicked.
- Updated the 'Closed Issues' badge link.

---

## The purpose of this change

The previous badge implementation used `&link=` parameters inside the `src` attribute, which caused browser navigation issues (the image itself opened in a new tab instead of the target page). This change ensures proper redirection and fixes broken/outdate marketplace badge sources.


Maxxxxior:

PR #4982, opened 2026-06-18
Fix vsce verify step to resolve vsce-sign from module tree

hanniavalera:
This pull request updates the way the pipeline locates the `vsce-sign` binary during release and prerelease jobs. Instead of assuming a fixed global path, it now dynamically resolves the binary location based on Node.js module resolution, making the process more robust and layout-agnostic. This change ensures compatibility with different installation structures and prevents errors when the binary is not hoisted to the global root.

Key changes:

**Improved vsce-sign binary resolution:**

* Updated the PowerShell script in both `jobs/release/prerelease.yml` and `jobs/release/release.yml` to resolve the `vsce-sign` binary using Node.js module resolution, mirroring how `vsce publish` locates it, instead of relying on a fixed global path. This prevents failures in environments where the module layout differs from expectations. [[1]](diffhunk://#diff-7b7c10c627921eaeb34e884afad4dd4b7707c9ef1ff15f9451a7af483c9b27dbL69-R76) [[2]](diffhunk://#diff-00a52e9268325fd97bdcd98dd8eeff56e4891b2a83b90d593dfe7dbc14ff3b07L93-R100)
* Added error handling to check if `npm root -g` fails or returns an empty value, and to ensure the resolved `vsce-sign` path is valid before proceeding. [[1]](diffhunk://#diff-7b7c10c627921eaeb34e884afad4dd4b7707c9ef1ff15f9451a7af483c9b27dbL69-R76) [[2]](diffhunk://#diff-00a52e9268325fd97bdcd98dd8eeff56e4891b2a83b90d593dfe7dbc14ff3b07L93-R100)
* Added log output to confirm the resolved location of the `vsce-sign` binary for easier debugging. [[1]](diffhunk://#diff-7b7c10c627921eaeb34e884afad4dd4b7707c9ef1ff15f9451a7af483c9b27dbL69-R76) [[2]](diffhunk://#diff-00a52e9268325fd97bdcd98dd8eeff56e4891b2a83b90d593dfe7dbc14ff3b07L93-R100)

### DIFF
--- a/jobs/release/release.yml
+++ b/jobs/release/release.yml
@@ -90,8 +90,14 @@ extends:
             npm_config_registry: https://pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages/npm/registry/
         - powershell: |
             $root = npm root -g
-            $sign = Join-Path $root '@vscode\vsce-sign\bin\vsce-sign.exe'
-            if (-not (Test-Path $sign)) { Write-Host "##vso[task.logissue type=error]Missing vsce-sign binary: $sign"; exit 1 }
+            if ($LASTEXITCODE -ne 0 -or [string]::IsNullOrWhiteSpace($root)) { Write-Host "##vso[task.logissue type=error]npm root -g failed"; exit 1 }
+            $root = $root.Trim()
+            # Resolve the vsce-sign native binary exactly as `vsce publish` does: from @vscode/vsce's own module tree.
+            # In a global install, @vscode/vsce-sign is nested under @vscode/vsce/node_modules (not hoisted to the
+            # global root), so a top-level path check is wrong. This mirrors Node resolution and is layout-agnostic.
+            $sign = node -e "const p=require('path'),fs=require('fs');const vsce=require.resolve('@vscode/vsce',{paths:[process.argv[1]]});const pkg=require.resolve('@vscode/vsce-sign/package.json',{paths:[p.dirname(vsce)]});const cmd=process.platform==='win32'?'vsce-sign.exe':'vsce-sign';const b=p.join(p.dirname(pkg),'bin',cmd);if(!fs.existsSync(b)){process.exit(2);}process.stdout.write(b);" "$root"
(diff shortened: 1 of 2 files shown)


jobs/release/release.yml
```diff
@@ -90,8 +90,14 @@ extends:
             npm_config_registry: https://pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages/npm/registry/
         - powershell: |
             $root = npm root -g
-            $sign = Join-Path $root '@vscode\vsce-sign\bin\vsce-sign.exe'
-            if (-not (Test-Path $sign)) { Write-Host "##vso[task.logissue type=error]Missing vsce-sign binary: $sign"; exit 1 }
+            if ($LASTEXITCODE -ne 0 -or [string]::IsNullOrWhiteSpace($root)) { Write-Host "##vso[task.logissue type=error]npm root -g failed"; exit 1 }
+            $root = $root.Trim()
+            # Resolve the vsce-sign native binary exactly as `vsce publish` does: from @vscode/vsce's own module tree.
```
snehara99, line 95: Why do we need to dynamically resolve the binary location now? Was there something breaking before?

hanniavalera: yes, the release job was failing at this exact check:

`##[error]missing vsce-sign binary: c:\npm\prefix\node_modules\@vscode\vsce-sign\bin\vsce-sign.exe`

the issue was the check, not necessarily the publish path.

the old check assumed `@vscode/vsce-sign` would be installed at the top-level global npm path:

`(npm root -g)\@vscode\vsce-sign\bin\vsce-sign.exe`

but `npm install -g @vscode/vsce` does not always hoist dependencies there. in this case, `vsce-sign` was nested under `vsce` itself:

`(npm root -g)\@vscode\vsce\node_modules\@vscode\vsce-sign\bin\vsce-sign.exe`

so the hardcoded path did not exist, which caused a false “missing binary” failure.

publishing itself should still be fine because `vsce publish` resolves `@vscode/vsce-sign` through node from `vsce`’s own install location, which finds the nested copy. the old check was just stricter than the actual publish behavior and failed before publish could run.

i updated the check to ask node to resolve `@vscode/vsce-sign` the same way `vsce publish` does, so it works whether npm hoists or nests the dependency. i tested the updated check locally, and it now finds the nested `vsce-sign.exe` path correctly instead of failing on the wrong hardcoded location.

snehara99 (approved):

PR #4993, opened 2026-06-30
Fix configuration regression

hanniavalera:
This pull request addresses a regression where the configure process could hang indefinitely due to compiler subprocesses not terminating properly. The main fix is to read compiler version telemetry from the CMake File-API toolchains data instead of launching the compiler during configure, which also improves reliability and performance. Several internal refactorings and type improvements support this change.

**Bug Fixes and Reliability Improvements:**

* Configure-time compiler version telemetry now reads the compiler id/version from the CMake File-API toolchains data instead of launching the compiler, preventing hangs caused by lingering compiler subprocesses. [[1]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR25) [[2]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fL1735-R1776) [[3]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fL1755-R1795)

**Internal Refactoring and Type Enhancements:**

* Introduced `getCompilerInfoForTelemetry` to derive compiler name/version from the File-API toolchains object, and refactored related logic for clarity and maintainability. [[1]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fR1498-R1529) [[2]](diffhunk://#diff-6645acdc0be3bcb29a2643c100233e69f82c7285cd04e83a0290eb2d9df96f4fL1467-R1481)
* Updated the `CodeModelToolchain` type to include optional `id` and `version` fields, and ensured these fields are populated when loading toolchain data. [[1]](diffhunk://#diff-a0051a7d09bfba52edd17da7d7026ced8580edac3ccc4c4e81afe1fcbda63694L16-R16) [[2]](diffhunk://#diff-7f4e58d932e1aff951164eed8488a529fd77b026451df2ea881e3cc36c4a7554R608-R613)
* Removed unused import of `getVSKitEnvironment` to clean up dependencies.


src/drivers/cmakeDriver.ts
```diff
@@ -1464,20 +1464,21 @@ export abstract class CMakeDriver implements vscode.Disposable {
         }
     ];
 
-    async getCompilerVersion(compilerPath: string): Promise<CompilerInfo> {
-        // Compiler name and path as coming from the kit.
-        const compilerName = path.parse(compilerPath).name;
-        const compilerDir = path.parse(compilerPath).dir;
-
+    private getCompilerAllowListEntry(compilerName: string) {
```
gcampbell-msft: Let's make sure that we have return types defined. 

Suggestion: We should actually also make sure that this is enforced via the linter, it's best to have our return types explicitly defined

src/drivers/cmakeDriver.ts
```diff
@@ -1464,20 +1464,21 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
         // Find an equivalent in the compilers allowed list.
         // To avoid finding "cl" instead of "clang" or "g++" instead of "clang++",
         // sort the array from lengthier to shorter, so that the find operation
         // would return the most precise match.
         // The find condition must be "includes" instead of "equals"
         // (which wouldn't otherwise need the sort) to avoid implementing separate handling
         // for compiler file name prefixes and suffixes related to targeted architecture.
-        const sortedCompilerAllowList = this.compilerAllowList.sort((a, b) => b.name.length - a.name.length);
-        const compiler = sortedCompilerAllowList.find(comp => compilerName.includes(comp.name));
+        const sortedCompilerAllowList = [...this.compilerAllowList].sort((a, b) => b.name.length - a.name.length);
```
gcampbell-msft: Any specific reason why we need the expansion operator here?

src/drivers/cmakeDriver.ts
```diff
@@ -1494,6 +1495,38 @@ export abstract class CMakeDriver implements vscode.Disposable {
...
+            return undefined;
+        }
+
+        return {
+            name: this.getCompilerNameForTelemetry(compilerPathForName),
+            version: toolchain?.version || "unknown"
+        };
+    }
+
+    async getCompilerVersion(compilerPath: string): Promise<CompilerInfo> {
```
gcampbell-msft: Is this method used anywhere now?

gcampbell-msft:
Do you think it's worth keeping the compiler invocation as a fallback if we don't find the information via the file-api? What version of the FileApi supports these version numbers? 

What if the user isn't using a FileApi supported version? At this point, this is not expected to be supported so it's okay if we say it's not a case we're targeting, but wanted to ask the question.

gcampbell-msft (dismissed):
Approving but with some feedback that I think should be updated

src/drivers/cmakeDriver.ts
hanniavalera: Not really, I removed it!

hanniavalera:
> Do you think it's worth keeping the compiler invocation as a fallback if we don't find the information via the file-api? What version of the FileApi supports these version numbers?
> 
> What if the user isn't using a FileApi supported version? At this point, this is not expected to be supported so it's okay if we say it's not a case we're targeting, but wanted to ask the question.

@gcampbell-msft Great questions. The compiler version/id comes from the file-API `toolchains` object, added in CMake 3.20. The relevant fields, `compiler.id`, `compiler.version`, `compiler.path`, and `compiler.target`, have been part of object v1 from the start and are optional.

For CMake < 3.20, we handle this cleanly. A missing `toolchains` object is already non-fatal in `cmakeFileApiDriver.ts`. We can still derive the compiler name from the kit compiler path, and only the version falls back to `"unknown"`. That only affects telemetry, not user-facing behavior.

I would not keep the compiler-invocation fallback. That probe caused this regression on Windows by leaving descendants like `mspdbsrv.exe` or `VCTIP.exe` holding inherited stdout/stderr pipes, which can make configure appear to hang. Reintroducing it is not worth the risk just to recover a telemetry-only version string for pre-3.20 CMake.

src/drivers/cmakeDriver.ts
hanniavalera: got it!

src/drivers/cmakeDriver.ts
hanniavalera: the spread was there to avoid mutating the shared list: the original `this.compilerAllowList.sort(...)` sorts in place, so it would reorder the member array on every call. `[...this.compilerAllowList]` made a throwaway copy to sort instead. 

that said, this is now handled better, i removed the per-call sort entirely. the list is sorted once at declaration: `private static readonly compilerAllowList: readonly CompilerAllowListEntry[] = [ ... ].sort(...)`

`getCompilerAllowListEntry` now just does a `.find()`, so there is no spread, no per-call sorting, and no mutation, this is addressed in the latest commit! :)

snehara99 (approved):

PR #4996, opened 2026-07-07
Add cmake.environmentSetupScript for preset mode

vasdal:
### This changes visible behavior + adds a setting

- Add a cross-platform `cmake.environmentSetupScript` setting. When using CMake
  Presets, CMake Tools sources (Linux/macOS) or calls (Windows) the script and
  uses its environment as the parent environment.
- The script env is applied to both `$penv{}`/`include` resolution and to
  per-preset expansion + the configure/build/test spawn, so a single script
  bootstraps everything. Cached by script path + mtime; re-applied on change.
- Reuses the kit's script-sourcing code (no behavior change for kits).

## The purpose of this change

Presets can only auto-bootstrap a parent environment on Windows (vcvarsall via
`tryApplyVsDevEnv`). On other platforms there's no way to source a setup script,
so environments whose variables are large, interdependent, or produced by a
tool's own setup script can't be used with presets: `$penv{...}` in
`include`/toolchain paths doesn't resolve (presets may fail to load) and
`find_package()` fails at configure time. Reconstructing such environments as
static `cmake.environment` entries isn't viable; sourcing the script is — which
is exactly what kits already do. This brings that to preset mode,
cross-platform.

## Other Notes/Information

- Same `__parentEnvironment` seam as the VS Dev Env; no new driver plumbing.
- Setting value supports `${workspaceFolder}`-style substitution.
- eslint clean; `yarn.lock` untouched; no dependency changes.


vasdal:

PR #4998, opened 2026-07-11
Fix Run/Coverage test profiles ignoring project in multi-root workspaces

danniesim:
## Summary
- In a multi-root workspace with more than one CMake project, the Test Explorer's **Run Tests** and **Run Tests with Coverage** profiles always dispatched through whichever project's `CTestDriver` happened to call `ensureTestExplorerInitialized()` first (since the `cmake-tools.CTest` `TestController` is a module-level singleton created lazily on first use, and the run-profile callbacks captured `this` from that first caller).
- Running any test that didn't belong to that "winning" project failed with `Could not determine the test program for test ...`, because the lookup checked the wrong project's cached CTest test list.
- The **Debug** profile already avoided this by resolving the owning project from the clicked test's URI (`request.include[0].uri.fsPath`) and delegating to that project's own `cTestController`. This PR applies the same resolution to the Run and Coverage profiles.

## Repro
1. Open a multi-root workspace with 2+ folders, each an independent CMake project with its own CTest tests, `cmake.ctest.testExplorerIntegrationEnabled: true`.
2. Let both projects configure/refresh so their tests populate the Test Explorer.
3. Click Run on a test belonging to whichever project did *not* win the initialization race.
4. Observe `Could not determine the test program for test <name>` even though the test builds and runs fine via `ctest` directly.

## Fix
Mirror the Debug profile's existing per-project resolution in the Run and Coverage profile callbacks in `src/ctest.ts`, falling back to the capturing driver's own handler when `request.include` is `undefined` (run-all) or resolution fails, exactly as Debug already does.

## Test plan
- [x] Manually verified against a local multi-root workspace with 3 independent CMake projects (each with its own CTest suite) that previously reproduced the bug reliably; after the fix, clicking Run on tests in every project correctly resolves and builds/runs that project's own test executable.
- [ ] Would appreciate guidance on whether a targeted multi-root end-to-end test (similar to `test/end-to-end-tests/multi-root-UI`) is wanted for this before merge — happy to add one.


danniesim:
Closing for now — thanks for bearing with the fork/branch setup. Might revisit later.

Issue #5005, opened 2026-07-24
[Bug] CMake version probe and build run outside the workspace, breaking version-manager (mise/asdf) shims


danieleformichelli:
### Brief Issue Summary

# Brief Issue Summary

When `cmake` is provided by a directory-based version manager (mise, asdf, vfox, ...), the extension fails with a bad/missing CMake executable, for example:
```
Bad CMake executable: "". Check to make sure it is installed or the value of the "cmake.cmakePath" setting contains the correct path
```

If I try to point cmakePath to the correct mise shim (`"cmake.cmakePath": "${userHome}/.local/share/mise/shims/cmake"`) I still get an error: 
```
[proc] Command output on standard error: mise ERROR No version is set for shim: cmake
```

even though `cmake` resolves fine from a terminal opened in the project.

## Related issues

- #4603 (mise, "Bad CMake executable, even though it's available on the path")
- #3359 (asdf, "Bad executable")
- #1731 (What is the Working Directory During Configure?)
- There is also a [third party extension](https://github.com/hverlin/mise-vscode/) that fixes it, but it would be much better for this to work out of the box

# Root cause

Several cmake child processes are spawned without a `cwd`, so they inherit the extension host's `process.cwd()`, which is not guaranteed to be the workspace folder. Version-manager shims resolve the tool version by walking up from the current working directory, so when the host cwd is outside the project the shim resolves no version and fails.

Source (v1.23.52 / current main):

- src/cmakeExecutable.ts `getCMakeExecutableInformation` runs the version and
  `-E capabilities` probes with `{ showOutputOnError, environment }` — no cwd.
- src/proc.ts `execute` only sets a spawn cwd when one is passed; otherwise Node
  `child_process.spawn` defaults to `process.cwd()`.
- The build path (src/drivers/cmakeDriver.ts `_doCMakeBuild`) and the task build
  path (src/cmakeTaskProvider.ts) also spawn cmake without a cwd. Configure works
  because it already passes `cwd: options?.cwd ?? binaryDir`.

This is launch-dependent, which is why it commonly appears on Remote-WSL when the editor is started from the Windows GUI (extension host cwd = home) rather than via `code .` inside the repo.

# Reproduction (self-contained, no mise/asdf needed)

[vscode-cmake-tools-shim-cwd-repro.zip](https://github.com/user-attachments/files/30339150/vscode-cmake-tools-shim-cwd-repro.zip)

[...]

PR #5006, opened 2026-07-24
Run cmake subprocesses in the project directory so version-manager shims resolve correctly

danieleformichelli:

- kits/kit.ts: probe cmake from the workspace folder during Clang/MSVC kit scanning (runs for any `clang` binary, so it hits shim users on Linux/macOS too).
- Run the cmake subprocesses launched by the extension inside the project directory instead of inheriting the extension host's `process.cwd()`.
- Thread an optional `cwd` through `getCMakeExecutableInformation(...)` (the `cmake --version` / `-E capabilities` probe) and pass the workspace/project folder from every caller (`extension.ts`, `cmakeProject.ts`, `debuggerScriptDriver.ts`).
- Default the working directory of driver subprocesses to `sourceDir`, the build invocation to `binaryDir`, and build tasks to `binaryDir || sourceDir || workspaceFolder`.
- Default the CMake script debugger child process cwd to the project folder.

## The purpose of this change

Users whose `cmake` is provided by a **directory-based version manager** (mise, asdf, vfox, ...) install a *shim* on `PATH`. These shims resolve the tool version by walking up from the **current working directory** to find the nearest config file. They do not have a "current" version baked in.

Today the extension launches `cmake` (the version/capabilities probe, the build, and the script debugger) without setting a `cwd`, so the child inherits the extension host's `process.cwd()`. Depending on how the editor was launched (e.g. from the OS GUI, Dock, or a Remote-WSL host started from the Windows GUI), that directory is often the user's home folder — outside the project. The shim then has no config to resolve against and fails, e.g.:
```
mise ERROR No version is set for shim: cmake
```

The result is that configuration intermittently succeeds (some paths already pass a cwd) while the version probe or the build fails, with an error that looks like a broken toolchain rather than a cwd issue.

Running these subprocesses in the project directory is the natural, correct behavior for a project-scoped build tool and makes shim-based toolchains resolve reliably, regardless of how the editor process was started.

## Other Notes/Information

- Scope is intentionally minimal: each call site defaults `cwd` only when the caller didn't already specify one, so existing behavior for explicitly-set working directories is unchanged.
- The change is a no-op for users on absolute cmake paths or non-shim installs, since the working directory doesn't affect how those binaries resolve.
- One related probe, `checkDebugger` in `src/debug/debugger.ts`, still runs without a cwd. It was deliberately left out of this PR to keep the change focused on the cmake execution paths; happy to include it if maintainers prefer.
- A minimal standalone reproduction (a fake cwd-sensitive `cmake` shim + a tiny CMake project + a script showing the shim succeeding in-project and failing from `$HOME`) is available and can be linked/attached on request.


danieleformichelli:
@microsoft-github-policy-service agree company="Bending Spoons"

src/drivers/cmakeDriver.ts
```diff
@@ -2015,7 +2021,7 @@ export abstract class CMakeDriver implements vscode.Disposable {
                     }
                 }
             } else {
-                const exeOpt: proc.ExecutionOptions = { environment: buildcmd.build_env, outputEncoding: outputEnc, useAutoEncoding: isAutoEncoding };
+                const exeOpt: proc.ExecutionOptions = { environment: buildcmd.build_env, outputEncoding: outputEnc, useAutoEncoding: isAutoEncoding, cwd: this.binaryDir || undefined };
```
hanniavalera: @danieleformichelli Could we use the source/project directory as the default cwd here rather than `binaryDir`?

A directory-based version manager searches upward from cwd for its configuration. If the user has an external build tree, for example, the source is under `~/src/project` but `binaryDir` is `~/build/project`, starting from `binaryDir` will never find the configuration in the source tree, so the shim still fails with the error this PR is intended to fix.

`cmake --build` already receives the build directory explicitly, so its process cwd does not need to be `binaryDir`. I suggest consistently defaulting direct and task builds to `sourceDir` (or the project folder), while continuing to preserve any explicitly supplied cwd.

src/cmakeExecutable.ts
hanniavalera, line 61: Now that this probe is cwd-sensitive, should `cmakeInfo` also include the normalized cwd in its cache key?

Different projects can use the same shim path while resolving different CMake installations. In a multi-root workspace, for example, both projects may invoke `~/.local/share/mise/shims/cmake`, but their project configuration files may select different CMake versions. With the cache keyed only by executable path, the first successful probe is returned for every later project without probing from that project's cwd.

That can produce incorrect version and capability information, including file-API or debugger support. Please either key the cache by both normalized path and normalized cwd, bypass it for contextual shim probes, or otherwise ensure results are re-probed when the project context changes.

src/cmakeProject.ts
```diff
@@ -1605,7 +1605,7 @@ export class CMakeProject {
 
     async getCMakeExecutable() {
         const cmakePath: string = await this.getCMakePathofProject();
-        const cmakeExe = await getCMakeExecutableInformation(cmakePath);
+        const cmakeExe = await getCMakeExecutableInformation(cmakePath, undefined, this.folderPath);
```
hanniavalera: Could this project-specific probe use the resolved `sourceDir` rather than `folderPath`?

For a single-project workspace, `folderPath` intentionally resolves to the workspace root, which may differ from `cmake.sourceDirectory`. If the version-manager configuration is inside a nested source directory, probing from the workspace root cannot discover it because these shims search upward from cwd, not downward into child directories.

Using `sourceDir` here, with the workspace folder only as a fallback if necessary, would make the probe run from the actual CMake project context.

hanniavalera:
@danieleformichelli Thank you for contributing this fix! It is quite appreciated! I reviewed the subprocess working-directory changes across CMake probing, builds, tasks, kit scanning, and script debugging, with particular attention to multi-root and nested-source projects, external build directories, version-manager shim caching, and regression risk for existing CMake workflows.

The overall direction is sound and directly addresses the underlying problem in #5005. I left a few comments about using the source directory for shim resolution and making the executable-information cache project-context aware. Thanks again for the clear explanation, focused implementation, and work to improve the experience!

danieleformichelli:
> Daniele Formichelli (@danieleformichelli) Thank you for contributing this fix! It is quite appreciated! I reviewed the subprocess working-directory changes across CMake probing, builds, tasks, kit scanning, and script debugging, with particular attention to multi-root and nested-source projects, external build directories, version-manager shim caching, and regression risk for existing CMake workflows.
> 
> The overall direction is sound and directly addresses the underlying problem in #5005. I left a few comments about using the source directory for shim resolution and making the executable-information cache project-context aware. Thanks again for the clear explanation, focused implementation, and work to improve the experience!

Thank you @hanniavalera 🙏 
I can't see any comment, are you sure you have submitted your review? 🙏 

Feel free to push your changes on the PR if you want!

hanniavalera (changes requested):
@danieleformichelli Sorry I forgot to submit the review, here ya go!

danieleformichelli:
Good morning @hanniavalera, could you please have a look at my changes when you have some time? 🙏

hanniavalera (approved):

snehara99 (approved):

PR #2835, opened 2022-11-01
CMake Multi Root and Multi Project support

elahehrashedi:
Addresses https://github.com/microsoft/vscode-cmake-tools/issues/1374

PR #5015, opened 2026-07-30
Fix kit selection targeting the wrong folder in multi-root workspaces (#5011)

### LINKED ISSUE #5011: [Bug] "Select a Kit for `<activeWorkspaceFolderName>` appears on workspace load instead of the valid/triggering workspace folder
### Brief Issue Summary

1. Open a workspace with many folders that use CMakeLists.txt, such as adding the vscode-cmake-tools repo to another workspace as another workspace folder.
2. Set the settings.json for one of the folders to 

```json
{
    "cmake.sourceDirectory": "pathToWorkspaceFolder/src"
}
```
which is already set for the vscode-cmake-tools repo case.
3. On workspace open, select a file in a folder that is not the same as the folder that is triggering the "Select" UI to appear, i.e. not vscode-cmake-tools.

Bug: The "Select a Kit for `<activeWorkspaceFolderName>` appears but the valid/triggering workspace folder is missing from the UI (e.g. vscode-cmake-tools) -- this makes it difficult to determine which folder in the workspace is causing the popup. In my repro, the actual root CMakeLists.txt is further up the directory hierarchy, not sure if that matters. The popup disappears when I remove that setting (not sure how it got added). i.e. when commenting out the "cmake.sourceDirectory": "$workspaceFolder}/test/fakeOutputGenerator" in the vscode-cmake-tools repo.

Image

### CMake Tools Diagnostics

```shell
...
```

### Debug Log

```shell
...
```

### Additional Information

_No response_

hanniavalera:
Fixes #5011.

## Problem
In a multi-root workspace, when a **non-active** workspace folder triggers configure-on-open and needs a kit, the "Select a Kit for `<name>`" prompt showed the **active** folder's name and applied the kit to the **active** project instead of the folder that actually needed it. Because the kit was never set on the triggering folder, its configure-on-open aborted and the prompt reappeared on every window reload. (The active file's folder — e.g. `App3` — was shown/used instead of the triggering folder — e.g. `vscode-cmake-tools`.)

## Root cause
`ExtensionManager.selectKit(folder)` resolved the folder's projects only for a null-check, then discarded them and operated on `getActiveProject()`:

```ts
const cmakeProject = this.getProjectsForWorkspaceFolder(folder); // only null-checked
const activeProject = this.getActiveProject();                    // <- ignores `folder`
const kitSelected = await activeProject?.kitsController.selectKit();
```

So the QuickPick placeholder (`Select a Kit for {0}` → `this.project.folderName`) and the kit assignment both targeted the active project. `ensureActiveConfigurePresetOrKit` then checked the *triggering* project's `activeKit`, which stayed unset → configure aborted → the prompt recurred. `selectKit` was the only folder-scoped selector that never adopted the `getProjectFromFolder(folder)` pattern already used correctly by `selectConfigurePreset`. (The `getActiveProject()` line dates back to the original multi-root support in #2835.)

## Fix
- `selectKit` now resolves the target via `getProjectFromFolder(folder, sourceDir)` and operates on **that** project for kit selection, the `activeKit` read, telemetry, and the return value.
- `ensureActiveConfigurePresetOrKit` passes the exact triggering project's `folderPath` as a `sourceDir` disambiguator to both `selectKit` and `selectConfigurePreset`, so **multi-project folders** (several `CMakeLists.txt` under one workspace folder) resolve to the precise project the post-check inspects — not just the first/active sibling.
- Status-bar writes in both `selectKit` and `selectConfigurePreset` are gated on the resolved project being the active one, so choosing a kit/preset for a non-active folder during configure-on-open no longer overwrites the active project's status bar.
- Removed the now-unused private `getProjectsForWorkspaceFolder` wrapper.

No behavior change for single-root workspaces, single-project folders, or manual command-palette invocation (`folder`/`sourceDir` undefined → resolves to the active project exactly as before).

## Validation
- `tsc -p test.tsconfig.json` — clean
- `webpack --mode development` — clean
- `node ./out/test/unit-tests/runTest.js` — **330 passing**
- Manual repro (multi-root workspace with `App1`/`App2`/`App3` plus a folder whose `.vscode/settings.json` sets `cmake.sourceDirectory` and `configureOnOpen`): with a non-triggering folder's file active, reload now shows **"Select a Kit for `<triggering-folder>`"** and applies the kit to that folder; the prompt no longer recurs on reload, and the active project's status bar is preserved.

## Notes
Behavior verified for both kits and presets modes; generator type (single- vs multi-config) is not relevant to this routing fix.


hanniavalera:
Accidentally double worked on the same fix as #5017

PR #5017, opened 2026-07-30
Fix multi-root Select a Kit prompt targeting the active folder instead of the triggering folder (#5011)

### LINKED ISSUE #5011: [Bug] "Select a Kit for `<activeWorkspaceFolderName>` appears on workspace load instead of the valid/triggering workspace folder

Omotola:
Fixes #5011.

### Problem
In a multi-root workspace, the automatic **Select a Kit** prompt raised during configure-on-open was titled for — and applied to — the folder of the **active editor**, not the folder whose configure-on-open actually triggered it. As a result:
- The prompt named the wrong folder, making it unclear which project needed a kit.
- The kit was set on the active project, so the triggering project never got a kit and kept re-prompting on every window reload.

### Root cause
`selectKit(folder)` in `src/extension.ts` resolved the correct project from the passed `folder` but then ignored it and ran `getActiveProject().kitsController.selectKit()`. The kit QuickPick title comes from that project's `folderName`, so both the title and the applied selection went to the active project.

### Fix
`selectKit` now runs the selection against the project that owns the passed `folder`, falling back to the active project only when no folder is given (the manual status-bar/command path). The status bar is updated only when the target is the active project, since it reflects the active project's kit.

### Manual testing
Multi-root workspace with two kit-mode CMake roots (`App3` active, `MainProj` triggering via `cmake.sourceDirectory`). Open a file in `App3`, reload window:
- **Before:** prompt reads "Select a Kit for App3"; `MainProj` never configured; prompt returns every reload.
- **After:** each folder's prompt is correctly titled for its own folder and the selection sticks.

Also verified the manual `CMake: Select a Kit` path still targets the active folder and updates the status bar as before.


src/extension.ts
```diff
@@ -1348,23 +1348,32 @@ export class ExtensionManager implements vscode.Disposable {
...
         }
 
+        // Select the kit for the project that owns the requested folder. Previously this always
+        // operated on the active project, which in a multi-root workspace caused the automatic
+        // configure-on-open kit prompt to target (and be titled for) whichever folder had the
+        // active editor rather than the folder that actually triggered configuration. 
         const activeProject = this.getActiveProject();
-        const kitSelected = await activeProject?.kitsController.selectKit();
+        const targetProject = (activeProject && cmakeProjects.includes(activeProject)) ? activeProject : cmakeProjects[0];
+        const kitSelected = await targetProject?.kitsController.selectKit();
```
hanniavalera: @Omotola Could we resolve the target using the triggering project’s source directory rather than choosing the active project or `cmakeProjects[0]`?

This still selects the wrong project when one workspace folder contains multiple CMake projects. If project B triggers configure-on-open while sibling A is active, `cmakeProjects.includes(activeProject)` selects A. If neither is active, the first-project fallback may also select A. B therefore remains without a kit and can continue prompting on reload.

The caller already has the exact triggering `CMakeProject`, so passing `cmakeProject.folderPath` into the existing `getProjectFromFolder(folder, sourceDir)` resolution path would preserve that identity and handle both multi-root and multi-project folders.

hanniavalera: Also the CI has a couple of lint errors! That's all :) Otherwise, fix is sound!

hanniavalera (approved):

PR #5023, opened 2026-08-03
Show all STRINGS choices in cache editor dropdown

quanzhuo:

- Replace the Cache Editor's native `datalist` control for `STRINGS` cache variables with a custom editable dropdown.
- Show all declared `STRINGS` choices whenever the dropdown is opened, instead of filtering choices by the current input value.
- Preserve manual text editing and invalid-choice highlighting for string cache entries.
- Add a regression test for rendering `STRINGS` choices as an unfiltered popup list.


quanzhuo:
The native datalist in VS Code webviews filters options based on the current input value and cannot be configured to always show all CMake STRINGS choices, so this replaces it with a small editable combobox implementation.

quanzhuo:
One behavior choice worth calling out: `cmake-gui` shows `STRINGS` cache entries as a non-editable dropdown. This PR keeps CMake Tools' existing editable input behavior and only fixes the filtered candidate list. I’m happy to adjust this to a non-editable dropdown if maintainers prefer matching `cmake-gui` exactly.

PR #5027, opened 2026-08-05
Show the CMake sidebar immediately with an initializing placeholder during activation

hanniavalera:
## What

The CMake Tools activity-bar sidebar, the container icon and all of its views (Project Status, Project Outline, Pinned Commands, Bookmarks), was hidden until the extension finished initializing. On machines where process/file I/O is intermittently slow, that can take **minutes**, during which the sidebar simply does not appear.

This PR introduces a **two-phase reveal**: the CMake activity-bar container now appears **immediately** with a lightweight, inert "CMake Tools is initializing…" placeholder, and hands off to the real views as soon as the backend is ready.

## Why

The container and every view are gated on the `cmake:enableFullFeatureSet` context key, which the extension only sets to `true` at the very end of `ExtensionManager.init()`, behind a serial chain of subprocess probes (`cmake --version` / `-E capabilities`, the kit file read, preset parse/expansion, and the Visual Studio developer-environment bootstrap (`vcvarsall`) for `cl.exe` presets).

When the environment makes process creation and file reads intermittently slow such as an antivirus scanning of every spawned executable, or extension-host file-handle exhaustion (`EMFILE`), those probes serialize into minutes of wall-clock time, and the entire sidebar stays hidden the whole time. `cmake:enableFullFeatureSet` also gates ~255 other contributions (commands, menus, keybindings, the status bar), so it cannot simply be flipped early without exposing project actions against a not-yet-initialized backend.

## How

- A new **`cmake:isInitializing`** context key backs a transient, inert placeholder view (`cmake.initializing`) whose content comes from `viewsWelcome`. The activity-bar container's `when` becomes `cmake:isInitializing || cmake:enableFullFeatureSet`, and the placeholder view is gated on `cmake:isInitializing && !cmake:enableFullFeatureSet`.
- The **real views, commands, menus, and status bar remain gated exclusively on `cmake:enableFullFeatureSet`**, nothing runs against a partially-initialized project. The transient placeholder and the real views are mutually exclusive, so there is no double-render and no empty-container flicker during the handoff.
- `cmake:isInitializing` is set right after the manager is constructed (via `setContextValue`, so it does not trigger a command recompute) and cleared in a `finally` after activation settles, so the placeholder can never get stuck on failure or when no project is found.
- The placeholder is shown only when a **cheap, fail-open preflight** determines the workspace has a CMake project: it does variable expansion plus a single `CMakeLists.txt` `stat` per configured source directory (mirroring `hasCMakeLists()`), with **no** subprocess spawn, kit/preset init, or recursive scan. Crucially it **fails open**, only `ENOENT`/`ENOTDIR` count as "no project"; transient errors like `EMFILE` still show the placeholder, so the fix works in exactly the resource-starved environments it targets. It is skipped entirely in language-server-only mode.

image

## Scope

This is a **UI-reveal** fix. It does not change the actual initialization work, its ordering, or timing, the environmental slowness itself (antivirus, `EMFILE`) is outside the extension's control. Deferring the individual probes (VS developer-environment acquisition, language-service asset loading) is a worthwhile but separate, higher-risk follow-up

## Testing

- `yarn backendTests` (386 passing) — new `activation.test.ts` (visibility policy + fail-open filesystem-error classification, incl. `EMFILE`) and `activation-contributions.test.ts` (container/view `when` gating, the "placeholder and real views are never simultaneously visible" invariant, and that `cmake:isInitializing` never gates any command/menu/keybinding).
- `yarn unitTests`, 330 passing.
- `tsc` + `gulp lint` clean.
- Behavior verified unchanged in presets and kits/variants modes, single- and multi-config generators, multi-root, language-server-only mode, configure-on-open on/off, and non-CMake / standalone `.cmake` workspaces (no placeholder shown).


hanniavalera:
Companion backend PR: #5035 (\Speed up CMake Tools activation: lazy language-service data and no activation-time kit-prune storm\). #5035 is the primary fix — it removes the two avoidable, environment-sensitive costs from the activation critical path (blocking ~594 KB language-service asset load; unbounded kit-compiler probe burst that also misread transient \EMFILE\ as a missing compiler). This PR (#5027) is complementary defense-in-depth: it shows an ''initializing'' placeholder so the sidebar appears immediately while the remaining irreducible subprocess latency (e.g. \cmake -E capabilities\, \cvarsall\ for cl.exe presets under antivirus) completes. Recommend landing #5035 first.

src/ui/initializingView.ts
```diff
@@ -0,0 +1,22 @@
...
+ * placeholder view.
+ *
+ * It intentionally has no children — the visible content is supplied by the view's
+ * `viewsWelcome` contribution in `package.json`. Its only purpose is to back the
+ * `cmake.initializing` view so the CMake activity-bar container can appear
+ * immediately during activation (via the `cmake:isInitializing` context key),
+ * before the real project views are ready. The real views, commands, and status
+ * bar remain gated on `cmake:enableFullFeatureSet`.
+ */
+export class InitializingViewProvider implements vscode.TreeDataProvider<vscode.TreeItem> {
```
gcampbell-msft: Nit: Is it possible there are other view providers we could use since we know we don't actually need a TreeDataProvider? Or is it possible we don't need it at all/

src/activation.ts
```diff
@@ -0,0 +1,46 @@
+/**
```
gcampbell-msft: Another feedback of let's try to trim comments to only what is necessary and ensures understanding is clear

gcampbell-msft (dismissed):
Approved with a couple of feedback points

src/activation.ts
hanniavalera: done!

src/ui/initializingView.ts
hanniavalera: Good call, removed it entirely. The `cmake.initializing` view only renders its `viewsWelcome` content and the container appears from the view's when clause, so no `TreeDataProvider` is needed. Deleted the provider and its registration.

snehara99 (approved):

PR #5035, opened 2026-08-11
Speed up CMake Tools activation: lazy language-service data and no activation-time kit-prune storm

hanniavalera:
## What

Two backend changes so extension activation no longer blocks on work that isn't needed to activate, and so CMake Tools stops adding to (and misreading) filesystem/handle pressure during startup:

1. **Load the bundled language-service data on first use** instead of during activation.
2. **Stop checking user kits for missing compilers on every activation**, and never treat a transient filesystem error (e.g. file-handle exhaustion) as a missing compiler.

## Why

On machines where process/file I/O is intermittently slow (aggressive antivirus scanning of spawned executables, or extension-host file-handle exhaustion / `EMFILE`), activation could take minutes and the CMake sidebar stayed hidden. Investigation showed the extreme latency is environmentally *triggered*, but the extension puts avoidable, environment-sensitive work on the activation critical path, and even contributes to the handle pressure itself:

- `ExtensionManager.init()` `await`ed `enableLanguageServices()` **first**, which serially read and parsed ~594 KB of bundled hover/completion JSON (`commands`/`variables`/`modules`/`policies`) and registered the providers only *after* that load. That data is only needed when the user hovers or completes in a CMake file — never for activation, the sidebar, or configure/build.
- Right after loading user kits, activation fire-and-forgot an **unbounded burst of compiler-existence probes** (up to ~2 per kit). Worse, those probes went through helpers that collapse *every* error to "missing", so under `EMFILE`/`EACCES` the extension both added handle pressure and could prompt to remove **valid** kits.

## How

**Language-service data (commit 1):**
- `LanguageServiceData.create()` is now synchronous; the data loads lazily on the first hover/completion via a memoized `ensureLoaded()` (a failed load is memoized so it doesn't retry the four files on every keystroke). `provideHover`/`provideCompletionItems` are async and honor cancellation/disposal before and after the load. The four reads stay serial (no `Promise.all`) and there is no background prewarm, so activation performs **zero** language-asset I/O.
- `enableLanguageServices()` registers the hover/completion providers and language configuration immediately; `disposeLanguageServices()` disposes the provider and clears its registrations. Language-server-only mode is unaffected.

**Kit pruning (commit 2):**
- Removed the stale-kit check from activation / user-kit-file reads. Stale kits are still detected during a kit/compiler **scan** (unchanged entry point), which is the expected maintenance boundary.
- Hardened the scan-time prune: it probes **sequentially** (no burst), dedupes repeated compiler paths, is single-flight with one coalesced follow-up, and prompts sequentially.
- Fixed the error classification (tri-state `present`/`absent`/`unknown`): absolute paths use `fs.stat` and treat only `ENOENT`/`ENOTDIR` as absent; relative names go through `which`, whose failures are treated as `unknown` (it synthesizes `ENOENT` for all failures and so cannot prove absence). A transient error can therefore **never** prompt removal of a valid kit.

## Scope / follow-ups

This is a backend activation-performance/correctness fix. It does not change the actual configure/build work or its timing. The remaining irreducible cost at activation is subprocess latency (e.g. `cmake -E capabilities`, and `vcvarsall` for `cl.exe` presets under antivirus); a companion UI change (#5027) shows an "initializing" placeholder so the sidebar appears immediately while that residual work completes: complementary to this PR, not a substitute for it. Further optimizations identified but intentionally deferred: single-spawn CMake interrogation (capabilities-first), in-process compiler lookup for VS-dev-env detection, and caching compiled JSON schema validators.

## Testing

- `yarn backendTests` (384 passing) - new `test/unit-tests/backend/kitTriState.test.ts` (tri-state classification incl. `EMFILE`/`ENFILE`/`EACCES`→unknown and `which`-error→unknown) and `test/unit-tests/backend/languageServiceLazyLoad.test.ts` (lazy memoization / failure / disposal semantics).
- `yarn unitTests` - 330 passing.
- `tsc` + `gulp lint` clean.
- Behavior verified in presets and kits/variants modes, single/multi-config, multi-root, and language-server-only mode. Stale-kit prompts now appear on explicit scan rather than on every activation (a deliberate behavior change, noted in the CHANGELOG).


src/kits/compilerPathState.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Tri-state classification of whether a kit's compiler binary exists on disk.
+ *
+ * This is kept as a small, dependency-free module (no `vscode` import) so it can
```
gcampbell-msft, line 4: Nit: Some of these comments are overly verbose. Copilot tends to get chatty in the comments, let's trim this down across this PR

gcampbell-msft (dismissed):

Omotola (approved):
Non-blocking, but: if the first hover/completion triggers the lazy load and it fails (e.g. under the file-handle exhaustion this PR targets), the failure is cached permanently and language services stay off until reload. Might be worth retrying on the next hover instead of caching the failure.

PR #5053, opened 2026-08-25
Update localized strings while preserving translation fixes

sean-mcmanus:
## Summary

Import the latest translated strings from the automated `localization` branch and preserve 15 reviewed translation corrections from #5042 on a durable branch that future localization force-pushes cannot overwrite.

The August 25 generated localization patch is identical to the August 24 patch: both change the same 299 localized values, with no added, removed, or updated localization keys or translations. The English source audit likewise found no changes across 1,021 `localize` calls or `package.nls.json`.

After this fix gets in, the localization branch will still include the unfixed issues so the commit with fixes needs to be cherry-picked into any future loc updates unless you want to try to get the loc team to fix the bugs, but it's very time consuming to file the bugs manually and they might not actually get fixed, so my recommendation is just to cherry-pick a commit which has all the known issues fixed and/or use Copilot to assist with that.

## Validation

- `yarn compile`
- `yarn lint`
- `yarn pretest-buildfakebin`
- `yarn unitTests` (333 passing, 29 pending)
- Parsed all replayed JSON bundles and verified interpolation/Markdown token preservation
- Compared generated localization patches and source localization definitions across the August 24 and August 25 snapshots

> This PR was re-investigated and updated by GitHub Copilot in VS Code. Any message starting with `✨Copilot` was sent by Copilot.


hanniavalera:
thank you! i appreciate you addressing this, will merge this in!

hanniavalera (approved):

PR #5079, opened 2026-09-08
Retry yarn install in CI to tolerate transient Azure Artifacts 429 throttling

hanniavalera:
## What

Wraps every CI `yarn install` that pulls from the private Azure Artifacts npm mirror in a retry-with-backoff, so a transient "429 Too Many Requests" no longer fails an otherwise-healthy build.

## Why

The mirror (`pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages`) intermittently returns `429 Too Many Requests` during `yarn install`. It happens only on a `node_modules` cache miss (for example a lockfile bump), and a single 429 kills the whole job — e.g. on 2026-09-08 the build job failed with `argparse-2.0.1.tgz: Request failed "429 Too Many Requests"` while the same commit's Linux build passed. There was no retry anywhere, and `build-vsix.yml` had no dependency cache at all (it always full-installs), so it is the most exposed. Re-running usually cleared it; this lets the pipeline self-heal instead.

## How

- New `.github/scripts/yarn-install.sh`: up to 5 attempts of `yarn install "$@"` with linear backoff (15s × attempt), a `::warning` per retry and a `::error` on exhaustion; exits 0 on success / 1 on failure. Arguments are forwarded verbatim (`--frozen-lockfile`, `--ignore-scripts`). Attempt count and backoff are overridable via env vars (used only for testing).
- `ci-main-linux.yml`, `ci-main-mac.yml`, `localization-readiness.yml`, `build-vsix.yml`: the `yarn install` steps now call that script (bash).
- `ci-main.win.yml` (windows-2022): an equivalent inline **pwsh** loop with `$ErrorActionPreference = 'Continue'`, so a native nonzero exit does not throw before the `$LASTEXITCODE` check.
- `build-vsix.yml`: the install is split into its own "Install dependencies" step (it was previously combined with compile/package in one step).
- `copilot-setup-steps.yml` is intentionally unchanged — it rewrites the registry to the public npm registry, so it never hits the throttled feed.

No install semantics change: on success the install runs exactly once and behaves identically; the retry only engages on failure, and `--frozen-lockfile` keeps it idempotent. A genuine (non-transient) failure still fails the job after the attempts are exhausted.

## Testing

- Verified the retry logic under **real bash** and **real pwsh 7** with a mocked `yarn`: succeeds-first-try runs once (exit 0); fails-twice-then-succeeds runs 3× (exit 0); always-fails exhausts the attempts (exit 1); arguments forward correctly.
- All five changed workflows pass `yaml.safe_load` and `actionlint` 1.7.12 with no new findings.
- The script is committed with LF endings (repo `.gitattributes` is `* eol=lf`).

## Notes

- Engineering / CI-only change; no user-facing behavior, so no `CHANGELOG` entry.
- `actionlint` surfaced a pre-existing, unrelated expression warning in `localization-readiness.yml:42` (a boolean `==` comparison, not shell injection — it already carries an explanatory comment). Left as-is to keep this PR scoped to the 429 fix.

### DIFF
--- a/.github/scripts/yarn-install.sh
+++ b/.github/scripts/yarn-install.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# Run `yarn install` with retries and exponential-ish backoff.
+#
+# CI installs dependencies from the private Azure Artifacts mirror
+# (pkgs.dev.azure.com/azure-public/VisualCpp/_packaging/cpp_PublicPackages). Under load that feed
+# intermittently returns "429 Too Many Requests", which fails an otherwise-healthy build. This only
+# happens on a node_modules cache miss (e.g. a lockfile bump), so retrying the install a few times
+# with backoff lets the transient throttle clear instead of failing the whole job.
+#
+# Any arguments are forwarded verbatim to `yarn install` (e.g. --frozen-lockfile --ignore-scripts).
+# Invoke as: bash ./.github/scripts/yarn-install.sh [yarn install args...]
+set -uo pipefail
+
+attempts="${YARN_INSTALL_ATTEMPTS:-5}"
+sleep_base="${YARN_INSTALL_RETRY_SLEEP_BASE:-15}"
+
+for attempt in $(seq 1 "$attempts"); do
+  if yarn install "$@"; then
+    exit 0
+  fi
+  if [ "$attempt" -ge "$attempts" ]; then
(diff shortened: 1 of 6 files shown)


Omotola (approved):
Only thing I would like to suggest is de-duplicating the retry logic in ci-main.win.yml but its non-blocking.

hanniavalera:
ty

PR #5084, opened 2026-09-09
Skip unreadable directories when scanning for kits (#5020)

### LINKED ISSUE #5020: [Bug] EACCES scanning for toolkits
### Brief Issue Summary

vscode on windows, connected to rocky9 wsl2 instance.

"scan recursively for kits in specific directories" -> "/opt/rh/"

```
[rollbar] Unhandled exception: Unhandled Promise rejection: selectKit Error: EACCES: permission denied, scandir '/opt/rh/gcc-toolset-13/root/root' {}
```

no kits are added

```
$ ls -ld /opt/rh/gcc-toolset-15/root/root
dr-xr-x--- 2 root root 4096 Oct 29  2025 /opt/rh/gcc-toolset-15/root/root
```

### CMake Tools Diagnostics

```shell
(see above)
```

### Debug Log

```shell
n/a
```

### Additional Information

_No response_

Omotola:
Fixes #5020.

### Problem
"Scan recursively for kits in specific directories" walked the chosen folder with `fs.readdir`/`fs.stat` and no error handling. A single unreadable subdirectory — e.g. a root-only `/opt/rh/gcc-toolset-13/root/root` (`dr-xr-x--- root root`) on a WSL/Linux host — rejected the whole scan with an unhandled `EACCES`, and **no kits were added**:

```
[rollbar] Unhandled exception: Unhandled Promise rejection: selectKit
Error: EACCES: permission denied, scandir '/opt/rh/gcc-toolset-13/root/root'
```

### Fix
Extract the recursive walk into a testable `accumulateDirsToScan()` helper that:
- catches errors from `fs.readdir` → an unreadable directory is skipped (logged at debug), not fatal;
- catches errors from `fs.stat` per entry → un-stat-able entries (broken symlink, races) are skipped;
- keeps honoring cancellation between entries.

The rest of the scan now completes even when some directories are unreadable.

### Tests
New unit suite in `kit-scan.test.ts` (sinon-stubbed `fs`): the scan no longer throws on an unreadable subdirectory, keeps readable siblings, returns `[]` cleanly when the root itself is unreadable, and skips un-stat-able entries.

### Verification
Reproduced end-to-end on WSL Ubuntu (non-root user) with the reporter's exact scenario. Before: unhandled `EACCES`, no kits. After (this branch, installed as a VSIX): the scan against `/opt/rh` completes and the kits file loads with no unhandled rejection.

`tsc` (src + tests) and `eslint` clean; new tests pass.

**Before**
image

**After**

image


src/kits/kitsController.ts
```diff
@@ -54,6 +54,46 @@ export function shouldKeepUserKitAfterScan(kit: Kit, discoveredKitNames: Readonl
...
+        }
+        const filePath = path.join(folder, file);
+        let isDirectory = false;
+        try {
+            isDirectory = (await fs.stat(filePath)).isDirectory();
+        } catch (e: any) {
+            log.debug(localize('skip.unstattable.entry', 'Skipping entry that could not be inspected while scanning for kits: {0} ({1})', filePath, e.code ?? e.message ?? String(e)));
+            continue;
+        }
+        if (depth > 0 && isDirectory) {
```
hanniavalera: @Omotola Could we preserve the previous short-circuit behavior by checking `depth > 0` before calling `fs.stat()`? At depth zero, we now stat every entry even though none will be traversed or added, introducing unnecessary filesystem calls in large directories.

hanniavalera (changes requested):

hanniavalera (approved):

PR #5095, opened 2026-09-14
Update CTest JSON output TypeScript definitions

Nerixyz:

- Make `backtrace`, `command`, and `properties` in `tests` optional to match CTest output and handle resulting TypeScript errors.

## The purpose of this change

CTest only guarantees that a test has a `name` (see [source](https://gitlab.kitware.com/cmake/cmake/-/blob/f5efe7a6b5fe2acbeceab5d79f397d58c25f3e01/Source/CTest/cmCTestMultiProcessHandler.cxx#L1452)).
When a project had a test that was missing some of these (e.g. because a target wasn't built yet), the log output would repeatedly show:
```
[testCodeLensProvider] Error in provideCodeLenses: TypeError: Cannot read properties of undefined (reading '0')
```
This could be traced back to `CTestDriver#getTestsForOutline` which unconditionally accesses `test.command[0]`. For this specific case, I opted to skip the entry.


src/ctest.ts
```diff
@@ -1558,16 +1558,18 @@ export class CTestDriver implements vscode.Disposable {
         if (this.tests) {
             const executableToSources = codeModelContent ? this.buildExecutableToSourcesMap(codeModelContent) : undefined;
 
-            return this.tests.tests.map(test => {
-                const { file: sourceFilePath, line: sourceFileLine } = this.resolveTestSourceLocation(test, executableToSources, this.tests!.backtraceGraph);
+            return this.tests.tests
+                .filter(test => test.command)
```
hanniavalera: `.filter(test => test.command)` in `getTestsForOutline` drops unbuilt tests from the outline entirely instead of showing them without a location. Was that intentional, or would you rather make `executablePath` optional so the test still shows up? Also, an empty `command: []` passes this filter but still gives undefined at `[0]`, so `!!test.command?.length` would be safer.

src/ctest.ts
```diff
@@ -1558,16 +1558,18 @@ export class CTestDriver implements vscode.Disposable {
...
-                return {
-                    name: test.name,
-                    executablePath: test.command[0],
-                    sourceFilePath,
-                    sourceFileLine
-                };
-            });
+                    return {
+                        name: test.name,
+                        executablePath: test.command![0],
```
hanniavalera: Related to the thread above: if the filter used a type predicate, this ! wouldn't be needed. Something like: `.filter((test): test is CTestInfo['tests'][0] & { command: string[] } => !!test.command?.length)`

hanniavalera (changes requested):
Thanks for catching this. The `CTestInfo` type was definitely over-promising here, and `testProgram`/`testArgs` already treated command as optional, so making these fields optional makes sense.

A few things before this can go in, let me know if you disagree with anything. 
I've attached a couple of comments, let me know if you disagree :) happy to chat about it, but there are other things that would be ideal to add on!

- Please add a CHANGELOG.md entry under the unreleased section for this fix.
- A unit test with a ctestInfo payload missing command/properties would help prevent regressions.
- The branch is also behind main, so it'll need a rebase.

src/ctest.ts
Nerixyz: > `.filter(test => test.command)` in `getTestsForOutline` drops unbuilt tests from the outline entirely instead of showing them without a location.

Every consumer requires the command to be there and in practice, there isn't much you can do with a test without a command. At least in my test, the test's name is `target-name_NOT_BUILT`, so you won't be able to run it in the first place.

Nerixyz:
> * A unit test with a ctestInfo payload missing command/properties would help prevent regressions.

Added one with both cases and an empty command.

> * The branch is also behind main, so it'll need a rebase.

I merged with main, since you seem to be squashing anyway and rebasing often destroys reviewer state.

src/ctest.ts
Nerixyz: Added a type predicate above.

src/ctest.ts
hanniavalera: i see, thank you for addressing everything!

hanniavalera (approved):

margaretselzer:
Independent reproduction confirmed

I hit this same crash on CMake Tools 1.24.42. The error flood originates from two unguarded accesses when a discovered test has a name but no command array:

- getTestsForOutline — executablePath: test.command[0] throws when test.command is undefined.
- resolveTestSourceLocation — test.properties.filter(...) throws when test.properties is undefined.

Environment:

- CMake Tools 1.24.42
- VS Code Remote (SSH server) on Linux
- Cross-compiled configuration (aarch64 / CM4 preset)

Repro: With a cross-compiled preset where CTest reports test entries lacking a populated command/properties, opening a file with test code triggers a continuous error flood and the Test Explorer / code lenses fail to populate.

Verified fix: The changes in this PR (#5095) resolve it — guarding getTestsForOutline to skip entries without a valid command, and making the properties access null-safe in resolveTestSourceLocation. Applying both locally stopped the flood and restored code lenses.

Thanks for putting this together — would be great to see it land.

PR #5101, opened 2026-09-17
Fix CTest discovery rescanning every source of a test executable for each test (#5093)

### LINKED ISSUE #5093: [Bug] Per-test CTest source resolution rescans every target source, freezing the extension host for minutes after each configure
### Brief Issue Summary

After 1.24.42 landed our workspace became unusable in VS Code.

On window open the status bar sits on "Configuring..." for several minutes with nothing further appearing in the CMake/Build output channel, and the extension host is unresponsive the whole time (other extensions stop answering, and opening a view shows "Activating Extensions..."). Launching a target looks the same. ninja reports no work to do, the log prints

```
[driver] Build completed: 00:00:00.748
[build] Build finished with exit code 0
```

and then nothing happens for minutes before the executable starts. This is every configure and every launch, not a one-off.

The project itself is fine. The same configure takes 2.1 s from the command line, a clean 662-step build takes 28.8 s, and the target is up in 0.26 s.

### What it is doing

I profiled the extension host while it was stuck. All of the time is in CTest source resolution:

```
36.9%  existsSync
35.4%  readFileUtf8
16.7%  readFileSync
 8.3%  parseDoctestCases                          dist/main.js:7834
 0.6%  tryResolveTestSourceFromExecutableSources  dist/main.js:7760
```

`tryResolveTestSourceFromExecutableSources` walks every source file of the test's target once per test, and `tryResolveDoctestLine` does `existsSync` + `readFileSync` + a regex parse for each of those files, with nothing cached between tests:

```js
for (const source of targetInfo.sources) {
    const candidateFile = path.resolve(targetInfo.sourceDir, source);
    const candidateLine = this.tryResolveDoctestLine(candidateFile, test.name);
    if (candidateLine !== undefined) { ... }
}
```

We use GoogleTest through `gtest_discover_tests()`, which registers 4093 CTest tests, and the test target has 250 sources. Those sources contain no doctest `TEST_CASE`s, so `parseDoctestCases` never matches, the loop never exits early, and every test ends up scanning all 250 files. That is up to 1,023,250 `existsSync` + `readFileSync` + parse calls, synchronously on the extension host, per configure.

Two things make it avoidable:

The expensive path runs before the cheap one. `resolveTestSourceLocation` calls `tryResolveTestSourceFromExecutableSources` first and only checks the `DEF_SOURCE_LINE` test property afterwards. `gtest_discover_tests` already writes `DEF_SOURCE_LINE` for every test it registers, so the correct file and line were available without touching the disk at all.

There is also no way to opt out. `cmake.ctest.testExplorerIntegrationEnabled: false` gates the post-build `refreshTests` call (#4909, #4914), but the post-configure call is unconditional:

```js
const result = await drv.configure(trigger, []);
if (result.exitCode === 0) { ... }
await this.cTestController.refreshTests(drv);   // not gated
```

`cmake.configureOnOpen: false` does not help either, since it still runs off the cached configuration. I measured both and neither changed the profile.

### Numbers

Same workspace, same settings, only the extension version changed. CPU time measured 240 s after opening the folder, with the machine otherwise idle:

| Version | Extension host process | All VS Code processes |
| --- | --- | --- |
| 1.24.42 | 220.0 s, still climbing (~92% of one core, indefinitely) | 242.8 s |
| 1.22.28 | 5.0 s, flat after 40 s | 14.2 s |

We have pinned 1.22.28 as a workaround, which is also why the diagnostics block below reports 1.22.28 rather than 1.24.42.

### Repro

1. A CMake project with a GoogleTest target registered through `gtest_discover_tests()`, where the test count multiplied by the number of sources in the test target is large. Ours is 4093 x 250, but it should be visible well below that.
2. Build once so CTest can enumerate the tests.
3. Open the folder with CMake Tools 1.24.42 and let it configure.
4. The extension host pegs a core for minutes after the configure has already finished.

### Suggested fix

Check `DEF_SOURCE_LINE` before falling back to scanning target sources, and memoize the parse per file, since the same 250 files are read and parsed 4093 times in a row. Honoring `cmake.ctest.testExplorerIntegrationEnabled` on the post-configure `refreshTests` call would at least give people a way out in the meantime.

### CMake Tools Diagnostics

```shell
{
  "os": "win32",
  "vscodeVersion": "1.137.0",
  "cmtVersion": "1.22.28",
  "configurations": [
    {
      "folder": "c:\Users\mewio\Documents\GitHub\Shalltear",
      "cmakeVersion": "4.2.0",
      "configured": true,
      "generator": "Ninja",
      "usesPresets": true,
      "compilers": {
        "C": "C:/Users/mewio/Documents/GitHub/Shalltear/External/Prebuilt/llvm-mingw/bin/x86_64-w64-mingw32-clang.exe",
        "CXX": "C:/Users/mewio/Documents/GitHub/Shalltear/External/Prebuilt/llvm-mingw/bin/x86_64-w64-mingw32-clang++.exe"
      }
    }
  ],
  "cpptoolsIntegration": {
    "isReady": false,
    "hasCodeModel": false,
    "activeBuildType": "",
    "buildTypesSeen": [],
    "requests": [],
    "responses": [],
    "partialMatches": [],
    "targetCount": 0,
    "executablesCount": 0,
    "librariesCount": 0,
    "targets": []
  },
  "settings": [
    {
      "useCMakePresets": "auto",
      "configureOnOpen": true
    }
  ]
}
```

### Debug Log

```shell
cmake.loggingLevel: trace, cmake.showTimestampsInOutput: true. Configure finishes at 13:38:24.481 and the last line of any kind is 13:38:25.092; the extension host then stays pegged at ~93% of one core for minutes with nothing further logged.

2026-09-13T13:38:23.441Z [info] [proc] Executing command: "C:\Program Files\CMake\bin\cmake.EXE" -DCMAKE_BUILD_TYPE=Debug -DSHALLTEAR_BUILD_SHARED=ON -DCMAKE_C_COMPILER=... -DCMAKE_CXX_COMPILER=... -DCMAKE_MAKE_PROGRAM=... -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -S C:/Users/mewio/Documents/GitHub/Shalltear -B C:/Users/mewio/Documents/GitHub/Shalltear/build -G Ninja
2026-09-13T13:38:24.478Z [trace] [cache] Read line in cache with name=GIT_EXECUTABLE-ADVANCED, typename=INTERNAL, valuestr=1
2026-09-13T13:38:24.481Z [trace] [cache] Parsed 161 cache entries
2026-09-13T13:38:24.481Z [trace] [cache] Parsed 161 entries from C:/Users/mewio/Documents/GitHub/Shalltear/build/CMakeCache.txt
2026-09-13T13:38:25.092Z [trace] [rollbar] Invoking function [$] with Rollbar wrapping [$Resolving process on "close" event]

<nothing further; UI frozen from here>
```

### Additional Information

The profile was taken by launching VS Code with `--inspect-extensions` and running a CPU profile against the extension host over the DevTools protocol while the freeze was in progress, on an otherwise idle machine. The A/B was run in a separate `--user-data-dir` and `--extensions-dir` so that nothing but the extension version differed between the two runs.

Environment: VS Code 1.137.0, Windows 11 build 26200, CMake 4.2.0, Ninja, clang 23.1.1 (llvm-mingw), 28 logical cores. Test target is 250 sources and 4093 CTest tests registered by `gtest_discover_tests`.

mewiof:
Fixes #5093.

### What happens today

After every configure and build, CTest discovery resolves a source location for each test. Since #4923 that starts with `tryResolveTestSourceFromExecutableSources()`, which reads and regex-parses every source file of the test's executable, per test, with nothing cached between tests; the `DEF_SOURCE_LINE` property is only consulted afterwards. With `gtest_discover_tests()` no source contains a `TEST_CASE`, so the scan never matches, every test walks every source, and a target with 4093 tests and 250 sources ends up at roughly a million `existsSync` + `readFileSync` + parse calls, synchronously on the extension host. The same resolution also runs for the Project Outline and the Test CodeLens provider through `getTestsForOutline()`.

### What this changes

- `DEF_SOURCE_LINE` is checked first. CMake's GoogleTest module writes it for every discovered test specifically so that this extension can locate the test (the comment in `Modules/GoogleTest.cmake` says so), and neither CMake itself, doctest nor Catch2 emit it, so no framework that relied on the source scan loses anything. This is also the priority that the doc comment of `resolveTestSourceLocation()` has described since #4780; #4923 inserted the scan above it.
- For tests without the property, the declarations found in the executable's sources are parsed once per discovery run and shared by all tests of that executable (`TestCaseDeclarationCache`). `parseDoctestCases` and `normalizeDiscoveredTestName` move to module scope unchanged.
- Name matching runs over all of the executable's declarations per tier (exact, normalized, containment, token overlap) instead of file by file, so an exact `TEST_CASE("beta")` in a later file is no longer shadowed by a `TEST_CASE("alpha beta")` in an earlier one. Within a tier the earliest declaration in source order still wins, as before.

The trailing line refinement for a location at line 1 is kept and goes through the same cache. No settings change.

### Numbers

Repro project from the issue (250 files × 16 `TEST()`s = 4000 tests, `gtest_discover_tests()`), Visual Studio 18 2026 generator, extension development host on VS Code 1.138.0, Windows 11, 28 logical cores. CPU time sampled over the 180 s after selecting the configure preset, with the project already built so configure-on-open runs discovery:

| | extension host process | all processes of the instance |
| --- | --- | --- |
| `main` (27ad50e2) | 115.6 s, one core pegged for about two minutes | 192.1 s |
| this branch | under 1 s | 3.8 s |

### Tests

New cases in `test/unit-tests/ctest.test.ts` drive `getTestsForOutline()` over real source files in a temporary directory and count the `fs` calls with sinon:

- `DEF_SOURCE_LINE` resolves without any file read, even when the sources contain a matching declaration (fails on `main`)
- the sources of an executable are read once per discovery run, whatever the number of tests (fails on `main`: six reads for three tests, and the wrong file for one of them)
- an exact declaration in a later source wins over a partial match in an earlier one (fails on `main`)
- the first-source fallback, and the line-1 refinement of a `DEF_SOURCE_LINE` location (the latter fails on `main` for the same shadowing reason)

`yarn lint`, `yarn backendTests` (460 passing) and the `ctest.test.ts` unit suite (27 passing) pass locally on Windows.


mewiof:

